PR #5923, opened 2022-05-30
Resolve "Add job support to schedules"

### LINKED ISSUE #3399: Add job support to schedules and to airflow
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3479

Originally created by @pandemicsyn on 2022-05-11 20:28:10

---

In https://gitlab.com/meltano/meltano/-/issues/2924 we're adding support for top level jobs concept in meltano.yaml along with adding support for jobs expansion to `meltano run`. We also want to add job support to the `meltano schedule` command, and likely also alter how airflow interacts with schedules ala https://gitlab.com/meltano/files-airflow/-/blob/master/bundle/orchestrate/dags/meltano.py. 

## Some clarification on things that caused confusion in [#2924](https://gitlab.com/meltano/meltano/-/issues/2924).

1. We are NOT deprecating support `elt` schedules in favor of `run`, and `meltano schedule` including `meltano schedule run` will support both.
2. Airflow calls `meltano schedule run` [per schedule](https://gitlab.com/meltano/files-airflow/-/blob/master/bundle/orchestrate/dags/meltano.py#L99), it does NOT use `meltano elt` directly,  `meltano schedule run` uses MeltanoInvoker to spawn a `meltano elt` subprocess.
   - at a minimum `meltano schedule run` will need to support calling both `meltano elt` and `meltano run` based on the scheduled its being asked to run.

## Latest `meltano schedule` UX spec
(ala @tayloramurphy in https://gitlab.com/meltano/meltano/-/issues/2924#note_937990262)

```
meltano schedule <schedule_name> <job_name> <interval> # This is the default
meltano schedule job add <schedule_name> <job_name> <interval> # This is the same behavior as the above but [job] and [add] are default
meltano schedule elt <schedule_name> <extractor> <loader> <interval> [--transform={run,skip,only}] # new syntax for elt workflows Q: Does this break anything in the UI?
meltano schedule elt add <schedule_name> <extractor> <loader> <interval> [--transform={run,skip,only}] # same as above with the default add
meltano schedule <schedule_name> set <job_name> <interval> <environment>
meltano schedule <schedule_name> set name=<job_name> # this is different from the description b/c of the key (same for the next 2) - Q: Is the key required?
meltano schedule <schedule_name> set interval=<interval>
meltano schedule <schedule_name> set environment=<environment>
meltano schedule list [--format=json] Does this work with jobs that are scheduled via the job definition?
meltano schedule run <schedule_name>
```

MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/merge_requests/2652

Originally created by @pandemicsyn on 2022-05-17 19:01:49

---

## Merge Request Checklist

- [x] Link to an issue
- [x] Include the proposed fix or feature
- [x] Include and update tests for the modified code
- [x] Include a documentation change
- [x] Add a [CHANGELOG.md](https://gitlab.com/meltano/meltano/blob/master/README.md#Changelog) entry in the Unreleased section for any user-facing changes
- [ ] Mention or assign a maintainer

Closes [#3479](https://gitlab.com/meltano/meltano/-/issues/3479) 

Spec:

```
# Both of these are okay. Same 'add' command but obviously different workloads being scheduled:
meltano schedule add SCHEDULE_NAME --job JOB_NAME --interval INTERVAL
meltano schedule add SCHEDULE_NAME --extractor EXTRACTOR --loader LOADER --interval INTERVAL

# This fails with a message like "You can specify `--job` or `--extractor` but not both."
meltano schedule add SCHEDULE_NAME --job JOB_NAME --extractor EXTRACTOR

meltano schedule set SCHEDULE_NAME <all the flags as optional from add>

# This fails with a message like "Nothing to schedule."
meltano schedule add SCHEDULE_NAME

# will behave as it does today when calling a legacy elt job
meltano schedule run SCHEDULED_ELT_NAME

#will call meltano run under the hood when the schedule is a job and effectively just be an alias for meltano run in that scenario.
#It will not call meltano run per task , it will simply call meltano run <job_name>
meltano schedule run SCHEDULED_JOB_NAME 

#The new schema output from meltano schedule list --format=json will clearly distinguish between legacy elt schedules and new jobs with tasks.
meltano schedule list
```

Implementation checklist:

- [x] meltano schedule add 
- [x] update CI sample_code tests to work with new cli def
- [x] meltano schedule list
- [x] meltano schedule remove
- [x] meltano schedule set
- [x] ScheduleService
   - [x] refactor handle both elt and job fields 
   - [x] refactor run to support invoking both patterns
- [x] Update Schedule API endpoints
- [x] add FF to allow scheduled job in API listings
   - [x] Updated associated tests

Telemetry:

- Add `meltano schedule set` 
- Update meltano schedule run to distinguish schedule types

Doc Updates:

* [x] Updated docs for `meltano schedule`
* [x] Update Getting Started guide
* [x] Update "Home Page" sample code


.gitlab/ci/sample_code.gitlab-ci.yml
```diff
@@ -271,12 +271,16 @@ getting_started_sample_code:
...
     ### https://docs.meltano.com/getting-started#run-a-data-integration-el-pipeline
 
-    - meltano schedule gitlab-to-postgres tap-gitlab target-postgres @daily
-
+    - meltano schedule add gitlab-to-postgres --extractor tap-gitlab --loader target-postgres --interval "@daily"
     - meltano schedule list
 
     - meltano add orchestrator airflow
 
+    ### quick job/run related sanity checks
```
pandemicsyn, line 279: could definitely remove this, but I snuck in a quick sanity check to make sure scheduling jobs actually works.

src/meltano/api/controllers/orchestrations.py
```diff
@@ -473,34 +475,50 @@ async def test_extractor():
...
     schedule_service = ScheduleService(project)
     schedules = list(map(dict, schedule_service.schedules()))
+
+    jobs_in_list = False
+    with ProjectSettingsService(project).feature_flag(
+        FeatureFlags.ENABLE_API_SCHEDULED_JOB_LIST, raise_error=False
+    ) as allow:
+        if allow:
+            jobs_in_list = True
+
```
pandemicsyn, line 493: @aaronsteers I wrapped that short-circuit to the API listing that we talked about in a feature flag. Mainly to make sure that we don't "forget" about it 😁

src/meltano/cli/schedule.py
```diff
@@ -27,57 +32,154 @@ def schedule(project, ctx):
...
         )
+    except ScheduleAlreadyExistsError:
+        click.secho(f"Schedule '{name}' already exists.", fg="yellow")
     finally:
         session.close()
 
 
+@schedule.command(short_help="[default] Add a new schedule.")
+@click.argument("name")
+@click.option("--interval", "-i", required=True, help="Interval of the schedule.")
```
pandemicsyn, line 90: I added both long and short flags, happy to remove the short flags - but fwiw - its definitely nice not having to type them out.

src/meltano/cli/schedule.py
```diff
@@ -27,57 +32,154 @@ def schedule(project, ctx):
...
+
+def _format_job_list_output(entry: Schedule, job: TaskSets) -> dict:
+    return {
+        "name": entry.name,
+        "interval": entry.interval,
+        "cron_interval": entry.cron_interval,
+        "env": entry.env,
+        "job": {
+            "name": job.name,
+            "tasks": job.tasks,
```
pandemicsyn, line 143: Since files-airflow will need it (because we're going to do a dag per job task), I fetch the job tasks and display them. We could skip this and just force files-airflow to call `meltano jobs list` to get a list of job/tasks.

docs/src/_guide/orchestration.md
```diff
@@ -16,13 +16,13 @@ When a new pipeline schedule is created using the [UI](/reference/ui) or [CLI](/
...
 ```bash
-meltano schedule [SCHEDULE_NAME] [EXTRACTOR_NAME] [TARGET_NAME] [INTERVAL]
+meltano schedule [SCHEDULE_NAME] --extractor [EXTRACTOR_NAME] --loader [TARGET_NAME] --interval [INTERVAL]
 ```
 
 Example:
 
 ```bash
-meltano schedule carbon__sqlite tap-carbon-intensity target-sqlite @daily
+meltano schedule carbon__sqlite --extractor tap-carbon-intensity --loader target-sqlite --interval="@daily"
```
aaronsteers, line 25: Should these use `schedule add`?

pandemicsyn:
Ok, @aaronsteers I think this is ready for your review. If it passes UX review and the cli looks good I'll ping Edgar/Cody for a review.

docs/src/_guide/orchestration.md
pandemicsyn: Ah yea good catch, let me give these a once over and make sure I actually updated them correctly.

tests/meltano/api/controllers/test_sql_controller.py
```diff
@@ -3,27 +3,33 @@
```
aaronsteers, line 3: For my own info, what is this file and should this functionality be phased out in the near future? (Not in this PR to be clear, but just want to better understand.)

src/meltano/cli/schedule.py
aaronsteers: @pandemicsyn - I do like the short flags. However, I want to be careful about the namespace of known flags and try to avoid collisions down the road as we add more.

- `-i` could be `--interval` or `--interactive`
- `-e` could be `--extractor` or `--environment`
- `-l` could be `--loader` or `--log-level`

What do you think? 

cc @tayloramurphy

src/meltano/cli/schedule.py
aaronsteers: @pandemicsyn - Thanks for calling this out. Both ways could work, but I think it's completely reasonable to include `tasks` here, as you have it - and I think that'd be my slight preference. Although it's redundant with what one would get from `meltano list jobs`, there's clear value in including it. And there's benefit in not needing to make two calls when one will suffice.

Certain applications will need to ignore it, like the proposed `--format=crontab` and/or a cron orchestrator which won't be able to invoke the tasks separately anyway. However, it seems reasonable to include this on default outputs, and callers can always ignore the `tasks` entry if they want to invoke directly by job name.

src/meltano/core/bundle/settings.yml
```diff
@@ -191,16 +191,19 @@ settings:
...
   value: false
-
-# Feature Flags
+# Control whether to use Uvicorn rather than Gunicorn as the API server.
 - name: ff.enable_uvicorn
   kind: boolean
   value: False
-
-# Feature Flags
+# Control whether the API includes scheduled jobs in listings.
```
aaronsteers, line 202: Possibly as a follow-on but would be useful if we could put the comment text as `description: <>`.

I don't know if this is supported though, and it's not worth blocking on. (Fwiw, I checked the yaml file and no 'description' entries exist.)

aaronsteers (changes requested):
A few relatively minor items for review or additional discussion.

tests/meltano/api/controllers/test_sql_controller.py
pandemicsyn: Yea its part of the old `/repos/sync` API path (not sure what that actually does), so would hope this is going to get phased in API 2.0.

Because I renamed the `schedule` fixture, which this test referenced, I had to update it. Then because it hadn't been touch recently, I also had to satisfy flake8/linter. 😮‍💨

aaronsteers:
There was a repo-level setting to enable this. Now we have the options:

image

aaronsteers:
@pandemicsyn - I've updated from `main`, which hopefully should resolve the lint CI failure.

tests/meltano/api/controllers/test_sql_controller.py
pandemicsyn: @aaronsteers fyi - cody is actually removing this as part of https://github.com/meltano/meltano/pull/5965/files#diff-0fc5534da41454b96e1a902d456ef8d91502d8269a306ffb19ba1d6e56560ce4, so this file is getting nuked anyway 🥳

edgarrmondragon (approved):

src/meltano/core/settings_service.py
```diff
@@ -31,6 +31,7 @@ class FeatureFlags(Enum):
 
     LOCKFILES = "lock_files"
     ENABLE_UVICORN = "enable_uvicorn"
+    ENABLE_API_SCHEDULED_JOB_LIST = "enable_api_scheduled_job_list"
```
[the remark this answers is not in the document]
edgarrmondragon: @pandemicsyn makes sense. That's my only question, this looks good 😄

src/meltano/cli/schedule.py
tayloramurphy: @pandemicsyn how about we just leave it off for now so as to unblock this. Can you then open a discussion (or even ADR) on how we want to use short flags?

src/meltano/core/bundle/settings.yml
pandemicsyn: > I don't know if this is supported though,

It supported in the sense that I can add it and it doesn't break anything. But I don't think we have anything that would automatically make use of it or expose it, at least that I'm aware of.

docs/src/_reference/command-line-interface.md
```diff
@@ -757,29 +759,49 @@ The interval argument can be a [cron expression](https://en.wikipedia.org/wiki/C
...
+meltano schedule set gitlab-sync --interval "@weekly"
+
+# Add a legacy ELT based schedule named "gitlab-to-jsonl" to run every minute
 # This specifies that the following command is to be run every minute:
 # meltano elt tap-gitlab target-jsonl --job_id=gitlab-to-jsonl
-
-meltano schedule run gitlab-to-jsonl --select=commits
-# This will run:
-# meltano elt tap-gitlab target-jsonl --job_id=gitlab-to-jsonl --select=commits
+meltano schedule gitlab-to-jsonl --extractor tap-gitlab --loader target-jsonl --interval="* * * * *"
```
aaronsteers, line 802: Ditto above. Should this use `add`?

```suggestion
meltano schedule add gitlab-to-jsonl --extractor tap-gitlab --loader target-jsonl --interval="* * * * *"
```

tests/meltano/cli/test_schedule.py
```diff
@@ -37,22 +39,165 @@ def test_schedule(self, get, session, project, cli_runner, schedule_service):
...
+
+        # verify that you can't use job and elt flags together
+        with mock.patch(
+            "meltano.cli.schedule.ScheduleService", return_value=schedule_service
+        ):
+            res = cli_runner.invoke(
+                cli,
+                [
+                    "schedule",
+                    "job-schedule-mock",
```
aaronsteers, line 81: Should this be "schedule", "add",...

```suggestion
                    "schedule",
                    "add",
                    "job-schedule-mock",
```

Or if we are supporting both methods, perhaps the test should test both ways?

.gitlab/ci/sample_code.gitlab-ci.yml
```diff
@@ -121,8 +121,8 @@ homepage_sample_code:
     ### https://docs.meltano.com/guide/orchestration
 
     # Schedule pipelines
-    - meltano schedule gitlab-to-jsonl tap-gitlab target-jsonl @hourly
-    - meltano schedule gitlab-to-postgres tap-gitlab target-postgres @daily --transform=run
+    - meltano schedule gitlab-to-jsonl --extractor tap-gitlab --loader target-jsonl --interval="@hourly"
+    - meltano schedule gitlab-to-postgres --extractor tap-gitlab --loader target-postgres --interval="@daily" --transform=run
```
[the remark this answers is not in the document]
pandemicsyn: Yea probably should since that's what I switched the docs too after the first review pass. The o.g. spec explicitly called out that add should be default behavior for schedule so I didn't update the use every where and just updated to use the appropriate flags.

aaronsteers (approved):
Thanks, @pandemicsyn. Exciting update! 🚢 when ready.

PR #5957, opened 2022-05-31
Remove add_related for TRANSFORMS

### LINKED ISSUE #3330: Fail adding `transform` plugin if no `transformer` is added
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3408

Originally created by @aaronsteers on 2022-04-18 04:50:25

---

Discovered from my dogfooding project (https://gitlab.com/meltano/meta/-/issues/234+), we currently auto-add the `model`, `transform`, and `transformer` plugins when adding extractors.

When adding extractors, we should not auto-add any other plugins:

- `model` plugins are being deprecated.
- `transform` plugins are being deprecated.
- `transformer` plugin can not be inferred to be `dbt`, since we are opting for adapter-specific transformer plugins now.

kgpayne:
This pr:

- removes the auto-add functionality that applied to Transforms (whereby the legacy 'dbt' Transformer was auto-added if it was missing)
- fails if a Transform is added without a Transformer plugin present in the users Project

As a side effect, this now _allows a Transform plugin to be added to a project with an adapter-specific dbt install_. **However** using an adapter-specific dbt with existing Transforms will fail due to a missing `DBT_SOURCE_SCHEMA` env var that was present as a setting in the legacy 'dbt' Transformer plugin definition but is not included in the new adapter-specific Transformer definitions. This should be addressed in a future issue if we wish to support the use of Transforms with adapter-specific dbt Transformers.


src/meltano/core/plugin_discovery_service.py
```diff
@@ -482,7 +482,8 @@ def find_related_plugin_refs(
 
         runner_ref = target_plugin.runner
         if runner_ref:
-            related_plugin_refs.append(runner_ref)
+            if runner_ref.type in plugin_types:
+                related_plugin_refs.append(runner_ref)
```
edgarrmondragon, line 486: How about something like

```python
             if runner_ref and runner_ref.type in plugin_types:
                 related_plugin_refs.append(runner_ref)
```

src/meltano/cli/utils.py
```diff
@@ -466,3 +468,34 @@ def _handler(sig, _):  # noqa: WPS430
         yield
     finally:
         signal.signal(signal.SIGTERM, original_termination_handler)
+
+
+def check_dependencies_met(plugins, plugins_service: ProjectPluginsService):
```
edgarrmondragon, line 473:
```suggestion
def check_dependencies_met(plugins, plugins_service: ProjectPluginsService) -> Tuple[bool, str]:
```

edgarrmondragon (dismissed):

src/meltano/core/plugin_discovery_service.py
kgpayne: Did this syntax change to be supported recently? I expect that in cases where `runner_ref = None` the `runner_ref.type` check will raise an `AttributeError`, which is why I have nested the `if` statements 🤔

edgarrmondragon: @kgpayne I don't know how recently the syntax is supported but this works in Python 3.7 😅 

```python
>>> def process_obj(obj):
    if obj and obj.x == 1:
        print(obj.x)
    else:
        print("It's none")
... 
>>> class MyClass:
    x = 1
>>> 
>>> process_obj(MyClass())
1
>>> process_obj(None)
It's none
```

I think everything after the `and` is lazily evaluated. So only if the first statement is truthy, the second one is processed.

edgarrmondragon (approved):

aaronsteers:
I've resolved the merge conflict by re-locking poetry.

src/meltano/cli/add.py
```diff
@@ -47,11 +54,6 @@
     is_flag=True,
     help="Add a custom plugin. The command will prompt you for the package's base plugin description metadata.",
 )
-@click.option(
-    "--include-related",
-    is_flag=True,
-    help="Also add transform plugins related to the identified discoverable extractor.",
-)
```
aaronsteers, line 54: TODO: We may want to call out in migration docs somewhere that the `--include-related` CLI arg is removed from `meltano add`.

src/meltano/cli/utils.py
```diff
@@ -495,3 +497,35 @@ def _handler(sig, _):  # noqa: WPS430
...
+
+    for plugin in plugin_refs:
+        if plugin.type == PluginType.TRANSFORMS:
+            # check that the `dbt` transformer plugin is installed
+            try:
+                plugins_service.get_transformer()
+            except PluginNotFoundError:
+                passed = False
+                messages.append(
+                    f"Plugin '{plugin.name}' requires a Transformer plugin."
```
aaronsteers, line 525:
```suggestion
                    f"Plugin '{plugin.name}' requires a Transformer plugin. "
                    "Please first add a transformer using `meltano add transformer ...`.
```

src/meltano/cli/utils.py
```diff
@@ -495,3 +497,36 @@ def _handler(sig, _):  # noqa: WPS430
...
+    for plugin in plugin_refs:
+        if plugin.type == PluginType.TRANSFORMS:
+            # check that the `dbt` transformer plugin is installed
+            try:
+                plugins_service.get_transformer()
+            except PluginNotFoundError:
+                passed = False
+                messages.append(
+                    f"Plugin '{plugin.name}' requires a Transformer plugin. "
+                    "Please first add a transformer using `meltano add transformer`.
```
aaronsteers, line 526: 🤦 

```suggestion
                    "Please first add a transformer using `meltano add transformer`."
```

src/meltano/cli/utils.py
```diff
@@ -495,3 +497,36 @@ def _handler(sig, _):  # noqa: WPS430
...
+    for plugin in plugin_refs:
+        if plugin.type == PluginType.TRANSFORMS:
+            # check that the `dbt` transformer plugin is installed
+            try:
+                plugins_service.get_transformer()
+            except PluginNotFoundError:
+                passed = False
+                messages.append(
+                    f"Plugin '{plugin.name}' requires a Transformer plugin. "
+                    "Please first add a transformer using `meltano add transformer`."
```
aaronsteers, line 526:
```suggestion
                    f"Plugin '{plugin.name}' requires a transformer plugin. "
                    "Please first add a transformer using `meltano add transformer`."
```

aaronsteers (approved):
Nice work, @kgpayne ! This looks ready to go.

PR #5978, opened 2022-06-02
Replace `GoogleAnalyticsTracker` with Snowplow-backed `MeltanoTracker`

### LINKED ISSUE #3366: Add Snowplow events at the block level
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3444

Originally created by @pnadolny13 on 2022-04-26 20:36:14

---

Relates to https://gitlab.com/meltano/meltano/-/issues/3150

We want to send events that tell us details about the execution of a plugin block. Many can be fired from a single CLI `run` command i.e. `run tap-csv target-postgres dbt:run dbt:test` could end up being an EL event and 2 command events (naming TBD).

Steps:
- determine where events should fire from
- determine all events we want to send
- design custom event and content
- create json schema for event
- register schema with snowcat cloud (our Snowplow hosting vendor)
- update meltano to fire events

### LINKED ISSUE #5941: Upgrade CLI Snowplow events and turn off GA telemetry
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3443

Originally created by @pnadolny13 on 2022-04-26 18:30:00

---

Relates to [#3150 (comment 925087374)](https://gitlab.com/meltano/meltano/-/issues/3150#note_925087374)

Just to recap the main challenges to solve:
> - Analysis is flaky because we need to parse command strings. We should split commands out logically into a structured format before sending.
> - Need more metadata about what plugin variants and versions are being run
> - Need more general metadata about users runtime environment i.e. OS, python version, meltano version, etc.
> - We dont know what plugins are failing vs succeeding.
> - We have no execution timing information. How long are commands running for?

Steps:
- try to solve all the challenges using Snowplow contexts
- if its possible
  - append the context to all CLI events as they are today
- if it isnt possible build custom CLI events
  - design
  - create json schema
  - update meltano to send self describing CLI events instead of structured events

Scope:
- Targeting only execution CLI events (run/elt/invoke/schedule run)

### LINKED ISSUE #5959: Telemetry signal if project ID is from a name hash
> @tayloramurphy, @pnadolny13, @WillDaSilva - I'll log a follow-on issue to distinguish between _actual_ UUIDs and hash-based IDs. For inputs like `meltano-hub-project` and `tap-gitlab--meltanolabs`, these will still be high probability of being unique.
>
> For inputs like `my_meltano_project` or `my_data_project`, I imagine we'll get collisions on those and we'll want a way to know the linkage is more likely to give false-positives.

_Originally posted by @aaronsteers in https://github.com/meltano/meltano/pull/5958#pullrequestreview-991017070_

WillDaSilva:
Closes #5941
Closes #5959


src/meltano/core/tracking/schemas/command_event.json
```diff
@@ -0,0 +1,73 @@
...
+    "properties": {
+        "event_uuid": {
+            "description": "A UUID that uniquely identifies an instance of this event",
+            "type": "string",
+            "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$",
+            "minLength": 36,
+            "maxLength": 36
+        },
+        "event": {
+            "description": "",
```
WillDaSilva, line 20: Descriptions are required by SnowcatCloud for most entries. Several are blank because I have yet to fill them in, but others will probably just remain blank as they are only there to satisfy SnowcatCloud's validator.

WillDaSilva:
There are 23 sites in the repository where a `GoogleAnalyticsTracker` method is called, and as best I can tell they will all be replaced by `MeltanoTracker.track_command_event`.

I'm thinking `Project` should have a `tracker` member. This will be the base tracker that only has the environment context. It comes with a context manager `MeltanoTracker.with_contexts` that can be used to track events with additional context. For instance:

```python
project.tracker.track_xyz_event({'name': 'qwerty', 'magnitude': 789})

plugin_context = SelfDescribingJson(
    'iglu:com.meltano/command_event/jsonschema/1-0-0',
    {...}
)

with project.tracker.with_context(plugin_context) as tracker:
    # `tracker` is only provided by the context manager for convenience - the same instance is used    
    assert project.tracker is tracker
    tracker.track_xyz_event({...}) # Tracks an xyz event as before, but this time with additional context
```

WillDaSilva:
> I'm thinking `Project` should have a `tracker` member. This will be the base tracker that only has the environment context. 

An alternative would be to decouple the project from the environment context (i.e. split the environment context into an environment context, and a project context), or to start the tracker off with no contexts and make it so that the caller needs to explicitly provide the environment context.

I'm leaning towards the first option, since we'll always want the environment context, and burdening the caller with obtaining it and passing it through every time seems entirely unnecessary.

src/meltano/core/tracking/schemas/command_event.json
aaronsteers: Thanks for calling this out. I'm totally fine with leaving these blank for now at least. We can open a new issue to fill those out after the fact, if needed.

aaronsteers:
@WillDaSilva - The meeting today was super helpful, I think. Ping me or send me a zoom invite if you need anything further. Thanks!

src/meltano/core/tracking/schemas/README.md
```diff
@@ -0,0 +1,14 @@
...
+SnowcatCloud.
+
+Login to https://app.snowcatcloud.com/ using your Meltano email address. Request permissions from
+an account admin if necessary. As of the time of writing this, the admins are:
+
+- pat@meltano.com
+- will@meltano.com
+- edgar@meltano.com
+- taylor@meltano.com
+- florian@meltano.com
```
WillDaSilva, line 11: It's satisfying how each name (listed in the order we signed up in) is one character longer than the last.

aaronsteers: We just need to add 'aj@' to the top of the list 😆 

Except I'm not actually an admin.

🤔 ...if only I knew who to ask... 🤔 

Okay, jk. I couldn't resist.

src/meltano/core/tracking/__init__.py
```diff
@@ -1 +1,172 @@
-from .ga_tracker import GoogleAnalyticsTracker
+"""Meltano telemetry."""
```
aaronsteers, line 1: I've been trained over time to keep `__init__.py` files as sparse as possible, reserving them mostly just for imports.

Wdyt of moving the bulk of this to something like `/src/meltano/core/tracking/tracker.py`?

aaronsteers (approved):

src/meltano/core/tracking/schemas/environment_context.json
```diff
@@ -0,0 +1,184 @@
...
+        },
+        "meltano_version": {
+            "description": "The version of Meltano in use",
+            "type": "string",
+            "maxLength": 256
+        },
+        "is_dev_build": {
+            "description": "Whether the installed version of Meltano is a dev build",
+            "type": "boolean"
+        },
```
aaronsteers, line 27: Can we tell `is_dev_build` with certainty?

src/meltano/core/tracking/environment.py
```diff
@@ -0,0 +1,124 @@
...
+# This file is only ever created in CI when building a release
+release_marker_path = Path(__file__).parent / ".release_marker"
+
+
+class EnvironmentContext(SelfDescribingJson):
+    def __init__(self):
+        logger.debug(
+            f"Initializing '{type(self).__module__}.{type(self).__qualname__}'"
+        )
+        ci_markers = ('GITHUB_ACTIONS', 'CI')
```
aaronsteers, line 34: I was going to say to remove this if we don't have confidence, but actually, this looks valid, and valuable.

- https://adamj.eu/tech/2020/03/09/detect-if-your-tests-are-running-on-ci/

src/meltano/core/tracking/environment.py
```diff
@@ -0,0 +1,124 @@
...
+        super().__init__(
+            "iglu:com.meltano/environment_context/jsonschema/1-0-0",
+            {
+                "context_uuid": uuid.uuid4(),
+                "meltano_version": meltano.__version__,
+                "is_dev_build": not release_marker_path.exists(),
+                "is_ci_environment": any(os.environ.get(x, '') == 'true' for x in ci_markers),
```
aaronsteers, line 41: Do we need to worry about case sensitivity or getting a value `'1'` or `1`?

Maybe this is too clever but not knowing the exact form of the variable's value, something like this seems safer:

```suggestion
                "is_ci_environment": any(str(os.environ.get(x, '')).lower()[:1] in ["1", "t"] for x in ci_markers),
```

src/meltano/core/tracking/schemas/environment_context.json
```diff
@@ -0,0 +1,184 @@
...
+            "maxLength": 256
+        },
+        "node": {
+            "description": "The (potentially not fully-qualified) network name of the computer",
+            "type": [
+                "string",
+                "null"
+            ],
+            "maxLength": 256
+        },
```
aaronsteers, line 76: Can we leave this and not collect? Seems like tmi.

src/meltano/core/tracking/schemas/environment_context.json
```diff
@@ -0,0 +1,184 @@
...
+            "type": "integer",
+            "multipleOf": 1,
+            "minimum": 1,
+            "maximum": 65535
+        },
+        "process_executable_path": {
+            "description": "The path to the executable used to create the process",
+            "type": "string",
+            "maxLength": 4096
+        },
```
aaronsteers, line 127: Ditto. Seems like TMI and my preference is to not collect.

src/meltano/core/tracking/schemas/environment_context.json
```diff
@@ -0,0 +1,184 @@
...
+                "description": "",
+                "type": "object",
+                "properties": {
+                    "process_id": {
+                        "description": "The PID of the process",
+                        "type": "integer",
+                        "multipleOf": 1,
+                        "minimum": 0,
+                        "maximum": 4294967295
+                    },
```
aaronsteers, line 141: Prefer to not collect.

src/meltano/core/tracking/schemas/environment_context.json
```diff
@@ -0,0 +1,184 @@
...
+                    "process_name": {
+                        "description": "The name of the process",
+                        "type": "string",
+                        "maxLength": 1024
+                    },
```
aaronsteers, line 146: Preference to hash or don't collect.

aaronsteers (dismissed):
More feedback coming but this is first pass.

src/meltano/core/tracking/schemas/environment_context.json
aaronsteers: Since I see the marker file is there. I think 'yes'.

WillDaSilva: What we learn is whether the installed version came from `pypi.org` (or e.g. conda-forge), or if it came from someone building/installing from source. Upon reflection, as the PR is currently, it'll also be true for the version of Meltano found in our official Docker images.

src/meltano/core/tracking/environment.py
WillDaSilva: I originally had pretty much that exact code, but swapped it out for the less-clever version thinking I should stick to something simple that works exclusively with GitLab CI and GitHub Actions. Though in that case the name is no longer accurate. I ought to use `GITLAB_CI` instead of `CI`, and should replace `is_ci_environment` with a string enum called `ci` that can be one of:

- `null`
- `"github_actions"`
- `"gitlab_ci"`

That way we can get more info, and extend it with more CI options by simply adding new enum values to newer versions of the schema.

WillDaSilva:
GitLab CI for this branch is currently failing hard because the Poetry lock file is out of sync with `pyproject.toml`.

It's out of sync because I added a couple of dependencies (`backports.cached_property` and `tzlocal`), but I haven't been able to get Poetry to update the lock file. Poetry just keeps taking more and more memory before ultimately getting killed by the OOM killer.

I've verified that no versions of those dependencies were installed by Meltano previously, and I've manually checked the transitive dependencies of the 2 dependencies I added. I don't see any conflicts, so I'm not sure what's going wrong here.

pandemicsyn:
Not sure why but I think its hanging on backports.cached-property:

```
$ poetry lock -vvv
<-- spam for infinity --->
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
   1: selecting backports.cached-property (1.0.1)
^C   1: Version solving took 12.258 seconds.
   1: Tried 1 solutions.
```

src/meltano/core/tracking/schemas/environment_context.json
aaronsteers: Thanks for the context, @WillDaSilva - For this, and for anything else that's kind of 'nice to have' but causing hangups, I think we can safely trim those from scope if needed and bring back in a future iteration.

WillDaSilva:
@pandemicsyn I'll switch to https://pypi.org/project/cached-property/ and see if Poetry likes that better.

WillDaSilva:
@pandemicsyn Switching to the other cached property package seems to have resolve the issue. I've used both in the past without much trouble, and we're only going to be using the basic features of it, and it's temporary until we drop Python 3.7 support, so I think we're fine with taking the fancier one for now that is more than just a backport.

src/meltano/core/tracking/schemas/environment_context.json
aaronsteers: Great - thanks very much!

pandemicsyn:
> so I think we're fine with taking the fancier one for now that is more than just a backport.

@WillDaSilva 👍 sounds good.

fwiw - its weird - has to be a conflict caused by another dep because this minimal pyproject.toml in a fresh project generates a lock file fine 😕 

```yaml
[tool.poetry]
name = "test"
version = "0.1.0"
description = ""
authors = ["pandemicsyn <syn@ronin.io>"]

[tool.poetry.dependencies]
python = "3.7,<3.10"
"backports.cached-property" = "^1.0.1"
meltano = "^1.105.0"

[tool.poetry.dev-dependencies]

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

src/meltano/core/tracking/__init__.py
WillDaSilva: Will do

src/meltano/core/tracking/environment.py
```diff
@@ -0,0 +1,124 @@
...
+                "python_version": platform.python_version(),
+                "python_implementation": platform.python_implementation(),
+            }
+            | self.system_info
+            | self.process_info,
```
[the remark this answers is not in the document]
WillDaSilva: Good catch - it is not. I'll update it to use `**` unpacking to merge them instead.

WillDaSilva:
Going back to our discussion about having a plugin event versus a plugins context, I'm thinking we should have a plugin context that is for a single plugin. That context is entered when the plugin in loaded. We'll then have a slim `plugin_loaded` event that is emitted within that context.

Though I have not used Snowplow before, I get the impression that it's ideal for us to rely primarily on contexts for gathering information, along with slim events whose main purpose is to provide info on what contexts were activate when the event was emitted.

Maybe I'm totally off-base with how Snowplow ought to be used. What do you all think? @aaronsteers @pnadolny13 @tayloramurphy @pandemicsyn

pnadolny13:
> Maybe I'm totally off-base with how Snowplow ought to be used. What do you all think? @aaronsteers @pnadolny13 @tayloramurphy @pandemicsyn

@WillDaSilva yeah that makes sense to me. I personally havent used Snowplow until now and its super flexible on how we can send data so its really up to us.

The only thing I'd add is that as part of a modeling discussion with Joao from Snowcat Cloud he was encouraging us to model the data in a way that encapsulates all our data in a few flexible and generally useful events/context (i.e. Command vs Meltano ELT or Meltano Run), each context and event has its own schema so making them generic reduces the need to constantly be updating the schemas or adding new ones. I think thats normal modeling/DRY principles anyways but still worth sharing.

edgarrmondragon:
> I'm not sure we wanna fire an event for every plugin seen/loaded

I agree with @pandemicsyn. There some inefficiencies in plugin loading a.t.m. (e.g. extra roundtrips) so events would be overcounted there.

src/meltano/core/tracking/schemas/project_context.json
```diff
@@ -0,0 +1,59 @@
...
+        },
+        "client_uuid": {
+            "description": "A UUID that uniquely identifies the client",
+            "type": "string",
+            "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$",
+            "minLength": 36,
+            "maxLength": 36
+        },
+        "environment_name": {
+            "description": "The name of the active environment for the Meltano project",
```
pandemicsyn, line 43: We'll want to hash this one.

src/meltano/core/tracking/environment.py
```diff
@@ -0,0 +1,125 @@
...
+            mask = DWORD_PTR()
+            if GetProcessAffinityMask(
+                GetCurrentProcess(), ctypes.byref(mask), ctypes.byref(DWORD_PTR())
+            ):
+                return bin(mask.value).count("1")
+            logger.debug("Call to 'GetProcessAffinityMask' failed")
+        logger.debug(
+            "Unable to determine the number of available CPU cores; "
+            "falling back to the total number of CPU cores"
+        )
```
edgarrmondragon, line 121: @WillDaSilva would these messages be useful for end users debugging their Meltano deployment? Might be for us in managed service setting, but then telemetry is not as useful 😅

src/meltano/core/tracking/schemas/plugins_context.json
```diff
@@ -0,0 +1,117 @@
...
+                    "category": {
+                        "description": "",
+                        "type": "string",
+                        "enum": [
+                            "extractor",
+                            "loader",
+                            "trasformer"
+                        ]
+                    },
+                    "type": {
```
pandemicsyn, line 35: drop "type"

src/meltano/core/tracking/schemas/plugins_context.json
```diff
@@ -0,0 +1,117 @@
...
+                        "description": "",
+                        "type": "string",
+                        "maxLength": 256
+                    },
+                    "name_hash": {
+                        "description": "",
```
pandemicsyn, line 41: From zoom: desc should describe "this is a 64char hash field" , but drop pattern, max, min constraints.

src/meltano/core/tracking/schemas/plugins_context.json
```diff
@@ -0,0 +1,117 @@
...
+                        "minLength": 64
+                    },
+                    "pip_url_hash": {
+                        "description": "",
+                        "type": "string",
+                        "pattern": "^[a-fA-F0-9]{64}$",
+                        "maxLength": 64,
+                        "minLength": 64
+                    },
+                    "version": {
```
pandemicsyn, line 79: From zoom: Drop version (not present in plugin today)

src/meltano/core/tracking/schemas/plugins_context.json
```diff
@@ -0,0 +1,117 @@
...
+                        "description": "",
+                        "type": "string",
+                        "maxLength": 256
+                    },
+                    "plugin_definition_hash": {
```
pandemicsyn, line 84: From zoom: renamed to plugin_lock_hash if accessible. make optional.

src/meltano/core/tracking/schemas/plugins_context.json
```diff
@@ -0,0 +1,117 @@
...
+                        "description": "",
+                        "type": "string",
+                        "pattern": "^[a-fA-F0-9]{64}$",
+                        "maxLength": 64,
+                        "minLength": 64
+                    },
+                    "containerized": {
```
pandemicsyn, line 91: From zoom: drop, should be invoker field on command event.

pandemicsyn:
Ok @aaronsteers @pnadolny13 and I just a had a group PR review over zoom to go over the schema's. Logged all the issues inline above. Next steps are:

1. resolve the circular import
2. address the PR feedback that came on the of the call
3. add cli tracking events to `meltano invoke`
4. merge this pr
5. profit!

src/meltano/core/tracking/schemas/plugins_context.json
```diff
@@ -0,0 +1,117 @@
...
+            "description": "An array containing information about each plugin in order",
+            "type": "array",
+            "items": {
+                "description": "",
+                "type": "object",
+                "properties": {
+                    "category": {
+                        "description": "",
+                        "type": "string",
+                        "enum": [
```
pandemicsyn, line 29: From zoom: drop enum, either use string or https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.9.5 (/cc @aaronsteers )

pandemicsyn:
This is gonna probably have *a lot* of linter & doc convention failures (both for the new additions and because theres adjecent updates in some files that haven't been touched during other linter fixups).

src/meltano/core/tracking/environment.py
```diff
@@ -0,0 +1,120 @@
+"""Environment context for the Snowplow tracker."""
+
+from __future__ import annotations
+
+import ctypes
+import ctypes.wintypes
```
pandemicsyn, line 6: This seems to not play nice with python 3.7: https://github.com/meltano/meltano/runs/6734175975?check_suite_focus=true , but seems to at least pass tests on 3.8+. Do we need to ensure that this only imports on Windows platforms?

The python docs don't explicitly mention that this should fail on non-windows systems but various posts like https://stackoverflow.com/questions/20897842/error-in-importing-ctypes-wintypes imply just that.

We're currently using this to get a core count on windows - so we could omit it if we really don't actually need the core count anyway.

pandemicsyn: Update. Ended up just nuking it because I wasn't sure how fragile ctypes where and we don't really test on windows (yet). Happy to put it back in (wrapping the import in try/except) though if we want it.

aaronsteers: @pandemicsyn - Yes, sounds like that is the right call. Sorry I didn't see this earlier.

.pre-commit-config.yaml
```diff
@@ -59,5 +59,6 @@ repos:
     args:
     - "--py3-plus"
     - "--py36-plus"
+    - "--keep-mock"
```
pandemicsyn, line 62: first time running into this but ended up having to add this because of https://github.com/asottile/pyupgrade/issues/314 (just didn't feel like i had the cycles to rewrite tests)

src/meltano/core/tracking/tracker.py
```diff
@@ -0,0 +1,196 @@
...
+            action = f"{action} --environment={hash_sha256(self.project.active_environment.name)}"
+
+        try:
+            self.snowplow_tracker.track_struct_event(
+                category=category,
+                action=action,
+                label=self.project_id,
+            )
+        except Exception as ex:
+            logger.warning(
```
pandemicsyn, line 167: @aaronsteers I think im gonna switch these to logger.debug. If we run into issues with telemetry I don't wanna spam the end-user with errors, and we can easily enough troubleshoot using --log-level=debug. wdyt ?

aaronsteers: Yeah - makes sense! 👍

src/meltano/core/tracking/tracker.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def track_command_event(self, event_json: dict[str, Any]) -> None:
+        """Fire generic command tracking event.
+
+        Args:
+            event_json: The event JSON to track. See cli_event schema for more details.
+        """
+        self.track_unstruct_event(
+            SelfDescribingJson(
+                "iglu:com.meltano/cli_event/jsonschema/1-0-0", event_json
```
aaronsteers, line 194: Here, and everywhere we have something like `"iglu:.../1-0-0"`, what do you think of moving these to constant string declarations at the top of the page.

As a single constant it might be something like `COMMAND_EVENT_SCHEMA_IGLU`.

Or as a two-part, it might be something like `COMMAND_EVENT_SCHEMA_IGLU` and `COMMAND_EVENT_SCHEMA_IGLU_VERSION`

src/meltano/core/utils/__init__.py
```diff
@@ -421,3 +422,15 @@ def human_size(num, suffix="B"):
...
+def hash_sha256(value: str) -> str:
+    """Get the sha256 hash of a string.
+
+    Args:
+        value: the string value to hash.
+
+    Returns:
+        The hashed value of the given string.
+    """
+    return hashlib.sha256(value.encode()).hexdigest()
```
aaronsteers, line 436: Great call to move this into `utils`. 👍

src/meltano/core/tracking/tracker.py
WillDaSilva: A constant dictionary mapping schema names to their version might make sense for this.

.gitlab/ci/publish.gitlab-ci.yml
```diff
@@ -7,14 +7,25 @@ publish:
...
   script:
-    # grab the built package
+    # Add the release marker: a file whose presence indicates that this copy of Meltano was
+    # obtained from a Python package index, rather than directly from source. This marker is used
+    # to improve Meltano's telemetry.
+    - pushd /meltano/dist
+    - ARCHIVE="$(sh -c 'echo "$1"' sh meltano*.tar.gz)"
+    - DIR="$(basename $ARCHIVE .tar.gz)"
+    - tar -xvzf "$ARCHIVE"
+    - touch "$DIR/src/meltano/core/tracking/.release_marker"
```
aaronsteers, line 17: Python packaging is weird sometimes about what actually ships and what doesn't. Can we double check this file gets bundled and installed with the operation? (Either in docs or through a 'real' test.)

Also - and I just realized this now - I think this breaks if a user does `pipx install git+https://github.com/meltano/meltano.git`

For now, I think this can be a bug we would live with, but perhaps we open an issue to track/prioritize in a future PR. 

Wdyt?

aaronsteers (approved):
I did a deep review of each file - minus the micro.conf file which I skimmed.

While I lost my original 15 other comments due to user error (🤦 ) , most were cosmetic and general feedback. 

I will make one more pass to see if I can recall/repost some of my other comments, but none of those were blocking issues.

@pandemicsyn and @WillDaSilva - Thank you both for your hard work on this, @WillDaSilva for starting it and @pandemicsyn for getting us to the finish line.

@pandemicsyn - You are clear to merge when ready.

src/meltano/core/tracking/README.md
```diff
@@ -0,0 +1,59 @@
...
+SnowcatCloud.
+
+Login to https://app.snowcatcloud.com/ using your Meltano email address. Request permissions from
+an account admin if necessary. As of the time of writing this, the admins are:
+
+- pat@meltano.com
+- will@meltano.com
+- edgar@meltano.com
+- taylor@meltano.com
+- florian@meltano.com
```
aaronsteers, line 25: I've got admin too now 😄 

```suggestion
Login to https://app.snowcatcloud.com/ using your Meltano email address. Request permissions from
an account admin if necessary. As of the time of writing this, the admins are:

- aj@meltano.com
- pat@meltano.com
- will@meltano.com
- edgar@meltano.com
- taylor@meltano.com
- florian@meltano.com
```

src/meltano/core/tracking/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+```
+docker run \
+  --mount type=bind,source=$(pwd),destination=/config \
+  -p 9090:9090 \
+  snowplow/snowplow-micro:1.2.1 \
+  --collector-config /config/micro.conf \
+  --iglu /config/iglu.json
+```
```
aaronsteers, line 40: Really appreciate these instructions being inline and included in the repo. Will be a huge help.

src/meltano/core/tracking/environment.py
```diff
@@ -0,0 +1,139 @@
...
+
+        Returns:
+            A dictionary containing the process information. Such as the hashed process name, pid, core counts, etc
+        """
+        process = psutil.Process()
+        with process.oneshot():
+            return {
+                "num_cpu_cores": psutil.cpu_count(),
+                "num_cpu_cores_available": self.num_available_cores,
+                "process_executable_path": process.exe() or None,
```
aaronsteers: @pandemicsyn called this out to me - sorry I missed it earlier.

`process_executable_path` can _definitely_ contain PII and is not permittable to be included.

Even this is PII:

`/usr/ajsteers/source/meltano`

And in addition to PII, there could be any amount of sensitive/non-public info in the path:

`/usr/ajsteers/projects/mysecretworldtakover/meltano/clone-army-analytics/`

aaronsteers (approved):
I'm approving (again) but the `process_executable_path` definitely should be removed. As noted.

pandemicsyn:
It lives!

Screen Shot 2022-06-04 at 2 36 36 PM

src/meltano/core/tracking/iglu-client-embedded/schemas/com.meltano/environment_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,169 @@
+{
+    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+    "description": "Schema for a Snowplow context describing the environment in which Meltano is operating",
```
aaronsteers, line 3: Total nit, but a slight tweak here makes it more clear we aren't talking about the 'environments' feature.

```suggestion
    "description": "Schema for a Snowplow context describing the environment in which the Meltano process is executing",
```

aaronsteers:
Done. 
image

@pandemicsyn - I went through all the files where my prior comments were lost. You should have all my feedback now.

Also - in case you haven't already seen, there are some merge conflicts noted that are pending resolution:

image

pandemicsyn:
Ok, with the latest change set the legacy tracking is working as well. Verified that the events are at least accepted as "good" by snowplow micro (can't speak to whether they look like they old ones 😁 ).

.gitlab/ci/publish.gitlab-ci.yml
pandemicsyn: 👍 I'll log an issue, im not as well versed with python packaging as y'all so gonna defer to y'all on how realiable this is.

pandemicsyn: https://github.com/meltano/meltano/issues/6021

PR #5999, opened 2022-06-03
fix: Add files from Hub for plugin requirements

### LINKED ISSUE #5997: Add from hub: file bundle requires reference not found should warn or error
When using whats currently on `main` to test the add from hub feature with `requires` file bundles, it looks like it just doesnt install the file bundle when the reference isnt found. I would expect it to do something something else, maybe error, or maybe just log a warning that it wasnt found. I'm not sure whats best.

I ran into this by pointing meltano to my local running instance of the hub and adding a non-existant name in requires setting.

cc @edgarrmondragon

edgarrmondragon:
- Code path fixed to use Hub
- Completely replace "add related" behavior

Closes #5997


edgarrmondragon:
cc @pnadolny13

pnadolny13:
> cc @pnadolny13

@edgarrmondragon I tested this locally and it fixed the bug I was seeing.

edgarrmondragon:
@aaronsteers this might bring even more conflicts with https://github.com/meltano/meltano/pull/5957

src/meltano/core/bundle/discovery.yml
```diff
@@ -3435,6 +3435,9 @@ transforms:
     pip_url: 'https://gitlab.com/meltano/dbt-tap-google-analytics.git@config-version-2'
     vars:
       schema: "{{ env_var('DBT_SOURCE_SCHEMA') }}"
+    requires:
+      files:
+        - name: dbt
```
aaronsteers, line 3440: I don't know if we should be making any changes to `discovery.yml`.

```suggestion
```

My expectation was that we would just add these to Hub. Wdyt?

aaronsteers (changes requested):
Looks great! 

Just one point on `discovery.yml`.

src/meltano/core/bundle/discovery.yml
edgarrmondragon: @aaronsteers correct but I'm using `discovery.yml` to mock the Hub, i.e. I'm feeding the mock with this catalog.

edgarrmondragon: https://github.com/meltano/meltano/blob/f564cf24c70add9cf12d8b1d5e94486cd117fe72/tests/conftest.py#L72

aaronsteers (approved):

src/meltano/core/bundle/discovery.yml
edgarrmondragon: @aaronsteers I'll merge this first thing tomorrow since it's blocking dev/testing of Hub features: https://github.com/meltano/meltano/issues/6046

PR #6002, opened 2022-06-03
feat: Make Hub URL a project setting

### LINKED ISSUE #5970: Make the Hub base URL configurable in Meltano
We should probably make this configurable, at least for testing purposes.

It would be nice to be able to stand up the Hub API from its main branch (or a feature one) and test Meltano against it directly instead of mocking it entirely.

_Originally posted by @edgarrmondragon in https://github.com/meltano/meltano/pull/5925#discussion_r885158727_

edgarrmondragon:
Closes #5970


docs/src/_reference/settings.md
```diff
@@ -237,10 +237,27 @@ meltano config meltano set project_readonly true
 export MELTANO_PROJECT_READONLY=true
 ```
 
+### <a name="hub-url"></a>`hub_url`
+
+- [Environment variable](/guide/configuration#configuring-settings): `MELTANO_HUB_URL`
+- Default: [`https://hub.meltano.com`](https://hub.meltano.com)
```
aaronsteers, line 243: Should we default to the versioned root of `https://hub.meltano.com/meltano/api/v1`?

From this sample url:

```
https://hub.meltano.com/meltano/api/v1/plugins/extractors/index
```

I can see wanting to overriding the version link, and I can see custom endpoints (even future meltano endpoits) not having `/meltano/api` in the string, e.g. we could have an alternate like `api.meltano.com/v2` or `api.meltano.com/v2-beta`.

docs/src/_reference/settings.md
```diff
@@ -237,10 +237,27 @@ meltano config meltano set project_readonly true
...
+
+Where Meltano can find the Hub that lists all [discoverable plugins](/concepts/plugins#discoverable-plugins).
+
+This manifest is used by [`meltano discover`](/reference/command-line-interface#discover) and [`meltano add`](/reference/command-line-interface#add), among others.
```
tayloramurphy, line 247: > among others

What else is doing the lookup? Don't need to add the caveat unless it's accurate.

docs/src/_reference/settings.md
DouweM: @aaronsteers If you're overriding the version, wouldn't you make that change directly in `client.py` in your feature branch, since you'd presumably also be updating the code there to account for the API changes? Changing the setting implies you can't change the code, but unchanged code would be unlikely to work with a new API version.

The other reason I kept it as just the root is so that my `https://localhost:4000` config would keep working once if the API version changes in my local meltano and hub. I wouldn't want the code to still be requesting v1 if it's actually expecting v2 endpoints.

docs/src/_reference/settings.md
DouweM: @tayloramurphy This was copied from the `discovery_url` description, fwiw. But the answer is: Any command that at some point needs the full plugin definition, if there is no lockfile or cached `discovery.yml`. So really any plugin-related command could need to do do the lookup. I listed `meltano discover` and `meltano add` here specifically because they're clearly related to discoverable plugins.

Do you think it's better to omit this entirely, or to expand it to cover what I just explained?

tayloramurphy:
```suggestion
This manifest is primarily used by [`meltano discover`](/reference/command-line-interface#discover) and [`meltano add`](/reference/command-line-interface#add). It is also used in cases where the full plugin definition is needed but no lock artifact or cached `discovery.yml` is found.
```

I like adding the above sentence. What do you think @DouweM?

DouweM: @tayloramurphy Works for me!

docs/src/_reference/settings.md
aaronsteers: What we ended up happening on `tap-gitlab` was that we ended up having to accept a bare domain reference _and_ a versioned endpoint: https://github.com/MeltanoLabs/tap-gitlab/blob/main/tap_gitlab/client.py#L38-L43

I would prefer to just take the versioned endpoint - but alternatively, if there are no URL path parts, we could try to smartly add `/meltano/api/v1` to that endpoint, but that seems very fragile to me.

If spinning up and endpoint on localhost (and especially if a 3rd-party is developing a custom endpoint), they definitely would not want to be constrained to keep our same version number, and they definitely would not want to be constrained to have a path ending in`/meltano/api/v1`.

These should all be compatible, I think, with this setting:

```
http://localhost:300/mytestapi
https://stitch.com/api-v1/meltanohub
https://api.meltano.com/hub/v2
https://api.meltano.com/v3
```

I think it's the responsibility of the software (and of the user if they are overriding it) to make sure a compatible endpoint is provided. I don't think the version and the path should be hard-coded and not able to be overriden.

DouweM: @aaronsteers I'm not sure I fully agree, but I don't feel strongly enough about this and trust your judgment, so go with whatever you think makes most sense!

DouweM (approved):

docs/src/_reference/settings.md
edgarrmondragon: > If you're overriding the version, wouldn't you make that change directly in `client.py` in your feature branch, since you'd presumably also be updating the code there to account for the API changes?

💯

edgarrmondragon: @aaronsteers If you have a strong opinion, do you want me to update the setting to include the API version?

aaronsteers: @edgarrmondragon - Yes, if you don't mind, I do feel strongly we should use the versioned endpoint. We took the other approach on tap-gitlab and ended up having to support _both_ versions. I'd like to avoid that here if we can and just take exactly the root of the API endpoint before additional `path` items are added.

🙏

aaronsteers: @edgarrmondragon - As it is named now (`MELTANO_HUB_URL`), I think I could be okay either way. 

I think I'd prefer to rename to `DISCOVERY_API_ROOT`,`MELTANO_DISCOVERY_API_ROOT` or similar, and use the full path including version, but we could later add an `API_ROOT` that overrides the other if/when this becomes an issue.

On tap-gitlab and ended up having to support _both_ conventions - the site base URL and _also_ the versioned path, which I think we can  avoid here if we can and just take exactly the root of the API endpoint before additional `path` items are added.

Thanks, @edgarrmondragon ! 🙏

PR #6020, opened 2022-06-04
Telemetry state change event

### LINKED ISSUE #6032: Add telemetry state change event
Adding this stub in reference to internal issue and discussion here: https://github.com/meltano/internal-general/issues/338

This will give better insight into users stopping using meltano (or changing their project IDs) vs choosing to opt out of anonymous usage stats.

aaronsteers:
Resolves https://github.com/meltano/internal-general/issues/338

This adds a one-time 'telemetry_state_change' event when the value of either project_id or send_anonymous_usage_stats changes.

~~Note: this borrows from #5941 and targets that branch.~~

~~Plan is that #5941 will merge first and then this will be updated to target `main`.~~ Done.

TODO:

- [x] Feature development
- [x] Define Snowplow schema for the new event type
- [ ] Add tests
- [x] Manual sanity check
- [x] Register the new event type with Snowcatcloud


src/meltano/core/tracking/tracker.py
```diff
@@ -44,6 +51,17 @@ def check_url(url: str) -> bool:
     return bool(re.match(URL_REGEX, url))
 
 
+class AnalyticsSettings(NamedTuple):
```
edgarrmondragon, line 54: Wdyt about calling this class `Telemetry settings`?

src/meltano/core/tracking/tracker.py
```diff
@@ -194,3 +263,103 @@ def track_command_event(self, event_json: dict[str, Any]) -> None:
...
+            )
+        )
+
+    @property
+    def _analytics_json_path(self) -> Path:
+        """Return path to the 'analytics.json' file."""
+        return self.project.meltano_dir() / "analytics.json"
+
+    @property
+    def _analytics_json_settings(self) -> AnalyticsSettings:
```
edgarrmondragon, line 310: IMO it's a bit weird to have a "private" computed property. What do you think of making this a proper method since it's only called once anyway?

@WillDaSilva or @pandemicsyn might have opinions too 🙂

src/meltano/core/tracking/tracker.py
```diff
@@ -3,10 +3,13 @@
...
 
 import datetime
+import json
 import locale
 import re
+import uuid
 from contextlib import contextmanager
-from typing import Any
+from pathlib import Path
+from typing import Any, NamedTuple, Optional, Tuple, Union
```
pandemicsyn, line 12: @aaronsteers  that from __future__ import annotations allows new style annotations, and we've been slowly moving the code base to those. That allows referring list , tuple, and friends without having to import anything from the typing package.

If you're running the pre-commit hook pyupgrade will catch a lot of these for you, but figured I'd give a ya heads up.

src/meltano/core/tracking/tracker.py
```diff
@@ -44,6 +51,17 @@ def check_url(url: str) -> bool:
     return bool(re.match(URL_REGEX, url))
 
 
+class AnalyticsSettings(NamedTuple):
+    """Settings which control telemetry and anonymous usage stats.
+
+    These are stored within `analytics.json`.
+    """
+
+    client_id: Optional[uuid.UUID]
```
pandemicsyn, line 60: for ref, these can convert to:

```
client_id: uuid.UUID | None
```

src/meltano/core/tracking/tracker.py
pandemicsyn: +1 I'd prefer this one be a method, IMHO this does a lot for being a "property" .

src/meltano/core/tracking/tracker.py
```diff
@@ -194,3 +263,103 @@ def track_command_event(self, event_json: dict[str, Any]) -> None:
...
+    def _analytics_json_settings(self) -> AnalyticsSettings:
+        """Get settings from the 'analytics.json' file."""
+
+        def _uuid_from_str(from_val: Optional[Any], warn: bool):
+            if not isinstance(from_val, str):
+                if from_val is None:
+                    return None
+
+                logger.warn(
+                    f"The value '{from_val}' in 'analytics.json' was of type "
```
pandemicsyn, line 319: can we use a structured log entry here. Something like:

```logger.warn("Unexpected analytics.json setting type", from_val=from_val, current_type=type(from_val))```

File: src/meltano/core/tracking/tracker.py [hunk 6]
```diff
@@ -194,3 +263,103 @@ def track_command_event(self, event_json: dict[str, Any]) -> None:
...
+
+            try:
+                return uuid.UUID(from_val, version=4)
+            except ValueError:
+                # Should only be reached if user manually edits 'analytics.json'.
+                log_fn = logger.debug
+                if warn:
+                    log_fn = logger.warning
+
+                log_fn(
```
pandemicsyn, line 332: Ditto the structlog entry here would be nice.

edgarrmondragon:
The commit history in this one has things it shouldn't, right? Does it need to be rebased?

src/meltano/core/tracking/tracker.py
aaronsteers: Sure! Initially it was named after the analytics.json file, but TelemetrySettings seems more appropriate, I think. 👍

src/meltano/core/tracking/tracker.py
aaronsteers: Agreed - this one grew and morphed quite a lot. I tidied it and moved the helper function out to its own private method.

src/meltano/core/tracking/tracker.py
aaronsteers: Updated per your suggestion. Thanks!

I considered putting them in the new style before, but my IDE linter is complaining about the new types. Going forward, I may need to find a fix for the mypy lint issue, perhaps using a higher Python version for my development, since it seems mypy prefers 3.9 or 3.10.

pandemicsyn:
@WillDaSilva heads up on https://github.com/meltano/meltano/pull/6028 which should hopefully land this morning, it'll probably give you some merge conflicts.

WillDaSilva:
@aaronsteers @pandemicsyn @pnadolny13 The new event type `telemetry_state_change_event` has been added to Snowcatcloud. Pending your approval, I'll put it into production mode.

pandemicsyn:
@WillDaSilva fyi - https://github.com/meltano/meltano/pull/6028 landed so probably have a little merge/commit clean up.

WillDaSilva:
@aaronsteers I am unable to add you as a reviewer.

aaronsteers:
@WillDaSilva - Because I authored it I guess I cannot be an approver. I'll review now and add any comments I see.

src/meltano/core/tracking/tracker.py
```diff
@@ -100,17 +118,80 @@ def __init__(
...
+        Returns:
+            True if anonymous usage stats are enabled.
+        """
+        if self.settings_service.get("send_anonymous_usage_stats", None) is not None:
+            return self.settings_service.get("send_anonymous_usage_stats")
+
+        if self.settings_service.get("tracking_disabled", None) is not None:
+            return not self.settings_service.get("tracking_disabled")
+
+        return True
```
aaronsteers, line 150: Yes, this is the bit that resolves the removal of the `env_alias` `MELTANO_TRACKING_DISABLED`.

aaronsteers (commented):
Looks good! ✅

pandemicsyn (approved):
lgtm! Also double checked,  I think i see the correct # of events at the correct times, in snowplow micro

PR #6024, opened 2022-06-05
remove sql module

aaronsteers:
Permanently removes the `sql` module

Follows from: https://github.com/meltano/meltano/issues/3218


aaronsteers:
Gitlab job is only complaining in the linter, because the diff is not able to find the matched files now that they are deleted.

aaronsteers:
Merging absent a review - no code changed (100% file deletions) and tests pass cleanly (minus 'lint diff' bug).

PR #6025, opened 2022-06-05
auto-store should always write a value with `meltano config set`

### LINKED ISSUE #6015: `meltano config meltano set` doesn't work as expected
```
~/Development/meltano-projects/superset-test ❯❯❯ meltano config meltano set discovery_url http://localhost:4000/discovery.yml
2022-06-04T01:36:33.566497Z [info     ] Environment 'dev' is active
Meltano setting 'discovery_url' was set in `meltano_environment`: 'http://localhost:4000/discovery.yml'
Current value is still: 'https://discovery.meltano.com/discovery.yml' (from the default)
```

I'd expect this to store `discovery_url: http://localhost:4000` in `meltano.yml`, but it's nowhere to be found.

aaronsteers:
Rather than check if the setting may already be true at any given level, we will just write the value when requested by the user.

### DIFF
--- a/src/meltano/core/settings_store.py
+++ b/src/meltano/core/settings_store.py
@@ -182,7 +182,7 @@ class SettingsStoreManager(ABC):
         path: list[str],
         value: Any,
         setting_def: SettingDefinition | None = None,
-    ) -> None:
+    ) -> dict:
         """Unimplemented set method.
 
         Args:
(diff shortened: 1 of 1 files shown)


aaronsteers:
@DouweM - Since you probably know this best and @kgpayne since you've had lots of recent exploration here, any strong reasons for us to not move forward with this change?

I still have tests to validate and perhaps update - but want to check in before I potentially change expected behaviors - in case there are any negative side effects you might foresee.

DouweM:
@aaronsteers [Chesterton's fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence) has me worried about removing code that was presumably added for a good reason.

In https://github.com/meltano/meltano/issues/6015, I think the more urgent issue is `meltano config meltano set discovery_url http://localhost:4000/discovery.yml` not working at all (first comment), not `meltano config meltano set send_anonymous_usage_stats false` being confused when the value is already `false` in the environment (second comment), which has presumably been a bug for as long as this defaults code has been here. So while we should carefully reason through the implications of this change before we merge it, I don't think it's the best use of our time to try to do so ahead of 2.0.

aaronsteers:
> In #6015, I think the more urgent issue is `meltano config meltano set discovery_url http://localhost:4000/discovery.yml` not working at all (first comment)

I'll test with your provide test case to make sure this resolves your issue. My hunch is this is all related to the same core issue(s) but of course we'll find this out as we test. 

@DouweM - So that I can test, can you provide any other context on how to reproduce this issue? I have your logs, but I don't know exactly from that context what specific env vars or other settings might have been set. Any repro instructions (and if you've tested with other settings besides that one) would be super helpful.

DouweM:
@aaronsteers Posted steps to reproduce in https://github.com/meltano/meltano/issues/6015#issuecomment-1146861685.

DouweM:
For what it's worth, this wasn't related to the fix for https://github.com/meltano/meltano/issues/6015, so we should consider the change on its merits and not as prompted by a bug elsewhere.

PR #6028, opened 2022-06-05
PluginsTrackingContext class and meltano run tracking

### LINKED ISSUE #3366: Add Snowplow events at the block level

pandemicsyn:
Given the time constraints, I deviated from the original idea of #3366 for this implementation a bit. I think an actual block level implementation (and by block level i mean `meltano.core.blocks`) is still the right long term fit, as it would allow for finer details around why a block failed. But a hurried version of that would have been more likely to miss block level failures.

This version generates the block level events from the cli's `_run_blocks` method instead, and so can easily start, stop, failure events for each block in the invocation chain.


src/meltano/core/tracking/plugins.py
```diff
@@ -0,0 +1,57 @@
...
+PLUGINS_CONTEXT_SCHEMA = "iglu:com.meltano/plugins_context/jsonschema"
+PLUGINS_CONTEXT_SCHEMA_VERSION = "1-0-0"
+
+
+def _from_plugin(plugin: ProjectPlugin, cmd: str) -> dict:
+    return {
+        "category": str(plugin.type),
+        "name_hash": hash_sha256(plugin["info"].get("name")),
+        # TODO: what if its "original" should be we hashing namespace and sending that as well?
+        "variant_name_hash": hash_sha256(plugin["info"].get("variant")),
```
pandemicsyn, line 23: @aaronsteers in a lot of cases the variant is set to "original" which I assume is ok. Curious though if we should snag a hashed version of plugin["info"]["namespace"] as well ?

src/meltano/core/tracking/plugins.py
```diff
@@ -0,0 +1,57 @@
...
+        "pip_url_hash": hash_sha256(plugin.formatted_pip_url),
+        # TODO: this is not the right field, but parent "name" isn't available in the plugin context
+        "parent_name_hash": hash_sha256(plugin.parent.executable),
```
pandemicsyn, line 26: @aaronsteers Should I be grabbing `plugin.parent.definition["name"]` for this ?


Screen Shot 2022-06-05 at 1 19 58 PM

src/meltano/cli/invoke.py
```diff
@@ -112,7 +112,10 @@ def invoke(
             print_var=print_var,
         )
     )
-    with tracker.with_contexts(cmd_ctx):
+
+    with tracker.with_contexts(
+        cmd_ctx, PluginsTrackingContext([(plugin, command_name)])
```
pandemicsyn, line 117: @aaronsteers quick and dirty example of how to attach a plugins context to a generic command event.

src/meltano/core/tracking/plugins.py
pandemicsyn: (fyi in more recent commit i do just that)

src/meltano/cli/run.py
```diff
@@ -112,9 +112,13 @@ async def run(
...
         else:
-            tracker.track_command_event(cli_tracking.FAILED)
+            tracker.track_command_event(cli_tracking.ABORTED)
             raise CliError("Some ExtractLoadBlocks set failed validation.")
-        await _run_blocks(tracker, parsed_blocks, dry_run=dry_run)
+        try:
+            await _run_blocks(tracker, parsed_blocks, dry_run=dry_run)
+        except Exception as err:
+            tracker.track_command_event(cli_tracking.FAILED)
+            raise err
```
pandemicsyn, line 121: @pnadolny13 in the vast majority of cases, you should now always see a tracking event signaling a CLI failure for `meltano run`. If a pre-flight check of the ExtractLoadBlock fail's I send an `aborted` instead. 

Would take an uncaught exception during validation or much earlier issue for that event to go missing.

src/meltano/core/tracking/plugins.py
```diff
@@ -49,6 +49,9 @@ def _from_plugin(plugin: ProjectPlugin, cmd: str) -> dict:
         "category": str(plugin.type),
         "name_hash": hash_sha256(plugin.name) if plugin.name else None,
         "namespace_hash": hash_sha256(plugin.namespace) if plugin.namespace else None,
+        "executable_hash": hash_sha256(plugin.executable)
+        if plugin.executable
+        else None,
```
pandemicsyn, line 54: Added plugin.executable. /cc @aaronsteers @pnadolny13

src/meltano/cli/invoke.py
```diff
@@ -90,29 +100,43 @@ def invoke(
...
-        plugin_name, plugin_type=plugin_type, invokable=True
-    )
+
+    try:
+        plugin = plugins_service.find_plugin(
+            plugin_name, plugin_type=plugin_type, invokable=True
+        )
+    except PluginNotFoundError as err:
+        with tracker.with_contexts(cmd_ctx):
+            tracker.track_command_event(cli_tracking.ABORTED)
```
pandemicsyn, line 110: @pnadolny13 similar to `meltano run` if you pass a bad plugin to invoke I'll fire a `aborted` event, reserving `failed` for actual failures stemming from errors.

src/meltano/cli/invoke.py
```diff
@@ -90,29 +100,43 @@ def invoke(
...
+                command_name,
+                containers,
+                print_var=print_var,
+            )
         )
-    )
-    with tracker.with_contexts(cmd_ctx):
+    except Exception as invoke_err:
+        with tracker.with_contexts(cmd_ctx):
+            tracker.track_command_event(cli_tracking.FAILED)
```
pandemicsyn, line 134: @pnadolny13 and I think this should catch most actual failures that happen while invoking a plugin.

src/meltano/core/utils/__init__.py
```diff
@@ -432,5 +432,25 @@ def hash_sha256(value: str) -> str:
 
     Returns:
         The hashed value of the given string.
+
+    Raises:
+        ValueError: If we are blindly passed a value that is None.
     """
+    if value is None:
+        raise ValueError("Cannot hash None.")
```
pandemicsyn, line 440: added this because its a little more of an explicit error if someone passes None here.

pandemicsyn:
@aaronsteers @pnadolny13 ok this should cover invoke and run. If this looks ok can we probably tactically add this to a few other commands tomorrow, long term I think https://github.com/meltano/meltano/issues/6041 is still the move (and would help DRY things up).

src/meltano/cli/run.py
```diff
@@ -85,58 +88,90 @@ async def run(
...
+                    block_type=blk_name,
                     comprised_of=f"{blk.string_id}:{blk.command}",
                 )
             continue
 
         try:
             await blk.run()
-        except RunnerError as err:
+        except Exception as err:
             logger.error(
```
pandemicsyn, line 155: @pnadolny13 this should catch any exceptions that come through from the lower level blocks. Previously we just captured RunnerError's only and let others bubble up.

src/meltano/core/tracking/plugins.py
```diff
@@ -0,0 +1,92 @@
...
+        plugins: list[(ProjectPlugin, str)] = []
+        for plugin_block in blk.blocks:
+            plugins.append((plugin_block.context.plugin, plugin_block.plugin_args))
+        return PluginsTrackingContext(plugins)
+    if isinstance(blk, PluginCommandBlock):
+        return PluginsTrackingContext([(blk.context.plugin, blk.command)])
+
+
+def _from_plugin(plugin: ProjectPlugin, cmd: str) -> dict:
+    if not safe_hasattr(plugin, "info"):
```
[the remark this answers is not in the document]
aaronsteers: @pandemicsyn - I appreciate that this is perhaps not ideal but I think your solution works well and solves the problem of consistent behavior with concise+readable code.

src/meltano/core/tracking/plugins.py
aaronsteers: Great, thanks!

src/meltano/core/utils/__init__.py
aaronsteers: Very nice. Thanks. 👍

aaronsteers (approved):
Nice work!

pandemicsyn:
@aaronsteers @pnadolny13 fyi - updated snowcat cloud with the latest schema's and asked Joao to delete the 3 registry entries we didn't end up using: https://meltano.slack.com/archives/C033Z4UNJ7R/p1654632323884469

The schema's are still not marked as "Production", @pnadolny13 do you want me to flip that that switch before or after this merges ?

aaronsteers (approved):

src/meltano/core/tracking/iglu-client-embedded/schemas/com.meltano/plugins_context/jsonschema/1-0-0
```diff
@@ -26,40 +26,46 @@
...
                         "description": "64 character hash of the plugin name",
                         "type": "string",
                         "maxLength": 64
                     },
-                    "parent_name_hash": {
-                        "description": "64 character hash of the parent plugin name",
+                    "namespace_hash": {
+                        "description": "64 character hash of the plugin namespace",
+                        "type": "string",
+                        "maxLength": 64
```
WillDaSilva, line 39:
```suggestion
                        "minLength": 64,
                        "maxLength": 64
```

src/meltano/core/tracking/iglu-client-embedded/schemas/com.meltano/plugins_context/jsonschema/1-0-0
```diff
@@ -26,40 +26,46 @@
...
+                    },
+                    "executable_hash": {
+                        "description": "64 character hash of the plugin executable",
+                        "type": "string",
+                        "maxLength": 64
```
WillDaSilva, line 44:
```suggestion
                        "minLength": 64,
                        "maxLength": 64
```

src/meltano/core/tracking/iglu-client-embedded/schemas/com.meltano/plugins_context/jsonschema/1-0-0
```diff
@@ -26,40 +26,46 @@
...
+                    },
+                    "variant_name_hash": {
+                        "description": "64 character hash of the plugin variant name",
                         "type": [
                             "string",
                             "null"
                         ],
                         "maxLength": 64
```
WillDaSilva, line 52:
```suggestion
                        "minLength": 64,
                        "maxLength": 64
```

WillDaSilva (approved):

src/meltano/core/tracking/iglu-client-embedded/schemas/com.meltano/plugins_context/jsonschema/1-0-0
pandemicsyn: @WillDaSilva we'll have to update these in a follow up, since it will require release a new version of the schema since its already live and in "production" on Snowcat cloud.

src/meltano/core/tracking/tracker.py
```diff
@@ -24,6 +24,18 @@
...
+BLOCK_EVENT_SCHEMA_VERSION = "1-0-0"
+
+
+class BlockEvents:
+    """Events describing a block state."""
+
+    INITIALIZED: str = "initialized"
+    STARTED: str = "started"
+    COMPLETED: str = "completed"
+    FAILED: str = "failed"
```
WillDaSilva, line 37:
```suggestion
class BlockEvents(Enum):
    """Events describing a block state."""

    initialized = auto()
    started = auto()
    completed = auto()
    failed = auto()
```

The string versions of each enum member can be retrieved through their `name` attribute.

src/meltano/core/utils/__init__.py
```diff
@@ -451,5 +451,25 @@ def hash_sha256(value: str) -> str:
...
+    This is a hacky workaround for the fact that `hasattr` is not allowed by WPS.
+
+    Args:
+        obj: The object to check.
+        name: The name of the attribute to check.
+
+    Returns:
+        True if the object has the attribute, False otherwise.
+    """
+    return getattr(obj, name, None) is not None
```
WillDaSilva, line 475: This doesn't work when the attribute being checked has the value `None`. Can we just do what `hasattr` does under the hood instead?

```suggestion
    try:
        getattr(obj, name)
    except AttributeError:
        return False
    return True
```

src/meltano/core/tracking/tracker.py
```diff
@@ -194,3 +207,17 @@ def track_command_event(self, event_json: dict[str, Any]) -> None:
...
+        """Fire generic block tracking event.
+
+        Args:
+            block_type: The block type.
+            event: The event string (e.g. "initialize", "started", etc)
+        """
+        self.track_unstruct_event(
+            SelfDescribingJson(
+                f"{BLOCK_EVENT_SCHEMA}/{BLOCK_EVENT_SCHEMA_VERSION}",
+                {"type": block_type, "event": event},
```
WillDaSilva, line 221:
```suggestion
    def track_block_event(self, block_type: str, event: BlockEvents) -> None:
        """Fire generic block tracking event.

        Args:
            block_type: The block type.
            event: The event string (e.g. "initialize", "started", etc)
        """
        self.track_unstruct_event(
            SelfDescribingJson(
                f"{BLOCK_EVENT_SCHEMA}/{BLOCK_EVENT_SCHEMA_VERSION}",
                {"type": block_type, "event": event.name},
```

PR #6031, opened 2022-06-05
Automatically generate Superset config file

DouweM:
Part of https://github.com/meltano/hub/issues/250

### DIFF
--- a/src/meltano/core/plugin/superset.py
+++ b/src/meltano/core/plugin/superset.py
@@ -1,0 +1,101 @@
+import logging
+import subprocess
+from typing import List
+
+from meltano.core.behavior.hookable import hook
+from meltano.core.error import AsyncSubprocessError
+from meltano.core.plugin_invoker import PluginInvoker
+
+from . import BasePlugin, PluginType
+
+
+class SupersetInvoker(PluginInvoker):
+    def env(self):
+        env = super().env()
+
+        env["SUPERSET_HOME"] = str(self.plugin_config_service.run_dir)
+        env["SUPERSET_CONFIG_PATH"] = str(self.files["config"])
+        env["FLASK_APP"] = "superset"
+
+        return env
+
+
+class Superset(BasePlugin):
+    __plugin_type__ = PluginType.UTILITIES
+
+    invoker_class = SupersetInvoker
+
+    @property
+    def config_files(self):
+        return {"config": "superset_config.py"}
+
+    def process_config(self, flat_config):
+        # TODO: This means that `meltano config superset` doesn't show settings that are handled by Meltano itself, e.g. `ui.port` in the `ui` command
+        return {key: value for key, value in flat_config.items() if key.isupper()}
+
+    @hook("before_configure")
+    async def before_configure(self, invoker: SupersetInvoker, session):  # noqa: WPS217
+        """Write plugin configuration to superset_config.py."""
+        config_path = invoker.files["config"]
+        with open(config_path, "w") as config_file:
+            config = invoker.plugin_config_processed
+            config_file.write(
+                "\n".join(
+                    [
+                        "import sys",
+                        "module = sys.modules[__name__]",
+                        f"config = {str(config)}",
+                        "for key, value in config.items():",
+                        "    setattr(module, key, value)",
+                    ]
+                )
+            )
+        logging.debug(f"Created configuration at {config_file}")
+
+    @hook("before_invoke")
+    async def db_upgrade_hook(self, invoker: PluginInvoker, exec_args: List[str]):
+        """Create or upgrade metadata database."""
+        handle = await invoker.invoke_async(
+            "db",
+            "upgrade",
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        exit_code = await handle.wait()
+
+        if exit_code:
+            raise AsyncSubprocessError(
+                "Superset metadata database could not be initialized: `superset db upgrade` failed",
+                handle,
+            )
+
+        logging.debug("Completed `superset db upgrade`")
+
+    @hook("before_invoke")
+    async def init_hook(self, invoker: PluginInvoker, exec_args: List[str]):
+        """Create default roles and permissions."""
+
+        handle = await invoker.invoke_async(
+            "init",
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        exit_code = await handle.wait()
+
+        if exit_code:
+            raise AsyncSubprocessError(
+                "Superset default roles and permissions could not be created: `superset init` failed",
+                handle,
+            )
+
+        logging.debug("Completed `superset init`")
+
+    @hook("before_cleanup")
+    async def before_cleanup(self, invoker: PluginInvoker):
+        """Delete the config file."""
+        config_file = invoker.files["config"]
+        try:
+            config_file.unlink()
+            logging.debug(f"Deleted configuration at {config_file}")
+        except FileNotFoundError:
+            pass
--- a/src/meltano/core/plugin/superset.py
+++ b/src/meltano/core/plugin/superset.py
@@ -2,11 +2,17 @@ import logging
 import subprocess
 from typing import List
 
+import structlog
+
 from meltano.core.behavior.hookable import hook
 from meltano.core.error import AsyncSubprocessError
+from meltano.core.plugin.error import PluginExecutionError
 from meltano.core.plugin_invoker import PluginInvoker
+from meltano.core.setting_definition import SettingDefinition
 
 from . import BasePlugin, PluginType
+
+logger = structlog.getLogger(__name__)
 
 
 class SupersetInvoker(PluginInvoker):
@@ -25,31 +31,51 @@ class Superset(BasePlugin):
 
     invoker_class = SupersetInvoker
 
+    EXTRA_SETTINGS = [SettingDefinition(name="_config_path")]
+
     @property
     def config_files(self):
         return {"config": "superset_config.py"}
 
-    def process_config(self, flat_config):
-        # TODO: This means that `meltano config superset` doesn't show settings that are handled by Meltano itself, e.g. `ui.port` in the `ui` command
-        return {key: value for key, value in flat_config.items() if key.isupper()}
-
     @hook("before_configure")
     async def before_configure(self, invoker: SupersetInvoker, session):  # noqa: WPS217
         """Write plugin configuration to superset_config.py."""
-        config_path = invoker.files["config"]
-        with open(config_path, "w") as config_file:
-            config = invoker.plugin_config_processed
-            config_file.write(
-                "\n".join(
+
+        config = invoker.plugin_config_processed
+
+        config_script_lines = [
+            "import sys",
+            "module = sys.modules[__name__]",
+            f"config = {str(config)}",
+            "for key, value in config.items():",
+            "    if key.isupper():",
+            "        setattr(module, key, value)",
+        ]
+
+        custom_config_filename = invoker.plugin_config_extras["_config_path"]
+        if custom_config_filename:
+            custom_config_path = invoker.project.root.joinpath(custom_config_filename)
+
+            if custom_config_path.exists():
+                config_script_lines.extend(
                     [
-                        "import sys",
-                        "module = sys.modules[__name__]",
-                        f"config = {str(config)}",
-                        "for key, value in config.items():",
-                        "    setattr(module, key, value)",
+                        "import imp",
+                        f'custom_config = imp.load_source("superset_config", "{str(custom_config_path)}")',
+                        "for key in dir(custom_config):",
+                        "    if key.isupper():",
+                        "        setattr(module, key, getattr(custom_config, key))",
                     ]
                 )
-            )
+
+                logger.info(f"Merged in config from {custom_config_path}")
+            else:
+                raise PluginExecutionError(
+                    f"Could not find config file {custom_config_path}"
+                )
+
+        config_path = invoker.files["config"]
+        with open(config_path, "w") as config_file:
+            config_file.write("\n".join(config_script_lines))
         logging.debug(f"Created configuration at {config_file}")
 
     @hook("before_invoke")
--- a/src/meltano/core/plugin/superset.py
+++ b/src/meltano/core/plugin/superset.py
@@ -1,3 +1,4 @@
+"""Plugin glue code for Superset."""
 import logging
 import subprocess
 from typing import List
@@ -16,7 +17,14 @@ logger = structlog.getLogger(__name__)
 
 
 class SupersetInvoker(PluginInvoker):
+    """Invoker that prepares env for Superset."""
+
     def env(self):
+        """Environment variables for Superset.
+
+        Returns:
+            Dictionary of environment variables.
+        """
         env = super().env()
 
         env["SUPERSET_HOME"] = str(self.plugin_config_service.run_dir)
@@ -27,6 +35,8 @@ class SupersetInvoker(PluginInvoker):
 
 
 class Superset(BasePlugin):
+    """Plugin glue code for Superset."""
+
     __plugin_type__ = PluginType.UTILITIES
 
     invoker_class = SupersetInvoker
@@ -35,12 +45,24 @@ class Superset(BasePlugin):
 
     @property
     def config_files(self):
+        """Return the configuration files required by the plugin.
+
+        Returns:
+            Dictionary of config file identifiers and filenames
+        """
         return {"config": "superset_config.py"}
 
     @hook("before_configure")
     async def before_configure(self, invoker: SupersetInvoker, session):  # noqa: WPS217
-        """Write plugin configuration to superset_config.py."""
+        """Write plugin configuration to superset_config.py.
 
+        Args:
+            invoker: the active PluginInvoker
+            session: metadata database session
+
+        Raises:
+            PluginExecutionError: if config file couldn't be found
+        """
         config = invoker.plugin_config_processed
 
         config_script_lines = [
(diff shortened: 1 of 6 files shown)


edgarrmondragon (approved):

aaronsteers (dismissed):
Wanted to voice a request to hold on merging this while I review.

@DouweM - For my info (and with the caveat that I still need to review the code changes), what are your thoughts on feasibility of implementing these via a file bundle instead of extending the core meltano codebase?

src/meltano/core/plugin/superset.py
```diff
@@ -0,0 +1,101 @@
...
+        with open(config_path, "w") as config_file:
+            config = invoker.plugin_config_processed
+            config_file.write(
+                "\n".join(
+                    [
+                        "import sys",
+                        "module = sys.modules[__name__]",
+                        f"config = {str(config)}",
+                        "for key, value in config.items():",
+                        "    setattr(module, key, value)",
```
aaronsteers, line 49: This being a local temporary file with cleartext secrets creates several problems, as we've run into with Airflow. 

Why not have a files bundle that creates a fixed `config.py`:

```
import sys
module = sys.modules[__name__]

ENV_VAR_PREFIX = f"SUPERSET_"

meltano_config = {k.rstrip(ENV_VAR_PREFIX: v for os.environ.items() if k.startswith(ENV_VAR_PREFIX)}
for key, value in meltano_config.items():,
    setattr(module, key, value)
```

DouweM:
@aaronsteers That was actually the original approach (see https://github.com/meltano/hub/issues/250#issuecomment-1146874661), but that didn't allow Superset to be configured through Meltano's own config and environments features, and would leave a lot of boilerplate code inside the user's repo that they shouldn't be touching or even knowing about.

src/meltano/core/plugin/superset.py
DouweM: @aaronsteers Interesting to hear you leaning towards the file bundle approach. I'm not actually aware of the Airflow problems you're talking about, and was modeling this approach after that one.

aaronsteers: @DouweM - I would prefer this be clear, explicit, and non-magic. If the file is in the repo, it can easily be packaged or volume-mapped into a docker container, and the user can optionally customize and control the behavior as needed.

The temporary file that deletes itself afterwards prevents alternative deployment options.

I'll try to get on a zoom to discuss the vision here.

DouweM: We decided that in this first iteration this approach that completely abstracts away configuration is acceptable, but it will not be sufficient for long, and we have to discuss more how prescriptive we want to be about how plugin configuration should be managed and how/when to leave the door open for users to do it the non-Meltano way.

aaronsteers: @DouweM - I'm okay with running the db upgrade as an action but I'd still prefer to move the `config.py` file back into the files bundle. Otherwise, we'll have a breaking change for all superset users when we do so later and this will be very difficult to unwind.

And as discussed, users who go to the docs expecting to have a config python file to modify will otherwise not have one.

aaronsteers:
@DouweM - If we open up `config.py`, the user could literally just copy-paste the oauth config from here and update their pointers. Without a `config.py` file to work with in the project, this become _a lot_ more work.

https://superset.apache.org/docs/installation/configuring-superset/#custom-oauth2-configuration

```
OAUTH_PROVIDERS = [
  {   'name':'egaSSO',
      'token_key':'access_token', # Name of the token in the response of access_token_url
      'icon':'fa-address-card',   # Icon for the provider
      'remote_app': {
          'client_id':'myClientId',  # Client Id (Identify Superset application)
          'client_secret':'MySecret', # Secret for this Client Id (Identify Superset application)
          'server_metadata_url': 'https://myAuthorizationServer/.well-known/openid-configuration'
      }
  }
]
```

And some customization options actually ask for subclassing:

import logging
from superset.security import SupersetSecurityManager

class CustomSsoSecurityManager(SupersetSecurityManager):

    def oauth_user_info(self, provider, response=None):
        logging.debug("Oauth2 provider: {0}.".format(provider))
        if provider == 'egaSSO':
            # As example, this line request a GET to base_url + '/' + userDetails with Bearer  Authentication,
    # and expects that authorization server checks the token, and response with user details
            me = self.appbuilder.sm.oauth_remotes[provider].get('userDetails').data
            logging.debug("user_data: {0}".format(me))
            return { 'name' : me['name'], 'email' : me['email'], 'id' : me['user_name'], 'username' : me['user_name'], 'first_name':'', 'last_name':''}
    ...

DouweM:
@aaronsteers Thanks, that's a great example of why users will need some way to provide that file directly. Thanks for looking that up.

That means we have two options:
1. Put `superset_config.py` in the project through a file bundle, with code injecting the Meltano-managed config at the top, so the user can add their own stuff below
2. Add a `config_path` setting to allow a user to point at their own `superset_config.py` only if they need it, which Meltano would inject into the auto-generated config file after the Meltano-managed config is injected

In both cases, custom config would override the Meltano-managed.

You prefer option 1 as (among other things) "users who go to the docs expecting to have a config python file to modify will otherwise not have one" which argues for having the file the tool's docs reference in the repo.

After reflecting on our earlier conversation for a bit, my primary reason for preferring option 2 is consistency with other plugins, specifically Singer and Airflow. As their config files (`config.json`, `airflow.cfg`) are not interpreted and are basically just dumped dicts, a file-bundle-provided config file in the project repo would not be able to inject any Meltano-managed config on the fly, and option 1 does not exist. The "users will expect to find the config file that the tool's docs describe" argument also exists here, and would point us to a third solution which is to only support config in the file-bundle-provided config file, and not support Meltano-managed config with these plugins at all. Here we've decided we'd rather tell users that "whenever the tool's docs say to put something in [config file], actually put it in `meltano.yml` or use `meltano config` instead", as we do we want users to be able to benefit from Meltano-managed config, especially in the case of Singer. Option 2 _is_ available here, and we could introduce it without a breaking change: we can add a new `config_path` property to Singer and Airflow and merge the Meltano-managed and user-provided configs together when generating the final config file that will be passed to the plugin.

I think the "let users use their existing config instead of forcing to rewrite it the Meltano way" argument is as strong with Singer and Airflow as it is with Superset, so I think that `config_path` property would be a great idea. And if we go with that option there (because it's the only acceptable option), I don't think it makes sense to go with a different option with Superset. I appreciate the argument that it's better to be explicit than implicit and let the user know exactly how their config is being generated, which option 1 would provide for Superset, but as that's simply impossible with the other plugins, I think it's better for the user if they get consistent behavior across plugins with a `config_path` property.

Even with that property, we have the option of leaving it unset by default, or providing an empty config file through a file bundle and pointing at that by default. I could be convinced that that makes sense in the case of Superset, since it has more complex settings than Singer and Airflow, but I don't think we'd be helping users by making that the default behavior for `extract/tap-foo/config.json` and `orchestrate/airflow.cfg` as well, as the Meltano-managed approach is clearly superior there and it's worth pushing people to use it instead of the static files. So here consistency (and the assumption that most tools will not have interpreted config files) point us to requiring `config_path: my/config.ext` to be set explicitly in `meltano.yml` in all cases.

So my suggestion is to add a `config_path` property to Superset now, and to Airflow and Singer and GE and everything else later, that Meltano will automatically merge into its own managed config if it's set. And documenting clearly what settings `meltano config` can be used for, and which it can't.

Dbt is a already somewhat inconsistent with Singer and Airflow in that the user has a full dbt project with `dbt_project.yml` living in the `transform` folder, and you can't set `dbt_project.yml` properties through Meltano-managed config at all. But I think `dbt_project.yml` and the other files in `transform` are more similar to `orchestrate/dags/meltano.py` (which is also checked into the project), in that it's a fundamental asset to the tool that users would change relatively frequently, and not "just" key-value configuration. So I think the reasons why we did _not_ put `airflow.cfg` and `tap.config.json` in the repo are more relevant to `superset_config.py` than the reason why we _did_ put `dbt_projecy.yml` in the repo.

aaronsteers:
@DouweM re:

> After reflecting on our earlier conversation for a bit, my primary reason for preferring option 2 is consistency with other plugins, specifically Singer and Airflow. 

This should be an argument in the opposite direction, since precedent actually goes the other direction. 

As I already mentioned, the Airflow behavior needs to be changed because it's already creating problems. The dbt implementation doesn't and shouldn't take this approach and neither do great expectations, sqlfluff, or Lightdash. The dbt implementation is the example for what we should be doing:

1. Use the tool's native paradigms to plug into Meltano-managed config - using code (yaml for dbt) that the user owns, even if they don't have to write it.
2. Do so in a way that the user doesn't need to understand the internals to start being productive.
3. Do so in a way that when they do want to customize or get better control, they can do this directly.

Singer isn't a good parallel because Meltano _is_ the EL tool (for now). Meltano (as the EL tool) gets its config from environment variables for the most part, especially when it comes to secrets.

aaronsteers:
Config path should be customizable, yes, just like DBT_PROFILE_DIR and (separately) the path to dbt project root containing dbt_project.yml. But the config path should just by default point to a file or directory that's in their repo, that they can easily see and modify if they need to.

The argument for a config path doesn't really get to the issue of

DouweM:
> The argument for a config path doesn't really get to the issue of whether we give them a scaffold or hide the scaffold.

@aaronsteers I touched on that here: (emphasis mine)

> Even with that property, we have the option of leaving it unset by default, or providing an empty config file through a file bundle and pointing at that by default. I could be convinced that that makes sense in the case of Superset, since it has more complex settings than Singer and Airflow, but **I don't think we'd be helping users by making that the default behavior for extract/tap-foo/config.json and orchestrate/airflow.cfg as well, as the Meltano-managed approach is clearly superior there and it's worth pushing people to use it instead of the static files**. So here consistency (and the assumption that most tools will not have interpreted config files) point us to requiring config_path: my/config.ext to be set explicitly in meltano.yml in all cases.

---

> But the config path should just by default point to a file or directory that's in their repo, that they can easily see and modify if they need to.

Am I understanding you correctly that you're saying that this goes for Superset, and also for Singer and Airflow? So we'd add `config_path` properties to both (as I quickly put together in https://github.com/meltano/meltano/pull/6045), with default values pointing at `extract/tap-foo/config.json` and `orchestrate/airflow.cfg` in the repo, which we'd scaffold as empty files?

DouweM:
> As I already mentioned, the Airflow behavior needs to be changed because it's already creating problems.

@aaronsteers Correct me if I'm misremembering or misunderstanding, but were you suggesting that users should have a full `airflow.cfg` in their project so that they can easily use it with a non-Meltano-managed Airflow containerized solution? And as `airflow.cfg` isn't interpreted but essentially just a serialized dict, we couldn't have the best-of-both-worlds option 1 from https://github.com/meltano/meltano/pull/6031#issuecomment-1148092898 where the user can use both Meltano-managed and custom properties side by side, and you're saying we would forego Meltano-managed config for Airflow entirely? (Just like `meltano config` doesn't really work for dbt, only for a handful of env vars that are referenced from `profiles.yml` etc, but we couldn't even do that with `airflow.cfg` since it can't read env vars...) 

Does it make a difference to portability that we have `meltano invoke --dump=config airflow` to dump the full generated `airflow.cfg`, which someone could then ship off to somewhere else?

If that's not what you're suggesting, I think we're agreeing that it's valuable to have Meltano combine the Meltano-managed config with the user-provided `airflow.cfg` when generating the config file, as in https://github.com/meltano/meltano/pull/6045, so that it's and-and rather than either-or. If that's the case, are we only disagreeing on "whether we give them a scaffold or hide the scaffold" as in https://github.com/meltano/meltano/pull/6031#issuecomment-1148194040?

pnadolny13:
> @aaronsteers That was actually the original approach (see https://github.com/meltano/hub/issues/250#issuecomment-1146874661), but that didn't allow Superset to be configured through Meltano's own config and environments features, and would leave a lot of boilerplate code inside the user's repo that they shouldn't be touching or even knowing about.

@DouweM I agree that it would leave boilerplate code but FWIW I think you could configure Superset through Meltano's own config and environments features. I was suggesting in https://github.com/meltano/files-superset/pull/12#discussion_r889574035 that a user could set environment specific settings using the Meltano environment feature as settings or plugin extras. Then they just customize the superset_config.py to receive as env vars and use them as they need. They could chose to have dev and prod superset_config.py files if they wanted and just toggle the config_path.

Side note: I found it relatively common in the Superset slack for people to have issues with Superset picking up their custom config file, my thought was that we're solving a pain point by handing them a config file and defaulting to reading it from the expected path so all they need to do is edit it.

I agree that its hard to generalize our plugins right now. My view is that Singer is a special case where hiding the config/catalog/state files is good because its so complicated while with Airflow and others its probably better to hide less and be more explicit.

> 1. Put superset_config.py in the project through a file bundle, with code injecting the Meltano-managed config at the top, so the user can add their own stuff below

I vote for this blended approach, although it sounded like @DouweM said this wasnt possible in https://github.com/meltano/meltano/pull/6031#issuecomment-1148092898 but I don't totally follow why not.

DouweM:
> I agree that its hard to generalize our plugins right now. My view is that Singer is a special case where hiding the config/catalog/state files is good because its so complicated while with Airflow and others its probably better to hide less and be more explicit.

@pnadolny13 I think "with Airflow and others its probably better to hide less and be more explicit" is where we simply disagree. I don't see a strong advantage in showing people `airflow.cfg` by default if they can do everything through `meltano config`/`meltano.yml` with tons of additional configuration functionality (layers, environments) added on top. "Punching through the code" as Taylor called it and using `airflow.cfg` directly is a advanced feature most people should never need. Providing separate `airflow.cfg` (or `superset_config.py` etc) for dev, staging and prod environments is strictly inferior in my eyes to using Meltano's more finegrained functionality where each setting can be set at the root or selectively overridden in an env.

> 
> > 1. Put superset_config.py in the project through a file bundle, with code injecting the Meltano-managed config at the top, so the user can add their own stuff below
> 
> I vote for this blended approach if possible but it sounded like @DouweM said there was a reason why we cant in [#6031 (comment)](https://github.com/meltano/meltano/pull/6031#issuecomment-1148092898), I dont know the details of why not though.

The blended approach is only possible with Superset, because `superset_config.py` can run actual code to mix in Meltano-managed config with custom overrides. But it's not possible with Airflow and Singer and most other tools that use static serialized dictionaries, so I'd rather go with option 2 that works with all 3, than making the behavior inconsistent.

---

> Both of these in my mind argue for the Option 2 in https://github.com/meltano/meltano/pull/6031#issuecomment-1148092898 - auto-generate the scaffolding (superset_config.py) out of the box but make it clear and easy to enhance or override the behavior and go fully custom ("punch through to the code" as I like to say sometimes).

@tayloramurphy The implementation in this branch currently does not put `superset_config.py` in your project by default, suggesting you use Meltano-managed config to manage all Superset config properties if possible, but lets you run `meltano config superset set _config_path analyze/superset/superset_config.py` so that you can put your own stuff in that new file, which Meltano will automatically combine with its Meltano-managed config when generating the file to be passed to Superset. That's what I meant with option 2, is that also what you had in mind with "auto-generate the scaffolding out of the box" and "go fully custom"?

tayloramurphy:
> The implementation in this branch currently does not put superset_config.py in your project by default, suggesting you use Meltano-managed config to manage all Superset config properties if possible, but lets you run meltano config superset set _config_path analyze/superset/superset_config.py so that you can put your own stuff in that new file, which Meltano will automatically combine with its Meltano-managed config when generating the file to be passed to Superset. That's what I meant with option 2, is that also what you had in mind with "auto-generate the scaffolding out of the box" and "go fully custom"?

@DouweM yes I believe this is correct. To repeat back to you:

* by default users won't have to provide a `superset_config.py` file and can instead manage everything via Meltano. Meltano will generate the file for them and pass it to Superset
* If users want, they can provide `config_path` and point it to their own `supserset_config.py` file
* If users have superset configuration in Meltano _and_ they've specified a `config_path` those two files will be combined via a simple dict merge
* If users do not have superset configuration in Meltano and they've specified a `config_path` then Meltano will only read from the users' `superset_config.py` (technically I guess it'll merge an empty dict into the custom file, but the behavior is correct).
* If users want to understand what the meltano-generated file looks like they can do `meltano invoke --dump=config superset` to inspect it themselves or possibly even use that at their `config_path` if they really wanted to. 

I quite like this approach and I'm game to make it happen like this.

pnadolny13:
@DouweM I'm not very familiar with the code itself but I was able to test this locally and it looks good to me! I overrode some configs in meltano.yml and also in a custom superset_config.py.

DouweM:
@aaronsteers @pandemicsyn Would you like to review the code before I merge this?

src/meltano/core/plugin/superset.py
```diff
@@ -0,0 +1,168 @@
...
+
+        Raises:
+            PluginExecutionError: if config file couldn't be found
+        """
+        config = invoker.plugin_config_processed
+
+        config_script_lines = [
+            "import sys",
+            "module = sys.modules[__name__]",
+            f"config = {str(config)}",
```
aaronsteers, line 71: Can we get these dynamically from env vars and avoid secrets being stored in clear text?

aaronsteers:
@DouweM - overall it looks okay.

I left a comment about clear text passwords in the generated file, but we're out of time to discuss further.

src/meltano/core/plugin/superset.py
DouweM: @aaronsteers I considered that, but env vars are all strings and `superset_config.py` expects values to have their proper types, e.g. integer. To read from env vars and convert to proper types, we'd need to hard-code the type of every setting, which is not reasonable and out of scope.

DouweM: @aaronsteers I see your concern, though. Do you know if we have a dedicated issue for that as it's also a problem with Singer and Airflow?

aaronsteers: It's a few more lines of code, but since this code path does have access to type info, we could have two declarations - one for secrets (kind=password) and the others in clear text, as is now.

We could also just encode conversions for all variables, but that's not going to be possible in this timeline.

Let's log as a follow-on issue.

DouweM: @aaronsteers Issue created: https://github.com/meltano/meltano/issues/6065. Please share your thoughts in there, I want to better understand the security implications and relevance to advanced deployments.

PR #6033, opened 2022-06-05
Let ProjectPlugin override capabilities and extend settings of parent

### LINKED ISSUE #6029: Settings/capabilities can't be overridden locally for plugin added from the hub with incomplete metadata
See https://github.com/meltano/meltano/issues/2717#issuecomment-1146363583 and @edgarrmondragon's response.

It means that adding a plugin from the Hub with missing/incorrect metadata leads to user to a bit of a dead end, as they won't be able tweak the `capabilities` and `settings` in their own project to find out what works 😬 They could directly edit the lockfile, but I don't think we want users to actually do that. 

There needs to be a way to for a user to try out metadata changes locally to get the plugin working, before they can contribute them to the Hub, and it doesn't look like there currently is.

We didn't have this issue with `meltano add --custom`, because that would always add a full plugin definition (with a `namespace`) to `meltano.yml` that didn't inherit anything from the Hub or discovery.yml and would allow `capabilities` and `settings` to be set.

DouweM:
Closes https://github.com/meltano/meltano/issues/6029


src/meltano/core/plugin/project_plugin.py
```diff
@@ -370,23 +398,30 @@ def venv_name(self) -> str:
...
+        return [
             ProjectPlugin(plugin_type=plugin_type, name=dep.name, variant=dep.variant)
-            for plugin_type, deps in self.requires.items()
+            for plugin_type, deps in self.all_requires.items()
             for dep in deps
         ]
-
-        plugins.extend(
-            ProjectPlugin(plugin_type=plugin_type, name=dep.name, variant=dep.variant)
-            for plugin_type, deps in self._parent.requires.items()
```
DouweM, line 388: We weren't correctly inheriting multiple layers of `requires` before, only the direct parents, but not the parent's parents('s parents).

pandemicsyn: can we construct a test case for this to make sure we don't drift again ?

edgarrmondragon: @pandemicsyn That might be addressed in https://github.com/meltano/meltano/pull/5999/ but I forgot we hand0t merged that 🤦

pandemicsyn: oh heh , yea - y'all are definitely gonna have a conflict 😁 @edgarrmondragon do you wanna just let this change roll first and rebuild on top of it?

edgarrmondragon: @pandemicsyn I can wait for this to merge, but I'm not sure if we're blocking @pnadolny13

pnadolny13: @edgarrmondragon no its not blocking me, I just needed it fixed in order to confirm that my changes in https://github.com/meltano/hub/pull/538 were working. Do whatever is best for you all.

edgarrmondragon:
@DouweM I can take it now that @pnadolny13 confirmed #5999 is not blocking him, so I'll fix CI here and update that one after this merges.

edgarrmondragon (approved):

PR #6035, opened 2022-06-06
snowplow env context python impl missing

pnadolny13:
@pandemicsyn I was getting bad events with `"message":"$.python_implementation: is missing but it is required"` when I ran `invoke` locally with snowplow micro. I see them pass validation after making this change.


pandemicsyn (approved):

WillDaSilva (approved):

edgarrmondragon (approved):
lgtm

PR #6043, opened 2022-06-06
Add feature flag for strict environment variable mode.

### LINKED ISSUE #3348: Allow defining required environment variables
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3426

Originally created by @pnadolny13 on 2022-04-20 18:09:54

---

Its common for projects to have a .env file with user or development secrets which is git ignored, along with a sample of that .env that is checked into git for users to know what environment variables to set when they clone the project.

Meltano could allow you to set these as env vars as required so it can fail if they arent set instead of relying on users to read the sample file and export them properly. They can be set at the environment level to allow for variation based on environment.

Squared Project Use Case:

For ELT development the Squared project has isolated environments inside Snowflake that are configured based on some local settings: 
- the [USER_PREFIX](https://gitlab.com/meltano/squared/-/blob/master/data/environments.meltano.yml#L116) set in the meltano.yml which is propagated throughout to create the isolation
- the snowflake password thats stored once in the [.env](https://gitlab.com/meltano/squared/-/blob/master/data/.env.template#L18) so it can be shared by dbt/target-snowflake/great expectations.

The setup instructions tell the user to edit the meltano.yml and set the USER_PREFIX to their name (i.e. pnadolny) but thats not ideal because theres always a diff when trying to commit something. The other option is to remove it from the meltano.yml and set it in the .env file, which works but it not ideal because someone can easily forget to add it and without any validation the dbt configuration ends up writing to `USERDEV._PROD.TABLE_NAME` instead of `USERDEV.PNADOLNY_PROD.TABLE_NAME` because USERDEV is null and is not validated.

```yaml
environments:
- name: userdev
  config:
    plugins:
      extractors:
        ...
  required_env_vars:
    - USER_PREFIX
  env:
    OTHER_HARD_CODED_ENVS: foo
```

```bash
# .env file
USER_PREFIX=pnadolny
```

cjohnhanson:
Adds a new `STRICT_ENV_VAR_MODE` feature flag which will raise a new `EnvironmentVariableNotSetError` if a setting attempts to expand using an environment variable which is not set in that environment.

Adds tests for `ProjectSettingsService` and `PluginSettingsService` to confirm that error is raised when flag is set and _not_ raised when flag is not set.

Also adds one additional xfail'ing test for environment inheritance behavior to reflect new expectations around environment variable behavior.

Closes #3348


src/meltano/core/utils/__init__.py
```diff
@@ -366,7 +366,23 @@ def set_at_path(d, path, value):
...
+        """
+        super().__init__(env_var)
+        self.env_var = env_var
+
+    def __str__(self) -> str:
+        """Return the error as a string."""
+        return f"{self.env_var} referenced but not set."
+
+
+def expand_env_vars(raw_value, env: Dict, raise_error: bool = False):
```
edgarrmondragon, line 385: Wdyt about calling the new keyword arg `raise_if_missing`?

cjohnhanson: Yeah I think that's more descriptive and clear. Updated.

tayloramurphy (approved):
Docs approval ✅

aaronsteers (approved):
Looks good!  🚢 

No need to change anything here, but if we can collect data on this with other integration tests and with the `sqaured` project, I'd opt to make strict mode enabled by default. 

The only reason not to do so is if there are unexpected side effects, which we'd want to test carefully for in real-world examples. (I'll spin a new issue out to discuss/evaluate after this merges.)

Thanks, @cjohnhanson!

PR #6050, opened 2022-06-07
bump discovery.yml version

### LINKED ISSUE #6049: Pre-2.0 versions should still work with `env_aliases`
We received a report this morning [in slack](https://meltano.slack.com/archives/C01TCRBBJD7/p1654603659407119?thread_ts=1654597663.385549&cid=C01TCRBBJD7) that the `env_aliases` were broken in pre-2.0 versions.

This was not intended to be broken in pre-2.0  versions.

aaronsteers:
This was missed earlier and was needed to prevent: #6049


DouweM:
@aaronsteers We should also update https://github.com/meltano/meltano/blob/main/src/meltano/core/plugin_discovery_service.py#L39, right?

PR #6051, opened 2022-06-07
Add default source schema to Transforms

### LINKED ISSUE #6048: Update Transforms with default source schema values
To enable Transforms to work with adapter-specific dbt installs, whilst continuing to work with legacy 'dbt' and `elt`, references to `env_var('DBT_SOURCE_SCHEMA')` can be updated to include a default value that corresponds to the Transforms' Extractors' default schema (e.g. `env_var('DBT_SOURCE_SCHEMA', 'tap_carbon_intensity')` for `tap-carbon-intensity`). This would not automatically populate non-default source schemas if they are configured by the Extractor, however users can already override the configuration of Transformer `vars` manually in their `dbt_project.yml`.

As per the note in #5957 we now allow Transform plugins to be added to projects without the legacy 'dbt' Transformer installed, as long as at least one Transformer (i.e. adapter-specific dbt) is present. However Transforms currently rely on `vars` that reference  `{{ env_var('DBT_SOURCE_SCHEMA') }}` in `discovery.yml`, which itself is set to `$MELTANO_LOAD__TARGET_SCHEMA` by default in the legacy 'dbt' Transformer. This is how `meltano elt` passes schema information from Loader to Transformer. However the `source_schema` setting is not present in adapter-specific installs, nor is the underlying `MELTANO_LOAD__TARGET_SCHEMA` env var when using `run` or `invoke`, breaking Transforms when used with those Transformers.

kgpayne:
Closes #6048


tayloramurphy:
@kgpayne this should probably be to the transforms definitions on the Hub, right? After today we shouldn't be making any more changed to `discovery.yml`.

edgarrmondragon (approved):

PR #6053, opened 2022-06-07
Add lint workflow with iglulint job

### LINKED ISSUE #6052: Use GitHub action to validate Snowplow schemas with igluctl lint
We can probably use [iglu-schema-lint from the action marketplace](https://github.com/marketplace/actions/iglu-schema-lint) for this.

WillDaSilva:
Closes #6052


WillDaSilva:
@pandemicsyn I understand the reasoning behind the new on-disk schema file layout, but why are they in a directory named "iglu-client-embedded"?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,22 @@
+name: Lint
+
+on:
+  pull_request: {}
+  push:
+    branches: [main]
```
pandemicsyn, line 6: I don't think we really need to run this workflow on every change, wdyt about limiting this to run only when a schema file actually changes?

pandemicsyn:
@WillDaSilva just incase you haven't come across it yet (https://github.com/snowplow-incubator/snowplow-micro#quick-start)

.github/workflows/lint.yml
aaronsteers: @pandemicsyn - Agreed. 👍

aaronsteers (approved):

edgarrmondragon (approved):

pandemicsyn (approved):
:shipit:

PR #6060, opened 2022-06-07
Remove lockfile feature flag

### LINKED ISSUE #5984: Remove Lockfiles Feature Flag and use feature by default

cjohnhanson:
This completely removes the lock files feature flag and all references. Lock files will now be used by default.

Closes #5984


tayloramurphy (approved):
Approving docs changes

aaronsteers (approved):
LGTM! 

Once tests pass, I think this is ready to go.

cjohnhanson:
@edgarrmondragon [this is the tests that's failing](https://github.com/meltano/meltano/blob/a881d74719bea9fbc1985d5ee376e29f955c0319/tests/meltano/core/test_project_plugins_service.py#L136) I'm not immediately grokking what's going on here or why turning on the lockfile feature by default would have impacted this. Anything immediately jump out at you?

cjohnhanson:
@aaronsteers @edgarrmondragon this should be good to go now but I updated/removed several tests and would like to get another quick review/approval before merge.

tests/meltano/core/test_project_add_service.py
```diff
@@ -103,8 +103,8 @@ def test_add_inherited(
             "inherit_from": "tap-mock-inherited",
         }
```
cjohnhanson, line 105: @aaronsteers @edgarrmondragon -- I commented all of the other changed tests to explain why lockfiles were being removed there.

The diffs in this file are because if I'm understanding correctly, now meltano will only reach out to the hub _once_ and then use the plugin definition it finds in the lockfile. So these should all be `1` now

edgarrmondragon: @cjohnhanson that's correct. Once the lockfile has been added, there should only be 1 trip to Hub for each endpoint.

edgarrmondragon (approved):
nice work @cjohnhanson! Those lingering lockfiles between tests were annoying 😬

PR #6068, opened 2022-06-07
Rough pass to fire elt tracking events

pandemicsyn:
This is a rough pass at firing new style cli tracking events from the elt command, with plugin tracking context. It differs from `meltano run` in that it does not distinguish between user caused failures which `run` flags as `aborted`  and actual errors like a loader failure that run flags as `failed`. In this implementation both scenarios will yield a `failed` cli event. 

Distinguishing between the two is totally possible in a future version, but its more invasive than I'm comfortable doing given the current time constraints.  

tl;dr i don't wanna break the elt command right before the 2.0 launch trying to catch specific exceptions in specific spots.


pandemicsyn:
@pnadolny13 @aaronsteers if either of y'all are around - heres the "eh, good enough?" telemetry for `elt`. @aaronsteers it does at least attach a PluginsContext to cli events after snagging the plugins from the created elb_context (so right before execution would start).

src/meltano/core/tracking/tracker.py
```diff
@@ -215,6 +215,14 @@ def timezone_name(self) -> str:
         except Exception:
             return datetime.datetime.now().astimezone().tzname()
 
+    def add_contexts(self, *extra_contexts):
+        """Add additional Snowplow contexts to the `Tracker`.
```
WillDaSilva, line 219:
```suggestion
        """Permanently add additional Snowplow contexts to the `Tracker`.
```

aaronsteers (approved):
Looks good!

WillDaSilva (approved):
LGTM!

src/meltano/cli/elt.py
```diff
@@ -237,10 +266,16 @@ async def _redirect_output(log, output_logger):
...
+async def _run_elt(
+    tracker: Tracker,
+    log: structlog.BoundLogger,
+    context_builder: ELTContextBuilder,
+    output_logger: OutputLogger,
+):
     async with _redirect_output(log, output_logger):
         try:
             elt_context = context_builder.context()
+            tracker.add_contexts(plugins_tracking_context_from_elt_context(elt_context))
```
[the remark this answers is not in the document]
pandemicsyn: @aaronsteers yup you got it.

pnadolny13:
@pandemicsyn I'm seeing `run` as the command and `transform` in the option_keys list when I run `elt tap-csv target-jsonl`. I think the `transform` is inferred but it would be nice to know if someone actually passed that arg vs it just being included because its there by default.

```
  "schema": "iglu:com.meltano/cli_context/jsonschema/1-0-0",
  "data": {
      "command": "run",
      "sub_command": null,
      "option_keys": [
          "transform"
      ]
  }
```

src/meltano/cli/elt.py
```diff
@@ -103,6 +108,24 @@ async def elt(
 
     \b\nRead more at https://docs.meltano.com/reference/command-line-interface#elt
     """
+    tracker = Tracker(project)
+    cmd_ctx = cli_context_builder(
+        "run",
```
pandemicsyn, line 113:
```suggestion
        "elt",
```

pandemicsyn:
> @pandemicsyn I'm seeing run as the command and transform in the option_keys list when I run elt tap-csv target-jsonl. I think the transform is inferred but it would be nice to know if someone actually passed that arg vs it just being included because its there by default.

Fixed the `run` copy and paste fail. As for transform, since skip is the default - could I just omit sending it to you if I see the default value of `skip` being passed in? Since `meltano elt tap target --transform=skip` and `meltano elt tap target` are the same ?

pnadolny13:
@pandemicsyn thanks I'm seeing that updated now! Yeah if its defaulted I can understand why that would be difficult to differentiate, but I always like sharing the ideal state in case its a simple change. Ideally we'd know how to differentiate `--transform=skip` vs no arguments and default skip vs `--tranform=run` but its definitely not a big deal if it doesnt get added tonight.

aaronsteers:
@pandemicsyn 

> As for transform, since skip is the default - could I just omit sending it to you if I see the default value of `skip` being passed in?

Yeah - this makes sense to me and does the job.

pnadolny13 (approved):
@pandemicsyn this looks awesome! Other than the thread about the `--transform` stuff thats an optional change, I wasnt able to find anything else.

src/meltano/cli/elt.py
```diff
@@ -103,6 +108,29 @@ async def elt(
...
+        state_id=state_id,
+        force=force,
+    )
+    tracker.add_contexts(cmd_ctx)
+    tracker.track_command_event(cli_tracking.STARTED)
+
+    # we no longer set a default choice for transform, so that we can detect explicit usages of the --transform option
+    # if transform is None we still need manually default to skip after firing the tracking event above.
+    if not transform:
+        transform = "skip"
```
pandemicsyn, line 132: @aaronsteers @pnadolny13 this should do it. Since `transform` was click.Choice it never actually printed what the default choice was in the help string as far as I could tell. 

So I just omit the default in the click.option definition and then override it here. But above we'll correctly only include the transform option in the tracking event if it was actually set by the user.

pandemicsyn:
@pnadolny13 @aaronsteers thing i was able to make it do our bidding , would one of mind y'all mind giving the new change a quick once over before I merge.

aaronsteers (approved):
Looks good! Thanks!

PR #6071, opened 2022-06-07
Make Poetry include `.release_marker`

### LINKED ISSUE #6021: Verify release marker behaves as intended
Per @aaronsteers  in https://github.com/meltano/meltano/pull/5978/files#r889606266 the release marker landing in #5978 maybe not always get packaged/distributed as intended:

> Python packaging is weird sometimes about what actually ships and what doesn't. Can we double check this file gets bundled and installed with the operation? (Either in docs or through a 'real' test.)

> Also - and I just realized this now - I think this breaks if a user does pipx install git+https://github.com/meltano/meltano.git

> For now, I think this can be a bug we would live with, but perhaps we open an issue to track/prioritize in a future PR.

WillDaSilva:
Assuming that Poetry doesn't need the full path for `include`, this should fix the problem of `.release_marker` not being installed by `poetry install`. Marked as draft because I have not had the time to test it yet.

Closes #6021

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,8 @@ classifiers = [
 ]
 include = [
   "LICENSE",
-  "*.m5o"
+  "*.m5o",
+  ".release_marker",
 ]
 
 [tool.poetry.dependencies]


WillDaSilva:
@aaronsteers @pandemicsyn @pnadolny13 This PR is not necessary. I've manually verified that the file is included when [this code](https://github.com/meltano/meltano/issues/6021#issuecomment-1149254365) is run in CI for the build of the release sdist. The issue around pipx installing Meltano directly from GitHub remains an issue. I'll address it in #6021.

PR #6082, opened 2022-06-08
add analyze page back in with superset docs

pnadolny13:
This has a lot of copied content from https://hub.meltano.com/utilities/superset so we can optionally pare it down and just basically link out to that hub page but I think this is a good starting point so we have content on release day.


aaronsteers (approved):

tayloramurphy:
Pushed a quick commit to change the ordering in the sidebar

tayloramurphy (approved):
I like it!

Issue #5986, opened 2022-06-03
Collect process exit events with Snowplow


WillDaSilva:
It would be useful to emit an event to Snowplow when the Python process running Meltano exits. It could let us know how long a process ran for (by comparing the exit timestamp to the `process_creation_timestamp` from the environment context), as well as the exit code.

We'll want to wrap the whole execution of the CLI in a try-finally block. Within the finally block the exit code can be obtained like so in `src/meltano/cli/__init__.py`:

```python
exit_code = None

def main():
    try:
        # <current definition of `main` goes here>
    finally:
        global exit_code
        ex = sys.exc_info()[1]
        if ex is None:
            exit_code = 0
        elif isinstance(ex, SystemExit):
            exit_code = 0 if ex.code is None else ex.code
```

Now that the exit code has been retrieved, we can use an [`atexit`](https://docs.python.org/3/library/atexit.html) handler to emit the Snowplow event. `Tracker.__init__` can register an `atexit` handler with itself as an argument, and that handler can check for an exit code in `meltano.cli`:

```python
def __init__(self, project: Project, ...):
    # <current definition of `Tracker.__init__` goes here>
    atexit.register(self.track_exit_event)

def track_exit_event(self):
    from meltano.cli import exit_code
    self.track_unstruct_event(
        SelfDescribingJson(
            "iglu:com.meltano/exit_event/jsonschema/1-0-0",
            {"code": exit_code, "timestamp": datetime.datetime.utcnow().isoformat() + "Z"}
        )
    )
```

Exit event schema (`src/meltano/core/tracking/schemas/exit_event.json`):

```json
{
    "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
    "description": "Schema for a Snowplow event of a process exiting",
    "self": {
        "vendor": "com.meltano",
        "name": "exit_event",
        "format": "jsonschema",
        "version": "1-0-0"
    },
    "type": "object",
    "properties": {
        "code": {
            "description": "The exit code of the process",
            "type": [
                "integer",
                "null"
            ],
            "minimum": 0,
            "maximum": 255
        },
        "timestamp": {
            "description": "The time ",
            "type": "string",
[...]

PR #6098, opened 2022-06-08
Update invoke to fire start event later

### LINKED ISSUE #6093: Telemetry - Invoke services dont have plugin context or structured event
When running a service plugin like superset that blocks for a long time and is then exited using CTRL+C I'm not able to track the plugin context. I think whats happening is:

- `meltano invoke superset:ui` (blocking)
- CLI started event fired when started up. Plugin context isnt available yet
- build dashboards
- CTRL+C
- <MISSING CLI completed/aborted/etc. with plugin context>
- <MISSING structured legacy event>

I think the fact that I need to CTRL+C to close the blocking service plugin means the structured event isnt fire because its not a successful exit and with the new unstructured events we dont get a plugin context. I dont get enough info for these events to be useful, all I know its that `invoke` ran but nothing about what plugin/command (i.e. `ui`)/etc. This is likely undercounting usage of Airflow also because we never get to the successful firing of structured events either. 

Would it be possible to include the plugin context on the first event?

cc @pandemicsyn @tayloramurphy @WillDaSilva

pandemicsyn:
This updates the telemetry for invoke to not fire the first event until *after* we've tried to obtain plugin context. If we're able to find the requested plugin, it will be present in the PluginsContext for the start event. 

This also cleans up the use of the tracker and does away `with` contexts use.

closes #6093

### DIFF
--- a/src/meltano/cli/invoke.py
+++ b/src/meltano/cli/invoke.py
@@ -113,8 +113,8 @@ def invoke(
         raise err
 
     if list_commands:
-        tracker.track_command_event(cli_tracking.COMPLETED)
         do_list_commands(plugin)
+        tracker.track_command_event(cli_tracking.COMPLETED)
         return
 
     invoker = invoker_factory(project, plugin, plugins_service=plugins_service)
--- a/src/meltano/cli/invoke.py
+++ b/src/meltano/cli/invoke.py
@@ -140,7 +140,6 @@ def invoke(
         tracker.track_command_event(cli_tracking.COMPLETED)
     else:
         tracker.track_command_event(cli_tracking.FAILED)
-        tracker.track_command_event(cli_tracking.FAILED)
     sys.exit(exit_code)
 
 
--- a/src/meltano/cli/invoke.py
+++ b/src/meltano/cli/invoke.py
@@ -78,6 +78,7 @@ def invoke(
     \b\nRead more at https://docs.meltano.com/reference/command-line-interface#invoke
     """
     tracker = Tracker(project)
+    # the `started` event is delayed until we've had a chance to try to resolve the requested plugin
     tracker.add_contexts(
         cli_context_builder(
             "invoke",
@@ -108,6 +109,7 @@ def invoke(
         tracker.add_contexts(PluginsTrackingContext([(plugin, command_name)]))
         tracker.track_command_event(cli_tracking.STARTED)
     except PluginNotFoundError:
+        # if the plugin is not found, we fire started and aborted tracking events together to keep tracking consistent
         tracker.track_command_event(cli_tracking.STARTED)
         tracker.track_command_event(cli_tracking.ABORTED)
         raise
--- a/src/meltano/cli/invoke.py
+++ b/src/meltano/cli/invoke.py
@@ -78,17 +78,17 @@ def invoke(
     \b\nRead more at https://docs.meltano.com/reference/command-line-interface#invoke
     """
     tracker = Tracker(project)
-    cmd_ctx = cli_context_builder(
-        "invoke",
-        None,
-        plugin_type=plugin_type,
-        dump=dump,
-        list_commands=list_commands,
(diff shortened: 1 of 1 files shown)


pandemicsyn:
@pnadolny13 so for something like `meltano invoke airflow:ui` you should see this in the started event:

```json
{
						"schema": "iglu:com.meltano/cli_context/jsonschema/1-0-0",
						"data": {
							"command": "invoke",
							"sub_command": null,
							"option_keys": []
						}
					},
					{
						"schema": "iglu:com.meltano/plugins_context/jsonschema/1-0-0",
						"data": {
							"context_uuid": "c5f7f99b-4bf8-443a-8fc2-a248185e8c3b",
							"plugins": [
								{
									"category": "orchestrators",
									"name_hash": "16df48936f837b402448fbbff41aaba51279da9c9b66d4b39412ab9d8237eedd",
									"namespace_hash": "16df48936f837b402448fbbff41aaba51279da9c9b66d4b39412ab9d8237eedd",
									"executable_hash": "16df48936f837b402448fbbff41aaba51279da9c9b66d4b39412ab9d8237eedd",
									"variant_name_hash": "d7e04cba6a8ba148c966860632ef3636c77935984791f36ce5457016d7c313da",
									"pip_url_hash": "ff76a19648e1e83e020ec374f5f8f86d38ea823dda382c596f32617872dbd10d",
									"parent_name_hash": "16df48936f837b402448fbbff41aaba51279da9c9b66d4b39412ab9d8237eedd",
									"command": "ui"
								}
```

src/meltano/cli/invoke.py
```diff
@@ -105,13 +105,16 @@ def invoke(
         plugin = plugins_service.find_plugin(
             plugin_name, plugin_type=plugin_type, invokable=True
         )
+        tracker.add_contexts(PluginsTrackingContext([(plugin, command_name)]))
+        tracker.track_command_event(cli_tracking.STARTED)
     except PluginNotFoundError as err:
```
WillDaSilva, line 110:
```suggestion
    except PluginNotFoundError:
```

src/meltano/cli/invoke.py
```diff
@@ -105,13 +105,16 @@ def invoke(
...
-        with tracker.with_contexts(cmd_ctx):
-            tracker.track_command_event(cli_tracking.ABORTED)
+        tracker.track_command_event(cli_tracking.STARTED)
+        tracker.track_command_event(cli_tracking.ABORTED)
         raise err
```
WillDaSilva, line 113:
```suggestion
        raise
```

Using a blank raise will re-raise the exception that is currently being handled.

src/meltano/cli/invoke.py
```diff
@@ -105,13 +105,16 @@ def invoke(
         plugin = plugins_service.find_plugin(
             plugin_name, plugin_type=plugin_type, invokable=True
         )
+        tracker.add_contexts(PluginsTrackingContext([(plugin, command_name)]))
+        tracker.track_command_event(cli_tracking.STARTED)
     except PluginNotFoundError as err:
-        with tracker.with_contexts(cmd_ctx):
-            tracker.track_command_event(cli_tracking.ABORTED)
+        tracker.track_command_event(cli_tracking.STARTED)
```
WillDaSilva, line 111: A comment explaining why we're starting late (and in this case, starting it only to immediately abort) would be appreciated.

WillDaSilva (approved):

pnadolny13 (approved):
@pandemicsyn this is working for me to fix the plugins_context! Although I still only get 1 event, is there a follow on to capture both the unstructured and structured (lower priority) exit event?

pandemicsyn:
@pnadolny13 yea that's expected, we don't capture keyboard interrupts or term signal's (and in a lot of cases won't be able to fire an exit event). We've got https://github.com/meltano/meltano/issues/5986 that might help in some scenarios.

Issue #2971, opened 2021-11-02
Prettier exception prints


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3037

Originally created by @aaronsteers on 2021-11-02 18:56:27

---

On the structlog site, there's some example screenshots of exceptions that are inherently very easy to read due to formatting and structure. 

Should we look at adopting something like [rich](https://rich.readthedocs.io/) or [better-exceptions](https://github.com/Qix-/better-exceptions) for Meltano and/or the SDK?

The docs for structlog seem to imply we only need to install the libraries, but I expect it may be slightly more work, and there could be reasons I'm not thinking of why we would not want to use them.

https://www.structlog.org/en/stable/development.html:

![image](http://gitlab.com/meltano/meltano/uploads/3e18f622a003275ef6358c18dab5977c/image.png)

MeltyBot:
[View 6 previous comments from the original issue on GitLab](https://gitlab.com/meltano/meltano/-/issues/3037)

aaronsteers:
@pandemicsyn, @edgarrmondragon, @kgpayne 

I'd like to pick this up - specifically to add support for prettier exception prints - and to do so behind a feature flag.

When we originally discussed this, we didn't have feature flags yet so the change was deemed excessively risky given people's various preferences and given that different execution environments would have different rendering experiences.

The overall dev cost is pretty low, and the potential improvement to user experience when doing local debug is very high.

@pandemicsyn, @edgarrmondragon - Anything we need to do to disambiguate this spec and size? Is `4` an appropriate weight estimate, given a minimal initial spec and given the expectation that we launch behind a feature flag, disabled by default?

aaronsteers:
Additional context:

To make a better experience for times when `structlog` is distracting and hard to read, I'd like to make a more viable experience when log level is set very high, such as to the `WARNING` level. If exceptions are clear and readable and containing the needed context, then we have less need for very verbose logging.

Not in any way a requirement that this be viable at the current stage, but I think this starts us in the right direction. Currently, our users have to make a choice every time they run meltano - "Do I think it might fail? If so, I need DEBUG logging turned on." But a more desirable situation would be that they almost never need DEBUG logging - instead relying on good exception logs and then clicking the a link to a detailed log file for further analysis when needed.

pandemicsyn:
[...]

PR #6103, opened 2022-06-08
Improve `meltano config <plugin> list`

### LINKED ISSUE #6122: Prettier CLI outputs with `rich`
Meltano currently uses the Click python library for our CLI definition and formatting. Whilst Click is excellent at defining CLI experiences, it is much less effective at formatting. Cross-platform libraries exist, such as rich, for better formatting of Click outputs and structlog logs. As Meltano is CLI-first, improvements in output formatting would positively impact most users.

Example improvements:

From #2971
---

image

With [rich-click](https://github.com/ewels/rich-click):
---

From #6103
---

Before

Screenshot 2022-06-08 at 17 44 58

After

Screenshot 2022-06-09 at 12 49 09

kgpayne:
`meltano config <plugin> list` is hard to parse, especially when settings include comprehensive descriptions (which are desired documentation artefacts, even if they reduce readability in the CLI). So I had a pass at using `rich` tables to improve readability.

**Note:** this is just a strawman example to try out [rich](https://github.com/Textualize/rich) in general. This may not ever be merged!

Before:

Screenshot 2022-06-08 at 17 44 58

After:

Screenshot 2022-06-08 at 17 45 30

FYI @tayloramurphy @aaronsteers and @edgarrmondragon (who originally suggested rich/textualize) 🙂


kgpayne:
@pandemicsyn one thing you might be able to point me in the direction of understanding and resolving. It seems `rich` has log formatting capabilities, which according to the docs have to be initialised. However, after simply installing rich it seems that DEBUG log formatting is enabled by default (or is maybe clashing with our log setup 🤔). It is pretty, but probably not what we want by default... but I can't find a way to turn it off 🤦‍♂️

Steps to repro: just a a `CliError("Test error.")` at the top of the `config list` command implementation and call list.

Screenshot 2022-06-08 at 18 25 02

edgarrmondragon:
+1 for more Rich interfaces

edgarrmondragon:
@kgpayne I think to disable rich tracebacks, you need to add 

```diff
+ from structlog.dev import plain_traceback

    structlog.configure(
        processors=[
            structlog.stdlib.add_log_level,
            structlog.stdlib.PositionalArgumentsFormatter(),
            TIMESTAMPER,
            structlog.processors.StackInfoRenderer(),
            structlog.processors.format_exc_info,
            structlog.stdlib.ProcessorFormatter.wrap_for_formatter,
+           structlog.dev.ConsoleRenderer(exception_formatter=plain_traceback),
        ],
```

in https://github.com/meltano/meltano/blob/688557554dd566c72a531ff2ea7bfb4bc7eb42d1/src/meltano/core/logging/utils.py#L132-L144

kgpayne:
[This comment](https://github.com/hynek/structlog/issues/394#issuecomment-1016199677) from the maintainer on a similar issue:

> You seem to have configured structlog to log rendered strings into stdlib logging. You have to use [structlog.org/en/stable/api.html#structlog.stdlib.ProcessorFormatter.wrap_for_formatter](https://www.structlog.org/en/stable/api.html#structlog.stdlib.ProcessorFormatter.wrap_for_formatter) in your structlog configuration as renderer (aka last processor), such taht your ProcessorFormatters within logging get raw data.

We already have `structlog.stdlib.ProcessorFormatter.wrap_for_formatter` as our last processor though 🤔

kgpayne:
@aaronsteers @tayloramurphy just looking to merge or close my lingering PR's. Are we in favour of adding `rich`, with this as the first usage? Otherwise I will close this PR 🙂

aaronsteers:
> @aaronsteers @tayloramurphy just looking to merge or close my lingering PR's. Are we in favour of adding `rich`, with this as the first usage? Otherwise I will close this PR 🙂

I think the challenge here is that we don't have much data yet that this will be a good experience for all use cases. Putting behind a feature flag would mitigate that risk and I like the idea of also making progress on 

- #2971 

@kgpayne - Do you think it would be feasible to put this behind a feature flag? That would presumably mean keeping our prior rendering logic for cases where the user wants `rich` disabled. 

If yes, I think tackling this and #2971 behind the same feature flag makes sense. We could tackle either this or #2971 first, or both together if its simple enough.

Wdyt?

edgarrmondragon:
> I'm generally in favor, but I think the challenge here is that we don't have much data yet that this will be a good experience for all use cases. Putting behind a feature flag would mitigate that risk and I like the idea of also making progress on

@kgpayne @aaronsteers how about putting it behind a different command that acts as preview (e.g. `list-rich`)? Then we can track command usage to see if folks prefer one or the other. If there's interest, we can later make `list` an alias of `list-rich` and deprecate the later.

pandemicsyn:
> > I'm generally in favor, but I think the challenge here is that we don't have much data yet that this will be a good experience for all use cases. Putting behind a feature flag would mitigate that risk and I like the idea of also making progress on
> 
> @kgpayne @aaronsteers how about putting it behind a different command that acts as preview (e.g. `list-rich`)? Then we can track command usage to see if folks prefer one or the other. If there's interest, we can later make `list` an alias of `list-rich` and deprecate the later.

@edgarrmondragon @kgpayne @aaronsteers Rather than doing a specific subcommand, I'd vote for adding a `--format` option. Else-where `list` (in jobs and schedules) already takes a `--format` option that controls whether the output is in a `plain text` or `json`. So IMHO this would keep things more consistent across commands. Especially if we can also add a `json` formatted option as well.

aaronsteers:
> 
> @edgarrmondragon @kgpayne @aaronsteers Rather than doing a specific subcommand, I'd vote for adding a `--format` option. Else-where `list` (in jobs and schedules) already takes a `--format` option that controls whether the output is in a `plain text` or `json`. So IMHO this would keep things more consistent across commands. Especially if we can also add a `json` formatted option as well. 
> 
> 

Was going to suggest the same. A `--format=rich` makes a lot of sense here and could be contrasted with --format=text and --format=json, etc.

kgpayne:
@aaronsteers @pandemicsyn @edgarrmondragon I love `--format` too. It reminded me of `--output` on the AWS CLI, which is configurable both in the config file (globally) or using/overriding `--output` for a specific command. Making the default `format` configurable in `meltano.yml` might make sense for us too.

One follow up question I have is where the `--format` should be defined. In existing cases, it is associated with a subcommand (and just the subcommands that currently benefit from formatting). We can continue that convention, but (following the AWS example), would it make sense to introduce a top-level `meltano --format=text ...` option and set the expectation that commands should all support `text` (the default) and `json` (machine readable) output as a minimum? I think this would be a great improvement to the CLI overall, but it maybe needs its own discussion 😅 Implementation wise, I would imagine a `BaseCliOutput` with `text`, `rich`, `json` abstract methods, so a developer can easily subclass and implement output formatting for each case with maximum code-reuse.

aaronsteers:
@kgpayne - I like the idea of either `--output=[json|no-color|text|rich]` _or_ `--format=[json|no-color|text|rich]` being a top-level option. But doing so significantly blows up the scope of this PR and would trigger a need for broader spec "sanity checks" for how this would perform across basically all commands and subcommands.

If we want to pause and consider as a global option, I think a spinoff discussion or issue would be appropriate for that.

pandemicsyn:
> @kgpayne - I like the idea of either `--output=[json|no-color|text|rich]` _or_ `--format=[json|no-color|text|rich]` being a top-level option. But doing so significantly blows up the scope of this PR and would trigger a need for broader spec "sanity checks" for how this would perform across basically all commands and subcommands.
> 
> If we want to pause and consider as a global option, I think a spinoff discussion or issue would be appropriate for that.
> 
> What do others think?

I think in the short term doing --format on the subcommand is my vote.

Long term, Instead of re-inventing the wheel, we do have standardized structured logging facilities we could try and use to control and format output that we could iterate on. No one says we have to keep our existing log formats or fields. We can make it do our bidding. IMHO could stop using `click.echo` and start using `log` calls and a `--format` option to control the output. 

I think we're past due for figuring out what we want our logging and output UX experience to be and coming up with a path forward for that. 

As an aside. Doing a `rich` table and cell based driven text user interface will hopefully not be our default for all the things - so things like `list` commands might just be one-offs that require custom output formatters.

kgpayne:
Closing as stale.

PR #6108, opened 2022-06-08
Add telemetry to `add` command

pandemicsyn:
This add's new style snowplow telemetry to `meltano add`.

### DIFF
--- a/src/meltano/cli/add.py
+++ b/src/meltano/cli/add.py
@@ -1,6 +1,5 @@
 """Plugin Add CLI."""
-
-from typing import List
+from __future__ import annotations
 
 import click
 
@@ -12,6 +11,9 @@ from meltano.core.plugin_install_service import PluginInstallReason
 from meltano.core.project import Project
 from meltano.core.project_add_service import ProjectAddService
 from meltano.core.project_plugins_service import ProjectPluginsService
+from meltano.core.tracking import PluginsTrackingContext, Tracker
+from meltano.core.tracking import cli as cli_tracking
+from meltano.core.tracking import cli_context_builder
 
 from . import cli
 from .params import pass_project
@@ -54,7 +56,7 @@ from .utils import (
 )
 @pass_project()
 @click.pass_context
-def add(
+def add(  # noqa: WPS238
     ctx,
     project: Project,
     plugin_type: str,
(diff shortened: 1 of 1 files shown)


src/meltano/cli/add.py
```diff
@@ -95,25 +112,38 @@ def add(
...
+                    plugin,
+                    inherit_from=inherit_from,
+                    variant=variant,
+                    custom=flags["custom"],
+                    add_service=add_service,
+                )
             )
-        )
-        tracker.track_meltano_add(plugin_type=plugin_type, plugin_name=plugin)
+        except Exception:
```
pandemicsyn, line 136: @pnadolny13 @aaronsteers this is the biggest gotcha, if one of the plugins is not known to meltano and the add command fails, we have no way to track what the requested plugin was. We do at least send along the PluginsContext for any plugins that we were able to resolve up until we hit the failure.

WillDaSilva: Does that information come through in some other way? Such as through the info in the CLI context?

aaronsteers: For this and also for `install`, I think in a future iteration I'd like to separate this into a thinner 'cli started' with all or no plugins, but then a separate 'start' and 'end' for each plugin's addition or installation.

Related:

Do we already have something like an "ExceptionContext" for trace info, call stack, error messages, or similar? I'm happy to log that though and keep it out of this round.

aaronsteers: So, in a future iteration... somthing like:

`
cli - start
add tap-foo - start
install tap-foo start
install tap-foo complete
add tap-foo - complete
add target-foo - start
install target-foo - start
install target-foo - failed
add target-foo - failed
cli - failed
```

WillDaSilva: The exception context is a great idea! We can easily get the current stack and exception info with `sys.exc_info` (or something like that - I'm on my phone and don't want to look it up). Then we can either send it with every event, or send it if and only if there is an exception currently being handled.

pnadolny13 (approved):
@pandemicsyn functionally this is working as I'd expect! One thing to call out is that I'm seeing a structured event fired even when the add fails (i.e. the actual python package install fails), which is fine if thats the status quo since we're keeping those basically as-is but its in consistent because all other struct events I know of are only fired on success.

cc @tayloramurphy

pandemicsyn:
@aaronsteers can i get a final 👍 or 👎 on this ?

pandemicsyn:
> @pandemicsyn functionally this is working as I'd expect! One thing to call out is that I'm seeing a structured event fired even when the add fails (i.e. the actual python package install fails), which is fine if thats the status quo since we're keeping those basically as-is but its in consistent because all other struct events I know of are only fired on success.

@pnadolny13 correct, i didn't alter how/when the legacy event fires. In the case of add it fired before the plugin was actually installed. I can change but it does mean its different than before 2.0.

aaronsteers (approved):

PR #6109, opened 2022-06-08
Add telemetry to `install` command

pandemicsyn:
This add's new style snowplow telemetry to meltano install.

### DIFF
--- a/src/meltano/cli/install.py
+++ b/src/meltano/cli/install.py
@@ -1,8 +1,14 @@
+"""CLI command `meltano install`."""
+from __future__ import annotations
+
 import click
 
 from meltano.core.legacy_tracking import LegacyTracker
 from meltano.core.plugin import PluginType
 from meltano.core.project_plugins_service import ProjectPluginsService
+from meltano.core.tracking import PluginsTrackingContext, Tracker
+from meltano.core.tracking import cli as cli_tracking
+from meltano.core.tracking import cli_context_builder
 
 from . import cli
 from .params import pass_project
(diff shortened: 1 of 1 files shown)


WillDaSilva (approved):
LGTM!

aaronsteers (approved):

pnadolny13 (approved):
@pandemicsyn this looks good to me! But I wanted to confirm what I'm seeing is expected behavior.

If I try to install a plugin that doesnt exist `meltano install extractor tap-foo-bar-baz` in my project I get the log `Installing 0 plugins...` and started/completed events + struct event. Is this expected? It seems like its not considered a failure, which is fine but wanted to double check.

If I try to install an extractor thats mis-configured i.e. pip_url is missing I get `Extractor 'tap-something-invalid' is not known to Meltano` and no events are fired at all. Would we expect a started/aborted set here?

edgarrmondragon (approved):

pandemicsyn:
> If I try to install a plugin that doesnt exist meltano install extractor tap-foo-bar-baz in my project I get the log Installing 0 plugins... and started/completed events + struct event. Is this expected? It seems like its not considered a failure, which is fine but wanted to double check.

Correct, thats not a failure. The command didn't do anything, but it didn't do anything successfully 😁 

> If I try to install an extractor thats mis-configured i.e. pip_url is missing I get Extractor 'tap-something-invalid' is not known to Meltano and no events are fired at all. Would we expect a started/aborted set here?

There's probably just a exception happening somewhere I don't know about let me poke around. We should be able to fire a started/failed pair for those.

pandemicsyn:
@pnadolny13 found it.

pnadolny13:
> @pnadolny13 found it.

@pandemicsyn I'm still not seeing any events come through when I try to `meltano install extractor tap-something-invalid` for:

```
plugins:
  extractors:
  - name: tap-something-invalid
    variant: meltanolabs
    pip_url: git+https://github.com/foo/tap-something-invalid.git
```

I acknowledge that this is a weird edge case, so I'm totally fine if we merge without fixing this.

pandemicsyn:
@pnadolny13 im gonna go ahead and merge but I'll pop an issue to follow up tomorrow if I have time. fwiw - testing it locally with your example yaml - i do get two events including 1 aborted: 
https://gist.github.com/pandemicsyn/df635c72e707b95c2c8034ec10192a62

PR #6111, opened 2022-06-08
Attach env and project contexts to structured events

### LINKED ISSUE #6057: Telemetry - Shared UUID to tie unstructured and structured events
As a follow on from https://github.com/meltano/meltano/pull/6028:

We're sending duplicate structured events along side our new unstructured events as a failsafe in case our new events have bugs. If we do end up having bugs and need to take all unstructured events and blend in structured events that are missing we need a way to tie them together. The SQL could say "give me all the unstructured events, and all the structured events where context_uuid isnt in that set of unstructured events".

The unstructured environment context has a `context_uuid` thats shared across all unstructured events in an execution session, could we append this custom context to the structured events also? Is there another way to add the `context_uuid`? 

cc @pandemicsyn @tayloramurphy @aaronsteers

pandemicsyn:
Originally, I was going to merge the LegacyTracker back into Tracker and just have a single unified tracker that just exposed a legacy method. But since the whole original intention was to have two separate paths in case we break things with the unstructured events... I ended up NOT going that route to keep them "seperated". 

This basically just steals the contexts present in Tracker() after initializations and uses those to initialize LegacyTracker(). Lastly, I also updated the `Tracker().track_struct_event` to include the ProjectContext and EnvironmentContext.

The end result is that both trackers have the same ProjectContext and EnvironmentContext. This allows @pnadolny13 to more easily tie both structured and unstructered events together into a single "cli session".

Freebie: This also means that tearing the LegacyTracker out and removing it once and for all is remains *really* easy to do.

closes #6057

### DIFF
--- a/src/meltano/core/tracking/tracker.py
+++ b/src/meltano/core/tracking/tracker.py
@@ -268,7 +268,7 @@ class Tracker:  # noqa: WPS214 - too many methods 16 > 15
                 category=category,
                 action=action,
                 label=str(self.project_id),
-                contexts=self.contexts,
+                context=self.contexts,
             )
         except Exception as err:
             logger.debug(
--- a/src/meltano/core/tracking/tracker.py
+++ b/src/meltano/core/tracking/tracker.py
@@ -10,7 +10,7 @@ import uuid
 from contextlib import contextmanager
 from enum import Enum, auto
 from pathlib import Path
-from typing import Any, NamedTuple, Optional
+from typing import Any, NamedTuple
 from urllib.parse import urlparse
 
 import tzlocal
@@ -76,7 +76,7 @@ class TelemetrySettings(NamedTuple):
 
 
 # TODO: Can we store some of this info to make future invocations faster?
-class Tracker:
+class Tracker:  # noqa: WPS214 - too many methods 16 > 15
     """Meltano tracker backed by Snowplow."""
 
     def __init__(
@@ -268,6 +268,7 @@ class Tracker:
                 category=category,
                 action=action,
                 label=str(self.project_id),
+                contexts=self.contexts,
             )
         except Exception as err:
             logger.debug(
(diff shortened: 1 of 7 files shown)


pandemicsyn:
@pnadolny13 will check double check me with `meltano invoke`.  This won't work yet for `meltano run`

Still need to update `meltano run` as well as anything else like `add/install` that is inflight.

src/meltano/core/tracking/tracker.py
```diff
@@ -268,6 +268,7 @@ def track_struct_event(self, category: str, action: str) -> None:
                 category=category,
                 action=action,
                 label=str(self.project_id),
+                context=self.contexts,
```
WillDaSilva, line 271: Instead of passing the contexts in when the legacy tracker is initialized, could we pass in the Snowplow tracker, and then have this take whatever the current contexts are?

pandemicsyn: I might be misunderstanding what you're asking, can you clarify/explain why? I could see having Tracker() accept an initialized Snowplow tracker make sense, but having Tracker() accept a snowplow tracker, to only use its contexts when firing a structured event seems odd?

Having a structured event always include its trackers context's seems like a good practice in general, its how I'd expect `Tracker().track_struct_event()` to behave, because that's also how `Tracker().track_unstruct_event()` behaves.

WillDaSilva: My apologies. I misunderstood what this change was doing. Please ignore my comment. I probably shouldn't be reviewing these PRs until I get home.

WillDaSilva (approved):

src/meltano/core/tracking/tracker.py
pandemicsyn: No worries! I'm super grateful you're around to 👀, even in passing. It's probably far too late for me to be writing quality stuff 😁 I'm definitely on the down slope after the last few days.

pnadolny13 (approved):
@pandemicsyn this looks good to me! I confirmed locally that I'm seeing the shared ID

edgarrmondragon (approved):

PR #6112, opened 2022-06-08
Additional docs for jobs

### LINKED ISSUE #6102: Improve documentation around (scheduled) jobs
Does https://docs.meltano.com/reference/command-line-interface#job put jobs in `meltano.yml`? What does that look like? It’s not documented under https://docs.meltano.com/concepts/project#meltanoyml-project-file.

https://docs.meltano.com/guide/orchestration#create-a-schedule also doesn’t mention scheduling a job, and should.

It’s also not clear to me what “Add a new job with multiple tasks by passing arrays in yaml format, where each item representing a run command.” means in https://docs.meltano.com/reference/command-line-interface#job. In a job with multiple tasks, do the tasks run in parallel when you call `meltano run <that-job>`?

cjohnhanson:
Closes #6102

### DIFF
--- a/docs/src/_guide/orchestration.md
+++ b/docs/src/_guide/orchestration.md
@@ -14,7 +14,8 @@ When a new pipeline schedule is created using the [UI](/reference/ui) or [CLI](/
 ## Create a Schedule
 
 ### Specifying extractors and loaders
-To regularly schedule your ELT to run, use the ["Pipelines" interface in the UI](/reference/ui#pipelines), or the [`meltano schedule`  command](/reference/command-line-interface#schedule):
+
+To regularly schedule your ELT to run, use the ["Pipelines" interface in the UI](/reference/ui#pipelines), or the [`meltano schedule` command](/reference/command-line-interface#schedule):
 
 ```bash
 meltano schedule add [SCHEDULE_NAME] --extractor [EXTRACTOR_NAME] --loader [TARGET_NAME] --interval [INTERVAL]
@@ -29,7 +30,9 @@ meltano schedule add carbon__sqlite --extractor tap-carbon-intensity --loader ta
 Now that you've scheduled your first pipeline, you can load the "Pipeline" page in the UI and see it show up.
 
 ### Scheduling predefined jobs
+
 If you've defined a [job](/reference/command-line-interface#job) within your project, you can schedule it directly without having to manually pass its plugins to the `meltano schedule add` command:
+
 ```
 # Define a job
 meltano job add tap-gitlab-to-target-postgres-with-dbt --tasks "[tap-gitlab target-postgres, dbt:run]"
--- a/docs/src/_guide/orchestration.md
+++ b/docs/src/_guide/orchestration.md
@@ -13,6 +13,7 @@ When a new pipeline schedule is created using the [UI](/reference/ui) or [CLI](/
 
 ## Create a Schedule
 
+### Specifying extractors and loaders
 To regularly schedule your ELT to run, use the ["Pipelines" interface in the UI](/reference/ui#pipelines), or the [`meltano schedule`  command](/reference/command-line-interface#schedule):
 
 ```bash
@@ -26,6 +27,16 @@ meltano schedule add carbon__sqlite --extractor tap-carbon-intensity --loader ta
 ```
 
 Now that you've scheduled your first pipeline, you can load the "Pipeline" page in the UI and see it show up.
+
+### Scheduling predefined jobs
+If you've defined a [job](/reference/command-line-interface#job) within your project, you can schedule it directly without having to manually pass its plugins to the `meltano schedule add` command:
+```
+# Define a job
+meltano job add tap-gitlab-to-target-postgres-with-dbt --tasks "[tap-gitlab target-postgres, dbt:run]"
+
+# Schedule the job
+meltano schedule add daily-gitlab-load --job tap-gitlab-to-target-postgres-with-dbt --interval '@daily'
+```
 
 ## Installing Airflow
 
--- a/docs/src/_reference/command-line-interface.md
+++ b/docs/src/_reference/command-line-interface.md
@@ -701,13 +701,12 @@ Note that if no environment is active, `meltano run` _does not_ generate a state
 
 ## `job`
 
-Use the `job` command to define one more related tasks. A job can contain a single task, or many tasks. Today all tasks
-are run sequentially. You can run a specified job by passing the job name as an argument to `meltano run`.
+Use the `job` command to define one or more related tasks. A job can contain a single task, or many tasks. Today all tasks
+are run sequentially. You can run a specified job by passing the job name as an argument to `meltano run`. You can also schedule jobs using `meltano schedule`.
 
 ### How to use
 
 ```bash
-
 # Add a job with a single task representing a run command
 meltano job add <job_name> --tasks "<tap_name> <mapping_name> <target_name> <command>"
 
@@ -730,13 +729,64 @@ meltano job list <job_name> --format=json
 meltano job remove <job_name>
 ```
 
+
 ### Examples
 
+##### Tasks and Blocks
+A task in a job consists of one or more _blocks_. A block is simply a set of related plugins that will be executed serially.
+For example, you can create a job that consists of a single tap and target:
+```bash
+meltano job add tap-gitlab-to-target-postgres --tasks "tap-gitlab target-postgres"
+```
+This would add the following to your `meltano.yml`:
+```yaml
+jobs:
+- name: tap-gitlab-to-target-postgres
+  tasks:
+  - tap-gitlab target-postgres
+```
+
+Plugins in the same block should be separated by a space and blocks should be separted by a comma.
+
+You can also provide tasks in YAML format. For example, the following is equivalent to the above:
+```bash
+meltano job add tap-gitlab-to-target-postgres --tasks "[tap-gitlab target-postgres]"
+```
+
+Note that any single block which includes a tap _must_ also include a target and vice versa.
+The tap must also precede the target.
+
+All of the following are invalid:
+```bash
+# A block with a tap must also include a target.
+meltano job add invalid-job --tasks "tap-gitlab"
+>>> Job 'invalid-job' has invalid tasks. block violates set requirements: Found no end in block set!
+
+# Within a block, a target must be preceded by a tap.
+meltano job add invalid-job --tasks target-postgres
+>>> Job 'invalid-job' has invalid tasks. block violates set requirements: Unknown command type or bad block sequence at index 1, starting block 'target-postgres'
+
+# Blocks are separated by commas, so Meltano considers this tap and target as two separate invalid blocks.
+meltano job add invalid-job --tasks "tap-gitlab, target-postgres"
+>>> Job 'invalid-job' has invalid tasks. Block tap-gitlab, not found
+```
+
+A job can consist of multiple blocks, which will be executed serially in the order they are provided:
+```bash
+# This job consists of two blocks:
+meltano add tap-gitlab-to-target-postgres-with-dbt --tasks "[tap-gitlab target-postgres, dbt:run]"
+
+# This job consists of one block, but runs the same three tasks in the same order as above. It is functionally equivalent to the comma-sparated approach.
+meltano add tap-gitlab-to-target-postgres-with-dbt-single-block --tasks "tap-gitlab target-postgres dbt:run"
+```
+
+##### Adding, listing, running, and removing:
 ```bash
 # Add a new job named "simple-demo" that contains two tasks
 # Task 1: tap-gitlab hide-gitlab-secrets target-mysql
 # Task 2: tap-gitlab target-csv
 meltano job add simple-demo --tasks "[tap-gitlab hide-gitlab-secrets target-mysql dbt:run, tap-gitlab target-csv]"
+
 # list the job named "simple-demo"
 meltano job list simple-demo --format=json
 # run the job named "simple-demo" using meltano run
--- a/docs/src/_reference/command-line-interface.md
+++ b/docs/src/_reference/command-line-interface.md
@@ -729,26 +729,30 @@ meltano job list <job_name> --format=json
 meltano job remove <job_name>
 ```
 
-
 ### Examples
 
 ##### Tasks and Blocks
+
 A task in a job consists of one or more _blocks_. A block is simply a set of related plugins that will be executed serially.
 For example, you can create a job that consists of a single tap and target:
+
 ```bash
 meltano job add tap-gitlab-to-target-postgres --tasks "tap-gitlab target-postgres"
 ```
+
 This would add the following to your `meltano.yml`:
+
 ```yaml
 jobs:
-- name: tap-gitlab-to-target-postgres
-  tasks:
-  - tap-gitlab target-postgres
+  - name: tap-gitlab-to-target-postgres
+    tasks:
+      - tap-gitlab target-postgres
 ```
 
 Plugins in the same block should be separated by a space and blocks should be separted by a comma.
 
 You can also provide tasks in YAML format. For example, the following is equivalent to the above:
+
 ```bash
 meltano job add tap-gitlab-to-target-postgres --tasks "[tap-gitlab target-postgres]"
 ```
@@ -757,6 +761,7 @@ Note that any single block which includes a tap _must_ also include a target and
 The tap must also precede the target.
 
 All of the following are invalid:
+
 ```bash
 # A block with a tap must also include a target.
 meltano job add invalid-job --tasks "tap-gitlab"
@@ -772,6 +777,7 @@ meltano job add invalid-job --tasks "tap-gitlab, target-postgres"
 ```
 
 A job can consist of multiple blocks, which will be executed serially in the order they are provided:
+
 ```bash
 # This job consists of two blocks:
 meltano add tap-gitlab-to-target-postgres-with-dbt --tasks "[tap-gitlab target-postgres, dbt:run]"
@@ -781,6 +787,7 @@ meltano add tap-gitlab-to-target-postgres-with-dbt-single-block --tasks "tap-git
 ```
 
 ##### Adding, listing, running, and removing:
+
 ```bash
 # Add a new job named "simple-demo" that contains two tasks
 # Task 1: tap-gitlab hide-gitlab-secrets target-mysql


docs/src/_guide/orchestration.md
```diff
@@ -13,6 +13,7 @@ When a new pipeline schedule is created using the [UI](/reference/ui) or [CLI](/
 
 ## Create a Schedule
 
+### Specifying extractors and loaders
 To regularly schedule your ELT to run, use the ["Pipelines" interface in the UI](/reference/ui#pipelines), or the [`meltano schedule`  command](/reference/command-line-interface#schedule):
```
aaronsteers, line 17: Sorry, total nit here but my markdown linter will always complain if there isn't a line between header and content.

```suggestion
### Specifying extractors and loaders

To regularly schedule your ELT to run, use the ["Pipelines" interface in the UI](/reference/ui#pipelines), or the [`meltano schedule`  command](/reference/command-line-interface#schedule):
```

What do you think of a blank line here and elsewhere as applicable?

docs/src/_reference/command-line-interface.md
```diff
@@ -730,13 +729,64 @@ meltano job list <job_name> --format=json
...
 ### Examples
 
+##### Tasks and Blocks
+A task in a job consists of one or more _blocks_. A block is simply a set of related plugins that will be executed serially.
+For example, you can create a job that consists of a single tap and target:
+```bash
+meltano job add tap-gitlab-to-target-postgres --tasks "tap-gitlab target-postgres"
+```
+This would add the following to your `meltano.yml`:
+```yaml
```
aaronsteers, line 742: Ditto comment above also for code blocks. I think each code block or paragraph should have a blank line between.

aaronsteers (approved):
This looks great to my eyes. Thanks, @cjohnhanson !

Just a small nit about blank lines surrounding headers and code blocks. Then I think this is ready to go unless others have feedback.

docs/src/_guide/orchestration.md
cjohnhanson: Good callout, I apparently had disabled running `prettier` on file save in my editor. I just ran these through `prettier` now and it added all the appropriate whitespace.

edgarrmondragon (approved):

docs/src/_guide/orchestration.md
```diff
@@ -27,6 +29,18 @@ meltano schedule add carbon__sqlite --extractor tap-carbon-intensity --loader ta
 
 Now that you've scheduled your first pipeline, you can load the "Pipeline" page in the UI and see it show up.
 
+### Scheduling predefined jobs
```
DouweM, line 32: Can we reorder these sections, since `meltano run`/jobs are now preferred over `meltano elt`?

docs/src/_reference/command-line-interface.md
```diff
@@ -732,11 +731,69 @@ meltano job remove <job_name>
...
+>>> Job 'invalid-job' has invalid tasks. Block tap-gitlab, not found
+```
+
+A job can consist of multiple blocks, which will be executed serially in the order they are provided:
+
+```bash
+# This job consists of two blocks:
+meltano add tap-gitlab-to-target-postgres-with-dbt --tasks "[tap-gitlab target-postgres, dbt:run]"
+
+# This job consists of one block, but runs the same three tasks in the same order as above. It is functionally equivalent to the comma-sparated approach.
```
DouweM, line 785: If this is the case, why explain both instead of just the simpler "all in one string" approach that matches the `meltano run` args? What's the benefit to the user of `[,]` and explicitly defining separate blocks?

docs/src/_reference/command-line-interface.md
```diff
@@ -732,11 +731,69 @@ meltano job remove <job_name>
...
+meltano add tap-gitlab-to-target-postgres-with-dbt-single-block --tasks "tap-gitlab target-postgres dbt:run"
```
DouweM, line 786: Does this actually form one "tap-gitlab target-postgres dbt:run" block as the comment suggests? I was under the impression that only "tap [transformer?] target" combinations form a block together, and our `meltano run` argument parsing logic would also interpret `"tap-gitlab target-postgres dbt:run"` as 2 blocks.

docs/src/_reference/command-line-interface.md
```diff
@@ -732,11 +731,69 @@ meltano job remove <job_name>
...
+This would add the following to your `meltano.yml`:
+
+```yaml
+jobs:
+  - name: tap-gitlab-to-target-postgres
+    tasks:
+      - tap-gitlab target-postgres
+```
+
+Plugins in the same block should be separated by a space and blocks should be separted by a comma.
```
DouweM, line 752:
```suggestion
Plugins in the same block should be separated by a space and blocks should be separated by a comma.
```

docs/src/_reference/command-line-interface.md
```diff
@@ -701,13 +701,12 @@ Note that if no environment is active, `meltano run` _does not_ generate a state
 
 ## `job`
 
-Use the `job` command to define one more related tasks. A job can contain a single task, or many tasks. Today all tasks
-are run sequentially. You can run a specified job by passing the job name as an argument to `meltano run`.
+Use the `job` command to define one or more related tasks. A job can contain a single task, or many tasks. Today all tasks
+are run sequentially. You can run a specified job by passing the job name as an argument to `meltano run`. You can also schedule jobs using `meltano schedule`.
```
DouweM, line 705:
```suggestion
are run sequentially. You can run a specified job by passing the job name as an argument to [`meltano run`](#run). You can also schedule jobs using [`meltano schedule`](#schedule).
```

docs/src/_reference/command-line-interface.md
```diff
@@ -732,11 +731,69 @@ meltano job remove <job_name>
 
 ### Examples
 
+##### Tasks and Blocks
+
+A task in a job consists of one or more _blocks_. A block is simply a set of related plugins that will be executed serially.
```
DouweM, line 736: They're plugin _commands_, not just plugins, e.g. `dbt:run`. Can we specifically refer to the `meltano run` syntax/docs since this (I think) is modeled after that?

docs/src/_guide/orchestration.md
```diff
@@ -27,6 +29,18 @@ meltano schedule add carbon__sqlite --extractor tap-carbon-intensity --loader ta
...
+
+If you've defined a [job](/reference/command-line-interface#job) within your project, you can schedule it directly without having to manually pass its plugins to the `meltano schedule add` command:
+
+```
+# Define a job
+meltano job add tap-gitlab-to-target-postgres-with-dbt --tasks "[tap-gitlab target-postgres, dbt:run]"
```
DouweM, line 38: If `--tasks "tap-gitlab target-postgres dbt:run"` would also work here (as the next doc suggests), I think we should use that as it's simpler and consistent with `meltano run`.

docs/src/_reference/command-line-interface.md
```diff
@@ -701,13 +701,12 @@ Note that if no environment is active, `meltano run` _does not_ generate a state
 
 ## `job`
 
-Use the `job` command to define one more related tasks. A job can contain a single task, or many tasks. Today all tasks
-are run sequentially. You can run a specified job by passing the job name as an argument to `meltano run`.
+Use the `job` command to define one or more related tasks. A job can contain a single task, or many tasks. Today all tasks
```
DouweM, line 704: "define one or more related tasks" makes it seem like this would add multiple task definitions to `meltano.yml`, but it adds a single job definition _consisting of_ one or more tasks that will be run sequentially, right? I think it's important to state the relation of jobs to tasks to blocks to plugin (commands) clearly, and now I'm a bit confused.

Also, I don't think "related" here helps, they're not just "related plugins", they're steps in a pipeline.

docs/src/_reference/command-line-interface.md
```diff
@@ -732,11 +731,69 @@ meltano job remove <job_name>
...
+meltano add tap-gitlab-to-target-postgres-with-dbt-single-block --tasks "tap-gitlab target-postgres dbt:run"
+```
+
+##### Adding, listing, running, and removing:
+
 ```bash
 # Add a new job named "simple-demo" that contains two tasks
 # Task 1: tap-gitlab hide-gitlab-secrets target-mysql
 # Task 2: tap-gitlab target-csv
 meltano job add simple-demo --tasks "[tap-gitlab hide-gitlab-secrets target-mysql dbt:run, tap-gitlab target-csv]"
```
DouweM, line 795: Is this equivalent to the following?

```
meltano job add simple-demo --tasks "tap-gitlab hide-gitlab-secrets target-mysql dbt:run tap-gitlab target-csv"
```

docs/src/_reference/command-line-interface.md
```diff
@@ -732,11 +731,69 @@ meltano job remove <job_name>
 
 ### Examples
 
+##### Tasks and Blocks
```
DouweM, line 734: Can we move this explanation of what the `--tasks` argument expects up into "How to use"? It's not just example code

docs/src/_reference/command-line-interface.md
```diff
@@ -732,11 +731,69 @@ meltano job remove <job_name>
...
+
+A task in a job consists of one or more _blocks_. A block is simply a set of related plugins that will be executed serially.
+For example, you can create a job that consists of a single tap and target:
+
+```bash
+meltano job add tap-gitlab-to-target-postgres --tasks "tap-gitlab target-postgres"
+```
+
+This would add the following to your `meltano.yml`:
```
DouweM, line 743: Can we please update the main `meltano.yml` documentation in https://docs.meltano.com/concepts/project#meltanoyml-project-file to cover the format of the `jobs` list as well? Since people will be finding that in their team's `meltano.yml` and want to learn what it's about without knowing `meltano job add` yet.

docs/src/_reference/command-line-interface.md
```diff
@@ -732,11 +731,69 @@ meltano job remove <job_name>
...
+
+```yaml
+jobs:
+  - name: tap-gitlab-to-target-postgres
+    tasks:
+      - tap-gitlab target-postgres
```
DouweM, line 749: Can we add a more complicated example with multiple items under `tasks` as well, here or in the `meltano.yml` docs? People will want to understand what the multi-line approach looks like.

docs/src/_reference/command-line-interface.md
DouweM: I think what's missing here is an explanation of why someone would want to use blocks, and why tap/target need to go together, but e.g. dbt or great_expectations are separate.

Note that I'm not sure they need to know about this level of detail at all, since it seems below that --tasks "tap-gitlab target-postgres dbt:run" also just works. If that's the case, can we simplify the docs by not explicitly explaining blocks, since we also don't in the meltano run docs?

docs/src/_reference/command-line-interface.md
DouweM: Also, I don't think "related" here helps, they're not just "related plugins", they're steps in a pipeline.

docs/src/_reference/command-line-interface.md
```diff
@@ -732,11 +731,69 @@ meltano job remove <job_name>
...
+```
+
+Plugins in the same block should be separated by a space and blocks should be separted by a comma.
+
+You can also provide tasks in YAML format. For example, the following is equivalent to the above:
+
+```bash
+meltano job add tap-gitlab-to-target-postgres --tasks "[tap-gitlab target-postgres]"
```
DouweM, line 757: I think it's good to introduce `[]`, but with a better example where it actually add something, like when you want to provide an array with multiple tasks separated by commas and need the brackets for it to work.

docs/src/_reference/command-line-interface.md
```diff
@@ -732,11 +731,69 @@ meltano job remove <job_name>
...
+# A block with a tap must also include a target.
+meltano job add invalid-job --tasks "tap-gitlab"
+>>> Job 'invalid-job' has invalid tasks. block violates set requirements: Found no end in block set!
+
+# Within a block, a target must be preceded by a tap.
+meltano job add invalid-job --tasks target-postgres
+>>> Job 'invalid-job' has invalid tasks. block violates set requirements: Unknown command type or bad block sequence at index 1, starting block 'target-postgres'
+
+# Blocks are separated by commas, so Meltano considers this tap and target as two separate invalid blocks.
+meltano job add invalid-job --tasks "tap-gitlab, target-postgres"
```
DouweM, line 775: Would this have worked with e.g. `--tasks "dbt:seed, dbt run"`? I'd expect not, because I think the `[]` are missing that make this an array.

docs/src/_reference/command-line-interface.md
DouweM: Above it says "Add a new job with multiple tasks by passing arrays in yaml format, where each item representing a run command.", so it seems that the comma/list format separates _tasks_, not _blocks_ as suggested here. Could that be part of why I'm confused?

"where each item representing a run command" also suggests that `--tasks "[foo bar, baz qux]"` is equivalent to `meltano run foo bar` followed by `meltano run baz qux`. It's still not clear to me why that is preferable to `--tasks "foo bar baz qux"` and `meltano run foo bar baz qux`, though.

DouweM (commented):
@cjohnhanson Thanks, it's better than what we had before so it's not a big deal that it was already merged, but I do have some questions that I want to see addressed ASAP.

I'm mostly confused about the relation between jobs, tasks, blocks, and plugin (commands), and the value to the user of providing tasks/blocks separately.

docs/src/_reference/command-line-interface.md
cjohnhanson: `--tasks "dbt:seed, dbt:run"` would be accepted as a valid job but then would fail at runtime because you're right, it wouldn't be parsed as an array. meltano would try to invoke the `seed,` (note the trailing comma) command for `dbt` and so would fail.

docs/src/_reference/command-line-interface.md
cjohnhanson: That's a good question. While these operate the same way, they _do_ result in different additions to `meltano.yml`, so I wanted to call out that they were equivalent to prevent confusion but I see how demonstrated both could cause more confusion than it prevents.

Here's the yaml that would result from `meltano job add with-commas --tasks "[tap-gitlab target-postgres, dbt:run]"` followed by `meltano job add without-commas --tasks "[tap-gitlab target-postgres dbt:run]"`:
```yaml
jobs:
- name: with-commas
  tasks:
  - tap-gitlab target-postgres
  - dbt:run
- name: without-commas
  tasks:
  - tap-gitlab target-postgres dbt:run
  ```

File: docs/src/_reference/command-line-interface.md (hunk 5 above)
Comment (cjohnhanson) [reply docs/src/_reference/command-line-interface.md:786]: Yeah I suppose I've overloaded the term "blocks" here when just sticking to "task" would mostly do the job. Under the hood, `"tap-gitlab target-postgres dbt:run"` definitely gets parsed into one `ExtractLoadBlock` for the tap+target and then one `InvokerCommand` for the `dbt:run`. But it gets written to a single array element in the `meltano.yml`, which is what I was considering synonymous with a "block" here, whereas maybe "task" is the better term for that.

File: docs/src/_reference/command-line-interface.md (hunk 12 above)
Comment (cjohnhanson) [reply docs/src/_reference/command-line-interface.md:734]: This is a great question and I'd be interested in @pandemicsyn's thoughts on this. The only benefit of splitting things out into multiple array elements/lines in `meltano.yml` that I can think of is readability and ease of editing in the `meltano.yml`. Since (unless I'm grossly misunderstanding how this works--which is definitely possible) everything is run serially and any `tap ... target` subset is going to be parsed as a block anyways.

File: docs/src/_reference/command-line-interface.md (hunk 11 above)
Comment (cjohnhanson) [reply docs/src/_reference/command-line-interface.md:795]: No it's not, that non-array space-delimited approach actually raises an error:
```bash
meltano@ip-172-31-30-47:~/meltano-projects/jobs$ meltano job add simple-demo --tasks tap-gitlab target-postgres dbt:run tap-gitlab target-csv
Usage: meltano job add [OPTIONS] JOB_NAME
Try 'meltano job add --help' for help.

Error: Got unexpected extra arguments (target-postgres dbt:run tap-gitlab target-csv)
```

This example actually clarifies the use case for arrays for me a little bit but I think potentially introduces some more confusion if we fully document it.

It looks like "tap to target blocks" (by which I mean the actual under-the-hood ExtractLoadBlock) within the same job _must_ be separated by commas and thereby split into separate tasks but plugin command blocks (i.e. PluginInvoker blocks) can be part of tasks alongside a tap to target block _or_ can stand alone as its own task, _or_ be grouped together into tasks.

pandemicsyn:
@cjohnhanson @DouweM gonna start with one giant comment rather than addressing each individual, since i think that might pull things together more easily. 

A block is our lowest discrete level of work in meltano run. `meltano run` doesn't care about singer plugins per say. It cares about things that are IO blocks (things that produce output or consume output, or both produce and consume output), or command blocks (things that don't care about IO they just do $something).  Technically, users might not need to care about our concept of "blocks", beyond the fact that they understand that:

1. `dbt:run` is a singular thing that can execute by itself
2. `tap-gitlab` is an IO block , it can't execute by itself
3. `tap-gitlab` together with `target-jsonl` from a higher level block that actually can be executed together by `run`

A single task is any **valid** meltano run command. So any blocks you would pass to meltano run. A job is composed of one or more tasks. Task's aren't super useful outside of Airflow **today**, meltano really only cares about tasks in the context of Jobs + Schedules + Airflow today. 

Creating a job definition like this:

```
meltano job add single --tasks="tap-gitlab target-postgres dbt:run tap-postgres target-salesforce"
```

Yields a yaml like this:

```yaml
# One job, with one task comprised of 3 blocks
# In airflow, one dag with ONE task.
jobs:
- name: single
  tasks:
   - tap-gitlab target-postgres dbt:run tap-postgres target-salesforce
```

When triggered in airflow that will create a DAG with one task:

```
meltano run  tap-gitlab target-postgres dbt:run tap-postgres target-salesforce
```

That's the simplest case. Given two tasks:

```
meltano job add two --tasks='["tap-gitlab target-postgres", "tap-gitlab target-salesforce"]'
```

Yields this yaml:
```
# One job, __Two__ tasks.
# In airflow, one dag with two tasks.
- name: two
  tasks:
  - tap-gitlab target-postgres
  - tap-gitlab target-salesforce
```

When triggered in airflow this will create 1 Dag, with 2 tasks:

```
#task 1
meltano run tap-gitlab target-postgres
#task 2
meltano run tap-gihub target-postgres
```

We also technically support one additional nesting level to support future features like parallel executions:

```
# One job, Two tasks.
# In airflow, one dag with two tasks.
- name: two
  tasks:
  - - tap-gitlab target-postgres
     - tap-github target-mysql
  - anothertap anothertarget
```

Today, that nested entry gets flattened into a single `meltano run` call that gets executed sequentially:

```
#task 1
meltano run tap-gitlab target-postgres tap-github target-mysql
```

In the future the intention is to have those two nested entries execute in parallel by meltano.

DouweM:
@pandemicsyn Thanks, that helps a lot. Since splitting up into tasks doesn't bring any advantages today, what do you think about not mentioning this in the docs at all for now, and only describing the "just provide what you'd provide to `meltano run` as a single argument to `--tasks`" case? Once we make tasks more valuable, we can explain "to opt into this behavior, consider splitting up your existing jobs into separate tasks".

docs/src/_reference/command-line-interface.md
pandemicsyn: 1. dbt:seed, dbt:run are independent commands. 
2. Thats still a single task, they'd be executed sequentially (assuming the first doesn't explode)

```yaml
# One task, two blocks
- name: demo
  tasks:
  - dbt:seed, dbt:run
```

```
(melty-3.8) ➜  rundev meltano run --dry-run demo
2022-06-09T00:31:28.773652Z [info     ] Setting 'console' handler log level to 'debug' for dry run
2022-06-09T00:31:28.855305Z [debug    ] expanding job to tasks         job_name=demo tasks=['dbt:seed,', 'dbt:run']
2022-06-09T00:31:29.584182Z [debug    ] Found plugin parent            parent=dbt plugin=dbt source=discovery
2022-06-09T00:31:29.592556Z [debug    ] plugin command added for execution command_name=seed, commands={0: 'seed,'} plugin_name=dbt
2022-06-09T00:31:29.592632Z [debug    ] found plugin in cli invocation plugin_name=dbt
2022-06-09T00:31:29.600597Z [debug    ] plugin command added for execution command_name=run commands={0: 'seed,', 1: 'run'} plugin_name=dbt
2022-06-09T00:31:29.600649Z [debug    ] found plugin in cli invocation plugin_name=dbt
2022-06-09T00:31:29.600718Z [debug    ] next block not extractor       offset=0 plugin_type=transformers
2022-06-09T00:31:29.600762Z [debug    ] found PluginCommand            offset=0 plugin_type=transformers
2022-06-09T00:31:29.651529Z [debug    ] next block not extractor       offset=1 plugin_type=transformers
2022-06-09T00:31:29.651617Z [debug    ] found PluginCommand            offset=1 plugin_type=transformers
2022-06-09T00:31:29.723809Z [debug    ] All ExtractLoadBlocks validated, starting execution.
2022-06-09T00:31:29.723977Z [info     ] Dry run, but would have run block 1/2. block_type=InvokerCommand comprised_of=dbt:seed,
2022-06-09T00:31:29.724066Z [info     ] Dry run, but would have run block 2/2. block_type=InvokerCommand comprised_of=dbt:run
```

docs/src/_reference/command-line-interface.md
pandemicsyn: I *think* my larger answers down below covers this. But tl;dr multiple tasks map to multiple tasks in airflow.

cjohnhanson:
@pandemicsyn 

> Given two tasks:
> 
> meltano job add two --tasks='["tap-gitlab target-postgres", "tap-gitlab target-salesforce"]'
> Yields this yaml:
> 
> # One job, __Two__ tasks.
> # In airflow, one dag with two tasks.
> - name: two
>   tasks:
>   - tap-gitlab target-postgres
>   - tap-gitlab target-salesforce
> When triggered in airflow this will create 1 Dag, with 2 tasks:
> 
> #task 1
> meltano run tap-gitlab target-postgres
> #task 2
> meltano run tap-gihub target-postgres
> 

Just to clarify, in this case task 2 would be downstream of task 1 in the generated DAG, right? 

@DouweM I think there are potentially still some benefits that users could get from splitting a job into multiple tasks:
1. It's more readable, particularly in diffs
1. It's more easily editable
1. If everything is in a single task then that means the corresponding Airflow task could only be run atomically. For users that still work directly in Airflow (particularly the UI) there might be benefit to, e.g., only retrying/rerunning individual tasks of the pipeline in Airflow. 

Do you think those benefits outweigh the potential cognitive overhead/confusion of having both approaches outlined in the docs?

docs/src/_reference/command-line-interface.md
pandemicsyn: Ah i see what your saying. Yea `dbt:seed,` is being used (with a dangling comma). That comma is being treated as a string,  and we shouldn't be documenting it as its not valid  we can try to be proactive and throw an error at job creation, but its no different than this typo:

```
meltano job add demo2 --tasks="dbt:seed- dbt:run"
Added job demo2: ['dbt:seed- dbt:run']
```

DouweM:
@pandemicsyn Have we considered automatically breaking `tap target dbt great_expectations` up into 3 tasks, over 3 lines? That way we get all the benefits @cjohnhanson points out without the user having to worry about this and manually break up that string in the appropriate bits.

docs/src/_reference/command-line-interface.md
cjohnhanson: Yeah I'm realizing my example of `meltano job add invalid-job --tasks "tap-gitlab, target-postgres"` doesn't fail because it parses into a `tap-gitlab` followed by a `target-postgres` block and so fails because `tap-gitlab` isn't valid as a block on its own. It fails because it parses into `tap-gitlab,` (note the trailing comma) followed by a `target-postgres` block and so fails because `target-postgres` isn't a valid block on its own.

docs/src/_reference/command-line-interface.md
pandemicsyn: > so it seems that the comma/list format separates tasks, not blocks as suggested here. Could that be part of why I'm confused?

Correct, blocks are not comma deliminated you run `meltano run tap target` not `meltano run tap, target` .

> It's still not clear to me why that is preferable to --tasks "foo bar baz qux" and meltano run foo bar baz qux, though.

Because in airflow this yields exactly that, two tasks:

1. `meltano run foo bar`
2. `meltano run baz qux`

pandemicsyn:
> @pandemicsyn Have we considered automatically breaking tap target dbt great_expectations up into 3 tasks, over 3 lines? That way we get all the benefits @cjohnhanson points out without the user having to worry about this and manually break up that string in the appropriate bits.

Think im gonna send up the bat signal for @aaronsteers @tayloramurphy at this point 😁  We might have ? But it wasn't any versions of the spec that I ended up building (we built multiple). I think we could switch to make that behavior as our default if y'all felt strongly, but I think it also knee-caps what @tayloramurphy and @aaronsteers had in mind.

DouweM:
@pandemicsyn We don't need to decide that right now, sorry for blowing up the scope here. I brought it up because if we can automate the breaking up into tasks / DAG steps in a future iteration, we don't need to document how users can do that manually now or at all.

aaronsteers:
In regards to this spec of allowing multiple tasks in one step, there are clear use cases where something like `dbt:compile dbt:test dbt:upload-artifacts` _needs_ to run on the same container because intermediate artifacts otherwise may not be preserved. In a DAG orchestration scenario, whether through Airflow or in any other platform, you will need to keep some steps contiguously running without a teardown/spinup in between. And otherwise, the teardown and spinup is a _good_ practice because it prevents inadvertant leakage from one step to another.

Example: `dbt:run dbt:upload-artifacts` need to be a single node. But `dbt:run and superset:build` should probably be isolated.

Second... the grain of the task also determines the grain at which steps can be restarted. So, especially when nodes are not idempotent, you'll want to have a low grain or restartability - except when the second step needs the first, or when a failure at the second step is actually more indicative of a failure at the first step . Like with dbt:run dbt:test - there's not point in continually rerunning and restarting `dbt:test` over and over again, but `dbt:run dbt:test` make sense to retry together (assuming a code push has updated the transformations).

We have users with use cases like these _today_, and not everyone is using the BashOperator - some are using K8s and ECS operators - so I don't think its accurate to say these use cases are not important.

Sure, the mainstream use case is "one execution per node", but because that is mainstream and as far as we currently want to dig ourselves does not imply our users don't immediately have those use cases.



[10:49](https://meltano.slack.com/archives/D013SUG5JJV/p1654753773357489)
Third... retry logic - when we get there, or when users today want to add this as custom to their Airflow dags. Retry logic will leverage the same principals in regards to how many retries are warranted - and which steps retry together. (edited)

Issue #6041, opened 2022-06-06
Refactor command tracking event/context to be initialized earlier


pandemicsyn:
To get things out the door in hurry, the current (although not actually merged yet as of this writing) command tracking events and context are initialized within the cli implementations of those commands. e.g.:

https://github.com/meltano/meltano/blob/762a2592c5da95fa6b1e91aa3c20d79725949a66/src/meltano/cli/invoke.py#L75-L80

But we could (should?) initialize it further up the stack when click is setting up, maybe https://github.com/meltano/meltano/blob/762a2592c5da95fa6b1e91aa3c20d79725949a66/src/meltano/cli/cli.py#L32 ? Then we could also add the global opts like `--log-level` to the tracking context. 

The initialized tracker with appropriate tracking context could then be embedded in the click context (`context` is suddenly a super loaded term for us), and down stream we could pull the tracker from the *click* context:

```python
@click.option("--some-option")
@click.pass_context
def some_cli_command(ctx, some_option):
    tracker = ctx.obj["tracker"] 
```

------

## Related work items

Noted this related issue may be solved at the same time (or not):

- https://github.com/meltano/meltano/issues/6095

If resolved, we can close the other at the same time. If not, we'll re-raise/re-prioritize that after this closes.

/cc @pnadolny13 @aaronsteers @WillDaSilva

pandemicsyn:
Cross post from https://github.com/meltano/meltano/issues/6057 , but that issue talks about tieing together both our new telemetry events and legacy tracking events under a single trace id. IMHO these two issues should be addressed at the same time, and a trace id generated and included in the click context along side already initialized tracker. 

That means that trace id could also be forwarded and configured to be used in things like structured logging and allow it to be used as an **actual** trace id.

WillDaSilva:
I think we can either hook into Click or something like the `pass_project` decorator that is currently used by many CLI commands to automatically instrument them.

With the project instance we could initialize a `Tracker`. The info needed to construct the CLI context could be obtained from Click. Once those are both available, the CLI started event can be fired. Then, once the decorated function returns, we can fire a CLI completed event if no exception was raised, and otherwise we can fire an aborted or failed event depending on what exception was returned. I'm not sure what to do with CLI skipped events - we don't currently seem to use them.

[...]

PR #6114, opened 2022-06-08
feat: Add meltano lock

### LINKED ISSUE #5961: Method to 'lock' plugins in a project and update existing lock files
So that we can fully move off of `discovery.yml`, we should add a method to lock all plugins.

> @aaronsteers from your last comment on the GitLab issue, will you create a relevant "lock all" issue? I'm not seeing one in https://github.com/meltano/meltano/issues?q=is%3Aissue+is%3Aopen+lock

Proposed spec details:

- https://github.com/meltano/meltano/issues/5961#issuecomment-1145378980

_Originally posted by @tayloramurphy in https://github.com/meltano/meltano/issues/5925#issuecomment-1141520551_

edgarrmondragon:
Closes #5961


src/meltano/cli/lock.py
```diff
@@ -0,0 +1,89 @@
...
+            plugin_type = PluginType.from_cli_argument(plugin_type)
+            plugins = plugins_service.get_plugins_of_type(plugin_type)
+            if plugin_name:
+                plugins = [plugin for plugin in plugins if plugin.name in plugin_name]
+        else:
+            # Make it a list so source preference is not lazily evaluated.
+            plugins = list(plugins_service.plugins())
+
+    tracked_plugins = []
+    tracker.track_command_event(CliEvent.started)
```
edgarrmondragon, line 68: @pandemicsyn @WillDaSilva this is my first time using the new tracking APIs. Does this look right to you?

WillDaSilva, line 68: The CLI started event should be fired right after the CLI contexts is added to the tracker.

WillDaSilva: > for each command is it expected to trigger exactly one started event, and exactly one of {completed, aborted, failed}?

Yes.

docs/src/_reference/command-line-interface.md
```diff
@@ -604,6 +604,12 @@ meltano invoke --print-var <PLUGIN_ENVIRONMENT_VARIABLE_1> <PLUGIN_NAME>
 meltano invoke --print-var <PLUGIN_ENVIRONMENT_VARIABLE_1> --print-var <PLUGIN_ENVIRONMENT_VARIABLE_2> <PLUGIN_NAME>
 ```
 
+## `lock`
+
+`meltano lock` creates lock files for all [non-custom](/concepts/plugins#custom-plugins) plugins in the project.
```
tayloramurphy, line 609: @edgarrmondragon per https://github.com/meltano/meltano/discussions/6018 and https://github.com/meltano/meltano/issues/5961#issuecomment-1146682249 we should have `--all` be the default flag so that `meltano lock` and `meltano lock --all` are equivalent.

docs/src/_reference/command-line-interface.md
```diff
@@ -604,6 +604,12 @@ meltano invoke --print-var <PLUGIN_ENVIRONMENT_VARIABLE_1> <PLUGIN_NAME>
...
+
+To update your lock artifacts with the latest plugin definitions in Meltano Hub, use the `--update` option.
```
tayloramurphy, line 611:
```suggestion
To update your lock artifacts with the latest plugin definitions in MeltanoHub, use the `--update` option.
```

In future iterations we'll want to figure out how to better version this, but I like this as a small first iteration!

tayloramurphy (changes requested):

src/meltano/cli/lock.py
edgarrmondragon: Thanks @WillDaSilva, I've moved the event up.

src/meltano/cli/lock.py
```diff
@@ -0,0 +1,94 @@
...
+                plugins = [plugin for plugin in plugins if plugin.name in plugin_name]
+        else:
+            # Make it a list so source preference is not lazily evaluated.
+            plugins = list(plugins_service.plugins())
+
+    tracked_plugins = []
+
+    for plugin in plugins:
+        descriptor = f"{plugin.type.descriptor} {plugin.name}"
+        tracked_plugins.append((plugin, None))
```
edgarrmondragon, line 77: @pnadolny13 Right now this is addding all matching plugins to the context (so either _all_, _all of a certain type_, or a _single one_). Do you think it's preferable to have this some other way? For example, only add plugins that were actually updated.

src/meltano/cli/lock.py
```diff
@@ -0,0 +1,94 @@
...
+
+    with plugins_service.use_preferred_source(DefinitionSource.HUB):
+        if plugin_type:
+            plugin_type = PluginType.from_cli_argument(plugin_type)
+            plugins = plugins_service.get_plugins_of_type(plugin_type)
+            if plugin_name:
+                plugins = [plugin for plugin in plugins if plugin.name in plugin_name]
+        else:
+            # Make it a list so source preference is not lazily evaluated.
+            plugins = list(plugins_service.plugins())
```
pandemicsyn, line 71: I know its ugly, but do you maybe wanna wrap 64-71 in a try except so we can fire a tracking event.(like when plugin or plugin type is not found)?

 `PluginType.from_cli_argument` and the `two plugins_service` calls can all raise exceptions directly or indirectly.

edgarrmondragon: @pandemicsyn Makes sense. Is `CliEvent.failed` the right event to push in the `except` block?

pandemicsyn: In this case probably `CliEvent.aborted`. I've been trying to use `CliEvent.aborted` to roughly signal when the issue was user induced (i.e. they specified a non-existent plugin) and `CliEvent.failed` for a meltano induced issue. Its imperfect though and we may just end up using one failure state in the long run.

WillDaSilva: > Its imperfect though and we may just end up using one failure state in the long run.

Related: https://github.com/meltano/meltano/issues/6041

docs/src/_reference/command-line-interface.md
edgarrmondragon: @tayloramurphy I've updated the docs to make usage options more clear: https://deploy-preview-6114--meltano.netlify.app/reference/command-line-interface#lock

So, to confirm: we want an `--all` flag that has no effect at the moment, i.e. `meltano lock --all` is exactly the same as `meltano lock`.

src/meltano/cli/lock.py
edgarrmondragon: @pandemicsyn I've added the try/except and used `CliEvent.aborted`

@WillDaSilva #6041 looks like something we need sooner rather than later.

src/meltano/cli/lock.py
edgarrmondragon: Related to

- #6201

pnadolny13: @edgarrmondragon I want to make sure I'm understanding correctly. Are the options to either:

1. send all the plugins being considered for locking i.e. if I say "lock all" I'd get every plugin even if they all already exist and no files end up changing in my project
2. or send only plugins that actually get locked i.e. if I say "lock all" I'd get ONLY plugins that ended up causing file changes in my project

I'd probably choose option 2. Option 1 is good to know but doesnt tell us much if its every plugin every time.

@tayloramurphy what do you think?

tayloramurphy: @pnadolny13 @edgarrmondragon I'd agree that option 2 would be best. In theory we could then understand the full set of a project's plugins vs what they had to separately lock, but I'm not sure that we'd a) prioritize that sort of analysis and b) make meaningful decisions with that distinction. So option 1 I think is fine if it's too cumbersome to do option 2.

edgarrmondragon: @pnadolny13 @tayloramurphy thanks for your input guys. I pushed the change so only plugins with updated locks are added to the context.

pandemicsyn (approved):

docs/src/_reference/command-line-interface.md
aaronsteers: @edgarrmondragon - Thanks for confirming - yes, that's the spec we were leaning towards, from this discussion: https://github.com/meltano/meltano/discussions/6018

I don't think this is a closed/settled issue, per se, but I felt good about what we were outlining in that discussion

I think there are two categories of commands:

1. Those that should default to noop (actually exit code 1) if no direct targets are passed. (e.g. `meltano run` as a silly example, but `meltano config set` as a more realistic one).
2. Those that should default to '--all' if no direct targets are passed.

What we think the reasonable default is of the above, may not be obvious to the user, and it could also change over time. For instance, we don't have `meltano config get --all` but in future, we could! To avoid writing a convention that changes or breaks according to whether `--all` is viable in the context, we were biasing towards accepting `--all` as an explicit way to say "yes, I really mean everything".

With the above context and framing, `meltano lock` could fail on it's own or succeed according to the decisions we make here in this PR, but `meltano lock --all` would be an intuitive way to run everything without having to remember the outcome of our decision on this point 😄 .

So, I'm actually fine with either of these behaviors:

- `meltano lock` is a noop with exit code 1
- `meltano lock` is a synonym for `meltano lock --all`

But `meltano lock my-plugin` definitely does not do `--all` and meltano lock --all my-plugin` is an illegal operation because `--all` and the named plugin are mutually exclusive.

edgarrmondragon: @aaronsteers

> But `meltano lock my-plugin` definitely does not do the "--all" behavior,

Ok, so the current interface in this PR uses `meltano lock <type>` and `meltano lock <type> <name> <name_two>`. Do we want to remove the `<type>` argument? I added it to make the command consistent with `meltano install` and friends.

> `meltano lock` is a noop with exit code 1

I like that 👍

aaronsteers: @edgarrmondragon - Sorry, yes. I think if it is not too late, I think _not_ having to declare the plugin type unless there's a naming conflict is probably the best path forward for new CLI commands.

edgarrmondragon: @aaronsteers sounds good, definitely not too late 🙂. I had forgotten that @cjohnhanson added a reference to clig.dev in the handbook:

> **Prefer flags to args.** It’s a bit more typing, but it makes it much clearer what is going on. It also makes it easier to make changes to how you accept input in the future. Sometimes when using args, it’s impossible to add new input without breaking existing behavior or creating ambiguity.

So this seems to be a good decision.

aaronsteers: @edgarrmondragon - Sorry we had to discover this after dev started - I dropped this discussion to try to more proactively and formally handle this in future developments: https://github.com/meltano/meltano/discussions/6222

aaronsteers (approved):
Looks great! Nice work and thanks for the updates!

tayloramurphy (approved):
Approving from a docs perspective.

PR #6118, opened 2022-06-09
Brute force catch all exceptions during install for telemetry

### LINKED ISSUE #6117: Verify `meltano install` telemetry fires for malformed plugin entries
In #6109 @pnadolny13 noticed that with the following entry no events where fired:

```yaml
  - name: tap-something-invalid
    variant: meltanolabs
    pip_url: git+https://github.com/foo/tap-something-invalid.git
```

I wasn't able to reproduce that at the time and did see two events (started/aborted) come across. We should double check though, its entirely possible that my local dev setup had a seperate issue that was triggering the `aborted` event.

pandemicsyn:
I shouldn't have tried to be so tactical trying to catch specific exceptions in specific places during my first pass.

This wraps the block leveraging the `plugins_service` to find plugins in a single try-except block. We throw a mix of  `ValueError` and `PluginNotFound` errors so I catch `Exception` to be safe. I suspect we may throw more than just `ValueError` and `PluginNotFound` so `Exception` seems like the safer bet.

closes #6117

### DIFF
--- a/src/meltano/cli/install.py
+++ b/src/meltano/cli/install.py
@@ -5,7 +5,6 @@ import click
 
 from meltano.core.legacy_tracking import LegacyTracker
 from meltano.core.plugin import PluginType
-from meltano.core.plugin.error import PluginNotFoundError
 from meltano.core.project_plugins_service import ProjectPluginsService
 from meltano.core.tracking import PluginsTrackingContext, Tracker
 from meltano.core.tracking import cli as cli_tracking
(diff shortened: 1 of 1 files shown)


src/meltano/cli/install.py
```diff
@@ -52,24 +51,18 @@ def install(project, plugin_type, plugin_name, clean, parallelism):
 
     plugins_service = ProjectPluginsService(project)
 
-    if plugin_type:
-        try:
+    try:
+        if plugin_type:
             plugin_type = PluginType.from_cli_argument(plugin_type)
-        except ValueError:
```
pandemicsyn, line 58: This was also a dumb bug on my part - i should have covered the `get_plugins_of_type` call with the exception out of the gate 😮‍💨

src/meltano/cli/install.py
```diff
@@ -52,24 +51,18 @@ def install(project, plugin_type, plugin_name, clean, parallelism):
...
+            plugins = plugins_service.get_plugins_of_type(plugin_type)
+            if plugin_name:
+                plugins = [plugin for plugin in plugins if plugin.name in plugin_name]
+        else:
             plugins = list(plugins_service.plugins())
-        except PluginNotFoundError:
-            tracker.track_command_event(cli_tracking.STARTED)
-            tracker.track_command_event(cli_tracking.ABORTED)
-            raise
+    except Exception:
```
edgarrmondragon, line 62: @pandemicsyn I'm not too familiar with the new tracking setup, but do we get information about which exception was raised here?

pnadolny13 (approved):
@pandemicsyn functionally this looks good to me!

src/meltano/cli/install.py
pandemicsyn: @edgarrmondragon theres also this pair of issues:

Refactor command tracking event/context to be initialized earlier - #6041
Collect process exit events with Snowplow - #5986 

Those two issue will help clean up the start/end of the tracker life cycles.

WillDaSilva (approved):

edgarrmondragon (approved):

PR #6092, opened 2022-06-08
Reference specific dbt transformer

pandemicsyn:

PR #6121, opened 2022-06-09
Address job docs issues

### LINKED ISSUE #6102: Improve documentation around (scheduled) jobs

cjohnhanson:
Starting a new PR off a clean branch for commit history reasons.

@pandemicsyn @DouweM Sorry, use this PR instead -- I know I could have fixed my other branch to get rid of the earlier merged changes but honestly this seemed a lot easier.

CC: @tayloramurphy @aaronsteers 

Closes #6102


docs/src/_concepts/project.md
```diff
@@ -251,33 +251,57 @@ Commands can specify a `container_spec` for containerized execution. To execute
...
 ### Schedules
 
 Your project's pipeline schedules,
 typically [created](/guide/orchestration#create-a-schedule)
 using [`meltano schedule`](/reference/command-line-interface#schedule),
- are defined under the `schedules` property.
+are defined under the `schedules` property.
 
-A schedule definition must have a `name`, `extractor`, `loader`, `transform` and `interval`:
+A schedule definition must have a `name`, `job` and `interval`:
```
pandemicsyn, line 279: I'm not sure if this intentional because we're steering people away from elt schedules, but a job is not required. A schedule must have a `name`, `interval`, and either `job` or "elt" definition. 

Maybe you could say something like:

"A scheduled _job_ must have..."

docs/src/_guide/orchestration.md
```diff
@@ -13,34 +13,37 @@ When a new pipeline schedule is created using the [UI](/reference/ui) or [CLI](/
...
+To regularly schedule your pipeline to run first define it as a [job](/reference/command-line-interface#job) within your project.
+Then you can schedule it directly using the `meltano schedule add` command:
 
 ```bash
-meltano schedule add [SCHEDULE_NAME] --extractor [EXTRACTOR_NAME] --loader [TARGET_NAME] --interval [INTERVAL]
-```
-
-Example:
+# Define a job
+meltano job add tap-gitlab-to-target-postgres-with-dbt --tasks "tap-gitlab target-postgres dbt:run"
```
pandemicsyn, line 23: does this need to be "dbt:run" or "dbt-prostgres:run" ? Mostly asking because I have this outstanding PR from my  Getting Started guide run through: https://github.com/meltano/meltano/pull/6092

docs/src/_concepts/project.md
cjohnhanson: Yeah this was intentional but agreed that specifying that it's for jobs is better.

docs/src/_guide/orchestration.md
```diff
@@ -13,34 +13,37 @@ When a new pipeline schedule is created using the [UI](/reference/ui) or [CLI](/
...
+meltano schedule add daily-gitlab-load --job tap-gitlab-to-target-postgres-with-dbt --interval '@daily'
 ```
 
 Now that you've scheduled your first pipeline, you can load the "Pipeline" page in the UI and see it show up.
 
-### Scheduling predefined jobs
+### Specifying extractors and loaders
 
-If you've defined a [job](/reference/command-line-interface#job) within your project, you can schedule it directly without having to manually pass its plugins to the `meltano schedule add` command:
+If you'd prefer not to define a job and just schedule a simple ELT pipeline instead you can pass an Extractor, Loader, and Transform directly via the CLI:
```
pandemicsyn, line 33: During the office hours I've heard us steer folks explicitly away from scheduling EL(T) tasks in 2.0. Should we maybe not directly mention the T portion here?

Is it worth calling out that this using `meltano elt` under the hood vs `meltano run`.  New users might not care, but more experienced users might?

/cc @aaronsteers @DouweM @tayloramurphy

docs/src/_guide/orchestration.md
```diff
@@ -13,34 +13,37 @@ When a new pipeline schedule is created using the [UI](/reference/ui) or [CLI](/
...
 
-# Schedule the job
-meltano schedule add daily-gitlab-load --job tap-gitlab-to-target-postgres-with-dbt --interval '@daily'
+Example:
+
+```bash
+meltano schedule add carbon__sqlite --extractor tap-carbon-intensity --loader target-sqlite --interval="@daily"
 ```
 
+You can also schedule pipelines using the ["Pipelines" interface in the UI](/reference/ui#pipelines).
```
pandemicsyn, line 45: We may want to note that the UI only supports scheduling legacy ELT tasks. It doesn't support jobs.

docs/src/_guide/orchestration.md
cjohnhanson: I think it would depend on which transformer you have installed in your project. I think you _can_ still use plain dbt instead of an adapter-specific version. But since in 2.0 we're also encouraging users to follow dbt best practices and use the adapter-specific version I think that yeah, changing this to `dbt-postgres` would be better, especially if that's what we're using in the getting started guide.

docs/src/_reference/command-line-interface.md
```diff
@@ -701,16 +701,18 @@ Note that if no environment is active, `meltano run` _does not_ generate a state
...
 
 ### How to use
 
 ```bash
 # Add a job with a single task representing a run command
 meltano job add <job_name> --tasks "<tap_name> <mapping_name> <target_name> <command>"
 
-# Add a new job with multiple tasks by passing arrays in yaml format, where each item representing a run command.
+# Add a new job with multiple tasks by passing arrays in yaml format, where each item represents a run command.
 meltano job add <job_name> --tasks "[<tap_name> <target_name>, <command>, <tap2_name> <target2_name>, ...]"
```
pandemicsyn, line 716: Should we illustrate how this breaks down into tasks? I think thats what we did in the cli reference and `--help` usage too. Something like:

```
# Add a job with a single task representing a single run command
meltano job add <job_name> --tasks "<tap_name> <mapping_name> <target_name> <command>"

# Add a job with multiple tasks
# task1: tap_name target_name
# task2: command
# task3: tap2_name target2_name
meltano job add <job_name> --tasks "[<tap_name> <target_name>, <command>, <tap2_name> <target2_name>, ...]"
```

docs/src/_concepts/project.md
pandemicsyn: 👍 cool makes sense in that case.

pandemicsyn:
Outside of that `dbt` vs `dbt-postgres` questions this lgtm! Would be interesting to get someone less familiar with jobs/schedules to read through it though - if there are pitfalls for a newbie here I'm not sure i'd notice them all.

pandemicsyn (approved):

docs/src/_guide/orchestration.md
```diff
@@ -13,34 +13,37 @@ When a new pipeline schedule is created using the [UI](/reference/ui) or [CLI](/
...
+# Define a job
+meltano job add tap-gitlab-to-target-postgres-with-dbt --tasks "tap-gitlab target-postgres dbt-postgres:run"
 
-```bash
-meltano schedule add carbon__sqlite --extractor tap-carbon-intensity --loader target-sqlite --interval="@daily"
+# Schedule the job
+meltano schedule add daily-gitlab-load --job tap-gitlab-to-target-postgres-with-dbt --interval '@daily'
 ```
 
 Now that you've scheduled your first pipeline, you can load the "Pipeline" page in the UI and see it show up.
```
cjohnhanson, line 29: @pandemicsyn re: https://github.com/meltano/meltano/pull/6121/files#r893034342

Since it's more of a legacy feature at this point, you think we should also just drop this line?

docs/src/_guide/orchestration.md
cjohnhanson: It's also occurring to me that we could maybe just drop this. Since this is more of a "How to Guide" rather than a comprehensive reference doc, I think it would be reasonable to only walk through the recommended approaches.

docs/src/_guide/orchestration.md
cjohnhanson: Same as above--maybe we just drop?

docs/src/_guide/orchestration.md
pandemicsyn: That's _**a really good point**_ actually. We probably could and even *should* be opinionated here right ?

docs/src/_guide/orchestration.md
pandemicsyn: I'd vote for that, not sure how other folks would feel about that though.

docs/src/_guide/orchestration.md
cjohnhanson: I went ahead and removed most references to UI and old-style EL(T) pipelines from this guide.  The `meltano.yml` and CLI reference docs both still mention the old-style EL(T) approach to running, so that material is still there for users that need to reference it. But this guide now explicitly recommends the job approach since it's the recommended and better supported approach.

CC @DouweM @aaronsteers @tayloramurphy

docs/src/_guide/orchestration.md
cjohnhanson: CC @DouweM @tayloramurphy in case y'all have opinions on dropping this.

docs/src/_reference/command-line-interface.md
```diff
@@ -729,15 +736,24 @@ meltano job list <job_name> --format=json
 meltano job remove <job_name>
 ```
 
-### Examples
+##### Tasks
+
+A task should be of the same format as arguments supplied to [the `meltano run` command](#run), which can be any valid sequence of extractors, mappers, loaders, and plugin commands.
```
tayloramurphy, line 741:
```suggestion
A task should be of the same format as arguments supplied to [the `meltano run` command](#run), which can be any valid sequence of plugins (e.g. extractors, mappers, loaders, utilities, etc.) and [plugin commands](/concepts/project#plugin-commands).
```

docs/src/_reference/command-line-interface.md
```diff
@@ -729,15 +736,24 @@ meltano job list <job_name> --format=json
...
+Note that such a sequence is only valid if it is one of:
 
-##### Tasks and Blocks
+1. An extractor followed directly by a loader. E.g. `tap-gitlab target-postgres`
+1. An extractor followed by one or more mappers and then a loader. E.g. `tap-gitlab hide-gitlab-secrets target-postgres`
+1. A plugin command invocation. E.g. `dbt-postgres:run`
```
tayloramurphy, line 746: @cjohnhanson I think it can just reference a full plugin as well if a user wants. It's just unlikely that they'll simple invoke the plugin as-is without a command / CLI args, right? @pandemicsyn can you validate?

tayloramurphy (approved):
@cjohnhanson minor comments, but looks great.

docs/src/_reference/command-line-interface.md
pandemicsyn: @tayloramurphy no today we require a command to run (would be a super easy change to remove that though, would just need to remove this check https://github.com/meltano/meltano/blob/dfd9650170ff97fd69fdff7ac8dc7dc0841bf663/src/meltano/core/block/singer.py#L97).

docs/src/_guide/orchestration.md
DouweM: We should keep existing functionality in the UI working, but don't need to add any other. So if using the UI for orchestration is no longer actually a good path to set users on, we should remove the mention.

cjohnhanson: @DouweM @alexmarple I'll go ahead and leave this mention of the UI out of the docs for now then.

docs/src/_reference/command-line-interface.md
cjohnhanson: @tayloramurphy @DouweM do we want to go ahead and allow runs as part of the scope of this issue/PR or make a follow-issue to enable?

tayloramurphy: @cjohnhanson definitely a follow-up issue! I was just making sure we documented it correctly for the current state.

docs/src/_concepts/project.md
```diff
@@ -290,13 +316,13 @@ To learn more about pipeline schedules and orchestration, refer to the [Orchestr
 
 As your project grows, and your `meltano.yml` with it, you may wish to break your config into multiple `.yml` files and to store those subfiles in various places in your Project folder hierachy.
 
-This can be done by creating new `.yml` files and adding them (directly or via a [glob pattern](https://en.wikipedia.org/wiki/Glob_(programming))) to the `include_paths` key of your `meltano.yml`:
+This can be done by creating new `.yml` files and adding them (directly or via a [glob pattern](<https://en.wikipedia.org/wiki/Glob_(programming)>)) to the `include_paths` key of your `meltano.yml`:
```
DouweM, line 319:
```suggestion
This can be done by creating new `.yml` files and adding them (directly or via a [glob pattern](https://en.wikipedia.org/wiki/Glob_(programming))) to the `include_paths` key of your `meltano.yml`:
```

docs/src/_guide/orchestration.md
```diff
@@ -5,45 +5,30 @@ layout: doc
...
-meltano schedule add carbon__sqlite --extractor tap-carbon-intensity --loader target-sqlite --interval="@daily"
-```
-
-Now that you've scheduled your first pipeline, you can load the "Pipeline" page in the UI and see it show up.
-
 ### Scheduling predefined jobs
 
-If you've defined a [job](/reference/command-line-interface#job) within your project, you can schedule it directly without having to manually pass its plugins to the `meltano schedule add` command:
+To regularly schedule your pipeline to run first define it as a [job](/reference/command-line-interface#job) within your project.
+Then you can schedule it using the `meltano schedule add` command:
```
DouweM, line 19:
```suggestion
Then you can schedule it using the [`meltano schedule add`](/reference/command-line-interface#schedule) command:
```

docs/src/_concepts/project.md
```diff
@@ -252,32 +253,57 @@ Commands can specify a `container_spec` for containerized execution. To execute
...
+jobs:
+  - name: tap-foo-to-target-bar-dbt
+    tasks:
+      - tap-foo target-bar dbt:run
+  - name: tap-foo-to-targets-bar-and-baz
+    tasks:
+      - tap-foo target-bar
+      - tap-foo target-baz
+```
+
```
DouweM, line 272:
```suggestion

You can learn more about how tasks are defined and run in the [`meltano job` documentation](/reference/command-line-interface#job).
```

DouweM (dismissed):
@cjohnhanson @pandemicsyn Thanks a lot, this looks great. Just a few small additions.

docs/src/_reference/command-line-interface.md
tayloramurphy: I made https://github.com/meltano/meltano/issues/6188 to track

Issue #2830, opened 2021-08-12
Support Python 3.10


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/2889

Originally created by @edgarrmondragon on 2021-08-12 15:46:09

---

### Problem to solve

Prepare for the eventual release of Python 3.10, to enable support as soon as possible.

### Target audience

Users of Meltano who want/need to run it on the latest Python.

### Further details

Hopefully we can release Meltano with Python 3.10 support a few days after it comes out.

### Proposal

- Add 3.10 (or a release candidate) to our testing matrix
- Make any necessary updates to the code
- Watch out for dependencies without support for 3.10 and follow their development towards eventual support.

### What does success look like, and how can we measure that?

Deprecations and other breaking changes are handled so Meltano can run on Python 3.10

### Regression test
(Ensure the feature doesn't cause any regressions)
- [ ] Write adequate test cases and submit test results
- [ ] Test results should be reviewed by a person from the team

### Links / references

MeltyBot:
[View 11 previous comments from the original issue on GitLab](https://gitlab.com/meltano/meltano/-/issues/2889)

PR #6135, opened 2022-06-09
Test python 3.10

edgarrmondragon:
I got this to a point where there's only 1 failing test for `tests/meltano/core/block/test_extract_load.py::TestExtractLoadBlocks::test_link_io`.

It seems like the test doesn't like `CliRunner`'s call to `cli.main` here: https://github.com/pallets/click/blame/9e9fe41a53d885d96e43dec7cd9eb69e352f801a/src/click/testing.py#L408.

Running the tests like

```bash
poetry run pytest \
  tests/meltano/core/test_environment_variables.py \
  tests/meltano/core/block
```

fails, but running them like

```bash
poetry run pytest \
  tests/meltano/core/block \
  tests/meltano/core/test_environment_variables.py
```

just works, so it's definitely something about `CliRunner` altering stdout/stderr streams for subsequent tests.

---

Other changes:

- Bump to [earliest](https://docs.pytest.org/en/latest/changelog.html#pytest-6-2-5-2021-08-29) `pytest` with support for Python 3.10
- Bump `pytest-asyncio` to support newer [`pytest` APIs](https://docs.pytest.org/en/stable/deprecations.html#node-construction-changed-to-node-from-parent)
- Fixed logic for getting the `site-packages` dir on 3.10 (was doing something like `"3.10"[:2]"`)


edgarrmondragon:
@WillDaSilva I see you picked up https://github.com/meltano/meltano/issues/2830. I started this a while ago but got stuck with what I described above. So, this can be a starting point if you're interested.

WillDaSilva:
> I got this to a point where there's only 1 failing test for tests/meltano/core/block/test_extract_load.py::TestExtractLoadBlocks::test_link_io.

This was caused by some tests and fixtures changing the root logging level during their execution, but not resetting it afterwards. That affects the `ExtractLoadBlocks` instances in their `_link_io` method: https://github.com/meltano/meltano/blob/de834382363782f30500b1a62cb65d4dbf0bb888/src/meltano/core/block/extract_load.py#L576

I'm closing this PR because it has been superseded by #6293, which includes the changes from this branch. Thanks for getting this started @edgarrmondragon!

PR #6138, opened 2022-06-09
chore: add sdist_public with release marker

aaronsteers:
Makes release marker code more portable

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,11 @@
 # - `make clean` deletes all the build artifacts
 # - `make docker_images` builds all the docker images including the production
 #   image
+#
+# To build and publish:
+#
+# > make sdist-public
+# > poetry publish --build
 
 ifdef DOCKER_REGISTRY
 base_image_tag = ${DOCKER_REGISTRY}/meltano/meltano/base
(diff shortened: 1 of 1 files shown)


aaronsteers:
@WillDaSilva, @edgarrmondragon, and @pandemicsyn - Can I get your eyes on this?

With this in place, the portable publish mechanism is:

```
make sdist_public
poetry publish --build
```

WillDaSilva (approved):
These changes themselves look good to me, but I don't this this release-specific step should be in the Makefile at all. I would prefer it only exist in a GitHub action yaml file as one of the steps, since in general (i.e. when the publish workflow is working) it should never be done manually.

edgarrmondragon (approved):

aaronsteers:
> These changes themselves look good to me, but I don't this this release-specific step should be in the Makefile at all. I would prefer it only exist in a GitHub action yaml file as one of the steps, since in general (i.e. when the publish workflow is working) it should never be done manually.

@WillDaSilva - Definitely I agree that we shouldn't have a manual release process at all, but in the case that CI is broken (today), I prefer to have a fallback method we trust.

aaronsteers:
@WillDaSilva and @edgarrmondragon - Thanks for reviewing.

I'll go ahead and merge and I'll use this for today.

As a manual check between `make sdist_public` and `poetry publish --build`, I think I will run `poetry build` and then check the artifacts by unzipping and confirming the presence of the UI files and the release marker file.

Does this sounds sensible to you?

Note to self:

- This content probably should get back into the handbook in case we have to repeat this.

WillDaSilva:
> Does this sounds sensible to you?

@aaronsteers Yup.

PR #6141, opened 2022-06-09
Add tracking for job command

### LINKED ISSUE #6191: Add telemetry for `meltano job` command
Update `meltano.cli.job` to support new style snowplow telemetry.

Spun out from https://github.com/orgs/meltano/discussions/392

cjohnhanson:
@pandemicsyn I don't have a ton of context here and this is mostly just following the pattern in [your config PR](https://github.com/meltano/meltano/pull/6137/files).

How have you been testing these? I'm looking into [snowplow micro](https://github.com/snowplow-incubator/snowplow-micro) but was also considering just doing some `assert_called_with()`s in our existing tests.


pandemicsyn:
> @pandemicsyn I don't have a ton of context here and this is mostly just following the pattern in [your config PR](https://github.com/meltano/meltano/pull/6137/files).

>How have you been testing these? I'm looking into [snowplow micro](https://github.com/snowplow-incubator/snowplow-micro) but was also considering just doing some assert_called_with()s in our existing tests.

Yea so far we've just been manually testing with snowplow micro. `cd src/meltano/core/tracking && the docker command from the README in that same dir` and then something like:

```
MELTANO_DISABLE_TRACKING=False MELTANO_SNOWPLOW_COLLECTOR_ENDPOINTS='["http://localhost:9090"]' meltano config tap-gitlab list 
```

There's a few issues (going back to october 😬 ) for setting up snowplow test setup, so I think @WillDaSilva was gonna try to coalesce the issues and plan into one single issue for the backlog. 

If you think you can get some basic tests working with some called_withs that would be dope.

WillDaSilva:
> There's a few issues (going back to october grimacing ) for setting up snowplow test setup, so I think @WillDaSilva was gonna try to coalesce the issues and plan into one single issue for the backlog.

https://github.com/meltano/meltano/issues/2955

src/meltano/cli/job.py
```diff
@@ -46,12 +51,14 @@ def _list_single_job(
         click.echo(
             json.dumps({"job_name": task_set.name, "tasks": task_set.tasks}, indent=2)
         )
-    tracker = LegacyTracker(project)
-    tracker.track_meltano_job("list", job_name)
+    legacy_tracker = LegacyTracker(project)
```
pandemicsyn, line 54: So `job` is a little bit different than config. Since `job` also uses the legacy tracker , you want initialize both the new style and legacy ones together:

https://github.com/meltano/meltano/blob/c41c2c153c00c0e01b8eabe5160429de6d6683b6/src/meltano/cli/install.py#L42-L43

This way you can seed the legacy tracker with the same starting contexts as the new tracker.

When you initialize a tracker `tracker = Tracker(project)` behind the scenes two contexts are immediately attached that describe the Project and Environment. Passing those to the legacy tracker by doing a `legacy_tracker = LegacyTracker(project, context_overrides=tracker.contexts) ` ensures that @pnadolny13 has shared context he can use to tie the two trackers together into a single session.

src/meltano/cli/job.py
```diff
@@ -278,8 +303,10 @@ def _validate_tasks(project: Project, task_set: TaskSets) -> bool:
             block_parser = BlockParser(logger, project, blocks)
             parsed_blocks = list(block_parser.find_blocks(0))
```
pandemicsyn, line 304: So what would be interesting here is if we attached a [PluginsContent](https://github.com/meltano/meltano/blob/main/src/meltano/core/tracking/plugins.py#L88) to tracker. The PluginsContext basically attaches info to every event that follows with a dict describing what plugins are in use.

Normally `PluginsContext()` wants a `tuple[ProjectPlugin, str]` but what you get from the block parser is a list of `blocks`. So I have:

https://github.com/meltano/meltano/blob/c41c2c153c00c0e01b8eabe5160429de6d6683b6/src/meltano/core/tracking/plugins.py#L41-L45

Which lets you get the PluginsContext for a `single` block (but here you've got a list of blocks). You could crib from that though to build a PluginsContext with all of the plugins. Something like:


```python
def the_thing(blocks: parsed_blocks: list[BlockSet | PluginCommandBlock]) -> PluginsTrackingContext
    plugins: list[(ProjectPlugin, str)] = []
    for blk in list_of_blocks:
        if isinstance(blk, BlockSet):
                for plugin_block in blk.blocks:
                    plugins.append((plugin_block.context.plugin, plugin_block.plugin_args))
            if isinstance(blk, PluginCommandBlock):
                plugins.append((blk.context.plugin, blk.command))
    return PluginsTrackingContext(plugins)

---

tracker.add_contexts(the_thing( parsed_blocks))

```

cjohnhanson: That's definitely doable. The idea being that we'd have telemetry on what plugins are being scheduled?  I have zero context on Snowplow and am just now getting local events firing to snowplow-micro to take a look at what these events actually look like, but I could definitely see having that granularity of data being useful. Do we currently get block-level/plugin-level data like that from `meltano run` or `meltano schedule` if either of those are running `job`s instead of parsing blocks directly from CLI args?

CC: @pnadolny13 in case you have thoughts on usefulness of knowing individual plugins being added to jobs

pandemicsyn: > The idea being that we'd have telemetry on what plugins are being scheduled? 

Yea exactly, so for example `meltano run` will report events for each block its executing, and as part of that its reporting what plugins are associated with that (even if you're asking it to exec a job)  

Similar behavior for invoke and config , they both attach a plugins context for the plugin thats being interacted with. 

I think for `jobs` this will be handy because it will help us figure out what sorta plugins/run commands are turning into jobs, but you're definitely right @pnadolny13 can call the shots on that and let us know whether or not to include plugin data.

pandemicsyn:
Note: https://github.com/meltano/meltano/pull/6148 has a bunch of clean up that will impact this PR.

cjohnhanson:
@pandemicsyn I took another pass at this with `PluginTrackerContext`s for the `add` and `set` subcommands and correcting to be in line with @WillDaSilva's recent changes. @WillDaSilva feel free to give this a review, but per our not-yet-formalized practices around PR review requests, I'm only formally requesting "required" reviews. Also cc @pnadolny13 if you want to triple-check what's being tracked here.

@pandemicsyn I'm curious what your experience with local testing has been using snowplow-micro. These events are definitely sending and when I look at the raw events they look correct, but they're all showing up as "bad" events with "Errors validating the event". This is also the case for other commands though, `meltano add` creates "bad" events just like `meltano  job`.

pandemicsyn (approved):
LGTM!

There was a few spots where I suggested switching to `CliEvent.aborted` for when we're positive the cause of the issue was a user error. We [might be moving away from that pattern](https://github.com/meltano/meltano/issues/6041#issuecomment-1156987341) shortly anyway but just incase it sticks around i dropped inline suggestions for them.

pandemicsyn:
Just quick mention, i did pull this PR down and verified I didn't see any bad events in a freshly booted instance of `snowplow-micro`.

WillDaSilva (approved):

pnadolny13:
@cjohnhanson I'm not seeing any bad events which is good but I'm curious what the spec is on this.

I ran `meltano job add foo --tasks "tap-csv target-jsonl"` and I got 6 events back, some of these questions might be unrelated to this PR but I'm throwing it all out there 😄 :

1. CLI started (unstruct)
1. Struct `meltano job add --environment=ef260e9aa3c673af240d17a2660480361a8e081d1ffeca2a5ed0e3219fc18567`
  - As part of this struct event we get environment/project/cli/ and 2 exception contexts. Is this expected? I know we added context to the struct events but I didnt know we added everything. Also the 2 different exception contexts seems weird, right?
1. Struct `meltano job foo add  --environment=ef260e9aa3c673af240d17a2660480361a8e081d1ffeca2a5ed0e3219fc18567`
  - Should we be getting multiple events for a single command? I'm not positive but I think we usually only send 1, I could be wrong though (thinking of `meltano add` specifically).
  - Also has all the same contexts including 2 exception contexts.
1. CLI completed
1. Exit Event 1
1. Exit Event 2

A couple of follow up questions:

- I'm curious about what the strategy is on adding new struct events, originally we were just keeping the existing ones firing as a fail safe but I believe this is adding a new struct event. Should we be adding more new struct events? I'd probably say no. cc @tayloramurphy what do you think?
- If we do want to add new legacy struct events, should we be firing 2 in this case or just 1?
- Are the double exit event a bug?
- Are the double exception contexts a bug?

tayloramurphy:
> Should we be adding more new struct events?

@pnadolny13 struct events are generally too limiting and we don't need to be adding new ones.

src/meltano/cli/job.py
```diff
@@ -111,6 +120,13 @@ def job(project, ctx):
     """
     ctx.obj["project"] = project
     ctx.obj["task_sets_service"] = TaskSetsService(project)
+    tracker = Tracker(project)
+    tracker.add_contexts(CliContext("job", ctx.invoked_subcommand or None))
+    legacy_tracker = LegacyTracker(project, context_overrides=tracker.contexts)
+    tracker.track_command_event(CliEvent.started)
+    legacy_tracker.track_meltano_job(ctx.invoked_subcommand or None)
```
pandemicsyn, line 127: Ah, i think this is the reason we're getting dupe structured events.  Sorry I missed this during the review. We're already firing this same event inline in all the subcommands. 

```suggestion
```

pnadolny13:
>Job's has had structured events [since Jobs was added](https://gitlab.com/meltano/meltano/-/merge_requests/2629). e.g. current master:

@pandemicsyn oh my bad I didnt realize jobs had a legacy event already, I thought this was the first time it was getting added.

pandemicsyn:
@pnadolny13 @cjohnhanson the double exit issue is unrelated to this pr - looks like @WillDaSilva resolved that in  https://github.com/meltano/meltano/pull/6203. @pnadolny13 if you're still see'ing double exit events with the last `main` than we probably need to reopen that issue for @WillDaSilva

pnadolny13:
> @pnadolny13 @cjohnhanson the double exit issue is unrelated to this pr - looks like @WillDaSilva resolved that in #6203. @pnadolny13 if you're still see'ing double exit events with the last `main` than we probably need to reopen that issue for @WillDaSilva

@pandemicsyn ah ok so those should be fixed once this branch gets merged into main where the bug is fixed?

pandemicsyn:
@pnadolny13 I just asked github to merge in main. I think https://github.com/meltano/meltano/pull/6203 landed a few hours after this PR last updated main.

cjohnhanson:
@pnadolny13 you mind taking another look? Now that I've committed @pandemicsyn 's suggestion I think that should have resolved the dupe structured events.

pnadolny13:
@cjohnhanson yep looks good to me!

pnadolny13 (approved):

PR #6142, opened 2022-06-09
Add CLI exit event

### LINKED ISSUE #5986: Collect process exit events with Snowplow

WillDaSilva:
I've tested this manually using Snowplow Micro. We could add automated testing for it by checking if Meltano subprocesses that exit with various exit codes behave properly. We'll probably want to `resolve` #2955 prior to adding such tests, as otherwise actually verifying the behaviour will be tricky, regardless of whether it works or not.

(`resolve` is in backticks to stop GitHub from automatically closing #2955)

Closes #5986


src/meltano/cli/__init__.py
```diff
@@ -60,3 +58,23 @@ def main():
     except CliError as cli_error:
         cli_error.print()
         sys.exit(1)
+
+
+exit_code = None
```
aaronsteers, line 63: What do you think of moving to the top of the file, adding a comment to explain what this is and how it will be used, and then also adding a type hint.

Something like:

```py
# Hold the exit code value for error reporting during termination/cleanup:
exit_code: None | int = None
```

src/meltano/core/tracking/tracker.py
```diff
@@ -427,3 +430,25 @@ def track_block_event(self, block_type: str, event: BlockEvents) -> None:
...
+
+        start_time = datetime.utcfromtimestamp(Process().create_time())
+
+        # This is the reported "end time" for this process, though in reality the process will end
+        # a short time after this time as it takes time to emit the event.
+        now = datetime.utcnow()
+
+        self.track_unstruct_event(
+            SelfDescribingJson(
+                "iglu:com.meltano/exit_event/jsonschema/1-0-0",
```
aaronsteers, line 445: Similar to other examples at top of this file, I think breaking this into two constants is a good idea - one for the schema path and one for the version number.

aaronsteers (approved):
A couple small changes requested.

If no other major changes, you can merge after those (and after your own testing).

If other large factors are changed/added, please let me know and I'll do another review round.

src/meltano/core/tracking/tracker.py
WillDaSilva: @aaronsteers How do you feel about me centralizing all of these schema-version pairs of constants into a dictionary in `src/meltano/core/tracking/schema.py` that maps from schema name to schema version with a function to get the full URL. That way we won't end up with any duplicates or version mismatches.

aaronsteers: I'm not sure. Do you mean, we'd have the constants for the version string, but the lookup to version would be by passing in the constant?

Honestly, I think might be better to safe for another iteration so we can discuss more.

WillDaSilva: I'll make a follow-up PR to illustrate my suggestion.

aaronsteers (approved):
Looks great! Thanks!

src/meltano/core/tracking/tracker.py
WillDaSilva: https://github.com/meltano/meltano/pull/6146

WillDaSilva:
Apologies for merging #6146 into this. I forgot that it wasn't targeting main.

aaronsteers:
> Apologies for merging #6146 into this. I forgot that it wasn't targeting main.

@WillDaSilva - not a problem. It happens. 

I like what you updated there so it shouldn't be an issue.

PR #6145, opened 2022-06-09
ci: Build Python artifacts

edgarrmondragon:


build.py
```diff
@@ -0,0 +1,31 @@
+import os
```
edgarrmondragon, line 1: @aaronsteers This is called during `poetry build` or `poetry build -f wheel`.

It's also called during `poetry install`, which means we get an updated UI bundle for free during local development.

edgarrmondragon:
Artifacts can be downloaded:

Screen Shot 2022-06-09 at 19 05 54

build.py
aaronsteers: @edgarrmondragon - Amazing! So cool!

aaronsteers: I checked the build artifacts and the wheels look perfect. I don't see those additional files in the sdist bundle though. Is that expected?

edgarrmondragon: TL;DR;

We could build the sdist after the wheel and forcefully shim the static files into the `.tar.gz` since those files are platform-agnostic anyway.

---

Yeah, only the wheels are expected to include "compiled" artifacts. The .tar.gz is the source distribution (sdist), that for example you'd clone and build from source.

That's why Poetry explicitly runs the build script only when building a wheel.

When pip or Poetry fails to find a wheel for your system, it downloads the source distribution and executes `setup.py`, which _usually_ builds any C dependencies, etc. But, because the static files that we want to ship with Meltano aren't built to be system-specific, they can be safely included.

If we don't, we probably want to document that installing from the sdist requires an additional step to build the static assets.

aaronsteers: @edgarrmondragon - Since `sdist` resources in the past did include all the resources, I think we've got to reach parity here also for sdist before I'd feel comfortable to change the process.

Went sleuthing around, and found this: https://github.com/python-poetry/poetry/issues/2740#issuecomment-860744196

> @Gobot1234, it seems it behaves differently for each distribution format: for sdist [the build method gets called](https://github.com/python-poetry/poetry-core/blob/58138c580210c473190f0513d6f1f6f0a86f6255/poetry/core/masonry/builders/sdist.py#L123-128) and for wheel [the module gets called](https://github.com/python-poetry/poetry-core/blob/58138c580210c473190f0513d6f1f6f0a86f6255/poetry/core/masonry/builders/wheel.py#L189-191). Also, for sdist, the file is only taken into account if generate-setup-file is True ([here is the call](https://github.com/python-poetry/poetry-core/blob/58138c580210c473190f0513d6f1f6f0a86f6255/poetry/core/masonry/builders/sdist.py#L95-96)), and on the other hand, while building wheel, the module is run [when generate-setup-file is False](https://github.com/python-poetry/poetry-core/blob/58138c580210c473190f0513d6f1f6f0a86f6255/poetry/core/masonry/builders/wheel.py#L122-31) but the build method gets called when generate-setup-file is True ([as SdistBuilder is used](https://github.com/python-poetry/poetry-core/blob/58138c580210c473190f0513d6f1f6f0a86f6255/poetry/core/masonry/builders/wheel.py#L134-135)).

A bit dated but wanted to share here in case helpful.

edgarrmondragon: @aaronsteers Status quo is exactly as the comment describes, so it seems that given our use case (including build artifacts in the `sdist`) we probably want to avoid using a build script with Poetry and just rely on a GitHub step. Wdyt?

edgarrmondragon: @aaronsteers 

> Sorry it was a dead-end but perhaps we can revise in future if we find a workaround

No worries.

> What do you think of keeping the other progress you've made here for the CI work and set the script to keep the `make` step immediately prior

So that means:

1. We get rid of the matrix (we're building a single multi-version pure-Python wheel)
2. We add a step that runs `make sdist_public`, before `poetry build`

Does that sound right?

> I think `workflow_dispatch` is probably the best trigger so this is executed manually.

Yeah, my original plan was to only trigger the very last step if the trigger was a GitHub release but a workflow dispatch is ok too.

.github/workflows/build.yml
```diff
@@ -0,0 +1,101 @@
...
+        overwrite: true
+        file_glob: true
+
+    - name: Publish
+      if: ${{ github.event.release }}
+      uses: pypa/gh-action-pypi-publish@v1.5.0
+      with:
+        user: __token__
+        password: ${{ secrets.PYPI_TOKEN }}
+        ## Test PyPI can be used here
```
aaronsteers, line 99: I would feel slightly more comfortable if we name the directory explicitly. Even though it is not required, it feels safer to be explicit here. What do you think of this?

```suggestion
        password: ${{ secrets.PYPI_TOKEN }}
        packages_dir: dist/
        ## Test PyPI can be used here
```

aaronsteers (approved):
Approved, with one suggestion (optional).

PR #6146, opened 2022-06-10
Centralize schema constants

WillDaSilva:


tests/meltano/core/tracking/test_plugins.py
```diff
@@ -38,10 +34,7 @@ def test_plugins_tracking_context_from_block(
     def test_plugins_tracking_context(self, tap: ProjectPlugin, dbt: ProjectPlugin):
 
         plugin_ctx = PluginsTrackingContext([(tap, None), (dbt, "test")])
-        assert (
-            plugin_ctx.schema
-            == f"{PLUGINS_CONTEXT_SCHEMA}/{PLUGINS_CONTEXT_SCHEMA_VERSION}"
-        )
+        assert plugin_ctx.schema == get_schema_url("plugins_context")
```
aaronsteers, line 37: I like this direction but ideally we would also remove the need for a string value here.

Would want to avoid having to remember (and possibly get wrong) the distinction between "plugins_context" or "plugin_context" or "plugins_contexts".

Two ways I can think of to address:

1. A constant `TELEMETRY_CONTEXT_PLUGINS` or `TELEMETRY_PLUGINS_CONTEXT` or similar.
2. A class name or object name.

Thanks for creating this by the way. Probably I won't get a chance to review this weekend, but I appreciate you thinking about this. 🙏

aaronsteers (approved):

tests/meltano/core/tracking/test_plugins.py
WillDaSilva: @aaronsteers Thanks for the review. I think I've addressed the problem with [e5bdb59](https://github.com/meltano/meltano/pull/6146/commits/e5bdb59c7d09ad44fbe6244cc35e7f95e9875ad3).

edgarrmondragon (approved):
nice!

aaronsteers (commented):
@WillDaSilva - Retroactively approving. 😄  

I really like what you did here. 👍 ✅

PR #6148, opened 2022-06-10
Move contexts into submodule & clean-up imports

WillDaSilva:
As we add more Snowplow contexts and more modules to build them, moving them all into a submodule should help keep things well-organized.

I've also done some misc other clean-up work, such as using a Python `Enum` to represent the `event` enum from the `cli_event`, and ensuring that public members of the tracking module and its submodules are being exposed in `__init__.py` so that the users of the module don't need to import from multiple levels down inside of it.

Nothing really important in this PR, but I want to get these changes into their own PR so that they don't clutter the diff for the environment context PR that I will open soon.


pandemicsyn (approved):
👍 lgmt!

edgarrmondragon (approved):

PR #6149, opened 2022-06-10
Add exception context

WillDaSilva:
This PR is a draft until the tests are passing, and additional tests have been added (e.g. to test that `ExceptionContext` handles nested exceptions and deep tracebacks properly).

The schema should be reviewed with special attention paid towards keeping the data anonymous. Currently we hash the `str` and `repr` of the exception, so we won't be getting e.g. error message. We do collect the unhashed paths of the Python source files across the traceback, but only if it can be made relative to the installation location of Python. 

For example:

- `/home/will/.pyenv/versions/meltano37/lib/python3.7/site-packages/meltano/core/utils/__init__.py` becomes `lib/python3.7/site-packages/meltano/core/utils/__init__.py`
- `/home/will/.pyenv/versions/meltano37/lib/python3.7/pathlib.py` becomes `lib/python3.7/pathlib.py`

Closes #6126


src/meltano/core/tracking/contexts/exception.py
```diff
@@ -0,0 +1,126 @@
...
+
+    path = Path(str_path)
+
+    for base_path in BASE_PATHS:
+        try:
+            return path.relative_to(base_path).as_posix()
+        except ValueError:  # Path could not be made relative to `base_path`
+            pass  # Try making it relative to the next base path
+
+    return None
```
aaronsteers, line 110: How about this as a fallback?

```suggestion
    for base_path in BASE_PATHS:
        try:
            return path.relative_to(base_path).as_posix()
        except ValueError:  # Path could not be made relative to `base_path`
            pass  # Try making it relative to the next base path

    # If path can't be safely scrubbed, return just the .py filename
    return f".../{path.basename()".
```

With this as a fallback, I think we can also remove the path hash.

https://docs.python.org/3/library/os.path.html#os.path.basename

src/meltano/core/tracking/iglu-client-embedded/schemas/com.meltano/exception_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,119 @@
...
+                                "maxLength": 1024
+                            },
+                            "path_hash": {
+                                "description": "64 character hash of the absolute path of the Python source file for this level of traceback",
+                                "type": [
+                                    "string",
+                                    "null"
+                                ],
+                                "maxLength": 64
+                            },
```
aaronsteers, line 64: I think I would rather not collect this, personally.

Especially with the proposed basename fallback (`.../filename.py`), what do you think of removing `path_hash`?

src/meltano/core/tracking/iglu-client-embedded/schemas/com.meltano/exception_context/jsonschema/1-0-0
```diff
@@ -0,0 +1,119 @@
...
+                    "description": "Traceback info from the exception",
+                    "type": [
+                        "array",
+                        "null"
+                    ],
+                    "items": {
+                        "type": "object",
+                        "properties": {
+                            "path": {
+                                "description": "The path of the Python source file for this level of the traceback, relative to the location Python was installed; null if the path could not be made relative",
```
aaronsteers, line 50: What do you think of renaming this to `file`?

Of course, we'll likely have `path` components if we're able to make them relative, but the name `file`, I think, shows priority to getting the codefile name.

```suggestion
                            "file": {
                                "description": "The path of the Python source file for this level of the traceback, relative to the location Python was installed; or the truncated format '.../filename.py' if the path could not be made relative",
```

src/meltano/core/tracking/contexts/exception.py
```diff
@@ -0,0 +1,126 @@
...
+        tb: The traceback from which to extract the path info.
+
+    Returns:
+        The hashed absolute path from the traceback, or `None` if the traceback is missing that
+        info.
+    """
+    try:
+        return hash_sha256(tb.tb_frame.f_code.co_filename)
+    except Exception:
+        return None
```
aaronsteers, line 126: Per my other comment, I think we can remove this.

aaronsteers (commented):
@WillDaSilva - Thanks again for your hard work on this! 

I have basically just a few requested changes and then from my perspective, this should be ready to go.

Do you mind if we wait until Monday to get an extra pair of eyes from @edgarrmondragon and/or @pandemicsyn ?

I'm _so excited_ about the brand new world of support and UX data analysis this will open up for us!

WillDaSilva:
Thanks for the review @aaronsteers! I agree with your suggestions, and have implemented them.

> Do you mind if we wait until Monday to get an extra pair of eyes from @edgarrmondragon and/or @pandemicsyn ?

I don't mind at all. I would greatly appreciate having an extra pair of eyes go over this.

WillDaSilva:
> nice, so in cases where we abort/fail due to an exception - as long as we fire an event (as we do in https://github.com/meltano/meltano/blob/main/src/meltano/cli/install.py#L65), we'll have the exception context attached along side any other contexts (env, cli, plugins, etc) that may have been added, correct ?

@pandemicsyn Correct. The exception context is automatically added to event event. If an event occurs while an error is being handled (i.e. within some `except` block) then we'll gain info about the exception.

Unfortunately the stack is unwound before the atexit handlers are run, so if an exit occurs within an exception block, we'll get an exit event, but the exception context it comes with will not have details about the exception that originally triggered the exit.

To get that info, we'll probably want to fire an event within [the finally block in `meltano.cli`](https://github.com/meltano/meltano/blob/3b404ac0575e5f6a5e48807b2aa5b6a4eae16bfa/src/meltano/cli/__init__.py#L75). This can be the exit event. We'll add a global flag for whether the exit event has been fired. If we fire the exit even within `meltano.cli`, then it will not be fired from the atexit handler later. This way we can still leverage atexit to ensure the exit event is fired in all possible cases, but also get more info about what caused the exit whenever possible.

edgarrmondragon (commented):
@WillDaSilva This looks good to me. Should I hold approval for after you add handling for all exceptions in `meltano.cli::main` or is that out of scope?

WillDaSilva:
Thanks @edgarrmondragon! I'll do that in a separate PR, and create an issue to track it now.

edgarrmondragon (approved):

WillDaSilva:
@edgarrmondragon @pandemicsyn This PR is targeting the branch from https://github.com/meltano/meltano/pull/6148. We'll want to either merge that PR in first, or if we decide against the changes it proposes, rebase this PR onto `main`.

pandemicsyn (approved):

PR #6155, opened 2022-06-10
ci: add `pytest` support on windows; change: add `meltano elt` guidance for windows users

### LINKED ISSUE #5940: Meltano ELT fails when running on native windows
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/2743

Originally created by @visch on 2021-05-05 18:18:00

---

<Issue description too long to migrate - see original GitLab issue>

visch:
Closes https://github.com/meltano/meltano/issues/5940

Old MR from Gitlab https://gitlab.com/meltano/meltano/-/merge_requests/2639/diffs

(Note this has been rebased since this MR)

Pipeline is running here https://gitlab.com/vischous/meltano/-/merge_requests/3/pipelines


tests/fixtures/core.py
```diff
@@ -471,7 +471,7 @@ def project(test_dir, project_init_service):
     # clean-up
     Project.deactivate()
     os.chdir(test_dir)
-    shutil.rmtree(project.root)
+    # shutil.rmtree(project.root)
```
visch, line 474: Should we do a full cleanup of all the pytest temp directories at the end of testing instead?

Also should we remove all the comments?

visch:
Still a few errors here https://gitlab.com/vischous/meltano/-/jobs/2576910300

.github/workflows/test.yml
```diff
@@ -11,10 +11,8 @@ jobs:
   tests:
     strategy:
       matrix:
-        include:
-        - { python-version: "3.7", os: "ubuntu-latest" }
-        - { python-version: "3.8", os: "ubuntu-latest" }
-        - { python-version: "3.9", os: "ubuntu-latest" }
+        python-version: ["3.7", "3.8", "3.9"]
```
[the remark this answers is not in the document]
edgarrmondragon: @visch This notation is easier to understand and it's ok as long as the combination is a cartesian product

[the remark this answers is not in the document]
visch: It does the cartesian product! :D

.gitlab/ci/global.gitlab-ci.yml
```diff
@@ -29,3 +29,8 @@
   parallel:
     matrix:
       - IMAGE_TAG_SUFFIX: ["", "-python3.7", "-python3.8", "-python3.9"]
+
+.windows:defaults:
```
visch, line 33: Not removing this only because of https://github.com/meltano/meltano/pull/5967/files

visch:
> Still a few errors here https://gitlab.com/vischous/meltano/-/jobs/2576910300

FAILED tests/meltano/core/test_plugin_install_service.py::TestPluginInstallService::test_install_all
FAILED tests/meltano/core/test_state_service.py::TestStateService::test_set_state

These work properly on my Windows machine, just not on Windows CI at gitlab for some reason.  

They are passing https://github.com/meltano/meltano/pull/6155/checks though? Which is curious, maybe that's enough to ignore them?

I'm removed the Gitlab Windows Checks and we'll just rely on the Windows Github checks.

tests/meltano/core/tracking/test_exception.py
```diff
@@ -97,6 +99,10 @@ def test_simple_exception_context():
         assert key in tb_data[0]
 
 
+@pytest.mark.skipif(
```
visch, line 102: This is pretty important as it's an example of a new Test was was written that doesn't pass on Windows. I went ahead and ignored this test. We'll probably want to at a minimum add to a guide that if a pass doesn't test on Windows you'll want to "ignore" the test, although that sounds terrible at least it won't impact development velocity as badly.

Punting these questions to https://github.com/meltano/meltano/issues/3444 is ok, but it does mean that we are currently telling everyone to just ignore all the Windows failures until we have a better solution here https://github.com/meltano/meltano/issues/3444 which will have to be some form of either always ignore the error, fix the error, or bring someone in to help fix the error.

visch:
@edgarrmondragon  actually ready for a review now, not a fake out this time!

.github/workflows/test.yml
```diff
@@ -52,12 +50,18 @@ jobs:
...
         poetry run coverage run --parallel -m pytest -m "$PYTEST_MARKERS"
+    
+    - name: Run pytest windows
+      if: (matrix.os == 'windows-latest')
+      run: |
+        poetry run coverage run --parallel -m pytest -m $Env:PYTEST_MARKERS
 
     - name: Upload coverage data
-      if: always() && (matrix.python-version == '3.9')
+      if: always() && (matrix.python-version == '3.9') && (matrix.os == 'ubuntu-latest')
```
edgarrmondragon, line 64: @aaronsteers any reason we don't want to combine coverage from multiple Python versions? If not, I guess there's also no reason not to combine coverage from multiple OSs. Wdyt?

> Often test suites are run under different conditions, for example, with different versions of Python, or dependencies, or on different operating systems. In these cases, you can collect coverage data for each test run, and then combine all the separate data files into one combined file for reporting.

https://coverage.readthedocs.io/en/latest/cmd.html#cmd-combine

tests/fixtures/cli.py
```diff
@@ -39,5 +39,5 @@ def project_files_cli(test_dir, compatible_copy_tree):
     # clean-up
     Project.deactivate()
     os.chdir(test_dir)
-    shutil.rmtree(project.root)
+    # shutil.rmtree(project.root)
```
edgarrmondragon, line 42: @visch do you know if these files would linger indefinitely in a dev's system when running tests locally?

.github/workflows/test.yml
```diff
@@ -52,12 +50,18 @@ jobs:
...
-    - name: Run pytest
+    - name: Run pytest linux
+      if: (matrix.os != 'windows-latest')
       run: |
         poetry run coverage run --parallel -m pytest -m "$PYTEST_MARKERS"
+    
+    - name: Run pytest windows
+      if: (matrix.os == 'windows-latest')
+      run: |
+        poetry run coverage run --parallel -m pytest -m $Env:PYTEST_MARKERS
```
edgarrmondragon, line 61: @visch Since the pytest markers string doesn't really need to be an env var, wdyt of adding `pytest_markers: ["not concurrent"]` to the matrix and using a single step with

```yaml
    - name: Run pytest
      run: |
        poetry run coverage run --parallel -m pytest -m "${{ matrix.pytest_markers }}"
```

edgarrmondragon (changes requested):

.github/workflows/test.yml
visch: Fixed, did it with Env variables looks like this is how do you "expansion" with github actions.

edgarrmondragon: @visch Ah, even better. Thanks!

visch:
From the latest Postgres add to github actions failure

```
https://github.community/t/how-to-run-services-on-windows-or-macos/16565
```

We could just roll with gitlab stuff for now, but I"d guess we want to spend the time to fix this. 

My gut "want" is some kind of inheritance  where the Windows tests don't use Postgres but can (somehow) still run the same Steps as the current github actions. Not sure how exactly but there may be an option. Worst case we make a `test-windows.yml`

visch:
> From the latest Postgres add to github actions failure
> 
> ```
> https://github.community/t/how-to-run-services-on-windows-or-macos/16565
> ```
> 
> We could just roll with gitlab stuff for now, but I"d guess we want to spend the time to fix this.
> 
> My gut "want" is some kind of inheritance where the Windows tests don't use Postgres but can (somehow) still run the same Steps as the current github actions. Not sure how exactly but there may be an option. Worst case we make a `test-windows.yml`

@edgarrmondragon  I pushed up a change removing Windows from the test with PostGres and adding a new `test-nopostgres.yml` file that works without it.  I know this is a bad solution (But it works), what do you think about installing the Postgres container manually (Not using a service, using a runner instead) so that we can workaround not having conditional services https://github.community/t/conditional-services-in-a-job/135301

.github/workflows/test-nopostgres.yml
```diff
@@ -0,0 +1,107 @@
...
+        poetry run coverage run --parallel -m pytest -m "${{ env.PYTEST_MARKERS }}"
+
+    - name: Upload coverage data
+      if: always() && (matrix.python-version == '3.9') && (matrix.os == 'ubuntu-latest')
+      uses: actions/upload-artifact@v3.1.0
+      with:
+        name: coverage-data
+        path: ".coverage.*"
+
+  coverage:
```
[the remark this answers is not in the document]
edgarrmondragon: @visch see https://github.com/meltano/meltano/pull/6268

visch:
> Windows Postgres failures
> 
> ```
> Run docker run -d -p "5432:5432" -e "POSTGRES_PASSWORD=postgres" --name postgres --health-cmd "pg_isready -d postgres -U postgres" --health-interval [1](https://github.com/meltano/meltano/runs/7042154323?check_suite_focus=true#step:7:1)0s --health-timeout 5s --health-retries 5 postgres:11
> Unable to find image 'postgres:11' locally
> 11: Pulling from library/postgres
> docker: no matching manifest for windows/amd64 10.0.2034[8](https://github.com/meltano/meltano/runs/7042154323?check_suite_focus=true#step:7:9) in the manifest list entries.
> See 'docker run --help'.
> Error: Process completed with exit code 1.
> ```
> 
> Few ways to handles this
> 
> 1. Ignore installing postgres on Windows, add if's in the GH actions
> 2. Barring some postgres image that works on Windows we could use WSL - [Add Windows Subsystem for Linux to Windows actions/virtual-environments#50](https://github.com/actions/virtual-environments/issues/50) which would mean:
>    
>    1. Install WSL
>    2. Install Docker on WSL2 (🤞 that this would work as it runs on HyperV)
>    3. Install Postgres via Docker on WSL2
>    4. Setup networking so that you can route to postgres, on my Windows box I do static IP mapping to the interface HyperV uses but there may be an easier way here
>    5. Write up some dev docs on how to use Docker Containers and interact with them via CI for folks that want to add mssql, mysql, and others ([Add broader system db support #6167](https://github.com/meltano/meltano/pull/6167))
> 3. Install postgres directly on windows (works pretty well and is simple, you have to worry a bit about dependencies on the local box impacting pytest in some unexpected way)
> 4. ~https://stackoverflow.com/questions/48066994/docker-no-matching-manifest-for-windows-amd64-in-the-manifest-list-entries - going to try this first.~ Nope :( and this might crush our wsl2 ideas) [Add Windows Agent with Docker in Linux container mode actions/virtual-environments#252 (comment)](https://github.com/actions/virtual-environments/issues/252#issuecomment-610370237)

Submitted https://github.com/meltano/meltano/issues/6281 instead of trying to fix this here.

tests/meltano/core/tracking/test_exception.py
visch: After chatting with @aaronsteers  it sounds like this is fine. I'll leave this open and let @aaronsteers  resolve just in case :D

WillDaSilva: I believe I've fixed the Windows incompatibility in my branch for Python 3.10 support. I expect this PR will be merged first, so it'll be easy for me to verify that.

WillDaSilva (changes requested):
The tests that fail on Windows, but shouldn't, should use `pytest.xfail` instead of `pytest.skipif`. The Pytest docs for xfail have an example for conditionally xfailing on Windows: https://docs.pytest.org/en/7.1.x/how-to/skipping.html#condition-parameter

Skipping is a neutral signal. It indicates that a test could not be run, but doesn't tell us whether anything is wrong with that or not. An xfail on the other hand tells us that there is a problem, but it's expected, and we ought to have an issue open to address it.

.github/workflows/test.yml
```diff
@@ -15,6 +15,9 @@ jobs:
         - { python-version: "3.7", os: "ubuntu-latest", backend-db: sqlite }
         - { python-version: "3.8", os: "ubuntu-latest", backend-db: sqlite }
         - { python-version: "3.9", os: "ubuntu-latest", backend-db: sqlite }
+        - { python-version: "3.7", os: "windows-latest", backend-db: sqlite }
+        - { python-version: "3.8", os: "windows-latest", backend-db: sqlite }
+        - { python-version: "3.9", os: "windows-latest", backend-db: sqlite }
```
edgarrmondragon, line 20: 👍 hopefully we can start testing Windows + Postgres at some point

visch:
> The tests that fail on Windows, but shouldn't, should use `pytest.xfail` instead of `pytest.skipif`. The Pytest docs for xfail have an example for conditionally xfailing on Windows: https://docs.pytest.org/en/7.1.x/how-to/skipping.html#condition-parameter
> 
> Skipping is a neutral signal. It indicates that a test could not be run, but doesn't tell us whether anything is wrong with that or not. An xfail on the other hand tells us that there is a problem, but it's expected, and we ought to have an issue open to address it.

I think you're right, but some of the tests on Windows lead to the job hanging the entire pytest process. I didn't take the time to figure out which tests fail and which ones hang I could do that here if you all want me to?

How do you feel about punting that to https://github.com/meltano/meltano/issues/3444 I added a comment there? I think that issue is going to end up being a bunch of sub issues addressing individual failures, and one of those should be switching to xfail for everything.

.github/workflows/test.yml
visch: https://github.com/meltano/meltano/issues/6281 :D yes, I'll add this issue link inline as well

WillDaSilva:
> How do you feel about punting that to https://github.com/meltano/meltano/issues/3444 I added a comment there? I think that issue is going to end up being a bunch of sub issues addressing individual failures, and one of those should be switching to xfail for everything.

Works for me, though an alternative that lets us use xfail and avoid the having the process hang would be adding something like this at the start of each of these failing tests:

```python
def test_something(...):
    if platform.system() == "Windows":
        pytest.xfail("reason...")
    # The original body of the test function continues here
```

tests/meltano/core/logging/test_output_logger.py
```diff
@@ -186,7 +211,8 @@ def test_logging_exception(self, log, subject, redirect_handler):
                     raise exception
 
         # make sure it let the exception through
-        assert exc.value is exception
+        # All code below here in this test cannot be reached
+        assert exc.value is exception  # noqa: WPS441
```
visch, line 215: Note I didn't fix all the linting issues I started to and gave up on it

visch:
> > How do you feel about punting that to #3444 I added a comment there? I think that issue is going to end up being a bunch of sub issues addressing individual failures, and one of those should be switching to xfail for everything.
> 
> Works for me, though an alternative that lets us use xfail and avoid the having the process hang would be adding something like this at the start of each of these failing tests:
> 
> ```python
> def test_something(...):
>     if platform.system() == "Windows":
>         pytest.xfail("reason...")
>     # The original body of the test function continues here
> ```



> > How do you feel about punting that to #3444 I added a comment there? I think that issue is going to end up being a bunch of sub issues addressing individual failures, and one of those should be switching to xfail for everything.
> 
> Works for me, though an alternative that lets us use xfail and avoid the having the process hang would be adding something like this at the start of each of these failing tests:
> 
> ```python
> def test_something(...):
>     if platform.system() == "Windows":
>         pytest.xfail("reason...")
>     # The original body of the test function continues here
> ```

Went ahead and added xfail instead of skip!

tests/meltano/cli/test_elt.py
```diff
@@ -219,6 +220,39 @@ def dbt_process(process_mock_factory, dbt):
...
+        assert (
+            "ELT command not supported on Windows. Please use the Run command as documented here https://docs.meltano.com/reference/command-line-interface#run"
+            in str(result.exception)
+        )
+        # Didn't use exception_logged() as result.stderr doensn't contain the error for some reason
+
+
+@pytest.mark.skipif(
+    platform.system() == "Windows",
+    reason="ELT doesn't work for Windows, don't run these tests on Windows as they fail in horrible ways.",
```
WillDaSilva, line 254:
```suggestion
    reason="ELT is not supported on Windows",
```

tests/meltano/cli/test_elt.py
```diff
@@ -1012,6 +1046,10 @@ def test_dump_loader_config(
             )
 
 
+@pytest.mark.skipif(
+    platform.system() == "Windows",
+    reason="ELT doesn't work for Windows, don't run these tests on Windows as they fail in horrible ways.",
```
WillDaSilva, line 1051:
```suggestion
    reason="ELT is not supported on Windows",
```

tests/meltano/cli/test_elt.py
```diff
@@ -1179,6 +1217,10 @@ def test_elt_transform_run_dbt_failure(
...
```
WillDaSilva, line 1222:
```suggestion
    reason="ELT is not supported on Windows",
```

WillDaSilva (approved):
Thanks for getting this done @visch! I'm glad we'll be able to make these Windows tests available in CI in time for me to run the Python 3.10 changes against them.

visch:
Code Coverage stuff isn't quite accurate here (Or maybe I misunderstand) as we are testing these things but it's dependent on our github actions. We could dive into this as a part of this PR as well? (this is new as well!)

edgarrmondragon:
> I guess my question here is do we need to fix code coverage here or not?

@visch not at all. It's currently only a guideline and not a strict requirement.

edgarrmondragon (approved):

WillDaSilva:
@visch The new tests failures are my bad: #6306 

Sorry about that. I'll have it sorted out as soon as possible.

.github/workflows/test.yml
aaronsteers: @edgarrmondragon and @visch - No reasons I'm aware to collect and combine coverage. This snippet was written before we had a combine step (or before I understood it. 😄 )

We can open this as a separate issue though rather than delay to PR.

tests/meltano/core/tracking/test_exception.py
aaronsteers: @visch and @WillDaSilva - Resolving this. Yes, makes sense. 👍

Issue #6062, opened 2022-06-07
Enable image security scanning in CI


WillDaSilva:
We can use (multiple) actions from the marketplace for this:
- https://github.com/marketplace/actions/setup-jfrog-cli
- https://github.com/marketplace/actions/container-image-scan
- https://github.com/marketplace/actions/anchore-container-scan

We will want to run these whenever new images are created, as well as whenever dependencies are updated. We'll also want to run this regularly (e.g. daily) as new security vulnerabilities to be discovered and reported despite us not changing anything.

Issue #3203, opened 2022-02-23
Slim Meltano Docker images


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3278

Originally created by @Stephen.Clarke.health.govt.nz on 2022-02-23 07:41:11

---

### Problem to solve
Reducing security vulnerabilities in Meltano Docker Images to meet IT Security Guidelines.

### Target audience
This will be beneficial for all Meltano Users who use the Meltano Docker Images - especially helpful for Containerized implementations of Meltano.

### Further details
When scanning the Meltano Image for vulnerabilities (via AWS Container Registry - ECR), there are many security vulnerabilities listed. They mainly appear to be attributed to the Docker Python base image.

My security team would like to minimize the number of identified vulnerabilities, and the only way to do this is to change the python base image Meltano is built on.  

### Proposal
Having come across this article, https://snyk.io/blog/best-practices-containerizing-python-docker/, they appear to recommend alternative base images for Python. For example I am building a Meltano Python 3.8 Image, the recommendation is
python:3.8.12-slim-bullseye
I'm wonder if it is possible to have alternative Meltano Docker builds which I could pull with a more hardened base image?

I have tested and use the slim-bullseye by making my own Custom Meltano Base image and it seems to work well. The other advantage is that it reduces the size of the overall Docker Image.

Please Note: I really don't want to maintain my own custom meltano base image as I need to update a number of the other dependency files e.g. pyproject.toml. This makes my builds brittle and subject to build failure when there are updated Meltano releases.

### What does success look like, and how can we measure that?
Success for me would be less security vunerabilities.

I feel this could be done either one of two ways which aligns to @aaronsteers thinking as well.

1. Completely swap out the Python Base image with the slim Python versions as suggested by the Synk article.
2. Have alternative builds / variants Meltano Docker Images which use the slim Python image for the Meltano Base build.

### Regression test
(Ensure the feature doesn't cause any regressions)
- [ ] Regression build some Meltano Docker images to ensure TAPS and TARGETS still work
- [ ] Consider a phased deployed e.g. Option 2 first and then Option 1. In that way users can confirm the new slim python builds work for them.

### Links / references

https://snyk.io/blog/best-practices-containerizing-python-docker/

The attached file is my example [Dockerfile](http://gitlab.com/meltano/meltano/uploads/435de894cd04eef11959b31a7b3c7a4d/Dockerfile) for my custom based build of the Meltano Base using a slim-python base image.

[...]

PR #6158, opened 2022-06-10
Add back docker publish process with lightweight image

### LINKED ISSUE #6134: Docker images for meltano 2.x
Opening this tracking issue for the 2.0 docker image.

Follow this issue for updates. 👂

aaronsteers:
Resolves #6134 by adding a new lightweight 'publish' image which doesn't need to include all of the dev-time dependencies for yarn/node, etc.

This has the following benefits:

- simplifies the build and publish pipeline
- speeds up CI (runtime <5 minutes for all images)
- reduces security vulnerabilities and begins to address  #3203 (related to #6062)

In future, this will also enable us to use lighterweight python base images, since we're not relying on dependencies that end users don't need.

For compatibility with prior images, all of the foundational expectations are still met:

- port 500 declared
- default workdir = `/project`
- entrypoint = `meltano`
- default cmd = `ui`

## Testing before merging

TIL you can test `workflow_dispatch` jobs before merging to `main`, but you have to do so in the CLI:

https://stackoverflow.com/a/67840292/4298208

```
gh workflow run workflow --ref branch-name -f myparameter=myvalue
```


aaronsteers:
The `SHA-**` labeled image is working well!

I'm moving forward with publishing the public images.

aaronsteers:
@pandemicsyn and @edgarrmondragon - I am going to go ahead and merge this now that the docker images are published - but I'd appreciate any feedback and advice for further improvements.

Thanks!

https://github.com/meltano/meltano/issues/6134#issuecomment-1152758274

docker/publish/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+# Probably not needed, but some devs may want these libraries
+RUN echo "Installing system deps" \
+    && apt-get update \
+    && apt-get install -y build-essential \
+    && rm -rf /var/lib/apt/lists/*
+
+# installing the application the same way our users
+# do when using PyPI
+RUN pip install meltano==${MELTANO_VERSION}
```
edgarrmondragon, line 25: @aaronsteers two changes I'd make here:

1. move `ARG` right before it's used, so the cache of he first `RUN` can be reused to rebuild on a different Meltano version,
2. make the `ARG` the whole version specifier, e.g. `meltano==2.0.1` instead of just the semver string. That way one can pass the a git URL, for example.

Issue #3238, opened 2022-03-07
Adding support for mysql/mysql like db's for the system db


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3315

Originally created by @pandemicsyn on 2022-03-07 19:31:26

---

During a recent meltano hack day event I wanted to try and get Meltano using Planetscale.com (mysql based) as system db. I ran into an issue straight out of the gate because we inconsistently define max string sizes in our migration/schemas (ala https://gitlab.com/meltano/meltano/-/blob/master/src/meltano/migrations/versions/b4c05e463b53_.py#L35). This is because sqlalchemy treats `String()` as varchar's and MySQL requires varchar's to have a max size set.

There maybe other issues beyond this but it might be worth exploring whether :point_up: is really the only blocker to supporting MySQL and friends.

MeltyBot:
[View 7 previous comments from the original issue on GitLab](https://gitlab.com/meltano/meltano/-/issues/3315)

tayloramurphy:
Duplicate of https://github.com/meltano/meltano/issues/6529

PR #6167, opened 2022-06-12
feat: Add systemdb support for MS SQL Server

JulesHuisman:
Relates to issue #3238. 

The main improvement is limiting all String types in the alembic migrations. This allows the possibility to use both MSSQL and MYSQL as a Meltano backend database.

Both databases were manually tested. It might be valuable to add automatic tests to the Github actions. 

To use MYSQL run: `pip install mysqlclient`
To use MSSQL run: `pip install pyodbc`


pandemicsyn:
Hi @JulesHuisman awesome to see you pick this back up, personally very excited to get some support for other DB's into Meltano 🥳 

@aaronsteers this is a continuation of this thread: https://gitlab.com/meltano/meltano/-/issues/3315#note_908238671, I think what @JulesHuisman has here plus some docs covering how to use "unofficial backends like MSSQL and MySQL" are a great start and easy way to test the water. 

We could follow it up and setup an actions workflow to test migrations against MySQL/MSSQL when a PR with new migrations is created. @aaronsteers In the original thread you'd expressed some concerns about being able to test against MSSQL, looks like the MSSQL on linux docker container is pretty popular and well-supported (https://hub.docker.com/_/microsoft-mssql-server), but just having a GH actions gate for migrations would go along way towards ensuring that future migrations don't having breaking changes for MySQL or MSSQL.

aaronsteers:
@JulesHuisman and @pandemicsyn - I would be happy to add CI tests for DBs. I am also okay to merge this and add those tests in a follow-on issue, depending on @JulesHuisman's availability and/or others who can help ship this in CI.

A quick google search shows tests may be fairly pretty easy in GitHub Actions - https://datastation.multiprocess.io/blog/2021-12-16-sqlserver-in-github-actions.html

src/meltano/migrations/versions/13e8639c6d2b_add_state_edit_to_job_state_enum.py
```diff
@@ -37,7 +37,7 @@ def upgrade():
         op.alter_column(
             table_name="job",
             column_name="state",
-            type_=sa.types.String,
```
aaronsteers, line 40: @edgarrmondragon - I'm not an Alembic guru by any measure. From your experience, do you it okay to modify these already-created migrations?

aaronsteers:
@pandemicsyn and @edgarrmondragon - I'm removing myself from reviewer role but please re-add me if you would like me to rereview. I'll defer to you both - when you feel this is ready to merge, I'm all for it. 👍 

And - again - since risk here is super low (adding str max when where they were otherwise omitted), I'm okay to merge without additional CI checks added.

The one thing I'd ask to watch out for is that I'd rather a stringlen max be overly permissive than to run a risk of truncating or failing on use cases that might introduce longer strings.

edgarrmondragon:
## Jobs

- [ ] [`job.state`](https://github.com/meltano/meltano/blob/4deafc883a9b0ffbf2fc87e4a4be28cb20034ce4/src/meltano/core/job/job.py#L103)
- [ ] [`job.trigger`](https://github.com/meltano/meltano/blob/4deafc883a9b0ffbf2fc87e4a4be28cb20034ce4/src/meltano/core/job/job.py#L109)

## Plugin Settings

- [ ] [`plugin_settings.label`](https://github.com/meltano/meltano/blob/4deafc883a9b0ffbf2fc87e4a4be28cb20034ce4/src/meltano/core/setting.py#L15)
- [ ] [`plugin_settings.name`](https://github.com/meltano/meltano/blob/4deafc883a9b0ffbf2fc87e4a4be28cb20034ce4/src/meltano/core/setting.py#L20)
- [ ] [`plugin_settings.namespace`](https://github.com/meltano/meltano/blob/4deafc883a9b0ffbf2fc87e4a4be28cb20034ce4/src/meltano/core/setting.py#L21)

## `embed_tokens`

NA

## Subscriptions

- [ ] [`subscriptions.recipient`](https://github.com/meltano/meltano/blob/4deafc883a9b0ffbf2fc87e4a4be28cb20034ce4/src/meltano/api/models/subscription.py#L20)
- [ ] [`subscriptions.event_type`](https://github.com/meltano/meltano/blob/4deafc883a9b0ffbf2fc87e4a4be28cb20034ce4/src/meltano/api/models/subscription.py#L21)
- [ ] [`subscriptions.source_type`](https://github.com/meltano/meltano/blob/4deafc883a9b0ffbf2fc87e4a4be28cb20034ce4/src/meltano/api/models/subscription.py#L22)
- [ ] [`subscriptions.source_id`](https://github.com/meltano/meltano/blob/4deafc883a9b0ffbf2fc87e4a4be28cb20034ce4/src/meltano/api/models/subscription.py#L23)

## User

These should all really have max length of 255, so migration needs to reflect that.

- [ ] [`user.username`](https://github.com/meltano/meltano/blob/4deafc883a9b0ffbf2fc87e4a4be28cb20034ce4/src/meltano/api/models/security.py#L8)
- [ ] [`user.email`](https://github.com/meltano/meltano/blob/4deafc883a9b0ffbf2fc87e4a4be28cb20034ce4/src/meltano/api/models/security.py#L9)
- [ ] [`user.password`](https://github.com/meltano/meltano/blob/4deafc883a9b0ffbf2fc87e4a4be28cb20034ce4/src/meltano/api/models/security.py#L10)

## Role

This should also have max length of 255, so migration  

- [ ] [`role.description`](https://github.com/meltano/meltano/blob/4deafc883a9b0ffbf2fc87e4a4be28cb20034ce4/src/meltano/api/models/security.py#L8)

## Role Permissions

- [ ] [`role_permissions.type`](https://github.com/meltano/meltano/blob/4deafc883a9b0ffbf2fc87e4a4be28cb20034ce4/src/meltano/api/models/security.py#L43)
- [ ] [`role_permissions.context`](https://github.com/meltano/meltano/blob/4deafc883a9b0ffbf2fc87e4a4be28cb20034ce4/src/meltano/api/models/security.py#L44)

# OAuth

- [ ] [`oauth.id_token`](https://github.com/meltano/meltano/blob/4deafc883a9b0ffbf2fc87e4a4be28cb20034ce4/src/meltano/api/models/oauth.py#L15)

src/meltano/migrations/versions/13e8639c6d2b_add_state_edit_to_job_state_enum.py
edgarrmondragon: @aaronsteers So, if MySQL/MSSQL can't even run the existing migrations then this is only way to get them started. We should confirm that.

Something I'd worry about is migrations auto-generated going forward so I think we should also change the data types in the SQLAlchemy models themselves, which I've linked in https://github.com/meltano/meltano/pull/6167#issuecomment-1154592468

edgarrmondragon:
@pandemicsyn we used to have a sort of test matrix running tests on both SQLite and PostgreSQL:

- https://github.com/meltano/meltano/blob/abb83158c443787d3c330558de29842d415d8e15/tests/conftest.py#L21-L38
- https://github.com/meltano/meltano/blob/1498bf59afa5a9eb44059a6a4f550caab4e07c31/.gitlab/ci/test.gitlab-ci.yml#L88-L94
- https://github.com/meltano/meltano/blob/1498bf59afa5a9eb44059a6a4f550caab4e07c31/.gitlab/ci/test.gitlab-ci.yml#L110-L113

I suggest we should migrate those (either before or after this PR merges) and include both MySQL and MSSQL to the mix.

JulesHuisman:
Thanks for taking a look guys, 

In some places it might be possible to make the max string length longer. I chose 128 because in some places MYSQL would throw errors when it was longer (450). 

I can also take a look at testing the migrations. The MSSQL docker container works quite well, I also used it when developing this PR.

src/meltano/migrations/versions/13e8639c6d2b_add_state_edit_to_job_state_enum.py
pandemicsyn: > So, if MySQL/MSSQL can't even run the existing migrations then this is only way to get them started. We should confirm that.

@edgarrmondragon when i started going down the path of adding mysql support for the hackday that was definitely the issue I ran into. The initial migrations failed because MySQL needs a size constraint on varchar's (String() here)  

We *could* retro fit the migrations and branch based on dialect or driver or something along those lines. e.g. sqlite/postgres get the old style create tables while everything else gets create/alter tables with defined string sizes. If we went this route , I think opting sqlite/postgres into the old style is probably better than explicitly opting MySQL/MSSQL into the new style.

JulesHuisman: @aaronsteers @edgarrmondragon Exactly, the problem is the fact that the alternate databases cannot run the existing migration. So creating an altering migration afterwards doesn't work in this case, as the migration cannot be reached.

JulesHuisman:
@edgarrmondragon @pandemicsyn Hi guys, what needs to be fixed or adjusted to get this PR through? 

And do we want the tests included in this PR, or can that be included in a separate PR?

Thanks!

JulesHuisman:
Ah alright, so it is running into issues with the string length. I will look into it!

edgarrmondragon:
@JulesHuisman an options seems to be off with the mssql service, and it's causing all workflows to fail.

We moved away from the `services` approach in favor of using `docker run` explicitly in a step (see #6268). What do you think of following a similar approach for mysql and mssql?

JulesHuisman:
> @JulesHuisman It seems like SQLAlchemy expects some extras/deps:
> 
> 
> 
> ```
> 
> ModuleNotFoundError: No module named 'pyodbc'
> 
> ```
> 
> 
> 
> and
> 
> 
> 
> ```
> 
> ModuleNotFoundError: No module named 'pymysql'
> 
> ```

Ah yes, I forgot to add them to the tests, I will take a look at it tomorrow.

edgarrmondragon:
> @edgarrmondragon I thought using the pip install would install it in the right environment. But that might not be the case. What is the best way to install the database drivers?

@JulesHuisman the project uses [`poetry`](https://python-poetry.org/docs/) so the best way is to add them explicitly with `poetry add`.

Though, I'm not sure if we want _all_ users to be forced install them and any system dependencies, which might be annoying. So, I'd opt for making them _optional_ dependencies and put them in pip extras that need to be requested explicitly, e.g.

```
pip install meltano[mysql]
```

It's a similar approach to that taken by [Airflow](https://airflow.apache.org/docs/apache-airflow/stable/extra-packages-ref.html#core-airflow-extras) and other complex applications and libraries. The way to accomplish that with poetry is to add the dependencies as optional:

```
poetry add pymysql --optional
poetry add pyodbc --optional
poetry lock --no-update
```

and then manually creating the extras in `pyproject.toml`:

```toml
[tool.poetry.extras]
infra = ["ansible"]
repl = ["ipython"]
mysql = ["pymysql"]
mssql = ["pyodbc"]
```

This is a bit of a product call too, so cc @tayloramurphy @aaronsteers

JulesHuisman:
@edgarrmondragon Let me know if I added it correctly.

I think we discussed on the original Gitlab issue if we should include these drivers by default. And we decided not to include them, but I personally like to the route of adding them as extras.

edgarrmondragon:
> A possibility would be to run the `mssql` inside a mac runner, as the install is way easier for mac. But that doesn't really feel that robust to me.

@JulesHuisman do you think using the [`pymssql`](https://docs.sqlalchemy.org/en/14/dialects/mssql.html#module-sqlalchemy.dialects.mssql.pymssql) driver instead of ODBC would be a problem? It seems to be more active and has [lots more wheels](https://pypi.org/project/pymssql/#files).

.github/workflows/test.yml
```diff
@@ -101,6 +101,12 @@ jobs:
         --health-retries 5
         mysql
         --default-authentication-plugin=mysql_native_password
+        --collation_server=utf8_general_ci 
+        --character_set_server=utf8
+        
+    - name: Wait for MYSQL Container
+      if: always() && (matrix.backend-db == 'mysql')
+      run: sleep 15
```
JulesHuisman, line 109: @edgarrmondragon The driver throws an authentication error if the mysql database is not started when the tests start. I am not sure if there is a better solution than just waiting x amount of seconds.

JulesHuisman:
@edgarrmondragon I switched out `pyodbc` for `pymssql`, in my fork the MSSQL testing now passes. 

There is still a weird bug when using the MySQL backend, when getting and setting does not match.

tayloramurphy:
> I'd opt for making them optional dependencies and put them in pip extras that need to be requested explicitly

@edgarrmondragon I think that makes sense. @aaronsteers would you chime in here as well?

edgarrmondragon:
MySQL still failing with

```
"Table 'pytest_warehouse.runs' doesn't exist"
```

aaronsteers:
> > I'd opt for making them optional dependencies and put them in pip extras that need to be requested explicitly
> 
> @edgarrmondragon I think that makes sense. @aaronsteers would you chime in here as well?

Yeah - this seems like the right approach to me:

https://github.com/meltano/meltano/pull/6167/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R67-R70

It would be a breaking change at this point, but similarly, we could have make `psycopg2-binary` a `postgres` extra, and in a major version bump we could consider that.

aaronsteers (changes requested):
This looks super close. @edgarrmondragon has called out a failing test and I would also request an update in the docs.

Regarding docs, approaches in the "Installation guide" that could work:

1. A section added to installation docs that says something like "Installing Optional Components", with guidance that by default only SQLite and Postgres libraries are included in the meltano install, with instructions to install optional components as extras and optionally including instructions on using pipx 'inject' to add more libraries into the base `meltano` venv.
2. Similar approach to the above, except the content would live under "Advanced Topics", just just a quick mention in the main Install guide, with a link to the appropriate section.

@JulesHuisman - Would you feel comfortable adding these docs, and what do you think about these approaches (or others)? 

cc @tayloramurphy, @afolson for their input on docs side.

tayloramurphy:
@aaronsteers @JulesHuisman @afolson on the docs side, we mention the system database in a few places:

* https://docs.meltano.com/guide/production#storing-metadata which has some repeated info from the next item
* https://docs.meltano.com/concepts/project#system-database
* https://docs.meltano.com/reference/settings#database_uri which has some repeated content
* https://docs.meltano.com/contribute/prerequisites#system-database

My preference for this PR and to keep it simple for @JulesHuisman would be to update https://docs.meltano.com/reference/settings#database_uri with details on how to use the other databases. We can open a separate issue to update the references in other places as a follow-on. Does that work?

joaopamaral:
Hi @JulesHuisman! First of all, thanks for this awesome work!

I just had to include a small change to make it work in my environment in the `src/meltano/migrations/versions/5b43800443d1_rename_job_to_job_run_and_job_id_to_job_.py` migration file to include the `existing_type` required by `alembic`. 

The new file looks like this after the changes:
```python
import sqlalchemy as sa
from alembic import op

from meltano.migrations.utils.dialect_typing import (
    get_dialect_name,
    max_string_length_for_dialect,
)

# revision identifiers, used by Alembic.
revision = "5b43800443d1"
down_revision = "13e8639c6d2b"
branch_labels = None
depends_on = None


def upgrade():
    dialect_name = get_dialect_name(op)
    max_string_length = max_string_length_for_dialect(dialect_name)

    op.alter_column("job", "job_id", 
        new_column_name="job_name",
        existing_type=sa.types.String(max_string_length)
    )
    op.rename_table("job", "runs")


def downgrade():
    dialect_name = get_dialect_name(op)
    max_string_length = max_string_length_for_dialect(dialect_name)
    
    op.rename_table("runs", "job")
    op.alter_column("job", "job_name", 
        new_column_name="job_id",
        existing_type=sa.types.String(max_string_length)
    )

```

After this change, I managed to make the db migration work with a MySQL instance.

edgarrmondragon:
@joaopamaral thanks for sharing! I've put your changes in diff format in case @JulesHuisman or someone else wants to try them out:

```diff
 import sqlalchemy as sa
 from alembic import op
 
+from meltano.migrations.utils.dialect_typing import (
+    get_dialect_name,
+    max_string_length_for_dialect,
+)
+
 # revision identifiers, used by Alembic.
 revision = "5b43800443d1"
 down_revision = "13e8639c6d2b"
 
 
 def upgrade():
-    op.alter_column("job", "job_id", new_column_name="job_name")
+    dialect_name = get_dialect_name(op)
+    max_string_length = max_string_length_for_dialect(dialect_name)
+
+    op.alter_column("job", "job_id", 
+        new_column_name="job_name",
+        existing_type=sa.types.String(max_string_length)
+    )
     op.rename_table("job", "runs")
 
 
 def downgrade():
+    dialect_name = get_dialect_name(op)
+    max_string_length = max_string_length_for_dialect(dialect_name)
+    
     op.rename_table("runs", "job")
-    op.alter_column("job", "job_name", new_column_name="job_id")
+    op.alter_column("job", "job_name", 
+        new_column_name="job_id",
+        existing_type=sa.types.String(max_string_length)
+    )
```

edgarrmondragon:
Heads up @JulesHuisman: I'm gonna merge `main` into this branch to try to resolve the conflicts.

edgarrmondragon (approved):

edgarrmondragon:
@pandemicsyn @aaronsteers this is ready for review for MSSQL support!

pandemicsyn (approved):
Awesome! 🥳

WillDaSilva (approved):

JulesHuisman:
Thanks @edgarrmondragon, I like the idea of moving the Mysql code to another branch. That created a really weird bug in testing which I could not figure out.

edgarrmondragon:
> Thanks @edgarrmondragon, I like the idea of moving the Mysql code to another branch. That created a really weird bug in testing which I could not figure out.

@JulesHuisman I started a draft PR #6528, but I (or someone else) need to take some to debug those failing state tests.

aaronsteers (approved):
Awesome work here! Very excited to get this in!

docs/src/_concepts/project.md
```diff
@@ -359,3 +359,19 @@ The remaining tables in the database are used exclusively by Meltano UI, mostly
...
+
+### Support for other database types
+
+Meltano currently supports the following databases as backends for state and configuration:
+
+* PostgreSQL
+* SQLite
+* MS SQL Server
+
+PostgresSQL and SQLite are supported out of the box, while the MS SQL Server requires installing Meltano with the [`mssql` extra](/guide/advanced-topics#installing-optional-components).
```
tayloramurphy, line 371: while extra is technically right here, I think it could be confusing with native Meltano extras. Can we specify this is a Python extra?

docs/src/_guide/advanced-topics.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Advanced Topics
+description: Learn about advanced topics in the Meltano.
+layout: doc
+weight: 10
+---
+
+## Installing Optional Components
+
+Most of Meltano's features are available without installing any additional packages. However, some niche or environment-specific features require the installation of [extras](https://peps.python.org/pep-0508/#extras). The following extras are currently supported:
```
tayloramurphy, line 10: Similar comment to above, adding a comment about this being a Python, not a Meltano, extra.

tayloramurphy (approved):
Minor comments but I'm approving assuming those can get fixed so it's not a blocker

JulesHuisman:
Thanks for taking this over the finish line @edgarrmondragon!

PR #6173, opened 2022-06-13
use run in the getting started guide

pnadolny13:
Use `meltano run` in getting started guide vs `meltano elt`. We want new people to start using `run` but keep the docs in good shape for `elt` because tons of users still use elt.

I had to update some incremental sync docs also because the getting started guide had lots of references around state ID. The guide now prefers run + state commands but still has a few reference to ELT. The integration guide still has lots of reference to ELT but has more details around managing state with `run`.


docs/getting-started.md
```diff
@@ -800,7 +791,7 @@ To help you achieve this, Meltano supports scheduled pipelines that can be orche
 
 _To learn more about orchestration, refer to the [Orchestration guide](/guide/orchestration)._
 
-1. Schedule a new [`meltano elt`](/reference/command-line-interface#elt) pipeline to be invoked on an interval using [`meltano schedule`](/reference/command-line-interface#schedule):
+1. Schedule a new pipeline to be invoked on an interval using [`meltano schedule`](/reference/command-line-interface#schedule):
```
pandemicsyn, line 794: In https://github.com/meltano/meltano/pull/6121/files @cjohnhanson is reworking things to bias towards a scheduled job. Should we do that here too, or do you think that bloats the Getting Started guide?

pandemicsyn (approved):

docs/getting-started.md
pnadolny13: @pandemicsyn yeah I thought about it and I figured we'd want to do that at some point but I was leaving it for another PR. I was trying to keep this schedule section the same but remove ELT references so more new users only use run/invoke.

tayloramurphy (approved):

PR #6186, opened 2022-06-14
Improve CLI performance by avoiding unnecessary imports

### LINKED ISSUE #6132: Analyze and address slow CLI start times, starting with `meltano --help`
Follow up from demo day today:

We'd like to improve start/spinup time of _all_ CLI commands, but a good place to start would be `meltano --help`.

WillDaSilva:
- Unused imports (outside of `__init__.py` files) have been removed.
- Some imports that were only used for type hints were moved into an `if TYPE_CHECKING` block so that they are not imported at runtime.
- Some imports that were potentially needed at runtime inside of a function were moved into the function.

These changes have cut the time it takes to run `meltano --help` roughly in half. I'm not a fan of having imports inside of functions, but given the significant performance improvement it seems worth it for these special cases.

Later this iteration, possibly for the hack day this Friday, I'm going to try to automatically defer imports by using a `LazyModule` type that is returned by import statements in place of regular modules when the import is originating from a Meltano module.

Closes #6132

### DIFF
--- a/src/meltano/api/app.py
+++ b/src/meltano/api/app.py
@@ -66,7 +66,7 @@ def create_app(config: dict = {}) -> Flask:  # noqa: WPS210,WPS213,B006
     from .json import setup_json
     from .mail import mail
     from .models import db
-    from .security import security, setup_security, users
+    from .security import setup_security
     from .security.oauth import setup_oauth
 
     db.init_app(app)
--- a/src/meltano/api/app.py
+++ b/src/meltano/api/app.py
@@ -111,7 +111,6 @@ def create_app(config: dict = {}) -> Flask:  # noqa: WPS210,WPS213,B006
         logger.debug("Notifications are disabled.")
 
     # Google Analytics setup
-    LegacyTracker(project)
 
     @app.before_request
     def setup_js_context():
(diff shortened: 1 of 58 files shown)


src/meltano/api/app.py
```diff
@@ -113,7 +111,7 @@ def create_app(config: dict = {}) -> Flask:  # noqa: WPS210,WPS213,B006
         logger.debug("Notifications are disabled.")
 
     # Google Analytics setup
-    tracker = LegacyTracker(project)
+    LegacyTracker(project)
```
pandemicsyn, line 114: We should  double-check whether we even need this. I don't recall actually seeing any API endpoints fire tracking events, but its possible I also haven't noticed them.

pandemicsyn: @WillDaSilva im pretty sure you could ditch/remove this completely.

WillDaSilva, line 114:
```suggestion
```

pandemicsyn (approved):
I think the imports being in lined (aiodocker, migration_service, etc) are pretty self-contained, so I'm personally ok with it, especially considering the perf improvement you've made. 

Though I do wonder if we should leave an inline comment about why they're func level imports.

I don't have a strong affinity for being very pythonic so would love to get a second opinion and get a quick 👍 or 👎 on this from @edgarrmondragon before merging this.

edgarrmondragon:
Taking a look now

edgarrmondragon (approved):
lgtm!

PR #6199, opened 2022-06-14
Restore `project_id` from `analytics.json` & improve `analytics.json` handling

WillDaSilva:
- Restore `project_id` from `analytics.json` if it is not found in `meltano.yml`.
- Overwrite `analytics.json` with a new one if it is malformed.
- Handle more errors when dealing with `analytics.json` (e.g. `JSONDecodeError`).
- Add additional debug logging.
- Do not update `meltano.yml` from `ProjectContext` (n.b. this would cause the `project_id` key to end up inside of the config of the active environment instead of the top-level config).


src/meltano/core/project_settings_service.py
```diff
@@ -61,6 +65,23 @@ def __init__(self, *args, config_service: ConfigService = None, **kwargs):
             **self.config_override,
         }
 
+        if self.get("project_id") is None:
```
WillDaSilva, line 68: There is still some weirdness if a user manually adds the `project_id` key at the environment-level. These changes will result in it being added to `meltano.yml` at the top level (but not removed from the environment config it originated from). 

Does it make sense to have per-environment project IDs? If so we can clean up this issue by checking `self.project.active_environment is not None` here before editing the config on-disk. The consequence however would be that we don't restore the project ID from `analytics.json` if there is no active environment. I'm not sure if that can ever occur.

WillDaSilva:
CI is failing with an error while setting up the fixtures, but it's passing locally on my computer. Will look into this later today.

src/meltano/core/project_settings_service.py
WillDaSilva: Actually, instead of erroring, we should just rewrite the config file to move `project_id` up. An error will be raised if there are multiple that conflict, but if there's only one or they all match there's no need for an error.

pandemicsyn:
> CI is failing with an error while setting up the fixtures, but it's passing locally on my computer. Will look into this later today.

👍 fwiw - was gonna ask about getting some explicit tests around state change behavior anyway. Right now we've got the one sort of canary test we added the last time around, but might be nice to get some explicit test cases to verify our behavior for:

(off the top of my head)

1. When there is an analytics.json and its "valid" (project id and client id match as expected)
   - Verify behavior when we have a project id in meltano yaml and tracking changes from enabled to disabled and vice versa.
    - Verify behavior when we lack a project id in meltano yaml and tracking changes from enabled to disabled and vice versa.

2. When there's not yet an analytics.json
    - Verify behavior when we have a project id in meltano yaml and tracking changes from enabled to disabled and vice versa.
    - Verify behavior when we lack a project id in meltano yaml and tracking changes from enabled to disabled and vice versa.

3. When there's an analytics.json but its contents (client id, project id) don't match the active project as if someone had copied an existing project skeleton.
    - Verify behavior when we have a project id in meltano yaml and tracking changes from enabled to disabled and vice versa.
    - Verify behavior when we lack a project id in meltano yaml and tracking changes from enabled to disabled and vice versa.

@aaronsteers in addition to these 3 scenarios, any other iffy/tricky ones we should have explicit tests for?

src/meltano/core/project_settings_service.py
WillDaSilva: In the end I decided against erroring at all. As you say it doesn't make sense, so silently getting rid of the env-level project IDs seems fine if necessary.

There's currently a bug that I'm not sure is worth fixing where the project ID that gets brought up from the env-level to the top-level will be turned into a UUID via hashing if it isn't one already. This is because `ProjectSettingsService` is initialized multiple times, and on the first time it will not have an active environment. As a result the project ID is taken from `analytics.json` instead of the env-level config. Later when `ProjectSettingsService` is initialized with an active environment, the UUID from `analytics.json` will be preferred over the one at the env-level since it's now the top-level `project_id`.

WillDaSilva:
>  When there's not yet an analytics.json
>  - Verify behavior when we have a project id in meltano yaml and tracking changes from enabled to disabled and vice versa.
>  - Verify behavior when we lack a project id in meltano yaml and tracking changes from enabled to disabled and vice versa.

@pandemicsyn I don't understand what you mean by "tracking changes" here. With no `analytics.json`, there is nothing to compare the current telemetry settings to, so we cannot identify a change. As implemented on `main` and in this PR, in the absence of `analytics.json` no telemetry state change event is ever fired.

pandemicsyn:
> @pandemicsyn I don't understand what you mean by "tracking changes" here. With no analytics.json, there is nothing to compare the current telemetry settings to, so we cannot identify a change. As implemented on main and in this PR, in the absence of analytics.json no telemetry state change event is ever fired.

Sorry, "tracking changes" here means a user actually changes there their tracking opt-in setting, but you're right this would apply to basically any command run after first updating 2.0.  Regardless of whether the opt-in setting changed. The behavior you described is what we should be verifying.   We should test that if there's no analytics.json and a user executes some command that no change events are fired and that an analytics.json gets created with the expected entries.

tests/meltano/core/tracking/test_tracker.py
```diff
@@ -1,22 +1,58 @@
...
+
+@contextmanager
+def delete_analytics_json(project: Project) -> None:
+    (project.meltano_dir() / "analytics.json").unlink()
+    try:
+        yield
+    finally:
+        # Recreate `analytics.json`, to avoid causing issues for other tests within the same class
+        # that expect it. Note that `project` is class-scoped.
+        Tracker(project)
```
pandemicsyn, line 41: It might be worth having an explicit test of `Tracker.can_track` or `Tracker.send_anonymous_usage_stats` to make sure that `MELTANO_SEND_ANONYMOUS_USAGE_STATS=FALSE`  and friends actually work. 

I realize thats getting tested indirectly by some state change tests but - since that's sort of the lowest level where a "send or don't send" decision happens its probably worth an explicit check that the config opts and envs are working as expected.

pandemicsyn (approved):
I know this addresses an outstanding bug so I think its ok to ship as-is and follow up with more tests later.

If not now, then in a follow up, I'd definitely love at least an explicit check somewhere to validate that `Tracker.can_track` or `Tracker.send_anonymous_usage_stats` is working as expected.

tests/meltano/core/test_tracking.py
```diff
@@ -10,8 +9,7 @@
 
 @pytest.mark.meta
 def test_tracking_disabled(project):
-    assert os.getenv("MELTANO_SEND_ANONYMOUS_USAGE_STATS") == "False"
-    assert ProjectSettingsService(project).get("send_anonymous_usage_stats") is False
+    assert not ProjectSettingsService(project).get("snowplow.collector_endpoints")
```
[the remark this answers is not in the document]
WillDaSilva: This meta test is there to ensure we aren't firing telemetry events during the Pytest executions. Having no collections set achieves that while also letting us set `send_anonymous_usage_stats` to true as needed during tests.

pandemicsyn:
@pnadolny13 @WillDaSilva this PR still fails test that @pnadolny13 describes in https://github.com/meltano/internal-general/issues/400 because we still fire an event when the project id changes , even if the saved `send_anonymous_usage_stats` entry is set to false. 

So we need a guard statement and matching test that:

- If the project id between meltano and analytics.json differs, AND the previous analytics.json entry's `send_anonymous_usage_stats` is set to False, then we should NOT fire a state change event.

WillDaSilva:
@aaronsteers @pandemicsyn What's up with the `tracking_disabled` setting? It was added to the new `Tracker` class, but I don't see it referenced anywhere else. Can it be removed? It's the same as `send_anonymous_usage_stats`, but negated. It's only respected in one place in the codebase.

WillDaSilva:
> What's up with the tracking_disabled setting? It was added to the new Tracker class, but I don't see it referenced anywhere else. Can it be removed? It's the same as send_anonymous_usage_stats, but negated. It's only respected in one place in the codebase.

I've gone ahead and removed it. Let me know if I shouldn't have done that.

WillDaSilva:
@pnadolny13 @pandemicsyn I've added the requested tests and bug fix for https://github.com/meltano/internal-general/issues/400: https://github.com/meltano/meltano/pull/6199/commits/2727df21a9862b0e1bba8967e669004c88b2ce57

WillDaSilva:
There's a test failure caused by some state hanging around from a previous non-tracking test. This might take a few minutes to fix because unless I can identify which test is leaving behind the bad state, I have to re-run the whole test suite each time to check if I've fixed the problem.

pandemicsyn:
> @aaronsteers @pandemicsyn What's up with the tracking_disabled setting? It was added to the new Tracker class, but I don't see it referenced anywhere else. Can it be removed? It's the same as send_anonymous_usage_stats, but negated. It's only respected in one place in the codebase.

That looks like it got introduced with the original state change event PR that you and @aaronsteers coauthored: https://github.com/meltano/meltano/blame/f58031c268982c6eed8032c80f7f75c63b89c3a3/src/meltano/core/tracking/tracker.py#L162

Originally, I thought that was actually related to the `MELTANO_DISABLE_TRACKING` setting but looking around I realize thats just a negated ENV_ALIAS for `send_anonymous_usage_stats` :https://github.com/meltano/meltano/blob/51881f94e93cb89a1ba69bb3339de037d18d1be1/src/meltano/core/bundle/settings.yml#L6

So removal might be fine but im wondering if @aaronsteers had some other intension behind including that in the original PR?

aaronsteers:
> > @aaronsteers @pandemicsyn What's up with the tracking_disabled setting? It was added to the new Tracker class, but I don't see it referenced anywhere else. Can it be removed? It's the same as send_anonymous_usage_stats, but negated. It's only respected in one place in the codebase.
> 
> That looks like it got introduced with the original state change event PR that you and @aaronsteers coauthored: https://github.com/meltano/meltano/blame/f58031c268982c6eed8032c80f7f75c63b89c3a3/src/meltano/core/tracking/tracker.py#L162
> 
> Originally, I thought that was actually related to the `MELTANO_DISABLE_TRACKING` setting but looking around I realize thats just a negated ENV_ALIAS for `send_anonymous_usage_stats` :
> 
> https://github.com/meltano/meltano/blob/51881f94e93cb89a1ba69bb3339de037d18d1be1/src/meltano/core/bundle/settings.yml#L6
> 
> So removal might be fine but im wondering if @aaronsteers had some other intension behind including that in the original PR?

When we removed `env_aliases`, this was one that was configured at the Meltano setting level and there was not a way to honor existing users' `MELTANO_DISABLE_TRACKING` env var while still removing support for `env_aliases` elsewhere in the product. The decision we landed on was to add a second setting which would emulate the prior behavior without still needing `env_aliases`.

aaronsteers (changes requested):
Just one requested change to bring back the `disable_tracking` setting, per my other comment in the PR.

Thanks!

WillDaSilva:
@aaronsteers I've updated `Tracker.send_anonymous_usage_stats` to use `disable_tracking` if `send_anonymous_usage_stats` is not set. To be clear, it's `disable_tracking`, and not `tracking_disabled`, right? The old code had `tracking_disabled`, but the env alias is for `disable_tracking`.

pandemicsyn (approved):
LGTM, but would definitely prefer if we wait to merge until @aaronsteers has a chance to confirm - https://github.com/meltano/meltano/pull/6199#issuecomment-1158394308

aaronsteers:
> To be clear, it's `disable_tracking`, and not `tracking_disabled`, right? ... the env alias is for `disable_tracking`.

@WillDaSilva - Good catch. Yes, it should match the env alias. This misspelling was possibly a bug on my side when I drafted that update.

aaronsteers (approved):

aaronsteers:
@WillDaSilva - Per your comment I did some digging and opened a new PR to make sure this is in a good place.

- https://github.com/meltano/meltano/pull/6199

You can proceed with merging this and we can take follow-on actions in #6199

PR #6211, opened 2022-06-15
Update debugging advice in contributor API doc

WillDaSilva:


tayloramurphy (approved):
Approved from a "is this grammatically correct English" standpoint, but would recommend an engineer approve as well!

edgarrmondragon (approved):
@WillDaSilva nice catch, we missed this when we dropped 3.6 👍

Issue #5983, opened 2022-06-02
Update environment variable inheritance to be in line with currently xfail-ing tests


cjohnhanson:
Environment variable inheritance should behave as illustrated by the `STACKED` variables in [this issue description](https://gitlab.com/meltano/meltano/-/issues/3173). There is a test written for this behavior in `tests/meltano/core/test_environment_variables.py` which is currently marked to `xfail`. Inheritance behavior should be updated so that this test passes.

## Related: 

- Exploratory issue:
  - https://github.com/meltano/meltano/issues/3100
  - https://github.com/meltano/meltano/pull/5938
- Also found:
  - https://github.com/meltano/meltano/issues/5982
- Xfail (expected-to-fail) tests to fix:  https://github.com/meltano/meltano/commit/17ce6c65d2ea3f4b5851528ec15de45e22843be8

aaronsteers:
@cjohnhanson - Thanks for logging this.

Can you say a bit more of the currently observed behavior you seeing on inheritance/stacking?

For instance, is it something of a name-conflict issue, or are higher-level values simply unavailable during evaluation at lower levels of the tree?

cjohnhanson:
@aaronsteers here's the current behavior, basically environment-level `env` takes precedence and doesn't inherit from variable definition in other "layers". See the `STACKED=3` output line here:
```bash
~/Projects/project-meltano/env-vars/env-vars-3 via 🐍 v3.9.12 (env-vars) on ☁️  cody@codyjhanson.com 
❯ export STACKED=1

~/Projects/project-meltano/env-vars/env-vars-3 via 🐍 v3.9.12 (env-vars) on ☁️  cody@codyjhanson.com 
❯ cat meltano.yml
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: meltano.yml
       │ Size: 667 B
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ version: 1
   2   │ default_environment: dev
   3   │ project_id: 044adc22-cf5b-4c51-9bef-46b512b848fe
   4   │ send_anonymous_usage_stats: false
   5   │ env: "${STACKED}2"
   6   │ plugins:
   7   │   utilities:
   8   │   - name: codys-test-plugin
   9   │     namespace: codys_test_plugin
  10   │     executable: pwd
  11   │     env:
  12   │       STACKED: "${STACKED}4"
  13   │     settings:
[...]

Issue #5982, opened 2022-06-02
Update terminal environment variable precedence to be in line with currently xfail-ing tests


cjohnhanson:
We want environment variable resolution to behave as described in [this miro board](https://miro.com/app/board/uXjVO3j_ADw=/).

This behavior is described in test cases written in `tests/meltano/core/test_environment_variables.py`which which currently have `xfail=True`. Behavior should be updated so that these tests pass. The relevant test cases will have a comment referencing this issue.

Related:

- Exploratory issue:
  - https://github.com/meltano/meltano/issues/3100
  - https://github.com/meltano/meltano/pull/5938
- Also found:
  - https://github.com/meltano/meltano/issues/5983
- Xfail (expected-to-fail) tests to fix:  https://github.com/meltano/meltano/commit/17ce6c65d2ea3f4b5851528ec15de45e22843be8

aaronsteers:
@cjohnhanson - Thanks for logging this.

Can you say a bit more of the currently observed behavior you seeing re: terminal environment variable precedence?

For instance, does it appear that terminal context is having highest-level resolution priority instead of the (proposed) lowest?

tayloramurphy:
@aaronsteers what's the weight on this and is this something that can be done for 2.0?

cjohnhanson:
@aaronsteers 
> For instance, does it appear that terminal context is having highest-level resolution priority instead of the (proposed) lowest?
^That's exactly right. Terminal context takes precedence over everything else. If its priority is correct, I believe that all 6 of these tests would start passing.

@tayloramurphy I would weight this at an 8, it definitely requires some exploration but I believe the current behavior is the result of several different parts of the code, there's no single "precedence order" set that we could just change.

cjohnhanson:
@aaronsteers (CC @kgpayne--you might have also come to this conclusion or otherwise have some insight here)

After banging my head against this for a long time, I think I've come to the embarrassing conclusion that current behavior is in line with the spec and I misinterpreted the precedence order when writing these tests.

[...]

PR #6217, opened 2022-06-16
Resolve #6207 "Add env as first-class attribute of MeltanoFile and ProjectPlugin classes (not an extra)"

### LINKED ISSUE #6207: Add `env` as first-class attribute of `MeltanoFile` and  `ProjectPlugin` classes (not an `extra`)
**Note:** `env` is already a first-class attribute of the `Environment` class.
**Note:** Just adding these will not make inheritance work (as per #5982 and #5983).

Spec:

- `env` is a first-class attribute of `MeltanoFile` accessible via top-level `env:` key in `meltano.yml`)
- `env` is a first-class attribute of `ProjectPlugin` accessible via plugin-scoped `env:` keys in base and environment-specific config
- `env` is settable via the CLI, including with the '.' syntax (e.g. `meltano --no-environment config meltano set env.MY_ENV_VAR value`)
- New attributes are added to jsonschema, tested and documented

Release:

As both #5982 and #5983 depend on an `env` attribute, they should include these changes in their PR to master. This issue will be closed when either of those issues is merged to `main`.

kgpayne:
Closes #6207

### DIFF
--- a/tests/meltano/api/conftest.py
+++ b/tests/meltano/api/conftest.py
@@ -1,4 +1,5 @@
 import pytest
+
 from meltano.api.models import db as _db
 
 
--- a/tests/meltano/api/controllers/test_orchestration.py
+++ b/tests/meltano/api/controllers/test_orchestration.py
@@ -5,6 +5,7 @@ from asynctest import CoroutineMock, mock
 from flask import Flask, url_for
 from flask.testing import FlaskClient
 from flask.wrappers import Response
+
 from meltano.core.plugin.settings_service import REDACTED_VALUE, SettingValueStore
 
 
--- a/tests/meltano/api/test_workers.py
+++ b/tests/meltano/api/test_workers.py
@@ -1,6 +1,7 @@
 from unittest import mock
 
 import pytest
+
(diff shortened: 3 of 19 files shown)


kgpayne:
@cjohnhanson (cc @pandemicsyn and @edgarrmondragon) before I go much further here, I would love to sense-check the direction. For context, I am working on this issue (adding `env` attributes to MeltanoFile and ProjectPlugin) as a precursor to working on #5982 and #5983. Whilst I am sure I will be able to get the `env` keys working and inheriting as expected by moving much of the functionality that was wedged into `env_override` into `env` properties on the various setting services, that approach has two problems: i) it doesn't make env var resolution any easier to reason about (and debug) and ii) I can't see yet how it will work when I come to implementing stacked inheritance and expansion. More and more I feel like the _storage_ of env vars should be on the `env` attributes of MeltanoFile and ProjectPlugin etc. but the _resolution_ and _expansion_ logic should be handled by a dedicated `EnvVarsService`, rather than spread across SettingService children `env` properties (as in this MR) 🤔 Its a bigger change, but feels like a more robust solution. WDYT?

PR #6245, opened 2022-06-20
Update workflow to add assignment column

tayloramurphy:


tayloramurphy:
I saw https://github.com/meltano/meltano/runs/6970010604?check_suite_focus=true didn't work. I _think_ this might fix it? @edgarrmondragon can you have a looksee?

edgarrmondragon (approved):
@tayloramurphy this _looks_ good but in the [error log](https://github.com/meltano/meltano/runs/6970010604?check_suite_focus=true) I'm also seeing that the project wasn't found:

```
jq: error (at <stdin>:4): Cannot index string with string "html_url"
No project was found.
jq: error (at <stdin>:4): Cannot index string with string "name"
Column name '' is not found.
```

tayloramurphy:
@edgarrmondragon I tried 3 things to see if it would fix it:

* Added MeltyBot to the project (probably not necessary but maybe since it's their token)
* Updated the URL to drop the trailing slash like the examples (though the bash on the command line seems like it should be fine)
* Verified it's a public project. 

So here's hoping I guess.

PR #6249, opened 2022-06-21
feat: Lazy modules & CLI

### LINKED ISSUE #6220: Make imports lazy by default
[PEP-0690](https://peps.python.org/pep-0690/) (draft) will make Python's imports lazy by default. This will significantly speed up the startup time of the Meltano CLI, as most of its time is spent importing modules, many of which it doesn't use for every command. In the interest of not waiting for this PEP that may or may not ever be accepted into Python, we can do it ourselves by using Python's import hooks.

WillDaSilva:
[Recording of the demo on YouTube](https://youtu.be/IqaPHGRUNTA?t=3266)

This PR implements [the lazy importing of modules](https://github.com/meltano/meltano/blob/404402abd1002922bf375c4cfc4fbd7b14e8b567/src/meltano/lazy_import.py) (only applies to modules in the Meltano package, excluding the CLI which relies on import-time side-effects), and a lazy CLI, which makes up for the fact that we would otherwise have to eagerly import the entire CLI by making it so that CLI commands/groups are registered early, but only imported when Click actually needs to run that specific command/group.

My rough measurements on my machine show that this brings the startup time of the Meltano CLI down from ~0.34 seconds to ~0.16 seconds. This is of particular importance for CLI tab-completion (https://github.com/meltano/meltano/pull/6250), wherein the startup time is incurred every time a completion is requested.

Closes #6220


pandemicsyn:
This is super cool, and having seen the video I get how everything works, why we're doing it, and realize this actually adds very little friction when adding a new command for contributors. 

However, I'm a little worried that for new contributors, this would be a little tricky to grok without context. Even if they come across LazyCommand early on, I'm not sure the doc string helps all that much. I wonder if we shouldn't put a basic overview doc comment in `cli.__init__.py` or the like. Might be worth to explicitly mention that if you're adding a new command that you may also need to add it to LazyCommand. 

Definitely curious what other folks think. I'd be happy to be overruled on this - was trying to approach this with a "newbie contributor" POV, but just might not be giving newbies enough credit. 

--
Separately, this also drives home for me that we really need to start considering collecting internal architecture docs. We've talked about a bunch but never got very far. I pop'd https://github.com/meltano/meltano/discussions/6313 to maybe resurrect that.

tayloramurphy:
I'm cool with this change from a Product / end user perspective, but it definitely needs thumbsup from Engineering. I'm not going to approve since there are no docs / Product-specific changes but know that I like the performance gains!

edgarrmondragon (commented):
The changeset looks good, but I worry that we might be making biggish changes for modest gains.

If it's not a big ask, can we have a comparison of different commands with and without these changes?

WillDaSilva:
It seems like the performance gains have been notably reduced now that more of the CLI is instrumented with telemetry. There are still gains, but I no longer feel that they are sufficient to justify the increase in complexity. More work could be done to re-gain the performance improvements, but this change highlights that the performance gains were not as stable as initially expected.

I'm going to close this PR (and its issue) for now. A better approach long-term will probably involve moving most of the core/business logic into an API which can run as a user-level daemon process, then having the CLI call out to it. Doing so should result in even better performance gains that what this PR achieves.

PR #6257, opened 2022-06-21
test: Run & test against Snowplow Micro under Pytest

WillDaSilva:
If Docker/docker-compose is not available, tests using the `snowplow` fixture will be skipped, and tests using the `snowplow_optional` fixture will receive `None` instead of a `SnowplowMicro` instance.

Currently all CLI tests that invoke a command are checked to ensure that no bad events are fired. We also check that an exit event is fired with an appropriate exit code in one case. In the future more tests can be added that check that the events fired have the expected events with their expected contexts.

Closes https://github.com/meltano/meltano/issues/2955


pandemicsyn (approved):
This looks ___great___ , would definitely love to merge this ahead of #6260 so I can test event coverage there. 

@WillDaSilva probably worth getting additional approval for this from @aaronsteers OR @edgarrmondragon

edgarrmondragon (approved):
@WillDaSilva Is it worth adding something to the contributing guide for the new docker-based tests?

WillDaSilva:
#6264 updated the version of `jsonschema` Meltano depends on. Unfortunately the newer version is incompatible with the `docker-compose` package that `pytest-docker` depends on. There's a PR to add support for `docker-compose` v2 (which is not a Python packge): https://github.com/avast/pytest-docker/pull/72

Once it has been merged in, we'll be able to use the newer version of `jsonschema`. Until then, we'll be stuck on an older version.

WillDaSilva:
> Is it worth adding something to the contributing guide for the new docker-based tests?

@edgarrmondragon Done! https://deploy-preview-6257--meltano.netlify.app/contribute/tests

WillDaSilva:
Looks like merging in `main` has caused tests failures on GitLab: https://gitlab.com/meltano/legacy-ci/meltano/-/pipelines/573076915

This PR will be a draft until those have been sorted out. Re-running the jobs does not appear to fix them in GitLab. One of the GitHub test jobs failed in the same way, but re-running it fixed it. Additional re-runs on the GitHub test jobs might result in additional failures. It is odd that it mostly passed on GitHub, but totally failed on GitLab.

docs/getting-started.md
```diff
@@ -21,7 +21,7 @@ _To learn more about the different installation methods, refer to the [Installat
 
 ### Local Installation
 
-If you're running Linux, macOS, or Windows and have [Python](https://www.python.org/) 3.7, 3.8 or 3.9 installed,
+If you're running Linux, macOS, or Windows and have a supported version of [Python](https://www.python.org/) installed,
```
afolson, line 24: I think we should be explicit in the versions we support. I'm not sure we mention supported versions elsewhere and I don't expect people to go look at the repo to figure it out.

docs/getting-started.md
```diff
@@ -780,10 +780,13 @@ This can be useful for debugging a failing extractor or loader.
 Now that you've successfully run your first data integration (EL) pipeline using Meltano,
 you have a few possible next steps:
 
-- [Schedule pipelines to run regularly](#schedule-pipelines-to-run-regularly)
-- [Transform loaded data for analysis](#transform-loaded-data-for-analysis)
-- [Containerize your project](#containerize-your-project)
-- [Deploy your pipelines in production](#deploy-your-pipelines-in-production)
+- [Example](#example)
```
afolson, line 783: What's the purpose of nesting these this way?

WillDaSilva: There is no purpose. Annoyingly, this change was made without my knowledge by the VS Code Markdown extension. It thought that the list was a table of content, and so it automatically reformatted it. I have disabled the `toc.updateOnSave` setting, and now it has been fixed. Good catch.

docs/getting-started.md
WillDaSilva: I prefer writing docs that aren't guaranteed to fall out of date 1-2 times per year as we add/remove supported Python versions, but I'll compromise here since having users need to check `pyproject.toml` or https://pypi.org/project/meltano/ for the version info while they're reading the getting-started guide is not ideal. In generally we should prefer not duplicating this information in many places if we can help it.

docs/getting-started.md
afolson: Haha well damn, I was hoping there was a cool secret here but that works too. Thanks!

PR #6260, opened 2022-06-22
chore: refactor to simplify command tracking

### LINKED ISSUE #6041: Refactor command tracking event/context to be initialized earlier

pandemicsyn:
Transition to tracking and firing an "automatic" start command. Doing it at the top level of the cli (e.g. `cli/cli.py or cli/__init__.py`) didn't end up as easy as I thought. How click builds up its context and parses options and args as each group and sub-command is encountered and resolved made that pretty annoying. This was complicated a bit by our need to snag option keys. Click does expose all the cli params via things like `click.utils.get_os_args()` but those are raw and unparsed:

```
click.utils.get_os_args()
Out[2]: ['--log-level', 'debug', 'run', '--dry-run', 'tap-gitlab', 'target-jsonl', 'dbt:test']
```

Trying to reparse/sanitize those just didn't seem worth it. So, I ended up settling on this approach:

1. I initialize the two trackers in `cli/cli.py` so that the two tracker objects are available for everything downstream, and individual commands no longer have to initialize them.
2. I added a `InstrumentedCmd` class that wraps `click.Command`. Commands we wish to instrument use this to override the default `click.Command` class. I stole this idea from the `DefaultGroup` class override we already use. It does two things:
   1. Add `CliContext` to our embedded tracker 
   2. Fires the cli start event  
3. Finally, I added `CliContext.from_click_context` that ...creates a `CliContext` from a `click.Context`. It builds up a list of **all** options/args encountered and passes those along as a new `Options` field (schema TBD, just using a dict for illustration) consisting of the key/value pairs of params. For ALL levels of click groups encountered, so this also includes `--log-level` and friends. If the value is a `bool`, `int`, or `NoneType` I pass the value along unhashed - but in all other cases I hash the value. 
4. Extends the cli event schema to add a new `inflight` status, so commands can periodically report context updates while in flight.

ex usage:

```python
@cli.command(
    cls=InstrumentedCmd, # <------ InstrumentedCmd use here
    short_help="very helpful message here"
    ...
)
@click.option(
    "--dry-run",
    is_flag=True,
)
@pass_project(migrate=True)
@click.pass_context
async def run(
    ctx: click.Context,
    project: Project,
    dry_run: bool,
):
```

This PR currently includes ~3~ 4 examples:

- Use in `meltano schedule` - this is net-new as `schedule` currently only has legacy tracking, with NO custom exit events.
- Use in `meltano run` - a command without a nested click group, keeping the existing custom exit events but switching to using `InstrumentedCmd`.
- Use in `meltano job` - a command with a nested click group, keeping the existing custom exit events but switching to using `InstrumentedCmd`.
- Use in `meltano elt` - illustrating using `inflight` cli command event to report ProjectContext early. 

If we like this pattern, I can update the rest of the code base to use `InstrumentedCmd` to fire cli start events. Not removing the existing custom exit events gives us a chance to validate that the generic exit events will work for us moving forward - before we remove all the custom events.

Side note:

I played around with overloading `click.Options` to provide a way to signal when an option is a "sensitive" flag that needs to have its values hashed in telemetry. Wasn't sure that was worth the hassle given we only have a few string flags that aren't what we now consider PII (log-level, elt transform flag, etc). We could probably just whitelist a few flags specific keys (like `log-level`) in `CliContext.from_click_context`.

closes #6041


src/meltano/core/tracking/iglu-client-embedded/schemas/com.meltano/cli_context/jsonschema/1-0-1
```diff
@@ -0,0 +1,54 @@
...
+        "option_keys": {
+            "description": "",
+            "type": "array",
+            "items": {
+                "description": "",
+                "type": "string",
+                "maxLength": 256
+            }
+        },
+        "options": {
```
pandemicsyn, line 34: @pnadolny13 prefs on how you'd like this new field to look? Here are two examples of how it currently looks:

```json
# from a `meltano --log-level=warning run --dry-run tap-gitlab target-jsonl`
"data": {
                            "command": "run",
                            "sub_command": null,
                            "option_keys": [],
                            "options": {
                                "dry_run": true,
                                "blocks": "e1c3e8486805f40d42c1887b4e8928c417d8d3a998b3e27a3e9b48b1dac4aa7a",
                                "full_refresh": false,
                                "no_state_update": false,
                                "force": false,
                                "database_uri": null,
                                "log_level": "4bd9354bb652233497a0f42647464b81b83a83654fcf1511263a5d0120d74c91",
                                "log_config": null,
                                "verbose": 0,
                                "environment": null,
                                "no_environment": false
                            }
                        }
```

```json
# from a `meltano --log-level=warning schedule list --format=text`
"data": {
                            "command": "schedule",
                            "sub_command": "list",
                            "option_keys": [],
                            "options": {
                                "format": "982d9e3eb996f559e633f4d194def3761d909f5a3b647d1a851fead67c32c9d1",
                                "database_uri": null,
                                "log_level": "4bd9354bb652233497a0f42647464b81b83a83654fcf1511263a5d0120d74c91",
                                "log_config": null,
                                "verbose": 0,
                                "environment": null,
                                "no_environment": false
                            }
                        }
```

WillDaSilva (commented):
This seems pretty great. Thanks for implementing this @pandemicsyn

src/meltano/cli/utils.py
```diff
@@ -510,3 +511,15 @@ def check_dependencies_met(
...
+class InstrumentedCmd(click.Command):
+    """A click.Command that automatically fires an instrumentation 'start' event, if a tracker is available."""
+
+    def invoke(self, ctx):
+        """Fire at start event and then invoke the requested command."""
+        ctx.ensure_object(dict)
+        if ctx.obj.get("tracker"):
+            ctx.obj["tracker"].add_contexts(CliContext.from_click_context(ctx))
+            ctx.obj["tracker"].track_command_event(CliEvent.started)
+        super().invoke(ctx)  # noqa: WPS608
```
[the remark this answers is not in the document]
pandemicsyn: Yea! When we go to yoink out the bespoke exit events we can definitely extend this.

src/meltano/cli/elt.py
```diff
@@ -279,6 +268,7 @@ async def _run_elt(
         try:
             elt_context = context_builder.context()
             tracker.add_contexts(PluginsTrackingContext.from_elt_context(elt_context))
+            tracker.track_command_event(CliEvent.inflight)
```
pandemicsyn, line 271: example usage of a new "inflight" cli event. 

/cc @pnadolny13

pandemicsyn:
Ok, I want to keep these PRs digestible for reviewers, so I'm breaking this change up into two. This PR updates all the **currently** instrumented commands to use this new cleaned up pattern. 

I'll send along another separate PR that updates all the remain commands to automatically start firing a start event, and also add `inflight` events where needed. If https://github.com/meltano/meltano/pull/6257 has landed by then I can also add some test coverage to make sure start events are being emitted.

@pnadolny13 heads up I updated `elt` , `invoke`, and `install` to emit `inflight` events so that you get a `PluginsContext` as soon as its available (and don't have to depend on the completed/aborted events).

src/meltano/core/tracking/iglu-client-embedded/schemas/com.meltano/cli_context/jsonschema/1-0-1
```diff
@@ -0,0 +1,54 @@
...
+        "version": "1-0-1"
+    },
+    "type": "object",
+    "properties": {
+        "command": {
+            "description": "",
+            "type": "string",
+            "maxLength": 1024
+        },
+        "sub_command": {
```
WillDaSilva, line 17: Shouldn't `sub_command` be a `$ref` to the schema for `command`, since they can be arbitrarily nested?

WillDaSilva (changes requested):

src/meltano/core/tracking/iglu-client-embedded/schemas/com.meltano/cli_context/jsonschema/1-0-1
pandemicsyn: Unless I'm misunderstanding, they're not arbitrarily nested atm - we're reporting one level.

WillDaSilva: On second thought we probably don't want to report a `sub_command` at all. Instead we probably want multiple command contexts: one for each (nested) command. We could add an `InstrumentedGroup` class that gives us info about a higher-level command, then use `InstrumentedCommand` to give info about the base command.

This would replace the approach that uses `CliContext.from_click_context`. Instead of building a `CliContext` from a Click context, we would have it built inside of `InstrumentedGroup`/`InstrumentedCommand`, and then stored in the Click context as a list under the key `snowplow_contexts` or something like that.

Note that `InstrumentedGroup` would be a subclass of `InstrumentedCommand`, just like how `click.Group` is a suclass of `click.Command`.

Something like `meltano --top-group-level-flag service --mid-group-level-flag start --command-level-flag service_name` would be broken into:
- a command context detailing `meltano --top-group-level-flag`
- a command context for `service --mid-group-level-flag`
- a command context for `start --command-level-flag service_name`

And then with those contexts in place (by `tracker.with_contexts(...)`) the CLI events would fire, and would only contain the event (e.g. `started`, `failed`, `inflight`, etc.). All the other data would be in the contexts.

WillDaSilva: > This also feels like very "late breaking" PR review feedback, I don't know if this approach is objectively better or pedantically better and just pop'd up because this PR's been open so long and you and I've gotten to think about it for so long.

What I described above is what I originally had in mind when I suggested that we hook into Click to get the CLI context. Sorry for the late feedback. This PR fell off my radar, and I've been a bit overwhelmed since - I have not been thinking about this PR until it popped up when you merged `main` into it.

I understand your reservations with the proposed change, and agree that at some point we need to accept what we have got and move on. That said, with the current approach, how can we deal with nested commands that have more than a single layer? I suppose we can use `CliContext.from_click_context::_recursively_collect_params` to collect it all, then flatten it, but I'm concerned that flattening it will make use lose valuable information. We can't associate options/arguments/flags with the command/sub-command that they were applied to.

The current schema has a `sub_command` field, but that would only work up to one layer deep if not for the special case in place (`if ctx.parent and ctx.parent.command.name != "cli":`) which lets it work two layers deep.

Since the commands have an inherently nested structure imposed by Click's concepts of groups/commands, the data we collect from them should also be nested. Whether it goes into one `CliContext` with a nested schema, or multiple with flat schemas does not matter to me, but the loss of information and inconsistency does.

pandemicsyn:
flipped back to draft for now.

src/meltano/core/tracking/iglu-client-embedded/schemas/com.meltano/cli_context/jsonschema/1-1-0
```diff
@@ -0,0 +1,44 @@
...
+        "version": "1-1-0"
+    },
+    "type": "object",
+    "properties": {
+        "command": {
+            "description": "",
+            "type": "string",
+            "maxLength": 1024
+        },
+        "parent_command_hint": {
```
pandemicsyn, line 17: we can infer it from list of the contexts, but thought it was handy to just explicitly include a ref to what the upstream parent was.

src/meltano/cli/utils.py
```diff
@@ -514,3 +516,37 @@ def check_dependencies_met(
...
+
+class InstrumentedGroup(click.Group):
+    """A click.Group that instruments its invocation by updating the telemetry context."""
+
+    def invoke(self, ctx):
+        """Update the telemetry context and invoke the group."""
+        ctx.ensure_object(dict)
+        if ctx.obj.get("tracker"):
+            ctx.obj["tracker"].add_contexts(CliContext.from_click_context(ctx))
+        super().invoke(ctx)  # noqa: WPS608
```
pandemicsyn, line 540: @WillDaSilva so rather than what you had mentioned:

> Instead of building a CliContext from a Click context, we would have it built inside of InstrumentedGroup/InstrumentedCommand, and then stored in the Click context as a list under the key snowplow_contexts or something like that.

> And then with those contexts in place (by tracker.with_contexts(...)) the CLI events would fire, and would only contain the event (e.g. started, failed, inflight, etc.). All the other data would be in the contexts.

I kept the add_contexts pattern rather than adding to the click context, and then having `with_contexts` since we create the tracker early on in `cli.py` anyway. 

As long as we have a tracker obj - any instrumented Group or Command will end up managing/adding its own CliContext, and things firing CliEvents don't have to think about contexts at all. 

If we kaboom halfway through the cli resolution chain, we might at least still have *some* cli context's this way as well.

pandemicsyn:
Update of what the new cli contexts looks like when attached to an event generated for an invocation of a command like `meltano --log-level=debug schedule list --format=json`:


Screen Shot 2022-06-29 at 6 11 08 PM

WillDaSilva (approved):
Looks great! I'm glad we can handle the CLI telemetry in the background with these changes, rather than within each CLI function.

src/meltano/cli/utils.py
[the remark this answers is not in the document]
WillDaSilva: https://github.com/meltano/meltano/issues/6324

pnadolny13:
@pandemicsyn this looks good to me! So if I understand correctly this will give us input options and an extra inflight event at the time a plugin is executed, which is particularly useful for invoke where we previously didnt get anything until the end.

I think walking through the schema changes really quick will help me understand better if I need to make changes to accomadate. I dont have any comment on the code changes itself, whatever is best for you all.

pnadolny13 (approved):
@pandemicsyn Looks good to me!

pandemicsyn:
Reminder to myself - need to update Snowcat cloud when I merge this.

edgarrmondragon (approved):

PR #6263, opened 2022-06-22
docs: Fix link in coverage badge

edgarrmondragon:
Link was pointing to `Test` workflow.


WillDaSilva (approved):
> Link was pointing to Test workflow.

That was intentional because that was where we collected our coverage info, and there's no easy way to link to the latest run of a particular job on a particular branch. I tried finding a link like this for Codecov, but [their front page](https://about.codecov.io/) is not particularly helpful in that regard.

Thanks!

tayloramurphy (approved):

PR #6278, opened 2022-06-23
feat(core): formalize settings alias behavior and disallow conflicting values between setting name and aliases

cjohnhanson:
This PR updates the settings alias behavior in the following ways:
- Updates the behavior to compare the environment variables and not the actual values themselves.
- Returns a more informative error if a setting's value is set via multiple names/aliases
- Updates documentation to better describe aliases behavior
- Updates the output of `meltano config` commands to include the specific environment variable that a setting's value is derived from (rather than just listing `(from the environment)`)

Closes #3135


docs/src/_guide/configuration.md
```diff
@@ -68,14 +68,72 @@ export <PLUGIN_NAME>_<SETTING_NAME>=<value>
...
+Plugins can also specify alternative variables ([aliases](#aliases) for their settings, to match existing usage or variables expected by plugin executables. You can use [`meltano config <plugin> list`](/reference/command-line-interface#config) to list all available settings for a plugin along with their variables, in order of precedence.
 
 Since environment variable values are always strings, Meltano will cast values to the appropriate type before passing them on to the plugin.
 
 To verify that any environment variables you've set will be picked up by Meltano as you intended, you can test them with [`meltano config <plugin>`](/reference/command-line-interface#config) before running [`meltano elt`](/reference/command-line-interface#elt) or [`meltano invoke`](/reference/command-line-interface#invoke).
 
 To learn how to use environment variables to specify pipeline-specific configuration, refer to the [Data Integration (EL) guide](/guide/integration#pipeline-specific-configuration).
 
+#### aliases
```
tayloramurphy, line 79:
```suggestion
#### Settings Aliases
```

docs/src/_guide/configuration.md
```diff
@@ -68,14 +68,72 @@ export <PLUGIN_NAME>_<SETTING_NAME>=<value>
...
+    - name: password
+      kind: password
+    - name: my_custom_tap_username
+      aliases: [custom_tap_username, username]
+```
+
+Within a given configuration layer, a setting can be set via only a single name, whether that name is its canonical name or one of its aliases.
+So given the custom extractor defined above, the `my_custom_tap_username` setting could be set via the `MY_CUSTOM_TAP_MY_CUSTOM_TAP_USERNAME` environment variable or either the `MY_CUSTOM_TAP_CUSTOM_TAP_USERNAME` or `MY_CUSTOM_TAP_USERNAME` variables.
+But if more than one of these variables is set in the terminal environment, then an exception will be raised--even if all the relevant environment variables have the same value.
+
```
tayloramurphy, line 108: Can we add a note in here about being able to reference the alias when using `meltano config`? Basically making it clear that 

`meltano config <plugin> set <name>`

and

`meltano config <plugin> set <alias>`

are both ok to do.

docs/src/_guide/configuration.md
```diff
@@ -68,14 +68,72 @@ export <PLUGIN_NAME>_<SETTING_NAME>=<value>
...
+Setting value set via multiple environment variables: ['MY_CUSTOM_TAP_CUSTOM_TAP_USERNAME', 'MY_CUSTOM_TAP_USERNAME']
+```
+
+If the values for the multiple environment variables differ, the error message will also list what the values are:
+
+```shell
+$ export MY_CUSTOM_TAP_USERNAME=some_username
+$ export MY_CUSTOM_TAP_CUSTOM_TAP_USERNAME=some_other_username
+$ meltano config my-custom-tap
+Conflicting values for setting found in: {'MY_CUSTOM_TAP_CUSTOM_TAP_USERNAME': 'some_other_username', 'MY_CUSTOM_TAP_USERNAME': 'some_username'}
```
tayloramurphy, line 134: Do we need to print the actual values to the command line? What if this is for a password and we've just printed it out into logs potentially? It would probably suffice that we just say they are in conflict, right?

docs/src/_guide/configuration.md
```diff
@@ -127,12 +185,14 @@ When Meltano invokes a plugin's executable as part of [`meltano elt`](/reference
...
 ```yml
 environments:
   - name: dev
     env:
       AN_ENVIRONMENT_VARIABLE: dev
 ```
+
 Any plugins run in that Meltano environment will then have the provided environment variables populated into the plugin's environment.
 
 ## Multiple plugin configurations
```
tayloramurphy, line 198: There's a yml section just below this that has highlighting of specific rows using `{8-18}`. Can we remove that here too since it currently doesn't work?

tayloramurphy: I made https://github.com/meltano/meltano/issues/6287 to track adding that feature back.

tayloramurphy:
@cjohnhanson @aaronsteers I wasn't able to get my mac setup in time to test this locally. I did a pass on the docs but would appreciate a thorough check on the behavior. Thoughts on recording a short video to show the desired behavior?

aaronsteers:
@cjohnhanson - We've updated PR naming guidelines. Do you mind updating the PR title here to match the conventional commit syntax. This is basically to changelog line with a prefix according to type ("feat", in this case, I think).

https://docs.meltano.com/contribute/merge#semantic-prs

cjohnhanson:
@edgarrmondragon you mind taking a look at this this afternoon?

CHANGELOG.md
```diff
@@ -12,10 +12,15 @@ This project adheres to [Semantic Versioning](http://semver.org/) and [Keep a Ch
...
 ### Changes
+
 - [#3174](https://gitlab.com/meltano/meltano/-/issues/3174) - We now recommend pipx over manually created virtual environments.
 - [#3450](https://gitlab.com/meltano/meltano/-/issues/3450) - Windows instllation instructions added
 - [#3367](https://gitlab.com/meltano/meltano/-/issues/3367) - Windows now covered in CI for some basic cases
+<<<<<<< 3135-alias-behavior
+- [#3135](https://github.com/meltano/meltano/issues/3135) - Formalize settings alias behavior and disallow conflicting values between setting name and aliases
+=======
 - [#2035](https://github.com/meltano/meltano/issues/2035) Rename `job` table and `job_id` column in backend db
+>>>>>>> main
```
aaronsteers, line 23: @cjohnhanson - fyi merge resolution artifacts ("<<<<<", "=====", ">>>>>>")

src/meltano/core/settings_store.py
```diff
@@ -50,6 +50,28 @@ def __str__(self) -> str:
...
+        self.names = names
+        super().__init__(names)
+
+    def __str__(self) -> str:
+        """Represent the error as a string.
+
+        Returns:
+            string representation of the error
+        """
+        return f"Setting value set via multiple environment variables: {self.names}"
```
aaronsteers, line 72: Read on its own, I wonder if a user might not immediately understand the context. 

Should we have the word "error"?:

> Error: Setting value set via multiple environment variables: ..."

or a normative assertion or complaint could work here too:

> Setting value cannot not be determined due to conflicting environment variables: ..."

cjohnhanson: Good point, I've updated it to add the word "Error".

aaronsteers:
@cjohnhanson - One question on scope here.

We previously made a decision during spec discussions that aliases and names would both be valid for _sourcing_ the setting value, but that the auto-generated context _sent_ to the plugin context would only use the env variable that was based on the `name`. As of this PR, do you know if we are sending _all_ variants of the env var down to the plugin context or just the `name`-based env var?

So, if `dbt-bigquery` has a setting with `{name: project_name, aliases: [database], ...}`, then when dbt-bigquery` is running, it would only get `DBT_BIGQUERY_PROJECT_NAME` and it _would not_ see `DBT_BIGQUERY_DATABASE`, even though both are valid for the user's perspective when sourcing the value.

I don't want to add scope here, so either answer is totally fine. I'm happy to log this (admittedly much lower pri) requirement into a new issue if still needed in follow-up.

aaronsteers:
> @aaronsteers - if the value for the setting is set via the "canonical" name, then _only_ the "canonical" environment variable gets passed through. If the value for the setting is set via an alias, then _both_ the alias and the canonical environment variable get passed through. Does that make sense?

Perfect. Thanks! That's the desired behavior, I think. 👍

aaronsteers (approved):
Would be helpful to have one additional approval besides mine, but for my part, this looks really good. Thanks!

cjohnhanson:
@pandemicsyn -- I know @edgarrmondragon is pretty busy with the release today do you have some extra capacity to give this a quick :eyes:? I'd love to get this merged in before I head out for PTO.

pandemicsyn (approved):
lgtm! also had to take it for a spin:

``` 
(melty-3.8) ➜ MELTANO_SEND_ANONYMOUS_USAGE_STATS=false meltano config meltano list | grep USAGE
send_anonymous_usage_stats [env: MELTANO_SEND_ANONYMOUS_USAGE_STATS] (default: True) current value: False (from the MELTANO_SEND_ANONYMOUS_USAGE_STATS variable in the environment)
```
Nice! Thats gonna be super handy for integration tests.

PR #6284, opened 2022-06-24
ci: auto changelog when creating release branch

aaronsteers:
Applies the GitHub action from here: https://github.com/saadmk11/changelog-ci

This was chosen because:

1. It has a relatively high number of stars (95 as of now)
1. It does not try to auto-detect version or auto-bump the version stored in files. (We have existing processes that control those.)
1. It is simple and written in Python, would be trivial to fork or extend.

As of now, a first implementation would be inserted into the release branch workflow, before running the bump version script.


.github/semantic.yml
```diff
@@ -12,6 +12,8 @@ types:
   - ci
   - chore
   - build
+  - break
```
aaronsteers, line 15: dummy change (will revert)

```suggestion
  - break  
```

aaronsteers:
This is low/no risk, since it is only affecting CI.

Also - the pipeline won't be testable until at least one merge to `main`.

I will merge and then continue tests in a newly-created release PR.

cc @edgarrmondragon, @pandemicsyn

edgarrmondragon:
@aaronsteers one trick I've used to test new workflows is to create a fork, add the remote  locally, make the PR branch the default in the fork and push commits to the fork to iterate

PR #6293, opened 2022-06-25
feat: add support for Python 3.10, including CI tests

### LINKED ISSUE #2830: Support Python 3.10

WillDaSilva:
Thanks @edgarrmondragon for getting this started in https://github.com/meltano/meltano/pull/6135

I would like to wait until #6155 has been merged in so that we can run the Windows tests in CI for these changes.

I'm not sure if `feat` is the correct semantic PR category to use for this, but using `feat` will cause this change to be listed under `New` in the changelog, which seems appropriate.

This PR addresses every warning that was being raised by Pytest.

Closes #2830


aaronsteers (commented):
@WillDaSilva - Do you mind also taking a pass at docs here? 

I think we should probably update references like this one that reference specific python versions: https://docs.meltano.com/guide/installation#python-37-38-or-39

Optionally, we could also add a suggestion that we still recommend 3.9 for best compatibility with the widest number of plugins.

edgarrmondragon (approved):

aaronsteers (approved):

PR #6297, opened 2022-06-27
fix: Require write lock when accessing Project.meltano attribute

### LINKED ISSUE #6240: Exception on /api/v1/orchestrations/pipeline-schedules AttributeError: 'NoneType' object has no attribute 'get'
I get this error sometimes when strating `meltano ui` on a windows machine.  The error also happens when I navigate between tabs in the UI.  The error is associtated with the following API call in the log
```
Exception on /api/v1/orchestrations/pipeline-schedules [GET]
```
Here is the ending bit of the error 

```
  File "C:\development\meltano\src\meltano\core\project_plugins_service.py", line 83, in current_plugins
    self._current_plugins = self.config_service.current_meltano_yml.plugins
  File "C:\development\meltano\src\meltano\core\config_service.py", line 35, in current_meltano_yml
    self._current_meltano_yml = self.project.meltano
  File "C:\development\meltano\src\meltano\core\project.py", line 230, in meltano
    return MeltanoFile.parse(self.project_files.load())
  File "C:\development\meltano\src\meltano\core\project_files.py", line 66, in load
    included_file_contents = self._load_included_files()
  File "C:\development\meltano\src\meltano\core\project_files.py", line 150, in _load_included_files
    for path in self.include_paths:
  File "C:\development\meltano\src\meltano\core\project_files.py", line 59, in include_paths
    include_path_patterns = self.meltano.get("include_paths", [])
AttributeError: 'NoneType' object has no attribute 'get'
```

It can't find the attribute `get` which I think is suppose to be a function of the class ?  I also have not been able to determine what the function `get` is suppose to do.  It seems like it is reading the `meltano.yaml` for the `include_paths` entries and return all the paths in a list.  In the `project_files.py` there are these back to back function you can see there is a `meltano` function.

```
    @property
    def meltano(self):
        """Return the contents of this projects `meltano.yml`."""
        if self._meltano is None:
            with open(self._meltano_file_path) as melt_f:
                self._meltano = yaml.safe_load(melt_f)
        return self._meltano

   @property
    def include_paths(self) -> List[Path]:
        """Return list of paths derived from glob patterns defined in the meltanofile."""
        include_path_patterns = self.meltano.get("include_paths", [])
        return self._resolve_include_paths(include_path_patterns)
```

In `project.py` there is a `meltano` function that returns a `MeltanoFile` class

```
    @property
    def meltano(self) -> MeltanoFile:
        """Return a copy of the current meltano config.

        Returns:
            the current meltano config
        """
        with self._meltano_rw_lock.read_lock():
            return MeltanoFile.parse(self.project_files.load())

```
The `MeltanoFile` class that is in `meltano_file.py` has a `get_plugins_for_mappings`.

cjohnhanson:
Closes #6240


BuzzCutNorman:
@cjohnhanson I am getting the same error.  I did the testing by switching to you PR in VS Code that showed up on my local fork after doing a `git fetch upstream`.  Then did a `poetry install` and `poetry shell`.  After navigating to the `test-meltano` project I ran `meltano --log-level ui` and navigated the UI.  I have attached the console output for your review.  I did not delete all the __pycache__ folders so their might be some left overs.  Please let me know if there is a clear way to test or this was an ok way.
[buzzcutnorman-meltano-ui-console-none-type-20220627.log](https://github.com/meltano/meltano/files/8995766/buzzcutnorman-meltano-ui-console-none-type-20220627.log)

BuzzCutNorman:
@cjohnhanson I was looking around and found this article about race conditions and locks on variables:

[How to use the Python Threading Lock to Prevent Race Conditions](https://www.pythontutorial.net/advanced-python/python-threading-lock/)

I made the following changes to the `project_files.py` and it seemed to resolve the issue.

I added this include 

`from threading import Lock`

Then edited the `__init__` to create a `self.lock`

```
    def __init__(self, root: Path, meltano_file_path: Path):
        """Instantiate ProjectFiles interface from project root and meltano.yml path."""
        self.lock = Lock()
        self.root = root.resolve()
        self._meltano = None
        self._meltano_file_path = meltano_file_path.resolve()
        self._plugin_file_map = {}
```

Then I added the `acquire()` and `release()` to the `load()` and `update()` class functions:
```
    def load(self) -> dict:
        """Load all project files into a single dict representation."""
        # meltano file may have changed in another process, so reset cache first
        self.lock.acquire()
        self.reset_cache()
        included_file_contents = self._load_included_files()
        self.lock.release()
        return deep_merge(self.meltano, included_file_contents)

    def update(self, meltano_config: dict):
        """Update config by overriding current config with new, changed config.

        Note: `.update()` will write blank entities for those no longer in use (i.e. contained config on load, but not on save).
        """
        self.lock.acquire()
        file_dicts = self._split_config_dict(meltano_config)
        for file_path, contents in file_dicts.items():
            self._write_file(file_path, contents)

        unused_files = [fl for fl in self.include_paths if str(fl) not in file_dicts]
        for unused_file_path in unused_files:
            self._write_file(unused_file_path, BLANK_SUBFILE)
        self.reset_cache()
        self.lock.release()
        return meltano_config
```

I had added them to the `property meltano()` , `property include_paths()`, and `reset_cache()` but with all five in place when you call `meltano --log-level debug ui` it wouldn't even start.  

I apologize in advance I know there is probably a better way to share this code but I just don't know how to do it yet.

BuzzCutNorman:
@cjohnhanson I was afraid I might have messed up the testing.  I ran the command you asked while in my `meltano` folder then while in that environment I changed the the `test-meltano` directory and ran `meltano --log-level debug ui` and then navigated the UI.  I still am getting the error but maybe again I didn't  setup the environment correctly.  Here is the console log containing what I did.
[buzzcutnorman-meltano-ui-console-none-type-20220628.log](https://github.com/meltano/meltano/files/9005603/buzzcutnorman-meltano-ui-console-none-type-20220628.log)

cjohnhanson:
@BuzzCutNorman thanks for digging into this, your explanation makes total sense to me. I applied your suggested changes in this branch. Assuming tests pass, I'll mark this as ready and get a review from someone else on the team and then hopefully we can get this out the door.

@visch if you had a chance to test this on your Windows machine, that would be super helpful too.

Thanks again for all your digging and hard work here!

BuzzCutNorman:
@cjohnhanson  I am glad I was on the right track and my rambling helped.  Thank you for the education on threads (context mangers, and fasteners).  I honestly thought the read_lock() and write_lock() were for files not threads and processes because the only times I saw them was when meltano files were being read from or written to.

cjohnhanson:
@pandemicsyn would you mind taking a look at this?

pandemicsyn:
@cjohnhanson @BuzzCutNorman dropped a note on the issue too but fwiw - i was able to repro this on my Mac. That aside,

So the exception is buried over in project_files.py:

```
  File "/Users/syn/projects/meltano/src/meltano/core/project_files.py", line 59, in include_paths
    include_path_patterns = self.meltano.get("include_paths", [])
AttributeError: 'NoneType' object has no attribute 'get'
```

The only thing that should be setting `self._meltano = None` is `reset_cache`:

https://github.com/meltano/meltano/blob/859083b8c76c1b5d88ba30dbefb1b475a08f1fc4/src/meltano/core/project_files.py#L84-L86

That in turn gets called in two spots `load` & `update`,

https://github.com/meltano/meltano/blob/859083b8c76c1b5d88ba30dbefb1b475a08f1fc4/src/meltano/core/project_files.py#L62-L67

https://github.com/meltano/meltano/blob/859083b8c76c1b5d88ba30dbefb1b475a08f1fc4/src/meltano/core/project_files.py#L69-L82

The load race is always present I think, since any concurrent calls via `self.project_files.load()` or `update()` could trigger it, and high request load or system latency would make it more likely to happen.

https://github.com/meltano/meltano/blob/859083b8c76c1b5d88ba30dbefb1b475a08f1fc4/src/meltano/core/project.py#L235-L236

@cjohnhanson the write lock does fix it/cover it up since it prevents concurrent reads, but feels like that's fixing a symptom and not a cause - and I wonder if that has perf implications for when you're just using cli.

Guessing at least part of the problem is how we patched in Uvicorn support and cheated to start asyncio event loops in parts of the Flask code base. 

Side note: Makes sense considering whats happening, but if increase the Uvicorn worker count it does happen less frequently.

pandemicsyn:
> Guessing at least part of the problem is how we patched in Uvicorn support and cheated to start asyncio event loops in parts of the Flask code base.

🤔 Maybe thats not entirely true, the exceptions actually happen even when calling paths outside of those that we patched when we added Uvicorn support. So might just be a side effect of uvicorn allowing more concurrency and exposing the underlying issue.

BuzzCutNorman:
@pandemicsyn I am glad you could reproduce and validate this on a non windows machine.

The  `GET /api/v1/orchestrations/pipeline-schedules ` call was the one that usually triggered the error for me and like you mentioned it doesn't have any of the asynio loop calls.  I read something in passing that flask v2 puts each new request in a new thread.  I have not done more research to see if that is true or not but wanted to mention it.

Here are two things I notice:

When I put a logger in the `load()` function to show how many times it is called I was surprised.  

When reading about the `fasteners.ReaderWriterLock()` locks the write_lock() makes sure all queued up calls are dealt with in a FIFO method writers first then readers.  While the read_lock() doesn't state the same guarantee.  Since they are all writes now the queue make sure they all get addressed in the order they came in.

pandemicsyn:
@BuzzCutNorman i think we've identified the issue, the problem is more how to properly resolve it. project_files.py in its current form isn't concurrency safe (and originally wasn't mean to), but forcing everything into a FIFO like you'd get by forcing a write lock is also not ideal.

cjohnhanson:
@BuzzCutNorman building on @pandemicsyn's comment--I think we need to do at least one of three things here:
1. Make `project_files.py` completely thread safe. This is a pretty heavy lift. It will be worth the investment long-term but is likely not a top priority for us in the short-term.
1. Fully determine the performance implications of the write lock that we currently have in this PR. If we can confirm that this locking approach doesn't have significant performance implications, then we'd be more comfortable rolling this change out to all users.
1. Protect the write lock we have here behind the `enable_uvicorn` feature flag so that if there _are_ significant performance implications, only users that are using this experimental uvicorn support will be impacted. This is the best short-term solution but still requires some small amount of additional effort beyond what's in this PR.

Currently the plan is to move forward with option (3) in the short term and then spawn follow-up issues for (1) and (2). That said, this will not be getting merged in time for this release or next week's release, so I'm going to move this PR back to `Draft` status for the time being.

cjohnhanson:
@BuzzCutNorman FYI I'm closing this out since this was resolved in https://github.com/meltano/meltano/pull/6466

Issue #6218, opened 2022-06-15
New process and guidelines for integration tests (replaces `demo-project` and `homepage_sample_code` tests in GitLab CI


aaronsteers:
From a call with @tayloramurphy and @DouweM, we started thinking of replacing `demo-project` with a set of sample Meltano projects in `meltano/meltano` that can serve as integration tests, perhaps along with specific scripts that would run - such as the invoking of specific plugins.

I'm sure there are better and more sophisticated approaches, but one approach would be a convention have a `meltano.yml`, a `script.sh`, and a `meltano.after.yml` which should match to the results. Another approach would be a set of pytest tests that perform a series of operations against (a copy of?) the source `meltano.yml` file and confirm the results (or errors), possibly including specific telemetry events that are expected to be sent over the course of those script commands.

pandemicsyn:
> I'm sure there are better and more sophisticated approaches, but one approach would be a convention have a meltano.yml, a script.sh, and a meltano.after.yml which should match to the results. Another approach would be a set of pytest tests that perform a series of operations against (a copy of?) the source meltano.yml file and confirm the results (or errors), possibly including specific telemetry events that are expected to be sent over the course of those script commands.

@aaronsteers Yep, even in the simple approach, we can spin up a snowplow-micro instance, and configure meltano to point telemetry at it. Run our `script.sh` integration script, and then at the end of the test check that snowplow micro reports only successful events. Probably wouldn't take much beyond a little grep/jq magic to also verify that we have the expected start/completed events for the commands we ran. That would make for a great general purpose safety net for telemetry. 

This setup  can also fairly easily grow to accommodate automated UI browser tests I'd think. Including checking that js fired telemetry gets emitted as expected.

pandemicsyn:
We've chatted a bit in the past about having more examples in the docs as well, and we had some issues with doc drift leading up 2.0. I've been thinking about that a bit and wanted to float an idea.  Using a setup like @aaronsteers described, I think we can actually kill two birds with one stone.

We can get more examples for users - that are actually maintained and always work, and new and improved integration tests. The only real change would be that instead of having a `script.sh` file we treat our test orchestrations as living and usable markdown docs.  Users can use them as examples, and we can use them as scripts to execute via [mdsh](https://github.com/bashup/mdsh).

Contrived example, but if you wanted test and document how to transition from using `elt` to `run` you might have something like:

````markdown
# transition-from-elt-to-run.md

[...]

PR #6303, opened 2022-06-28
ci: basic integration test framework

### LINKED ISSUE #6181: Add folder of `meltano.yml` samples to validate against JSON Schema rules
> Another thing we should consider, I think, is having some reference samples in a `tests/resources/meltano-yml/` folder which themselves are validated against the JSON Schema definition linked above using some pytest code which would fail CI if examples do not conform to the schema.
>
> cc @pandemicsyn, @tayloramurphy, @pnadolny13

_Originally posted by @aaronsteers in https://github.com/meltano/meltano/issues/6170#issuecomment-1154646911_

The way we'd use this, I think, is that any time we add a new major feature which effects what is stored in yaml, we'd _also_ add/update one or more sample files, and those sample files would be tested against the JSON Schema.

pandemicsyn:
A modified (imho slighty improved) take on the integration framework proposed in #6218.

Example run: https://github.com/meltano/meltano/actions/runs/2578001545
And inline below:

Screen Shot 2022-06-28 at 3 12 42 PM

Dir/file layout:

- `docs/example-library/<test-name>` holds the markdown guide/test and the meltano YAML files. I opted to include both the starting and ending YAML files here. Including the "ending" or resulting YAML config was something that I saw https://buf.dev did in their guides that was subtle but appreciated (since it lets impatient users like me jump ahead). Since we need this file to diff against anyway I went a head included there.
  -  **Open Question**: Do we want to make these actual pages on the docs site ? If so I'll probably need to adjust the paths some and include the Jekyll preamble in the index.md's.
- `integration/` holds the integration framework. 
  - `integration/validate.sh` the main entry point for all integration tests. It takes care of producing the script from the markdown, injecting a logging YAML, diffing the meltano yaml's, and calling the additional per-test validations when required.
  - `integration/<test-name>/validate.sh` - optional additional tests to run for a test. Often, this won't even be needed, but the ability is available if you want to run extended checks (i.e. grepping logs, running sql, etc)

---

 Status: ~~still a WIP (hence being a draft still) so don't need an exhaustive review, but looking for early feedback (cc @aaronsteers @afolson ) to make sure I'm on the right path.~~ Good to go!

---

Updated to now also include:

- shell linter check based on [`shellcheck`](https://github.com/koalaman/shellcheck) for integration scripts
- jsonschema validation check of sample meltano.yml's. closes #6181


docs/example-library/meltano-run/index.md
```diff
@@ -0,0 +1,120 @@
...
+```shell
+meltano install
+```
+
+## Configure an extractor
+
+The installed meltano.yml already includes a configured instance of the `tap-gitlab` extractor for this example. You can view
+the configuration of the extractor by running:
+
+```shell !
```
pandemicsyn, line 15: example of how to skip a shell block. Adding a ` !` preserves syntax highlighting but signals to mdsh that this block shouldn't included.

docs/example-library/meltano-run/index.md
```diff
@@ -0,0 +1,120 @@
...
+
+Lets go ahead add a `job` for this that we can reference directly:
+
+```shell
+meltano job add gitlab-to-postgres --tasks "tap-gitlab target-postgres"
+```
+
+You can reference jobs in `meltano run` invocations directly:
+
+```shell !
```
pandemicsyn, line 59: Another one we skip, its useful for the user to validate they're following instructions and having a working config, but for our test it would just waste cycles.

docs/example-library/meltano-run/index.md
```diff
@@ -0,0 +1,120 @@
...
+meltano config dbt-postgres set dbname warehouse
+meltano config dbt-postgres set schema analytics
+```
+
+### Prep the `dbt` transform for this demo
+
+```shell
+mkdir ./transform/models/tap_gitlab
+touch  ./transform/models/tap_gitlab/source.yml
+```
```
pandemicsyn, line 82: We could just include these files in test directory, and it would "just work" for our tests. But I wanted to illustrate that even sort of tricky operations work.

integration/meltano-run/validate.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
pandemicsyn, line 1: Completely contrived example just to illustrate what you could do here.

.github/workflows/check_jsonschema.yml
```diff
@@ -0,0 +1,31 @@
...
+        architecture: x64
+
+    - name: Install check-jsonschema
+      run: |
+        pipx install check-jsonschema
+        check-jsonschema --version
+
+    - name: Validate example meltano yaml files
+      run: |
+        check-jsonschema docs/example-library/*/meltano.yml docs/example-library/*/ending-meltano.yml --schemafile schema/meltano.schema.json
```
aaronsteers, line 31: Nice 👍

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,68 @@
...
+  workflow_dispatch:
+    inputs: {}
+
+jobs:
+  validate:
+    strategy:
+      matrix:
+        include:
+        - { integration_test: "meltano-basics", needs_postgres: false }
+        - { integration_test: "meltano-run", needs_postgres: true }
```
aaronsteers, line 17: Can we add a reminder somewhere in `CONTRIBUTING.md`?

 or perhaps contributor-related docs in `docs/example-library/README.md` that mention registering the example here in the GitHub CI matrix?

aaronsteers (approved):
One request on my side to perhaps beef up contributor docs - either in the examples readme or in the contributors guide. 

At least one more approval would be helpful here in case I've missed anything important. 

Thanks!

pandemicsyn:
> One request on my side to perhaps beef up contributor docs - either in the examples readme or in the contributors guide.

Absolutely, just wanted to make sure i was on the right path before going through and documenting all the things.

.github/workflows/integration_tests.yml
pandemicsyn: 👍 the readme already links out to the contributor docs - was planning to document it there if this all looked good.

aaronsteers:
> Theres still a larger open question of exactly when these should run and what all tests we want to add out of the gate, but i think as a "get some scaffolding up and running" this is ready to review/merge.

Given the (elapsed) runtime is just 2.5 minutes, I think it's totally fine to run on each build.

Based on: https://github.com/meltano/meltano/actions/runs/2584765141

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,68 @@
...
+      env:
+        PIP_CONSTRAINT: .github/workflows/resources/constraints.txt
+      run: |
+        pip install pip
+        pip --version
+
+    - name: Install Dependencies
+      run: |
+        poetry env use "3.8"
+        poetry install
```
aaronsteers, line 49: In the long run would be good to replace `poetry install` with something like `pip install .` or `pipx install .`. Could expose issues related to the build process and/or dev dependencies being installed along with the prod dependencies.

In the long run, again, we could also sequence this behind the job @edgarrmondragon is working on which pre-builds the dist wheels and then install from the built wheels.

pandemicsyn:
> Given the (elapsed) runtime is just about 2.5 minutes, I think it's totally fine to run on each build.

Yea was actually more concerned about the total volume of CI minutes used if we have a ton of tests , but I just realized this isn't a private repo so the smaller billing limits don't apply 🥳

edgarrmondragon (approved):
This is really nice! 🚀

.github/workflows/integration_tests.yml
aaronsteers: @edgarrmondragon also found when working on the SDK that a migration from `tox` to `nox` would give an option to run in an environment that only uses the release libs. (Best in a separate issue.)

Issue #3277, opened 2022-03-29
Improve Makefile behavior


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3354

Originally created by @pandemicsyn on 2022-03-29 03:16:59

---

Our Makefile doesn't have a help option AND defaults to executing a build if invoked without arguments. Makefile help/docs can be generated if you run `make explain_makefile` but you have to know that exists (or read the Makefile, in which case you no longer need the explainer). 
With some very minor changes (essentially this bit from https://gitlab.com/pandemicsyn/llamalake/-/blob/main/Makefile#L7 + a few inline comments), we could provide a help option, and use that as the default target:

```
(melty-3.8) ➜ make
api                            build api
base_image                     builds meltano/base
build                          build the ui and api
bundle                         clean up static assets and rerun UI build
clean                          clean statically generated assets
docker_images                  build docker base_image and prod_image
docker_sdist
docs_image
explain_makefile               launch a web server further explaining this makefile.
freeze_db
lint                           run python and eslint tasks
lint_eslint                    run eslint
lint_python                    run json/yaml validation and execute tox to perform python lint tasks
lock                           run poetry lock
prod_image                     builds meltano/meltano the AIO production image that includes static UI artifacts
release                        perform a release, should only be performed as part of the CI pipeline
show_lint                      show python and eslint ralated commands
test                           run tests
ui                             build the UI
```

And explicitly support `help` arg via

```

(melty-3.8) ➜ make help
api                            build api
base_image                     builds meltano/base
build                          build the ui and api
bundle                         clean up static assets and rerun UI build
clean                          clean statically generated assets
docker_images                  build docker base_image and prod_image
docker_sdist
docs_image
explain_makefile               launch a web server further explaining this makefile.
freeze_db
lint                           run python and eslint tasks
lint_eslint                    run eslint
lint_python                    run json/yaml validation and execute tox to perform python lint tasks
lock                           run poetry lock
prod_image                     builds meltano/meltano the AIO production image that includes static UI artifacts
release                        perform a release, should only be performed as part of the CI pipeline
show_lint                      show python and eslint ralated commands
test                           run tests
ui                             build the UI
```

PR #6305, opened 2022-06-28
chore: Delete outdated files & merge configs into `pyproject.toml`

### LINKED ISSUE #6262: Clean-up, delete, and consolidate top-level files/directories in the repository
The top-level of the Meltano repository is cluttered. This serves to hide our readme when one visits https://github.com/meltano/meltano. For that reason and more, we should probably delete the outdated/unused files and directories, and consolidate configs where possible.

To (potentially) delete:
- [x] `.gitlab/` once the migration to GitHub Actions is complete
- [x] `cloud/`
- [x] `license_finder/` if it is no longer used after the migration away from GitLab CI is complete. Even if we still need it, it probably shouldn't be at the top level. @kgpayne
- [ ] `.adr-dir` @pandemicsyn what is the point of this file? Does some tool we use depend on it?
- [x] `.dockerignore` if it is not doing anything. If I'm not mistaken the `.dockerignore` file needs to be at the root of the build context to have any effect, and as far as I know we don't use the top-level of the Meltano repository as a Docker build context.
- [x] `.env.dev`
- [x] `.env.example`
- [x] `.gitlab-ci.yml` once the migration to GitHub Actions is complete
- [ ] `.tmuxinator.yml`
- [x] `Caddyfile`
- [x] `Makefile_explain.sh`
- [ ] `alembic.ini` if Alembic is no longer needed
- [x] `docker-compose.yml`
- [x] `setup.cfg`

Aside from that, we can move some config into `pyproject.toml` (n.b. `pyproject.toml` aims to store the configs of more or less all Python dev tools):
- If we are willing to write [`pflake8`](https://github.com/csachs/pyproject-flake8) instead of `flake8`, then we can put the flake8 config (`.flake8`) in `pyproject.toml`
- [`pytest.ini` can be moved into `pyproject.toml`](https://docs.pytest.org/en/latest/reference/customize.html#pyproject-toml)
- [`.coveragerc` can be moved into `pyproject.toml`](https://coverage.readthedocs.io/en/latest/config.html)
  > Coverage.py will read settings from other usual configuration files if no other configuration file is used. It will automatically read from “setup.cfg” or “tox.ini” if they exist. In this case, the section names have “coverage:” prefixed, so the [run] options described below will be found in the [coverage:run] section of the file. If coverage.py is installed with the toml extra (pip install coverage[toml]), it will automatically read from “pyproject.toml”. Configuration must be within the [tool.coverage] section, for example, [tool.coverage.run].

@aaronsteers @tayloramurphy @pandemicsyn @edgarrmondragon

WillDaSilva:
Closes #6262 

I've gone ahead and deleted `.gitlab-ci.yml` with this. Do we still depend on the GitLab pipelines for anything important? If so, are there issues open to track the migrate of these CI elements over to GitHub workflows/actions?


WillDaSilva:
Coverage is handled separately in https://github.com/meltano/meltano/pull/6304

edgarrmondragon:
@WillDaSilva This removes the lint checks that were running on gitlab ci, but I'm not sure if we want to simply adapt the old implementation (which had some drawbacks) or pursue a different approach.

WillDaSilva:
> This removes the lint checks that were running on gitlab ci, but I'm not sure if we want to simply adapt the old implementation (which had some drawbacks) or pursue a different approach.

@edgarrmondragon I'll open an issue to address that, and switch this PR to a draft.

pandemicsyn (changes requested):
You'll want to remove the refs for tmuxinator from the docs too: https://docs.meltano.com/contribute/prerequisites#tmuxinator 

I don't think we actually want to nuke `Makefile_explain.sh` ,its used by the Makefile to "generate help". Side note: https://github.com/meltano/meltano/issues/3277 exists to make the `make` experience a little better. 

It's probably worth having a look and making sure that some of these other files also aren't referenced else where in the project.

WillDaSilva:
Superseded by:
- https://github.com/meltano/meltano/pull/6837
- https://github.com/meltano/meltano/pull/6838
- https://github.com/meltano/meltano/pull/6839
- https://github.com/meltano/meltano/pull/6842
- https://github.com/meltano/meltano/pull/6843
- https://github.com/meltano/meltano/pull/6847

PR #6308, opened 2022-06-29
feat: Preserve comments and flow style in `meltano.yml`

### LINKED ISSUE #2133: Preserve comments and flow style when programmatically modifying `meltano.yml` (Ruamel)
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/2177

Originally created by @DouweM on 2020-07-17 16:23:10

---

PyYAML doesn't support this currently (https://github.com/yaml/pyyaml/issues/90), but https://pypi.org/project/ruamel.yaml does.

Note:

Our usage of the Canonical class may also be a factor we need to consider in our implementation. Noted by @kgpayne below: https://gitlab.com/meltano/meltano/-/issues/2177#note_708071594

edgarrmondragon:
Progress:

- [x] Multi-line strings are preserved.
- [x] Key order is preserved
- [x] Comments are preserved in all data structures

Closes #2133


edgarrmondragon:
> I've added one more task to the description: "Key order is preserved"

@DouweM Thanks, I've marked it as completed and changed the fixture a bit to confirm.

edgarrmondragon:
@aaronsteers

> Environments, schedules, etc., which get shuffled around from multiple files, are losing their comments, e.g. a comment at the `environments` level. Fixing that will be the true heavy lifting.

continuing our conversation from slack, I'll try to time-slot support for those shuffled bits but I agree that if that does not work it may be fine to split the work in two PRs.

edgarrmondragon:
@kgpayne tagging you for review since `ProjectFiles` is your art 😃

tests/fixtures/multifile_project/subconfig_2.yml
```diff
@@ -1,3 +1,4 @@
+# Top-level comment
```
kgpayne, line 1: Is it worth adding comments to the array elements of the subfile, to make sure comments don't go missing when arrays are merged in the single-dict representation?

src/meltano/core/project_files.py
```diff
@@ -214,11 +295,39 @@ def _split_config_dict(self, config: dict):
...
+        self._copy_yaml_attributes(file_dicts)
         return file_dicts
 
-    @staticmethod
-    def _write_file(file_path, contents):
+    def _copy_yaml_attributes(  # noqa: WPS210
+        self,
+        file_dicts: dict[str, CommentedMap],
+    ):
+        for file, file_dict in file_dicts.items():
```
kgpayne, line 309: There is a bit going on in here, as this seems to do most of the comment preservation work. Do you mind adding a docstring with a brief description for future skimming?

kgpayne (commented):
This is amazing 🤘Just a couple of small suggestions.

I tried to introduce breakages in a local project too and couldn't generate any failures. Had forgotten that we raise an explicit error if users try to spread environments over multiple files `[critical ] Plugin with path "environments:dev" already added in file /Users/kp/Projects/demo/2133-ruamel/subconfig_2.yml.
Duplicate plugin name found.` which stopped me from creating intentional merge conflicts (in comments and values) 😈. Seems robust to me 👏

tests/fixtures/multifile_project/subconfig_2.yml
edgarrmondragon: > Is it worth adding comments to the array elements of the subfile, to make sure comments don't go missing when arrays are merged in the single-dict representation?

@kapayne good call. I'll add that to the fixture 👍

edgarrmondragon: K, so I found that this was actually a real good test case, because it breaks the order in the second file, since the order itself is just copied from the main file. Gotta fix that.

tayloramurphy (changes requested):
@edgarrmondragon I'm testing this locally and I'm not seeing key order preserved. (I'm basically following my speedrun https://github.com/meltano/internal-marketing/issues/388#issuecomment-1155389364 )

My test case was to set the job using the CLI, then I manually moved that key above environments. When I set a schedule (or any other config), the `jobs` key is then moved to its original location.

I can confirm that that comments are preserved. Multi-line seems to work too, I tested with the `|` literal block and it kept it. I also used `>` and it did not keep the newline but replaced it with a space, which I think is all we can really expect.

edgarrmondragon:
> When I set a schedule (or any other config), the `jobs` key is then moved to its original location.

@tayloramurphy note that

- #6312

So, I was hesitant to add support for `jobs` in this PR. But, I just pushed a commit to support that key, so this will also close the issue linked above. Can you try again?

tayloramurphy:
@edgarrmondragon I'll test it, but it's not clear to me based on the scope of the issue or the PR description that this was a limitation of the implementation. What isn't in scope for this issue?

tayloramurphy:
@edgarrmondragon that doesn't seem to have fixed it. 

I've added my job and manually moved the key above `environments` to generate the following `meltano.yml`:

```yaml
version: 1
default_environment: dev
project_id: 90715e56-ec79-4ca8-b1d3-123a2f74fd49
plugins:
  extractors:
  - name: tap-github
    variant: singer-io
    pip_url: tap-github
    select:
    - commits.*
  loaders:
  - name: target-snowflake
    variant: transferwise
    pip_url: pipelinewise-target-snowflake
    config:
      account: epa06486
      dbname: USERDEV_RAW
      user: TAYLOR
  mappers:
  - name: transform-field
    variant: transferwise
    pip_url: pipelinewise-transform-field
    mappings:
    - name: hash-url-data
      config:
        transformations:
        - field_id: html_url
          tap_stream_name: commits
          type: HASH
jobs:
- name: github_data
  tasks:
  - tap-github hash-url-data target-snowflake
environments:
- name: dev
  config:
    plugins:
      extractors:
      - name: tap-github
        config:
          repository: >
            meltano/meltano  meltano/hub
          start_date: '2022-06-02T00:00:00Z'
- name: staging
- name: prod
```

Running `meltano schedule add run_github_commits --job github_data --interval "@daily"` generates:

```yaml
version: 1
default_environment: dev
project_id: 90715e56-ec79-4ca8-b1d3-123a2f74fd49
plugins:
  extractors:
  - name: tap-github
    variant: singer-io
    pip_url: tap-github
    select:
    - commits.*
  loaders:
  - name: target-snowflake
    variant: transferwise
    pip_url: pipelinewise-target-snowflake
    config:
      account: epa06486
      dbname: USERDEV_RAW
      user: TAYLOR
  mappers:
  - name: transform-field
    variant: transferwise
    pip_url: pipelinewise-transform-field
    mappings:
    - name: hash-url-data
      config:
        transformations:
        - field_id: html_url
          tap_stream_name: commits
          type: HASH
schedules:
- name: run_github_commits
  interval: '@daily'
  job: github_data
environments:
- name: dev
  config:
    plugins:
      extractors:
      - name: tap-github
        config:
          repository: >
            meltano/meltano  meltano/hub
          start_date: '2022-06-02T00:00:00Z'
- name: staging
- name: prod

jobs:
- name: github_data
  tasks:
  - tap-github hash-url-data target-snowflake
```

Curiously if I move `version`, `default_environment`, `project_id` elsewhere in a file they get moved back to their original location when I add the schedule as well.

edgarrmondragon:
> Curiously if I move `version`, `default_environment`, `project_id` elsewhere in a file they get moved back to their original location when I add the schedule as well.

@tayloramurphy ok so in the case of Meltano settings, I made the call to keep those always at the top, otherwise when doing `meltano config meltano set <setting> <value different from default>` they will be added at the bottom of `meltano.yml`, but seems to me that we wanna call attention to those at first glance by putting them at the top, but if that's not the case then the latest commit should address that.

tayloramurphy:
@edgarrmondragon so I'd say you're right that the default structure of `meltano.yml` for Meltano settings should be at the top. My confusion is that the user is required to keep it there even if they move the keys or re-order them. 

I think my expectation for the ruamel feature would be that if a user changes the order of their `meltano.yml` file that we would make a best effort attempt to respect that ordering no matter the key. That said, I think it's fine to carve out Meltano-specific settings and have an explicit requirement that they be at the top of the root meltano.yml file in a  given order - I'm guessing that makes it a bit easier to implement? 

I would like for all _other_ keys though to have their positions respected. My goal here is in having sane diffs for a user. If they're adding a schedule that shouldn't create a crazy diff because `jobs` was also moved by us. 

So `plugins`, `environments`, `schedules`, and `jobs` (plus all of their sub keys) are in scope for preserving the positioning. 

Is that reasonable? Do we need to do some of this in a separate PR?

edgarrmondragon:
> I think my expectation for the ruamel feature would be that if a user changes the order of their `meltano.yml` file that we would make a best effort attempt to respect that ordering no matter the key.

@tayloramurphy That's the right expectation with ruamel. However, the way `include_paths` work is by merging all the contents into a single object, and then splitting that into the corresponding original files, so they're essentially rebuilt every time a command updates `meltano.yml`. That messes up comments and ordering for all the top-level keys (`plugins`, `default_environment`, etc.) and required special handling to make sure they're both preserved.

> That said, I think it's fine to carve out Meltano-specific settings and have an explicit requirement that they be at the top of the root meltano.yml file in a given order - I'm guessing that makes it a bit easier to implement?

It's actually easier to implement if we don't care about new project settings being added at the bottom of the file by `meltano config meltano set ...`.

> So `plugins`, `environments`, `schedules`, and `jobs` (plus all of their sub keys) are in scope for preserving the positioning.
> 
> Is that reasonable? Do we need to do some of this in a separate PR?

That's reasonable. In retrospect, it would've made sense as a separate PR but the work there is almost complete.

tayloramurphy:
> It's actually easier to implement if we don't care about new project settings being added at the bottom of the file by meltano config meltano set ....

@edgarrmondragon Well in that case then let's just drop the carve out then 😄  I don't think it's a bad user experience to say "running `meltano config meltano set ...` will add the entry to the bottom of your file but we encourage you to organize it how you see fit" and then we do our best to not actually move anything. 

Thanks for all the context. Let me know when you're ready for another review 👍

edgarrmondragon:
> Let me know when you're ready for another review

@tayloramurphy ok this is ready for another pass 🙏

tayloramurphy:
@edgarrmondragon it looks good based on my testing. I tested moving jobs around and adding schedules and it worked. I also moved the `project_id` to a random spot and added a schedule and that is preserved. 

I also tested moving a schedule to a new file called `/schedules/schedules.meltano.yml` and added 

```yaml
include_paths:
- ./schedules/*.meltano.yml
```

to my `meltano.yml`. When I added a new schedule it created a new `schedule` key in the main `meltano.yml` instead of adding it to the existing key. And when I went to remove the schedule in the `schedules.meltano.yml` I got this:

```bash
❯ meltano schedule remove run_github_commits
2022-07-07T19:46:50.212402Z [info     ] Environment 'dev' is active
2022-07-07T19:46:50.695048Z [critical ] Could not update meltano.yml: '/Users/taylor/Documents/Projects/dev/meltano/testruamel/schedules/schedules.meltano.yml'
'/Users/taylor/Documents/Projects/dev/meltano/testruamel/schedules/schedules.meltano.yml'
```

Do you think that's related to this PR or a separate issue?

edgarrmondragon:
> When I added a new schedule it created a new `schedule` key in the main `meltano.yml` instead of adding it to the existing key.

That is expected. Multi-files has always worked like that: new items go to `meltano.yml`.

> Do you think that's related to this PR or a separate issue?

Definitely related. I just pushed a commit with a test case and a fix for that.

@tayloramurphy thank you for QA'ing this thoroughly 🙏

pandemicsyn (approved):
not gonna lie, had to pull this one down and do some manual tests because following this one along only via the GH diff was a little tricky 😁 

looks solid though! Being able to add inline comments to the meltano.yml's in the example library/integrations is gonna super nice for users.

tayloramurphy:
@edgarrmondragon ok, one more question. `meltano schedule remove` did remove the schedule but the `meltano.yml` went from 

```yaml
schedules:
- name: run_github_commits
  interval: '@daily'
  job: github_data
```

to

```yaml
plugins: {}
schedules: []
jobs: []
environments: []
```

which I guess makes sense b/c presumably you have empty values for each of those? I guess I would expect that it would either have just `schedules: []` or just an empty file. When I remove all schedules from the main file the key just disappears. But perhaps referencing an empty meltano.yml file from the main one wouldn't work?

I'm okay to not fix that particular quirk if we can pop an issue on this weird UX.

tayloramurphy (approved):
Approving as the functionality looks good though there are some UX quirks we'll want to follow-up on afterwards.

edgarrmondragon:
@tayloramurphy yeah, I kept the behavior of files left empty after changes mostly the same. So, for sure if we want to do something different that's probably best left for a follow-up

tayloramurphy:
@edgarrmondragon I made https://github.com/meltano/meltano/issues/6389 to track.

Issue #6170, opened 2022-06-12
Add 'jobs' to the meltano project json schema


JulesHuisman:
The `jobs` key currently shows an error in `meltano.yaml`. This is because it is missing in the [schema definition](https://github.com/meltano/meltano/blob/main/schema/meltano.schema.json).

<img width="222" alt="image" src="https://user-images.githubusercontent.com/26525996/173252897-3f0c4827-785c-4d15-9ee8-c0b84ac491df.png">

aaronsteers:
@JulesHuisman - I ran into this over the weekend too. Thanks for logging!

vicradon:
@aaronsteers 
I'll love to take on this issue. Do you mind guiding me on how I can contribute?

pnadolny13:
@JulesHuisman @aaronsteers I also noticed this last week. A couple others I saw that might be worth fixing all in one go are: mappers and "commands" keys. I think for "commands" we allow a dict of args/executable or a plain old command string but the json schema is requiring the args/executable, which isn't correct.

aaronsteers:
Hi, @JulesHuisman . This inline validation, and also the help text and auto-complete features, are driven by our JSON schema definitions. 

Specifically this file: https://github.com/meltano/meltano/blob/main/schema/meltano.schema.json

JulesHuisman:
Hi @aaronsteers, I can create a PR for this. Unless you want to pick this up @vicradon?

pnadolny13:
>Yes, I think so. And possibly also the things that @pnadolny13 mentioned above.

@vicradon @JulesHuisman it would be awesome if you got all of them fixed up but its up to you whether you want to pull in those additions I mentioned. I also agree with @aaronsteers that it would we great to have a way to assert some examples to prove our schemas are working as expected, since theyre quite large.

vicradon:
Working on it now @pnadolny13 🙂

JulesHuisman:
> @JulesHuisman Sorry to ask, but which object should I add the 'jobs' key to? Is it under `properties > definitions > environment > properties > config > properties > plugins`? I took a screenshot to verify.

I think the jobs are at the root level, so on the same level as `plugins`, `base_settings` and `schedules`. But it might be better to ask someone from the meltano team, as they probably know the schema inside out. :)

pnadolny13:
@vicradon following up on our [our slack thread](https://meltano.slack.com/archives/C013Z450LCD/p1655311751483699). Again thanks for taking this on!

[...]

PR #6309, opened 2022-06-29
fix: update schedule schema entry to support job ref and add a jobs entry

pandemicsyn:
Update's the schedule schema to support jobs and adds a shiny new jobs entry.

Should address:

-  https://github.com/meltano/meltano/issues/6170

### DIFF
--- a/schema/meltano.schema.json
+++ b/schema/meltano.schema.json
@@ -103,6 +103,13 @@
         "$ref": "#/definitions/schedules"
       },
       "description": "Scheduled ELT jobs, added using 'meltano schedule'"
+    },
+    "jobs": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/jobs"
+      },
+      "description": "Jobs, added using 'meltano job'"
     }
   },
   "definitions": {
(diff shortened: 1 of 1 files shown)


pandemicsyn:
Passes a local check-jsonschema validation, which checked all the random job and schedule permutations i've accumulated so far 😁 :

```
 meltano git:(fix-jobs-and-schedule-mel) check-jsonschema ~/projects/meltano-projects/rundev/meltano.jobs.yml --schemafile schema/meltano.schema.json
ok -- validation done
```

schema/meltano.schema.json
```diff
@@ -762,6 +776,45 @@
...
+      ],
+      "description": "https://docs.meltano.com/reference/command-line-interface#job",
+      "additionalProperties": false,
+      "properties": {
+        "name": {
+          "type": "string",
+          "description": "This jobs unique name",
+          "examples": ["gitlab-to-jsonl"]
+        },
+        "tasks": {
```
pandemicsyn, line 793: this is actually lifted straight from the validation we do when adding a task via the cli: https://github.com/meltano/meltano/blob/main/src/meltano/core/task_sets.py#L17

schema/meltano.schema.json
```diff
@@ -717,13 +724,15 @@
...
-      "required": [
-        "name",
-        "extractor",
-        "loader",
-        "transform",
-        "interval",
-        "start_date"
+      "oneOf": [
+        {"required": ["name", "job", "interval"]},
+        {"required": [
```
vicradon, line 729: @pandemicsyn please why does it take one of the two?

JulesHuisman:
Looks good, thanks for picking this up!

edgarrmondragon (approved):

PR #6327, opened 2022-06-30
feat: Resolve #3419 "Change default meltano config behavior to --no-environment"

### LINKED ISSUE #3419: Change default `meltano config` behavior to `--no-environment`
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3503

Originally created by @aaronsteers on 2022-05-18 03:06:00

---

I keep running into this and I think it's worth reconsidering the defaults here.

In CI - my CLI commands looked fine and only after 'cat'ing my `meltano.yml` file was I able to diagnose the issue here.

```console
$ meltano config tap-gitlab set _pip_url git+https://github.com/MeltanoLabs/tap-gitlab.git@main
2022-05-18T00:23:28.361467Z [info     ] Environment 'dev' is active
Extractor 'tap-gitlab' setting '_pip_url' was set in `meltano_environment`: 'git+https://github.com/MeltanoLabs/tap-gitlab.git@main'
$ cat meltano.yml
version: 1
default_environment: dev
send_anonymous_usage_stats: false
plugins:
  extractors:
  - name: tap-gitlab
    variant: meltanolabs
    pip_url: git+https://github.com/MeltanoLabs/tap-gitlab.git
environments:
- name: dev
  config:
    plugins:
      extractors:
      - name: tap-gitlab
        pip_url: git+https://github.com/MeltanoLabs/tap-gitlab.git@main
```

This one is particularly sticky because 'pip_url' at the environment level seems to be a noop.

kgpayne:
Closes #3419

- #3419

This PR:
- adds `deactivate_environmnent` to the `Project` class
- deactivates the active environment for the `meltano config` ~~and `meltano select`~~ commands if the current active environment was sourced from the `default_environment` key (i.e. no `--environment` was chosen explicitly)
- adds an `info` log line to inform users of this behaviour and to direct them to use `--environment` to target an environment
- updates the docs with the new behaviour


afolson (approved):

kgpayne:
@aaronsteers @tayloramurphy as per the updated PR description, this also opts `meltano select` out of the `default_environment` (in addition to `meltano config`). I found when testing that `select` is often used alongside `config` and is essentially just another kind of config. We don't _have_ to treat these commands the same, and its simple to revert if prefer that `select` should still adhere to the `default_environment` 👍

edgarrmondragon (approved):

src/meltano/cli/config.py
```diff
@@ -81,6 +84,13 @@ def config(  # noqa: WPS231
         tracker.track_command_event(CliEvent.aborted)
         raise
 
+    if ctx.obj["is_default_environment"]:
+        logger.info(
+            f"Deactivated Environment '{project.active_environment.name}' (the default) as the `meltano config` command does not adhere to your projects `default_environment`. "
```
edgarrmondragon, line 89: This wording sounds a little weird to me. It makes it seem like the specific value of `default_environment` is incompatible with `meltano config`, while it's just that the setting is ignored for `meltano config`. So how about

```suggestion
            f"The default environment ({project.active_environment.name}) will be ignored for `meltano config`."
```

src/meltano/cli/select.py
```diff
@@ -75,6 +80,13 @@ async def select(
 
     \b\nRead more at https://docs.meltano.com/reference/command-line-interface#select
     """
+    if ctx.obj["is_default_environment"]:
+        logger.info(
+            f"Deactivated Environment '{project.active_environment.name}' (the default) as the `meltano select` command does not adhere to your projects `default_environment`. "
+            + "To configure a specific Environment, please use option `--environment=<environment name>`."
+        )
```
edgarrmondragon, line 87: Ditto the comment above

edgarrmondragon (approved):
@kgpayne this looks to me. I just left some comments on the logged messages.

tayloramurphy (changes requested):
@kgpayne I think it could make sense to change the default for `select` to be no-environment as well, but I'd like to keep this PR focused on just `config`. If you could open a discussion on this topic with the argument around `select` and possibly other topics that'd be helpful. We can pull each command into separate issues as we figure that out.

For this PR, can you also update https://docs.meltano.com/reference/command-line-interface#using-config-with-environments to make it clear the default behavior?

tayloramurphy (changes requested):

docs/example-library/meltano-basics/index.md
```diff
@@ -17,8 +17,8 @@ an explicit start_date from which you would like tap-gitlab to start importing d
 that the resulting jsonl files NOT actually include timestamps in their file names:
 
 ```shell
-meltano config tap-gitlab set start_date 2022-04-25T00:00:01Z
-meltano config target-jsonl set do_timestamp_file false
+meltano --environment=dev config tap-gitlab set start_date 2022-04-25T00:00:01Z
```
tayloramurphy, line 20: @kgpayne I'd like this test case to be updated to include setting the config globally. I think you could drop https://github.com/meltano/meltano/blob/main/docs/example-library/meltano-basics/meltano.yml#L11 and then add the configuration to validate it does it globally. 

As it stands the integration tests aren't capturing the new behavior.

tayloramurphy (changes requested):
I've tested locally and it seems to work. I've opened:

- https://github.com/meltano/meltano/issues/6390 which highlights a new thing I think we should do. 

I'm also not too keen on having:

```
2022-07-08T19:48:50.200082Z [info     ] Environment 'dev' is active
2022-07-08T19:48:50.566516Z [info     ] The default environment (dev) will be ignored for `meltano config`. To configure a specific Environment, please use option `--environment=<environment name>`.
``` 

which seem to be saying conflicting things. I'll file a follow-up issue to address that UX. We may need to move the invocation of that not at the start of the CLI but only once we know what command you're trying to do.

docs/example-library/meltano-basics/index.md
tayloramurphy: Looks like you just added it! Approving!

tayloramurphy (approved):
Test case is added!

PR #6341, opened 2022-07-01
refactor: update to automatically fire start AND end events

### LINKED ISSUE #6041: Refactor command tracking event/context to be initialized earlier

### LINKED ISSUE #6324: Handle the firing of CLI exit events automatically
Context: https://github.com/meltano/meltano/pull/6260#discussion_r903905100

pandemicsyn:
This PR updates the `InstrumentedCmd` wrapper to automatically fire exit events in addition to start events. 

The 10 existing commands that have bespoke exit events (i.e. they fire custom `CliEvent.aborted` or `CliEvent.failed`) are updated to use a new `PartialInstrumentedCmd` wrapper that continuous to *only* fires start events automatically.  Logged https://github.com/meltano/meltano/issues/6342 for a future cleanup pass. That covers these commands specifically:

- src/meltano/cli/add.py
- src/meltano/cli/config.py
- src/meltano/cli/elt.py
- src/meltano/cli/environment.py
- src/meltano/cli/install.py
- src/meltano/cli/invoke.py
- src/meltano/cli/job.py
- src/meltano/cli/lock.py
- src/meltano/cli/run.py
- src/meltano/cli/schedule.py

All other commands are updated use `InstrumentedCmd` which now fires both start AND exit events.

closes #6324
closes #6041 

> **Note**
Here's a sample of a CliEvent.failed that was fired automatically when I executed a `meltano state move not-an-id okthx`: https://gist.github.com/pandemicsyn/dc780f4a0d1cd9983042105ba656ba6b


edgarrmondragon:
@pandemicsyn do bespoke exit events in commands not conflict with the new base command class? Don't we receive duplicate events? Or is #6342 meant as a _quick_ follow-up?

Nit: should the semantic type be `refactor` instead of `chore`?

pandemicsyn:
>@pandemicsyn do bespoke exit events in commands not conflict with the new base command class? Don't we receive duplicate events? Or is https://github.com/meltano/meltano/issues/6342 meant as a quick follow-up?

Nope we're all good, theres no duplication of exit events for these. The 10 commands with bespoke exit events where already reliant on automatic start events, so I switched them to use a `PartialIntstrumentedCmd` class that *only* fires start events, it doesn't fire exit events.

When we remove the bespoke exit events we can switch the commands over to use `InstrumentedCmd` and remove `PartialInstrumentedCmd`.

https://github.com/meltano/meltano/issues/6342 is not meant/required to be a quick follow up. I'd actually rather we take our time with #6342 because while we're removing it - we shoud also take the time and see which of those try/except that where added to capture telemetry events should raise clearer errors or generate more end-user friendly errors.

edgarrmondragon (approved):

WillDaSilva (approved):
Looks good to me!

Something we may want to do in a future PR: right now we have some of our CLI comments implemented by `async` functions. What if we were to implement them all using `async` functions, and then run them in a wrapper that yields from them, but also fires an inflight event automatically each time the CLI function yields? This further takes the burden of implementing telemetry away from the dev who is adding/updating a CLI command.

If we're worried about sending too many inflight events, we could limit them in some way, such as only sending one if it has been X seconds since the previous one was sent.

What do you think @pandemicsyn @aaronsteers? Should I open an issue for this?

pnadolny13 (approved):
@pandemicsyn lgtm!

PR #6344, opened 2022-07-01
ci: Publish image based on Python 3.10

edgarrmondragon:


.github/workflows/docker_publish.yml
```diff
@@ -36,6 +36,8 @@ jobs:
           is-default-python: true  # will be used in 'latest' images
         - python-version: "3.9"
           is-default-python: false
```
WillDaSilva, line 38: Should we make Python 3.9 the default now?

WillDaSilva (approved):

.github/workflows/docker_publish.yml
edgarrmondragon: I think we should. Wdyt @aaronsteers

aaronsteers: @edgarrmondragon, @WillDaSilva  Works for me! 👍 

Do you see this as a project-wide bump to our default, or just for Dockerfile?

I'm open to both, but the other probably would need it's own PR.

aaronsteers (approved):

.github/workflows/docker_publish.yml
edgarrmondragon: @aaronsteers The others are purely internal, so I'm happy to make 3.9 the default image here and update the rest in a future PR

pandemicsyn: @edgarrmondragon heads up you'll probably want to update the docs. Since 3.8 is our current recommended version and thats referenced in a few spots.

pandemicsyn: Yea looks like even the docker image [docs](https://docs.meltano.com/guide/installation#using-pre-built-docker-images) mention 3.8:

> By default, these images come with a version of Python chosen for a balance of stability and compatibility, currently Python 3.8. 

We could probably make that a little more generic so we don't have to keep bumping it 😁

PR #6345, opened 2022-07-01
docs: fix outdated plugin names in `meltano add files` references

### LINKED ISSUE #6197: Airflow docs incorrect for adding files
This was raised in slack https://meltano.slack.com/archives/C01TCRBBJD7/p1655231194958339

The docs say `meltano add files airflow` but now we prefix the file bundles with `files-` so it should be updated to `meltano add files files-airflow`.

Its probably worth while to search for other references that might be incorrect also. The file bundle readme's look like they could use an update as well https://github.com/meltano/files-airflow.

aaronsteers:
Resolves #6197

Method:

- Searched for all references to `meltano add files`.
- If the named plugin was not prefixed with `files-`, I added a `files-` prefix.

### DIFF
--- a/docs/src/_guide/containerization.md
+++ b/docs/src/_guide/containerization.md
@@ -87,7 +87,7 @@ you can add the appropriate `docker-compose.prod.yml` file to your project by ad
 docker-compose --version
 
 # Add Docker Compose files to your project
-meltano add files docker-compose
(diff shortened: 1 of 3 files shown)


pandemicsyn (approved):
👍 also had a quick grep and think you got'em all.

PR #6355, opened 2022-07-05
fix: Swap add_related for add_required in api v1 plugins

BuzzCutNorman:
This PR swaps the call of `.add_related()`, which was recently removed, to its replacement function `.add_required()` .  The variables `related_plugin` variables used were changed to be `required_plugins` to match the function swap. 

Closes #6301

### DIFF
--- a/src/meltano/api/controllers/plugins.py
+++ b/src/meltano/api/controllers/plugins.py
@@ -147,7 +147,7 @@ def install_batch():  # noqa: WPS210
     )
 
     add_service = ProjectAddService(project, plugins_service=plugins_service)
-    related_plugins = add_service.add_related(plugin)
+    related_plugins = add_service.add_required(plugin)
 
     # We will install the plugins in reverse order, since dependencies
     # are listed after their dependents in `related_plugins`, but should
--- a/src/meltano/api/controllers/plugins.py
+++ b/src/meltano/api/controllers/plugins.py
@@ -58,8 +58,8 @@ def _handle(ex):
     return (jsonify({"error": True, "code": str(ex)}), 502)
 
 
-@pluginsBP.route("/all", methods=["GET"])  # noqa: WPS125
-def all():
+@pluginsBP.route("/all", methods=["GET"])
+def all():  # noqa: WPS125
     """Plugins found by the PluginDiscoveryService.
 
     Returns:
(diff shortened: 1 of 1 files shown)


src/meltano/api/controllers/plugins.py
```diff
@@ -147,12 +147,12 @@ def install_batch():  # noqa: WPS210
...
     add_service = ProjectAddService(project, plugins_service=plugins_service)
-    related_plugins = add_service.add_related(plugin)
+    required_plugins = add_service.add_required(plugin)
 
     # We will install the plugins in reverse order, since dependencies
-    # are listed after their dependents in `related_plugins`, but should
+    # are listed after their dependents in `required_plugins`, but should
     # be installed first.
-    related_plugins.reverse()
+    required_plugins.reverse()
```
pandemicsyn, line 155: I *think* this can be omitted when using add_required. Can you double check if you still actually need to reverse the list?

/cc @edgarrmondragon

pandemicsyn (changes requested):

src/meltano/api/controllers/plugins.py
edgarrmondragon: @pandemicsyn yeah, I think it should be safe to omit it

edgarrmondragon (approved):

src/meltano/api/controllers/plugins.py
BuzzCutNorman: I think I did something wrong when I merged the change.  😞

BuzzCutNorman:
@pandemicsyn and @edgarrmondragon Sorry I did something really wrong when I committed the requested change.  I tired to add the most recent changes to the branch and I obviously did that wrong. 😞   I believe I have the issue resolved.  Please take a look.  I apologize for the mess.  🙏

pandemicsyn:
> @pandemicsyn and @edgarrmondragon Sorry I did something really wrong when I committed the requested change. I tired to add the most recent changes to the branch and I obviously did that wrong. 😞 I believe I have the issue resolved. Please take a look. I apologize for the mess. 🙏

@BuzzCutNorman No worries! Looks like you got it all sorted and looking great.

pandemicsyn (approved):
🚀

PR #6376, opened 2022-07-07
fix: Extras of kind `object` not honored in environments

### LINKED ISSUE #3322: `metadata` and other extras of kind `object` not honored when using environments
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3400

Originally created by @tbriggs2 on 2022-04-14 20:18:50

---

If the metadata extra is defined in the base configuration of an inherited plugin it is not respected - the metadata from the base configuration of the parent plugin is used instead.

The metadata extra from the inherited plugin definition should be used.

### Steps to reproduce
Using the attached [meltano.yml](http://gitlab.com/meltano/meltano/uploads/3ddeb9658b37693e2edfe65519e970c2/meltano.yml) run

`meltano --environment=prod run tap-mysql2 target-jsonl`

SomeTable2 will get replicated but with replication-method=LOG_BASED. If instead you run

`meltano --environment=dev run tap-mysql2 target-jsonl`

SomeTable2 will be replicated using replication-method=FULL_TABLE as expected. The point being that this appears to be a problem specifically with metadata defined on the base inherited plugin only.

Note also that if you change the metadata in the base parent plugin to not use a wildcard, i.e.

```   
    metadata:
      SchemaName-SomeTable:
        replication-method: LOG_BASED
```

The pipeline will fail with an error about the replication method not being defined for SomeTable2. I think that's relevant because it shows that the wildcard isn't overriding the inherited plugin configuration.

### Further regression test
_Ensure we automatically catch similar issues in the future_

- [ ] Write additional adequate test cases and submit test results

edgarrmondragon:
## Context

Extras like tap `metadata` and transform `vars` were not being incorporated into the list of plugin settings.

Consider the configuration

```yaml
plugins:
  extractors:
  - name: my-tap
    metadata:
      my_stream_1:
        replication-key: INCREMENTAL
environments:
- name: dev
  config:
    plugins:
      extractors:
      - name: my-tap
        metadata:
          my_stream_2:
            replication-key: INCREMENTAL
```

Meltano flattens object settings, so this would result in `_metadata.my_stream_1.replication-key` and `_metadata.my_stream_2.replication-key`. For Meltano, these then become two separate settings, so the second setting only has a defined value in the `dev` environment but not in the top-level plugin definition. This in turn makes the setting an "orphan", and subject to https://gitlab.com/meltano/meltano/-/merge_requests/2570. However, that GitLab MR failed to take into account orphan extras. This PR should address that.

Closes #3322


src/meltano/core/environment.py
```diff
@@ -92,7 +92,7 @@ def get_orphan_settings(
         Returns:
             List of orphan settings.
         """
-        return SettingDefinition.from_missing(existing, self.config)
+        return SettingDefinition.from_missing(existing, self.config_with_extras)
```
pandemicsyn, line 95: sneaky

pandemicsyn (approved):
lgtm.

src/meltano/core/environment.py
edgarrmondragon: Indeed!

edgarrmondragon:
@DouweM in theory the changelog generation can now be automated from the semantic commits landing in `main`

DouweM:
@edgarrmondragon Ah OK, I didn't realize we weren't adding those manually anymore. I noticed https://github.com/meltano/meltano/blob/main/CHANGELOG.md#changes already listed a change, but not this bugfix yet. Do we autogenerate those as part of the release process now?

PR #6365, opened 2022-07-06
chore: Enable pre-commit hooks in CI

WillDaSilva:
In addition to changing the argument for pyupgrade, this PR also functions to test pre-commit CI integration.

Closes #6307

PR #6381, opened 2022-07-07
ci: Use pre-commit action instead of `pre-commit.ci`

### LINKED ISSUE #6307: Add lint workflow for GitHub CI

WillDaSilva:
~~This PR only exists to test `pre-commit.ci`, since https://github.com/meltano/meltano/pull/6365 was accidentally merged before the issues with `pre-commit.ci` were totally resolved.~~

From `.pre-commit-config.yml`:

```
# pre-commit.ci cannot be used because it runs on all files, rather than just
# the files added/modified by the PR. Instead, we use the deprecated pre-commit
# action with arguments provided to limit it to the added/modified files.
```

I had a chat with the maintainer of `pre-commit.ci`, and they said that the ability to run only on changed files may be available for the pro version of `pre-commit.ci` in the future. This PR implements a workaround. Since we weren't planning on using the PR autofixes provided by `pre-commit.ci`, we don't lose much by using this workaround.

Closes #6307


.github/workflows/precommit.yml
```diff
@@ -0,0 +1,31 @@
...
+        with:
+          python-version: 3.7
+          architecture: x64
+
+      - name: Get added/modified files
+        id: file_changes
+        uses: trilom/file-changes-action@v1.2.4
+        with:
+          prNumber: ${{ github.event.number }}
+          output: ' '
```
edgarrmondragon, line 26: These public actions are a gift 😄

edgarrmondragon (approved):
lgtm!

PR #6385, opened 2022-07-08
change: Disallow Meltano-level settings within environments

### LINKED ISSUE #6247: Disallow Meltano-level settings within environments
From https://github.com/meltano/meltano/issues/6034#issuecomment-1160993462_:

> @DouweM, @tayloramurphy, @kgpayne, @edgarrmondragon  - I think this needs to be reverted so that `meltano` settings are disallowed in environments. 

Re: 

- https://github.com/meltano/meltano/issues/6034

(Please dedupe if this issue exists already elsewhere.)

The rationale is that most Meltano settings are designed to be globally set and may not load in time for this. Also - until we update the default location to be global, the docs related to `meltano config meltano set...` are not properly paired with these settings landing in an environment by default. The proper and recommended way to get environment-specific config into Meltano-owned environments would be to use a `dotenv` file or explicitly set the env var within the specific container or hosting layer.

Implementation options:

1. Hard fail if `meltano config set meltano ...` would land settings in an environment.
2. Infer `--no-environment` instead of the default environment if no environment is explicitly requested during `meltano config set meltano ...`.
   - This would likely require a warning message: `Falling back to --global / --no-environment behavior for Meltano-owned settings.`
   - We could still hard fail if an enviornment is specifically requested via command-line flags.

kgpayne:
Closes #6247

- #6247


src/meltano/core/project_settings_service.py
```diff
@@ -66,15 +66,6 @@ def ensure_project_id(self) -> None:
...
         If it is not present, it will be restored from `analytics.json` if possible.
         """
-        if self.project.active_environment and "project_id" in self.environment_config:
-            # `project_id` is only valid at the top-level of the config, so we move it up there.
-            env_config = self.environment_config.copy()
-            self.update_meltano_yml_config(
-                # Potentially overwrite it with the existing top-level `project_id`
-                {"project_id": env_config.pop("project_id"), **self.meltano_yml_config}
-            )
-            self.update_meltano_environment_config(env_config)
```
pandemicsyn, line 76: Just making sure I grok the implication of removing this.

So if you update meltano to a version containing this PR, and you happen to have previously ended up with a `project_id` an env config, AND you didn't have a `project_id` in analytics.json, we'd no longer restore project_id from an entry found in the env config. 

Which would result with you having a new correctly located project_id at the top level, and an old/stale project_id left behind in your env config? 

That very much seems like it's going to be a very rare edge case, so probably ok. Any other more common scenarios I'm missing?

WillDaSilva: We used to have a bug where we were writing the project ID to the env-level config in `meltano.yml`, and it's unclear how many `meltano.yml` files are affected by this. This code that raises the project ID up to the top-level addresses that problem, but now that some time has passed it might not be necessary anymore.

I say we should clean it up, as you say.

kgpayne: Great, thanks 🙏

edgarrmondragon (changes requested):
@kgpayne can a test be added to `tests/meltano/core/test_project_settings_service.py` to validate that configuring Meltano in an environment should fail?

kgpayne:
@edgarrmondragon thanks for the test nudge! I have added one to make sure a meltano/project setting makes it to the `meltano_yml` store even if an environment is active. This is the desired behaviour, rather than a hard fail 🙂

src/meltano/core/project_settings_service.py
```diff
@@ -146,10 +138,10 @@ def meltano_yml_config(self):
     def environment_config(self):
         """Return current environment configuration in `meltano.yml`.
 
-        Returns:
-            Current environment configuration in `meltano.yml`
+        Raises:
+            NotImplementedError: Meltano config not supported in Environments.
         """
-        return self.config_service.current_environment_config
+        raise NotImplementedError("Meltano config not supported in Environments.")
```
edgarrmondragon, line 144: @kgpayne Would it make sense to remove these methods from the ABC?

kgpayne: @edgarrmondragon I did think about that, and it is totally possible. My thought was a `NotImplementedError` with a clear description might be more insightful than a generic `AttributeError` if a future feature tries to access these properties on the wrong subclass of the `SettingsService` 🤔 That said, I am not sure how likely that is to happen, and agree its probably neater to just remove both properties from the ABC and the `ProjectSettingsService`. WDYT?

edgarrmondragon: An `AttributeError` might then indicate that the method has to be pushed up to the ABC, so yeah I'd vote for removing them from the ABC and the project-level implementation 👍

edgarrmondragon (approved):

tests/meltano/core/test_project_settings_service.py
```diff
@@ -165,3 +170,16 @@ def test_warn_if_default_setting_is_used(self, subject, monkeypatch):
...
+    ):
+        # make sure that meltano setting values are written to the root of `meltano.yml`
+        # even if there is an active environment
+        with monkeypatch.context() as mpc:
+            mpc.setattr(subject.project, "active_environment", environment)
+            assert subject.project.active_environment == environment
+            subject.set("database_max_retries", 10000)
+            value, source = subject.get_with_source("database_max_retries")
+            assert source == SettingValueStore.MELTANO_YML
+            assert value == 10000
```
WillDaSilva, line 185: The `monkeypatch` fixture will automatically roll-back changes made after the test function returns. An explicit context is generally not required. See [the note about `monkeypatch.context` here](https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html#global-patch-example-preventing-requests-from-remote-operations).

```suggestion
        monkeypatch.setattr(subject.project, "active_environment", environment)
        assert subject.project.active_environment == environment
        subject.set("database_max_retries", 10000)
        value, source = subject.get_with_source("database_max_retries")
        assert source == SettingValueStore.MELTANO_YML
        assert value == 10000
```

WillDaSilva (approved):

PR #6395, opened 2022-07-10
ci: Use Nox to test installed package

### LINKED ISSUE #6277: How to catch dev dependencies that actually need to be included in the distributed presentations of Meltano?
Just got bit by this with `packaging` in https://github.com/meltano/meltano/pull/6276:

It's a transitive dependency required by `tox`, but `tox` is only a dev dependency. So, in CI tests, things look good because dev dependencies are installed by `poetry install`. But, when meltano is installed from an upstream source or by another tool like `pipx`, which is the case in end-to-end tests [here](https://gitlab.com/meltano/legacy-ci/meltano/-/jobs/2632994538#L63) (`pipx install ..`), then the transitive dev dependency `packaging` is not installed and errors are encountered: https://gitlab.com/meltano/legacy-ci/meltano/-/jobs/2632994539#L1237.

```console
$ meltano add orchestrator airflow
2022-06-23T19:10:30.061252Z [info     ] Environment 'dev' is active
No module named 'packaging'
```

I guess one easy way to ensure we don't fall in that scenario is to have tests that don't rely on Poetry to install Meltano.

edgarrmondragon:
## Notes for reviewers

- Nox and `nox-poetry` ensure tasks run against an installed form of the package. This is accomplished under the hood by building a wheel and installing the package and any dependencies with a constraints file exported (via `poetry export`) from the project's pinned dependencies.

- I had to add an extra check to `UpgradeService._upgrade_package` to fail when running inside a Nox `tests` session. That's because the package is not installed in editable mode, so it would've tried and "succeeded" in upgrading the package.

- Tests are failing for Windows on Python 3.7 and 3.10, these failures were not introduced but rather uncovered by this PR. See #6479.

  My suggestion is that we fix those in a different PR.

Closes #6277


edgarrmondragon:
Failing jobs due to:

- On Windows + Python 3.7, asyncio's `create_subprocess_exec` is seemingly broken
- On Windows + Python 3.10, the `docker` dev dependency is not compatible

pandemicsyn (approved):

WillDaSilva (commented):
The changes look great, but if we merge this in it'll result in CI failures for all PRs that merge in `main`. I'd rather consider this blocked by
- https://github.com/meltano/meltano/issues/6479

and mark it as a draft until that is resolved.

edgarrmondragon:
> The changes look great, but if we merge this in it'll result in CI failures for all PRs that merge in `main`. I'd rather consider this blocked by

@WillDaSilva what do you think we should fix before merging this PR?

1. the fact that we're not running tests on Windows x Python 3.7,3.10 or
2. or the failing tests themselves

I feel like `1` is fairly easy, while `2` is a more complex and I don't think we're in a position to solve quickly. Wdyt?

cc @aaronsteers wdyt about this? 👆

WillDaSilva:
> what do you think we should fix before merging this PR?
> 
>     1. the fact that we're not running tests on Windows x Python 3.7,3.10 or
> 
>     2. the failing tests themselves
> 
> 
> I feel like `1` is fairly easy to tackle on its own (or by merging this PR), while `2` is a more complex and I don't think we're in a position to solve quickly. Wdyt?

@edgarrmondragon I was referring to 2, but upon second thought, we could just have this PR comment out the Windows tests from the matrix for 3.7 and 3.10. That way we can merge this in immediately, and not have it cause failures for all other PRs, *and* be in a better position than we were before wrt Windows test coverage. There's no need for us to wait till those tests are actually fixed to get this PR merged in.

.github/workflows/test.yml
```diff
@@ -24,46 +24,48 @@ jobs:
...
         - { python-version: "3.9",  os: "ubuntu-latest",  backend-db: postgresql }
         - { python-version: "3.10", os: "ubuntu-latest",  backend-db: postgresql }
-        - { python-version: "3.7",  os: "windows-latest", backend-db: sqlite } # We'd like to run Windows tests for all backend-dbs see https://github.com/meltano/meltano/issues/6281
-        - { python-version: "3.8",  os: "windows-latest", backend-db: sqlite }
-        - { python-version: "3.9",  os: "windows-latest", backend-db: sqlite }
-        - { python-version: "3.10", os: "windows-latest", backend-db: sqlite }
+        - { python-version: "3.7",  os: "windows-2022",   backend-db: sqlite } # We'd like to run Windows tests for all backend-dbs see https://github.com/meltano/meltano/issues/6281
+        - { python-version: "3.8",  os: "windows-2022",   backend-db: sqlite }
+        - { python-version: "3.9",  os: "windows-2022",   backend-db: sqlite }
+        - { python-version: "3.10", os: "windows-2022",   backend-db: sqlite }
```
WillDaSilva, line 30:
```suggestion
        # We'd like to run Windows tests for all backend-dbs see https://github.com/meltano/meltano/issues/6281
        # Windows tests for Python 3.7 and 3.10 are temporarily disabled - see https://github.com/meltano/meltano/issues/6479
        # - { python-version: "3.7",  os: "windows-2022",   backend-db: sqlite }
        - { python-version: "3.8",  os: "windows-2022",   backend-db: sqlite }
        - { python-version: "3.9",  os: "windows-2022",   backend-db: sqlite }
        # - { python-version: "3.10", os: "windows-2022",   backend-db: sqlite }
```

WillDaSilva (changes requested):

WillDaSilva (approved):

Issue #6393, opened 2022-07-08
Enable dependency license scanning in CI


WillDaSilva:
This issue tracks the migration of the functionality provided by the GitLab CI job `license_scanning`

WillDaSilva:
As per https://github.com/meltano/meltano/pull/6403#issuecomment-1181281741, we should begin building up a list of permitted licenses (identified by their [SPDX ID](https://spdx.org/licenses/)). @DouweM @tayloramurphy do either of you have any recommendations for what we should accept from our (direct and transitive) dependencies? A simple start would be:

- 0BSD
- BSD-1-Clause 
- MIT
- MIT-0
- Apache-2.0
- ISC

Beyond that we can run the license check to see what we're missing, and add the missing licenses to the list of permitted licenses one by one as we deem them acceptable, and if we do not deem them acceptable, we can remove the dependency that introduced the unacceptable license.

tayloramurphy:
@WillDaSilva https://gitlab.com/meltano/meltano/-/issues/3206 has some additional context when we looked at this previously. I've used https://about.gitlab.com/handbook/engineering/open-source/#acceptable-licenses as prior art and a pretty good starting point for us.

kgpayne:
Related to https://github.com/meltano/gitops/issues/12

PR #6403, opened 2022-07-12
ci: Add dependency security review GitHub workflow

WillDaSilva:


WillDaSilva:
In a future PR, once we've compiled a list of permitted licenses, we can resolve #6393 by using the `allow-licenses` option: https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/configuring-dependency-review#configuring-the-dependency-review-github-action

The licenses that we can add to the list can be found here: https://spdx.org/licenses/

edgarrmondragon (approved):
nice!

PR #6404, opened 2022-07-12
fix(core): Extras of kind `object` not honored in inherited plugins

### LINKED ISSUE #6380: `metadata` and other extras of kind `object` not honored when using inheritance
Consider the following inheritance scheme:

```yaml
plugins:
  extractors:
  - name: tap-postgres
    variant: transferwise
    pip_url: pipelinewise-tap-postgres
    select:
    - raw_github-issues.*
    metadata:
      raw_github-issues:
        replication-method: INCREMENTAL
        replication-key: updated_at
  - name: tap-postgres--inherited
    inherit_from: tap-postgres
    select:
    - raw_github-commits.*
    metadata:
      raw_github-commits:
        replication-method: INCREMENTAL
        replication-key: commit_timestamp
```

The expectation is that `tap-postgres--inherited` uses its own metadata object to patch the Singer catalog, and the output of `meltano config` would suggest that everything is working as expected:

```console
$ meltano config --extras tap-postgres--inherited
{
  "_catalog": null,
  "_state": null,
  "_load_schema": "tap_postgres__inherited",
  "_select": [
    "raw_github-commits.*"
  ],
  "_metadata": {
    "raw_github-issues": {
      "replication-method": "INCREMENTAL",
      "replication-key": "updated_at"
    }
  },
  "_schema": {},
  "_select_filter": [],
  "_metadata.raw_github-issues.replication-method": "INCREMENTAL",
  "_metadata.raw_github-issues.replication-key": "updated_at",
  "_metadata.raw_github-commits.replication-method": "INCREMENTAL",
  "_metadata.raw_github-commits.replication-key": "commit_timestamp"
}
```

> **Note**
> The bug seems to be that `_metadata` above does not contain everything that starts with `_metadata` prefix 🤔

However, invoking `tap-postgres--inherited` fails as its own metadata is being ignored:

```
  File "/Users/edgarramirez/Code/meltano-personal/.meltano/extractors/tap-postgres/venv/bin/tap-postgres", line 8, in <module>
    sys.exit(main())
  File "/Users/edgarramirez/Code/meltano-personal/.meltano/extractors/tap-postgres/venv/lib/python3.9/site-packages/tap_postgres/__init__.py", line 435, in main
    raise exc
  File "/Users/edgarramirez/Code/meltano-personal/.meltano/extractors/tap-postgres/venv/lib/python3.9/site-packages/tap_postgres/__init__.py", line 432, in main
    main_impl()
  File "/Users/edgarramirez/Code/meltano-personal/.meltano/extractors/tap-postgres/venv/lib/python3.9/site-packages/tap_postgres/__init__.py", line 421, in main_impl
    do_sync(conn_config, args.catalog.to_dict() if args.catalog else args.properties,
  File "/Users/edgarramirez/Code/meltano-personal/.meltano/extractors/tap-postgres/venv/lib/python3.9/site-packages/tap_postgres/__init__.py", line 293, in do_sync
    sync_method_for_streams(streams, state, default_replication_method)
  File "/Users/edgarramirez/Code/meltano-personal/.meltano/extractors/tap-postgres/venv/lib/python3.9/site-packages/tap_postgres/__init__.py", line 107, in sync_method_for_streams
    raise Exception(f"Unrecognized replication_method {replication_method}")
Exception: Unrecognized replication_method None
```

Removing the `metadata` section from the _parent_ plugin solves the problem, but this not how things should work.

edgarrmondragon:
Related:

- Closes #6380
- Similar PR for environments #6376

### DIFF
--- a/src/meltano/core/settings_service.py
+++ b/src/meltano/core/settings_service.py
@@ -374,12 +374,14 @@ class SettingsService(ABC):  # noqa: WPS214
                 value = expanded_value
 
         if setting_def:
+            # Expand flattened config values if the root value is the default
+            # or inherited empty object.
             if setting_def.kind == SettingKind.OBJECT and (
                 metadata["source"]
                 in {SettingValueStore.DEFAULT, SettingValueStore.INHERITED}
             ):
                 object_value = {}
-                object_source = SettingValueStore.DEFAULT
+                object_source = metadata["source"]
                 for setting_key in [  # noqa: WPS335
                     setting_def.name,
                     *setting_def.aliases,
--- a/src/meltano/core/settings_service.py
+++ b/src/meltano/core/settings_service.py
@@ -374,9 +374,9 @@ class SettingsService(ABC):  # noqa: WPS214
                 value = expanded_value
 
         if setting_def:
-            if (
-                setting_def.kind == SettingKind.OBJECT
-                and metadata["source"] is SettingValueStore.DEFAULT
+            if setting_def.kind == SettingKind.OBJECT and (
+                metadata["source"]
+                in {SettingValueStore.DEFAULT, SettingValueStore.INHERITED}
             ):
                 object_value = {}
                 object_source = SettingValueStore.DEFAULT
--- a/tests/meltano/core/plugin/test_plugin_settings.py
+++ b/tests/meltano/core/plugin/test_plugin_settings.py
@@ -876,6 +876,27 @@ class TestPluginSettingsService:
             SettingValueStore.MELTANO_ENV,
         )
 
+        subject.project.active_environment = None
+
+        try:
+            inherited = project_add_service.add(
+                PluginType.TRANSFORMS,
+                "tap-mock-transform--inherited",
+                inherit_from=transform.name,
+            )
+        except PluginAlreadyAddedException as err:
+            inherited = err.plugin
+
+        inherited_subject = plugin_settings_service_factory(inherited)
+        inherited_subject.set("_vars", {"new": "from_inheriting"})
+
+        assert inherited_subject.get_with_source("_vars") == (
+            {
+                "new": "from_inheriting",
+            },
+            SettingValueStore.MELTANO_YML,
+        )
+
     def test_find_setting_raises_with_conflicting(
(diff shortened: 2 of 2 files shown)


src/meltano/core/settings_service.py
```diff
@@ -374,9 +374,9 @@ def get_with_metadata(  # noqa: WPS210, WPS615
                 value = expanded_value
 
         if setting_def:
-            if (
-                setting_def.kind == SettingKind.OBJECT
-                and metadata["source"] is SettingValueStore.DEFAULT
+            if setting_def.kind == SettingKind.OBJECT and (
+                metadata["source"]
+                in {SettingValueStore.DEFAULT, SettingValueStore.INHERITED}
```
edgarrmondragon, line 379: I can't believe how many hours I put in for a fix that seems too simple 😳

cjohnhanson (commented):
@edgarrmondragon holding off on approval due to a couple questions about the logic here, but mostly seems reasonable to me.

src/meltano/core/settings_service.py
edgarrmondragon: > Would there potentially be other places where nested (or "object") extras are being ignored by this? For instance, should this also include `SettingValueStore.DB`? I can't immediately think of a case where that could happen but this logic isn't totally clear to me.

@cjohnhanson AFAICT, other stores like env vars and database are safe because the value comes in a string and cast into an object by `SettingDefinition.cast_value`. All of these extras of kind `object` have a default value of `{}`, and the original check was added to allow things like `_metadata.stream_id.replication-key` to be expanded after trying to find values for `_metadata` in all other sources before `DEFAULT`.

A comment is a good idea 👍

src/meltano/core/settings_service.py
```diff
@@ -374,9 +374,9 @@ def get_with_metadata(  # noqa: WPS210, WPS615
...
             ):
                 object_value = {}
                 object_source = SettingValueStore.DEFAULT
```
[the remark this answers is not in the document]
edgarrmondragon: Makes sense

cjohnhanson (approved):
:shipit:

Issue #6398, opened 2022-07-11
feat: Refactor Airflow integration as an external Python plugin


tayloramurphy:
_This is a specific implementation issue for https://github.com/meltano/meltano/issues/6130 and is related to https://github.com/meltano/meltano/issues/6397_

### Problem to Solve

Much of the problems documented in #6397 are applicable to Airflow. Files and functionality related to Airflow are spread out across multiple repos:

* Meltano https://github.com/meltano/meltano/blob/main/src/meltano/core/plugin/airflow.py
* Files bundle https://github.com/meltano/files-airflow/
* Metadata definition on MeltanoHub https://hub.meltano.com/orchestrators/airflow

This is cumbersome and not contributor-friendly. Additionally, it's difficult to iterate on as it's not clear _where_, _how_, or even _what_ users should do to extend functionality - do they add a command? do they add something to Meltano? Should it be in the file bundle?

There's also the challenge of hooking into the capabilities of Meltano. As we expand what users can do with Meltano across multiple plugins, we dramatically increase the number of places we need to make an update when we update a capability or add additional functionality. 

We need to make it easy to add more plugins to Meltano and to iterate on their capabilities without users having to be aware of the internals of Meltano specifically. We'll start with Airflow & dbt. 

### Prior Art / Things to Keep in Mind

Refer to #6397 for some of the prior art around Superset, but there is a longer history of challenges users have with understanding our Airflow implementation.

* Original GitLab Epic https://gitlab.com/groups/meltano/-/epics/134#note_981098666
* Main meta-issue https://github.com/meltano/meltano/issues/6130 
* https://github.com/meltano/meltano/issues/6110

### Definition of Done

For this issue, there are a few items we'd need to see out of a first iteration. 

* All of the Airflow-specific code is in its own package and repository. We'll worry about the monorepo (whether extension code should live in the meltano/meltano repo) question separately
* This is built with the the skeleton of a true Meltano SDK in mind - bonus if the skeleton is started. This should also be built with Superset (#6397) and other generic plugins in mind.
* User functionality with Airflow is basically the same but there will likely be new generic capabilities either used or at least proposed
* Docs are updated in Meltano and the Hub on how the Airflow integration is used and changed 
* Hub updated to use new integration

pandemicsyn:
@aaronsteers @tayloramurphy given:

[...]

Issue #6130, opened 2022-06-09
Python-based Plugin Architecture


aaronsteers:
I think we have a path forward on python plugin architecture (https://gitlab.com/groups/meltano/-/epics/134#note_981098666) which does not require modifying Meltano itself (at least not in the first iterations).

Since `pip_url` for any repo can include multiple libraries, we can extend any plugins' capabilities by _adding_ a meltano-managed python tool to any plugin that needs capabilities. Then those capabilities just need to registered as CLI-executable and be mapped to `commands` and `capabilities` in the Meltano plugin definition.

## Example: registering dbt packages in `packages.yml`

Today we have custom code in Meltano that adds entries to `packages.yml` for dbt during `meltano add transform ...`. We could replace it as follows:

1. Create a new python tool called `meltano-dbt-helpers` with a registered CLI command `meltano-dbt-util add package`.
2. Add that package to the pipurl for all dbt plugins. E.g. `pip_url: dbt-snowflake meltano-dbt-helpers`.
3. Register commands in the package that leverage the helper. E.g. 
    ```
   commands:
     add-dep:
       command: meltano-dbt-util
       args: add package
   ```

Then from the CLI as a user:

```
meltano run dbt:add-dep dbt-util
```

Would add `dbt-util` to the package.

For "smarter" integrations, we just need a contract that pairs `capabilities` like `service` with expected commands, like `ui`, `up`, and `down`.

## Comparison with other approaches

The best part of this approach is that it doesn't require _any_ code changes in Meltano - we and community members could implement this today with existing features.

DouweM:
@aaronsteers Great suggestion, and injecting custom code into the venv will go a long way to adding new Meltano-specific functionality and commands to tools. 

Our current plugin glue code (e.g. for [Airflow](https://github.com/meltano/meltano/blob/main/src/meltano/core/plugin/airflow.py#L134), [Singer taps](https://github.com/meltano/meltano/blob/main/src/meltano/core/plugin/singer/tap.py), [dbt](https://github.com/meltano/meltano/blob/main/src/meltano/core/plugin/dbt/base.py), [superset](https://github.com/meltano/meltano/blob/main/src/meltano/core/plugin/superset.py)) mostly sets up the execution environment by setting env vars, running initialization commands, and generating config files. All of that could likely be done from a custom executable that wraps the main executable and has access to the configuration environment variables/files injected by Meltano. That's similar to the `entrypoint: init.sh` paradigm seen in Docker(-compose).

[...]

PR #6409, opened 2022-07-13
feat: Allow setting the working directory for plugin invocation

rabidaudio:
Hello folks!

I really wanted to set up Dagster with a `test` command that runs pytest on the project:

```yaml
utilities:
  - name: dagster
    namespace: dagster
    pip_url: -e orchestrate
    executable: dagster
    settings:
    - name: home
      env: DAGSTER_HOME
    config:
      home: $MELTANO_PROJECT_ROOT/orchestrate
    commands:
      scheduler:
        args: run -w $DAGSTER_HOME/workspace.yaml
        executable: dagster-daemon
      ui:
        args: -w $DAGSTER_HOME/workspace.yaml -h 0.0.0.0
        executable: dagit
      test:
        args: --rootdir=$DAGSTER_HOME
        executable: pytest
```

However, despite setting `--rootdir` in the `pytest` args, pytest insisted on picking up tests and `conftest.py` files from elsewhere in our project, like `transforms/dbt_modules`.

I figured it made sense to allow plugins and commands to define a working directory to be invoked in, instead of the current working directory:

```yml
utilities:
  - name: dagster
    namespace: dagster
    pip_url: -e orchestrate
    executable: dagster
    cwd: orchestrate # also supports variable expansion, e.g. $MELTANO_PROJECT_ROOT/orchestrate although that's redundant
    commands:
      test:
        executable: pytest
        cwd: orchestrate/tests # commands can override the plugin config
```

This simplified `DbtPlugin`, as it no longer needs a custom `PluginInvoker`, it just needs to set the cwd to the project directory. This could be further simplified if we defined it in the discovery configuration, but to avoid backward compatibility issues I explicitly implemented the method.

This could use some docs, but wanted to get y'alls thoughts first.

### DIFF
--- a/src/meltano/core/plugin/base.py
+++ b/src/meltano/core/plugin/base.py
@@ -243,6 +243,7 @@ class Variant(NameEq, Canonical):
         settings: list | None = None,
         commands: dict | None = None,
         requires: dict[PluginType, list] | None = None,
+        cwd: str | None = None,
         env: dict[str, str] | None = None,
         **extras,
     ):
@@ -261,6 +262,7 @@ class Variant(NameEq, Canonical):
             settings: The settings of the variant.
             commands: The commands of the variant.
             requires: Other plugins this plugin depends on.
+            cwd: The working directory to use for invocation.
             env: Environment variables to inject into plugins runtime context.
             extras: Additional keyword arguments.
         """
@@ -277,6 +279,7 @@ class Variant(NameEq, Canonical):
             settings=list(map(SettingDefinition.parse, settings or [])),
             commands=Command.parse_all(commands),
             requires=PluginRequirement.parse_all(requires),
+            cwd=cwd,
             env=env or {},
             extras=extras,
         )
@@ -721,6 +724,7 @@ class StandalonePlugin(Canonical):
         settings: list | None = None,
         commands: dict | None = None,
         requires: dict[PluginType, list] | None = None,
+        cwd: str | None = None,
         env: dict[str, str] | None = None,
         **extras,
     ):
@@ -741,6 +745,7 @@ class StandalonePlugin(Canonical):
             settings: The settings of the plugin.
             commands: The commands of the plugin.
             requires: Other plugins this plugin depends on.
+            cwd: The working directory to use for invocation.
             env: Environment variables to inject into plugins runtime context.
             extras: Additional attributes to set on the plugin.
         """
@@ -759,6 +764,7 @@ class StandalonePlugin(Canonical):
             settings=list(map(SettingDefinition.parse, settings or [])),
             commands=Command.parse_all(commands),
             requires=PluginRequirement.parse_all(requires),
+            cwd=cwd,
             env=env or {},
             extras=extras,
         )
--- a/src/meltano/core/plugin/base.py
+++ b/src/meltano/core/plugin/base.py
@@ -243,7 +243,7 @@ class Variant(NameEq, Canonical):
         settings: list | None = None,
         commands: dict | None = None,
         requires: dict[PluginType, list] | None = None,
-        cwd: str | None = None,
+        workdir: str | None = None,
         env: dict[str, str] | None = None,
         **extras,
     ):
@@ -262,7 +262,7 @@ class Variant(NameEq, Canonical):
             settings: The settings of the variant.
             commands: The commands of the variant.
             requires: Other plugins this plugin depends on.
-            cwd: The working directory to use for invocation.
+            workdir: The working directory to use for invocation.
             env: Environment variables to inject into plugins runtime context.
             extras: Additional keyword arguments.
         """
@@ -279,7 +279,7 @@ class Variant(NameEq, Canonical):
             settings=list(map(SettingDefinition.parse, settings or [])),
             commands=Command.parse_all(commands),
             requires=PluginRequirement.parse_all(requires),
-            cwd=cwd,
+            workdir=workdir,
             env=env or {},
             extras=extras,
         )
@@ -724,7 +724,7 @@ class StandalonePlugin(Canonical):
         settings: list | None = None,
         commands: dict | None = None,
         requires: dict[PluginType, list] | None = None,
-        cwd: str | None = None,
+        workdir: str | None = None,
         env: dict[str, str] | None = None,
         **extras,
     ):
@@ -745,7 +745,7 @@ class StandalonePlugin(Canonical):
             settings: The settings of the plugin.
             commands: The commands of the plugin.
             requires: Other plugins this plugin depends on.
-            cwd: The working directory to use for invocation.
+            workdir: The working directory to use for invocation.
             env: Environment variables to inject into plugins runtime context.
             extras: Additional attributes to set on the plugin.
         """
@@ -764,7 +764,7 @@ class StandalonePlugin(Canonical):
             settings=list(map(SettingDefinition.parse, settings or [])),
             commands=Command.parse_all(commands),
             requires=PluginRequirement.parse_all(requires),
-            cwd=cwd,
+            workdir=workdir,
             env=env or {},
             extras=extras,
         )
--- a/src/meltano/core/plugin/dbt/base.py
+++ b/src/meltano/core/plugin/dbt/base.py
@@ -15,8 +15,8 @@ logger = logging.getLogger(__name__)
 
 
 class DbtInvoker(PluginInvoker):
-    def Popen_options(self):
-        return {**super().Popen_options(), "cwd": self.plugin_config["project_dir"]}
+    def popen_options(self):
+        return {**super().popen_options(), "cwd": self.plugin_config["project_dir"]}
 
 
 class DbtPlugin(BasePlugin):
--- a/src/meltano/core/plugin/dbt/base.py
+++ b/src/meltano/core/plugin/dbt/base.py
@@ -20,7 +20,8 @@ class DbtInvoker(PluginInvoker):
         self, command: Optional[str] = None, env: Optional[dict] = None
     ) -> Optional[Path]:
         return (
-            super().workdir(command=command, env=env) or self.plugin_config["project_dir"]
+            super().workdir(command=command, env=env)
+            or self.plugin_config["project_dir"]
         )
 
 
--- a/src/meltano/core/plugin/dbt/base.py
+++ b/src/meltano/core/plugin/dbt/base.py
@@ -7,22 +7,18 @@ from meltano.core.plugin import BasePlugin, PluginType
 from meltano.core.plugin.error import PluginNotFoundError
 from meltano.core.plugin.project_plugin import ProjectPlugin
 from meltano.core.plugin_install_service import PluginInstallReason
-from meltano.core.plugin_invoker import PluginInvoker
 from meltano.core.setting_definition import SettingDefinition, SettingKind
 from meltano.core.transform_add_service import TransformAddService
 
 logger = logging.getLogger(__name__)
 
 
-class DbtInvoker(PluginInvoker):
-    def popen_options(self):
-        return {**super().popen_options(), "cwd": self.plugin_config["project_dir"]}
-
-
 class DbtPlugin(BasePlugin):
     __plugin_type__ = PluginType.TRANSFORMERS
 
-    invoker_class = DbtInvoker
+    @property
+    def cwd(self) -> str:
+        return self.plugin_config["project_dir"]
 
 
 class DbtTransformPluginInstaller:
--- a/src/meltano/core/plugin/dbt/base.py
+++ b/src/meltano/core/plugin/dbt/base.py
@@ -1,24 +1,33 @@
 """Defines DBT-specific plugins."""
 import logging
 from pathlib import Path
+from typing import Optional
 
 from meltano.core.error import PluginInstallError
 from meltano.core.plugin import BasePlugin, PluginType
 from meltano.core.plugin.error import PluginNotFoundError
 from meltano.core.plugin.project_plugin import ProjectPlugin
 from meltano.core.plugin_install_service import PluginInstallReason
+from meltano.core.plugin_invoker import PluginInvoker
 from meltano.core.setting_definition import SettingDefinition, SettingKind
 from meltano.core.transform_add_service import TransformAddService
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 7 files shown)


pandemicsyn:
👍 I like the idea, a `cwd` option for plugins makes a lot of sense. Especially in the context of enabling a robust external plugin architecture (e.g. https://github.com/meltano/meltano/issues/6130 which leads to https://github.com/meltano/meltano/issues/6434 , https://github.com/meltano/meltano/issues/6398). Feels like this could end up being pretty useful.

@aaronsteers curious how you feel about this ?

tests/meltano/core/test_plugin_invoker.py
```diff
@@ -167,3 +176,19 @@ async def test_expand_nonpip_command_exec_args(
 
         assert "VIRTUAL_ENV" not in env
         assert "PYTHONPATH" not in env
+
+    def test_cwd(self, alternate_cwd_plugin_invoker):
+        root = alternate_cwd_plugin_invoker.project.root
+        assert alternate_cwd_plugin_invoker.cwd() == root / "path"
+        assert alternate_cwd_plugin_invoker.cwd(command="relative") == root / "path2"
+        assert alternate_cwd_plugin_invoker.cwd(command="absolute") == Path("/root")
```
pandemicsyn, line 184: Heads up - we run tests on windows now as well, which are currently failing:

https://github.com/meltano/meltano/runs/7313447720?check_suite_focus=true#step:9:648

```
_________________________ TestPluginInvoker.test_cwd __________________________
self = <test_plugin_invoker.TestPluginInvoker object at 0x000001DAA9868310>
alternate_cwd_plugin_invoker = <meltano.core.plugin_invoker.PluginInvoker object at 0x000001DAADEFCBB0>
    def test_cwd(self, alternate_cwd_plugin_invoker):
        root = alternate_cwd_plugin_invoker.project.root
        assert alternate_cwd_plugin_invoker.cwd() == root / "path"
        assert alternate_cwd_plugin_invoker.cwd(command="relative") == root / "path2"
>       assert alternate_cwd_plugin_invoker.cwd(command="absolute") == Path("/root")
E       AssertionError: assert WindowsPath('C:/root') == WindowsPath('/root')
E        +  where WindowsPath('C:/root') = <bound method PluginInvoker.cwd of <meltano.core.plugin_invoker.PluginInvoker object at 0x000001DAADEFCBB0>>(command='absolute')
E        +    where <bound method PluginInvoker.cwd of <meltano.core.plugin_invoker.PluginInvoker object at 0x000001DAADEFCBB0>> = <meltano.core.plugin_invoker.PluginInvoker object at 0x000001DAADEFCBB0>.cwd
E        +  and   WindowsPath('/root') = Path('/root')
D:\a\meltano\meltano\tests\meltano\core\test_plugin_invoker.py:184: AssertionError
```

aaronsteers:
> 👍 I like the idea, a `cwd` option for plugins makes a lot of sense. Especially in the context of enabling a robust external plugin architecture (e.g. #6130 which leads to #6434 , #6398). Feels like this could end up being pretty useful.
> 
> @aaronsteers curious how you feel about this ?

@rabidaudio - Love this direction and I agree the feature can be very helpful. I may want to hold while we settle the spec here but hopefully not leaving you in a holding pattern for long.

A few thoughts come to mind:

1. `cwd` is not an obvious acronym for non-*nix folks out there. I wonder if something like `workdir: __` would be more readable for a broader audience. Inspired by Docker's naming of a similar concept: https://www.educative.io/answers/what-is-the-workdir-command-in-docker
2. `dbt` plugins are doing something similar today with the `project_dir` config option, which I think at least in theory could be replaced by this new feature. Thoughts? Example here: https://github.com/meltano/hub/blob/main/_data/meltano/transformers/dbt-snowflake/dbt-labs.yml#L17-L19
3. We should plan ahead for users to override this and make sure we implement in a way that's as easy as possible to debug and not feel too magical. Like pip_url, we may want to this to be one of the settings that auto-registers itself into `meltano.yml` during the `add` process, and is therefor easy and intuitive for users to override if/when they start moving things around.

@pandemicsyn, @rabidaudio, @tayloramurphy, @pandemicsyn - Penny for your thoughts. Thanks! 😄

src/meltano/core/plugin/dbt/base.py
```diff
@@ -15,8 +16,12 @@
...
 
 class DbtInvoker(PluginInvoker):
-    def Popen_options(self):
-        return {**super().Popen_options(), "cwd": self.plugin_config["project_dir"]}
+    def cwd(
+        self, command: Optional[str] = None, env: Optional[dict] = None
+    ) -> Optional[Path]:
+        return (
+            super().cwd(command=command, env=env) or self.plugin_config["project_dir"]
+        )
```
aaronsteers, line 24: Superceding my prior comment about `cwd` replacing the `project_dir` setting of dbt, I see here that yes, absolutely this appears to replace that functionality - and I like that we're doing it in a generic way for all plugins, which would not require special per-plugin logic internally within Meltano.

Also - Thanks for keeping this backwards compatible. As much as I like the generic approach, we also don't want this to break existing dbt capabilities/behaviors.

src/meltano/core/plugin/base.py
```diff
@@ -243,6 +243,7 @@ def __init__(
         settings: list | None = None,
         commands: dict | None = None,
         requires: dict[PluginType, list] | None = None,
+        cwd: str | None = None,
```
aaronsteers, line 246: I'm slightly preferring a more explicit naming like [`workdir`](https://www.educative.io/answers/what-is-the-workdir-command-in-docker) over `cwd`. (Here and elsewhere as noted in my prior comment.) 

Curious what others think: @edgarrmondragon, @pandemicsyn, @tayloramurphy, @pandemicsyn.

aaronsteers (changes requested):

rabidaudio:
> cwd is not an obvious acronym for non-*nix folks out there. I wonder if something like workdir: __ would be more readable for a broader audience.

I agree, I think I like `workdir` better 👍

tayloramurphy:
@aaronsteers @rabidaudio I like this proposal and excited about the PR. Responding to AJ's points:

1. Agreed, `workdir` is better than `cwd`.
1. this makes me think about #6130 as well - seems like the plugin specific behavior should be there and not in Meltano
1. Agreed also that users should be able to override. 

I don't agree that it should auto-register itself in `meltano.yml` during the add process. Likely the default behavior should stay the same but we should document in multiple places about the availability of this plugin-level property. 

 >  Side note, our documentation on plugin properties versus config versus settings versus extras is a bit confusing. The addition of a `workdir` property can actually be thought of as a plugin extra that happens to apply to all plugins (aka a property!). While testing this, I also realized it's not possible to know what environment variables Meltano will be injecting from the plugin-level based on the property names https://docs.meltano.com/guide/configuration#available-environment-variables I'll make an issue around this. 

All that said, I agree with the spec that users should be able to set `workdir` at the plugin level and override it at the command level. This value should also be available to users at `MELTANO_<PLUGIN_TYPE>_WORKDIR` based on the documentation in https://docs.meltano.com/guide/configuration#available-environment-variables. 

Question: Would env var expansion in the commands work as well such that:

```yaml
utilities:
  - name: dagster
    namespace: dagster
    pip_url: -e orchestrate
    executable: dagster
    workdir: orchestrate # also supports variable expansion, e.g. $MELTANO_PROJECT_ROOT/orchestrate although that's redundant
    commands:
      test:
        executable: pytest
        workdir: $MELTANO_UTILITY_WORKDIR/tests
```

would evaluate to `$MELTANO_PROJECT_ROOT/orchestrate/tests` for the `test` command?

Perhaps that's a separate, larger issue on if and how commands are dealing with environment variables.

aaronsteers:
@tayloramurphy - Thanks! Yeah, sounds like we are agreed on most points. A couple thoughts on the points you raised:

> this makes me think about https://github.com/meltano/meltano/issues/6130 as well - seems like the plugin specific behavior should be there and not in Meltano

Agreed. 👍 From the diff here, I think we are moving in the right direction, which will helps us towards #6130 because it removes/de-emphasizes the custom dbt behavior in favor of this more generic approach. 

> I don't agree that it should auto-register itself in `meltano.yml` during the add process.

To confirm, all I meant here is that if the remote plugin definition (on the hub, for instance) is defining a custom `workdir`, we should try to make that custom `workdir` as transparent and easy to override as possible for an end user, so they are not confused when the move a directory around and everything breaks. If that is the case (that the plugin defaults to a specific subdirectory) then I think it is a good idea to bring that into the `meltano.yml` file so they can see+modify what was provided to them by the plugin developer.

For plugins which do not require a custom `workdir` by default, there's no reason to bring that into the `meltano.yml` file by default. Also, in a first iteration, I'd be fine with either behavior. I do think bringing it into `meltano.yml` is probably the better long-term path because it is makes everything obvious and easy-to-debug.

> This value should also be available to users at `MELTANO_<PLUGIN_TYPE>_WORKDIR` based on the documentation in [docs.meltano.com/guide/configuration#available-environment-variables](https://docs.meltano.com/guide/configuration#available-environment-variables).

In that case (and if resolution order permits it), dbt's `project_dir` setting definition might actually just be declared with a default `value: $MELTANO_UTILITY_WORKDIR`. I had not considered this previously, and you bring up a very good point here and in your comment re env var expansion. 

Optional:

For extra stability to the plugin developer, we might consider also populating `workdir` (and it's env var) with the value of `MELTANO_PROJECT_ROOT` if no explicit `workdir` is provided. This would allow the plugin to 'not care' how it is getting its `workdir` context, but trust it is there for referencing regardless regardless of how/if it was overridden. 

@rabidaudio - No need to tackle everything in the above in a first iteration but I'm curious your thoughts on these points. For my part, the only things I see as strictly required are (1) agreement on the field name (`workdir` vs `cwd` or alternatives) and (2) enough exploration on behaviors like env var expansion that we can provide docs that tell users what to expect.

Have you seen the new process regarding [integration tests](https://docs.meltano.com/contribute/tests#integration-tests)? Adding this functionality to an existing integration test, or writing a new one would be super helpful. Let us know if you feel comfortable adding yourself or if you'd like assist from a team member on that.

tayloramurphy:
> To confirm, all I meant here is that if the remote plugin definition (on the hub, for instance) is defining a custom workdir, we should try to make that custom workdir as transparent and easy to override as possible for an end user, so they are not confused when move a directory breaks the functionality.

Ah ok, makes sense 👍 

> For extra stability to the plugin developer, we might consider also populating workdir (and it's env var) with the value of MELTANO_PROJECT_ROOT if no explicit workdir is provided. 

I quite like that! Basically just have it as a default property of every plugin than can be overridden.  Nice idea @aaronsteers 😄

rabidaudio:
> This value should also be available to users at MELTANO_<PLUGIN_TYPE>_WORKDIR based on the documentation in https://docs.meltano.com/guide/configuration#available-environment-variables.

> In that case (and if resolution order permits it), dbt's project_dir setting definition might actually just be declared with a default value: $MELTANO_UTILITY_WORKDIR.

@tayloramurphy @aaronsteers 

`MELTANO_<PLUGIN_TYPE>__WORKDIR` (note the double underscores) works out of the box, although currently it's not expanded to an absolute path. Expanding it and changing the name should be easy to include. After that the `pytest` example should work as written!

My thought was doing things like `workdir: $MELTANO_TRANSFORM_PROJECT_DIR` but I like the other way better, where the setting is derived from the workdir setting rather than the other way around.

rabidaudio:
> > For extra stability to the plugin developer, we might consider also populating workdir (and it's env var) with the value of MELTANO_PROJECT_ROOT if no explicit workdir is provided.
>
> I quite like that! Basically just have it as a default property of every plugin than can be overridden. 

I like this two. Two concerns:

1. It's probably a breaking change, as currently all plugins execute in the shell working directory by default, not the project root directory
2. It complicates the override for `DbtInvoker`, as it's hard to tell if the value has been overridden

aaronsteers:
> > > For extra stability to the plugin developer, we might consider also populating workdir (and it's env var) with the value of MELTANO_PROJECT_ROOT if no explicit workdir is provided.
> >
> > I quite like that! Basically just have it as a default property of every plugin than can be overridden. 
> 
> I like this too. Two concerns:
> 
> 1. It's probably a breaking change, as currently all plugins execute in the shell working directory by default, not the project root directory
> 2. It complicates the override for `DbtInvoker`, as it's hard to tell if the value has been overridden

Thanks for calling this out! It's a really good point and based on your points above, I withdraw my suggestion :) - specifically on the basis that the working directory will often need to match the shell directory. For example, if invoking sqlfluff on a specific file in the shell's directory (e.g. `cd transforms/marketing && meltano invoke sqlfluff:fix ./my_table.sql`) we would not want the shell's own working directory to get lost in that case.

I'll think on this a bit more, but per the above, it makes sense to only have an explicit `workdir` if the plugin explicitly defines it.

src/meltano/core/plugin/base.py
edgarrmondragon: 👍 for `workdir`. GitHub actions use `working-directory` and docker uses `workdir`.

tayloramurphy:
@aaronsteers @rabidaudio what's the status on this? Keen to get it in 😄

rabidaudio:
> @aaronsteers @rabidaudio what's the status on this? Keen to get it in 😄

@tayloramurphy  AFAIK this ready to go. We may want to add some documentation, not sure where is best for that as it's more for plugin creators than end users.

edgarrmondragon:
@rabidaudio it might be good to update the JSON schema in `schema/meltano.schema.json` to know about this field.

aaronsteers:
@rabidaudio, @tayloramurphy - It looks like we still need a docs entry and some CI tests are passing.

- [ ] Docs entry for new `workdir` config. (Perhaps in the plugin reference page?)
- [ ] CI tests passing or signoff. (I see dependency review, which we can ignore, but pre-commit tests should pass and we should make a call on if the coverage alarm can be ignored or if we should add additional tests.)
- [ ] Merge conflicts to resolve. (Optionally we can wait for the above to resolve so as not to have to go through multiple merge conflict passes.)

@cjohnhanson - If any of the above need assist, would you be able to jump in?

aaronsteers:
@cjohnhanson - As discussed, feel free to jump in on the above to get this over the line - as time permits.

PR #6410, opened 2022-07-13
ci: Scan Docker images

### LINKED ISSUE #6062: Enable image security scanning in CI

WillDaSilva:
Since we weren't previously doing any container scanning it comes as no surprise that several hundred issues have been flagged by the container scan for each of our 4 containers (one for each supported Python version). We will want to address those in a separate issue.

Closes #6062

### DIFF
--- a/.github/actions/docker-build-scan-push/action.yml
+++ b/.github/actions/docker-build-scan-push/action.yml
@@ -17,6 +17,9 @@ inputs:
   dockerhub-password:
     description: 'The Docker Hub password to use when pushing the image'
     required: true
+  python-version:
+    description: 'The Python version installed on the image'
+    required: true
 
 runs:
   using: 'composite'
@@ -60,6 +63,7 @@ runs:
       uses: github/codeql-action/upload-sarif@v2
       with:
         sarif_file: ${{ steps.anchore-scan.outputs.sarif }}
+        category: grype-python-${{ inputs.python-version }}
 
     - name: Login to Docker Hub
       uses: docker/login-action@v2
--- a/.github/actions/docker-build-scan-push/action.yml
+++ b/.github/actions/docker-build-scan-push/action.yml
@@ -69,6 +69,7 @@ runs:
     - name: Provide a link to results
       shell: bash
       run: |
+        cat ${{ steps.anchore-scan.outputs.sarif }}
         echo "View scan results at:"
         echo "https://github.com/meltano/meltano/security/code-scanning?query=ref:${GITHUB_REF}"
         npx -y violations-command-line -print-violations false -max-violations 0 -v "SARIF" "." ".*sarif$" "" > /dev/null 2>&1
--- a/.github/actions/docker-build-scan-push/action.yml
+++ b/.github/actions/docker-build-scan-push/action.yml
@@ -61,11 +61,11 @@ runs:
 
     - name: Provide a link to results
       shell: bash
+      continue-on-error: true
       run: |
         echo "View scan results at:"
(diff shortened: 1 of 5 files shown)


WillDaSilva:
I've enabled [Scanitizer](https://github.com/apps/scanitizer) for the Meltano repository. It's a tool developed by GitHub to make it easy to update the `Azure/container-scan` allow-list.

tayloramurphy (approved):
Approving docs since it's a small change!

edgarrmondragon:
@WillDaSilva is it safe to trigger the workflow in dry run mode to test it?

afolson (approved):

WillDaSilva:
> is it safe to trigger the workflow in dry run mode to test it?

@edgarrmondragon It should be safe since it uses the same dry-run logic as previously (which is a bit misleading imo since even in dry-run mode it still uploads one of the images to Docker Hub). At least it's safe if we consider the previous implementation of dry-run safe.

Unfortunately GitHub does not allow you to manually run a workflow using workflow dispatch from a PR, so I'll update it to run on PRs for the sake of testing it. I'll also update it to not push an image to Docker Hub in any scenario. Once we can verify that it's working as intended in this PR, I'll reset the configuration to run on workflow dispatch instead of PRs.

WillDaSilva:
Using https://github.com/meltano/meltano/runs/7367915153?check_suite_focus=true as an example, we can see how the container scan fails. Whether it passes or fails, it provides a link to the results. It'll fail if any of the results have a severity greater than the cutoff we provide, which is currently `4.0`, i.e. medium.

Here's a link directly to the scan results: https://github.com/meltano/meltano/security/code-scanning?query=ref:refs/pull/6410/merge

I think actually addressing the many issues found falls outside of the scope for this PR.

This PR also makes a few changes to how the Docker publish workflow works beyond adding security scanning:
- The `dry-run` workflow input now totally prevents the upload of images, rather than allowing the SHA tagged image to be uploaded.
- Only 1 image is built per Python version, rather than 2 or 3.
- Image tags are now unique. Previously some tags were being set/pushed multiple times.

aaronsteers (changes requested):
@WillDaSilva - Thanks for creating this PR. Only because I'm not confident of the implications here, I'll ask to put a hold request to give time for review and discussion.

aaronsteers:
@WillDaSilva - My reading of sample outputs in your prior comment was that currently the scanning fails with critical findings - and that this would block publish of images to Docker Hub. Is that expected or am I misreading perhaps?

Thanks!

aaronsteers:
@WillDaSilva - As a path to resolve some/most of the critical findings, what do you think of incorporating 

- #3203?

WillDaSilva:
@aaronsteers @edgarrmondragon I've committed [be9a74c](https://github.com/meltano/meltano/pull/6410/commits/be9a74ca1e8bdeeea92e10864bd6e6b48201341d) to make this PR not fail the workflow if vulnerabilities (regardless of their severity) are detected. We can merge this in now, and take care of those issues (i.e. by [improving the base image](https://github.com/meltano/meltano/issues/3203) + additional fixes) later.

aaronsteers (approved):
Looks great! Merge when ready.

edgarrmondragon (approved):

PR #6413, opened 2022-07-13
ci: Add dependency license allow-list

### LINKED ISSUE #6393: Enable dependency license scanning in CI

WillDaSilva:
Closes #6393


WillDaSilva:
`actions/dependency-review-action@v2` only checks the dependencies changed by a PR. Before merging this PR I want to somehow apply the check to every dependency so that we can ensure we have the licenses we need in the allow-list, and to ensure that we don't have any incompatible dependencies.

PR #6415, opened 2022-07-13
test: Snowplow fixture coverage

WillDaSilva:
Per our discussion on Slack.

### DIFF
--- a/tests/fixtures/cli.py
+++ b/tests/fixtures/cli.py
@@ -24,9 +24,9 @@ class MeltanoCliRunner(CliRunner):
 
     def invoke(self, *args, **kwargs) -> Any:
         results = super().invoke(*args, **kwargs)
-        if self.snowplow:  # nocov
-            assert self.snowplow.all()["bad"] == 0  # nocov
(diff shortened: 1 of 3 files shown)


tests/fixtures/cli.py
```diff
@@ -24,8 +24,9 @@ def __init__(self, *args, snowplow: SnowplowMicro | None = None, **kwargs):
 
     def invoke(self, *args, **kwargs) -> Any:
         results = super().invoke(*args, **kwargs)
-        if self.snowplow:
-            assert self.snowplow.all()["bad"] == 0
+        if self.snowplow:  # nocov
```
edgarrmondragon, line 27: I can't find `nocov` in the coverage docs, does it work the same as `# pragma: no cover`?

https://coverage.readthedocs.io/en/6.3.3/excluding.html#excluding-code-from-coverage-py

edgarrmondragon: Ah I just saw you'd added to the coverage config, sorry about that. `pragma: no cover` is built-in so it's probably preferred anyway 👍

edgarrmondragon (approved):

PR #6430, opened 2022-07-15
Docs: Add feedback widget

afolson:
Adds a widget to the docs that allows us to collect feedback:
Screen Shot 2022-07-15 at 12 36 48 PM

This appears on every page:

Screen Shot 2022-07-15 at 12 37 34 PM

### DIFF
--- a/docs/_includes/head.html
+++ b/docs/_includes/head.html
@@ -18,4 +18,14 @@ j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
 'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
 })(window,document,'script','dataLayer','GTM-KS47SNT');</script>
 <!-- End Google Tag Manager -->
+
+<!-- Freddy Feedback Widget -->
+<script>
+    var ffWidgetId = '21957808-3a88-445b-88f8-4530d7c60cef';
+    var ffWidgetScript  = document.createElement('script');
+    ffWidgetScript.type = 'text/javascript';
+    ffWidgetScript.src = 'https://freddyfeedback.com/widget/freddyfeedback.js';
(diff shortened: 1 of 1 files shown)


afolson:
@tayloramurphy I can't seem to get it to anchor to the bottom of the TOC. It looks like we can have more control over the CSS stuff once we upgrade to a paid account.

There's also no way to make it look exactly like the one on the homepage since they host all of the JS for it. I could probably do a dirty hack but I don't know if that might break in the future.

emilylucie (approved):

tayloramurphy:
@afolson can we make the TOC in a smaller column then? I definitely don't want it to overlap the text at all.

pandemicsyn (approved):
lgmt - from a "its valid js, isn't throwing console errors, and netlify preview works as i'd expect" standpoint 😁

tayloramurphy:
Actually @afolson, what if we just scoot it down the page and have it about 1/4 up the page from the bottom? I don't like the TOC overlap and scooting it down seems easier than making than making the TOC smaller.

afolson:
@tayloramurphy I'm having a hard time placing it manually without it clobbering something or shifting the page somehow. Can we just accept it as an iteration and I'll chat with their support team to see when they're adding more options? Ideally we'd be able to control the whole widget but it looks like they're still building some of that out.

tayloramurphy (changes requested):
@afolson the latest iteration in the preview has it pop-up after scrolling for a bit, but it seems like it's going to happen for any page which feels like it'll get annoying very quick. 

Can we do a more stable iteration where's it's statically appended to the bottom of every page near the `Edit this page` text? Then we can experiment with more dynamic placing and interaction. I get it's not ideal from the perspective of making users more aware of it, but I'd rather that than an obtrusive interaction pattern.

afolson:
I'll just find another solution. Thanks!

PR #6431, opened 2022-07-15
fix: Allow `meltano run` to run without providing subcommand

### LINKED ISSUE #6188: meltano run should not require a command to execute a plugin
Per https://github.com/meltano/meltano/pull/6121#discussion_r895929389 it should be a simple change.

From @pandemicsyn the code is likely in https://github.com/meltano/meltano/blob/dfd9650170ff97fd69fdff7ac8dc7dc0841bf663/src/meltano/core/block/singer.py#L97 and we would need a docs update with this as well.

cjohnhanson:
Closes #6188

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ This project adheres to [Semantic Versioning](http://semver.org/) and [Keep a Ch
 
 - [#3419](https://github.com/meltano/meltano/issues/3419) Change default `meltano config` behavior to --no-environment.
 
+- [#6188](https://github.com/meltano/meltano/issues/6188) Allow `meltano run` to run without provided subcommand
+
 ### Fixes
 
 ### Breaks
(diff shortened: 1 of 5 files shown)


edgarrmondragon (approved):

tayloramurphy (changes requested):
@cjohnhanson this needs a doc change as well - see pt. 3 https://docs.meltano.com/reference/command-line-interface#tasks

pandemicsyn (approved):
lgtm. Small nit (feel free to ignore). Similar to what Taylor mentioned, wonder if we should clarify usage. Might be worth updating `cli/run.py` usage to say:

```
      Run a set of command blocks in series.

      Blocks are specified as a list of plugin names, e.g.     `meltano run
      some_extractor some_loader some_plugin:some_optional_command` and are run in the
      order they are specified from left to right. A failure in any block
      will cause the entire run to abort.
```

To flag that the command is optional?

tayloramurphy (approved):

PR #6432, opened 2022-07-15
feat: meltano now surfaces full stderr output for catalog discovery failures.

### LINKED ISSUE #6150: Catalog discovery errors are cumbersome to debug
When a catalog discovery attempt fails, the error message does not give enough information about what happened:

```
$ meltano select tap-postgres --list --all
2022-06-10T14:46:22.487369Z [info     ] Environment 'dev' is active
Cannot list the selected attributes: Catalog discovery failed: command ['/home/juanlu/Projects/Orchest/Singer/test-replication-none/.meltano/extractors/tap-postgres/venv/bin/tap-postgres', '--config', '/home/juanlu/Projects/Orchest/Singer/test-replication-none/.meltano/run/tap-postgres/tap.c331e588-15d0-4ec6-aa62-90ec0c99c76f.config.json', '--discover'] returned 1
```

In fact, the intermediate `.meltano/run/tap-postgres/tap.c331e588-15d0-4ec6-aa62-90ec0c99c76f.config.json` config file gets automatically deleted, which makes debugging more cumbersome.

The current workaround is to:

1. Generate a config manually: `meltano config tap-postgres > config.json`
2. Rerun the command with the custom config, assuming it's the same as the temporary one generated on the fly: `/home/juanlu/Projects/Orchest/Singer/test-replication-none/.meltano/extractors/tap-postgres/venv/bin/tap-postgres --config config.json --discover`
3. See actual error

It would be nice if the "returned 1" message would also give the output of the command, for example:

```
$ /home/juanlu/Projects/Orchest/Singer/test-replication-none/.meltano/extractors/tap-postgres/venv/bin/tap-postgres --config config.json --discover
time=2022-06-10 16:49:52 name=tap_postgres level=CRITICAL message=Config is missing required keys: ['dbname']
  File "/home/juanlu/Projects/Orchest/Singer/test-replication-none/.meltano/extractors/tap-postgres/venv/bin/tap-postgres", line 8, in <module>
    sys.exit(main())
  File "/home/juanlu/Projects/Orchest/Singer/test-replication-none/.meltano/extractors/tap-postgres/venv/lib/python3.8/site-packages/tap_postgres/__init__.py", line 435, in main
    raise exc
  File "/home/juanlu/Projects/Orchest/Singer/test-replication-none/.meltano/extractors/tap-postgres/venv/lib/python3.8/site-packages/tap_postgres/__init__.py", line 432, in main
    main_impl()
  File "/home/juanlu/Projects/Orchest/Singer/test-replication-none/.meltano/extractors/tap-postgres/venv/lib/python3.8/site-packages/tap_postgres/__init__.py", line 393, in main_impl
    args = parse_args(REQUIRED_CONFIG_KEYS)
  File "/home/juanlu/Projects/Orchest/Singer/test-replication-none/.meltano/extractors/tap-postgres/venv/lib/python3.8/site-packages/tap_postgres/__init__.py", line 384, in parse_args
    utils.check_config(args.config, required_config_keys)
  File "/home/juanlu/Projects/Orchest/Singer/test-replication-none/.meltano/extractors/tap-postgres/venv/lib/python3.8/site-packages/singer/utils.py", line 190, in check_config
    raise Exception("Config is missing required keys: {}".format(missing_keys))
Exception: Config is missing required keys: ['dbname']
```

cjohnhanson:
When invoking discovery for Singer taps, we previously redirected stderr to `asyncio.subprocess.DEVNULL` _unless_ debug logging was enabled. With this PR, we instead redirect that stderr stream into a `BytesIO` and then that buffer is tacked on to the error message. So anywhere that a catalog discovery error is surfaced, it will now include the specific stderr output from the underlying plugin invocation. 

We had explicit tests around the existing stream redirection logic, so those tests have been changed and assertions removed to reflect he new logic.

Closes #6150

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/) and [Keep a Ch
 ### Changes
 
 - [#3419](https://github.com/meltano/meltano/issues/3419) Change default `meltano config` behavior to --no-environment.
+- [#6150](https://github.com/meltano/meltano/issues/6150) Print full stderr for catalog discovery failures.
 
 ### Fixes
 
--- a/docs/example-library/meltano-run/ending-meltano.yml
+++ b/docs/example-library/meltano-run/ending-meltano.yml
@@ -14,7 +14,7 @@ plugins:
   transformers:
   - name: dbt-postgres
     variant: dbt-labs
-    pip_url: dbt-core~=1.0.0 dbt-postgres~=1.0.0
+    pip_url: dbt-core~=1.1.0 dbt-postgres~=1.1.0
 environments:
 - name: dev
   config:
--- a/src/meltano/core/plugin/singer/tap.py
+++ b/src/meltano/core/plugin/singer/tap.py
@@ -9,7 +9,7 @@ import shutil
 import sys
 from asyncio.streams import StreamReader
 from hashlib import sha1
-from io import BytesIO
+from io import StringIO
 from pathlib import Path
 from typing import Tuple
 
@@ -38,23 +38,24 @@ logger = structlog.getLogger(__name__)
 
 
 async def _stream_redirect(
-    stream: asyncio.StreamReader, file_like_obj, write_str=False
+    stream: asyncio.StreamReader, *file_like_objs, write_str=False
 ):
     """Redirect stream to a file like obj.
 
     Args:
         stream: the stream to redirect
-        file_like_obj: the object to redirect the stream to
+        file_like_objs: the objects to redirect the stream to
         write_str: if True, stream is written as str
     """
     encoding = sys.getdefaultencoding()
     while not stream.at_eof():
         data = await stream.readline()
-        file_like_obj.write(data.decode(encoding) if write_str else data)
+        for file_like_obj in file_like_objs:
+            file_like_obj.write(data.decode(encoding) if write_str else data)
 
 
 def _debug_logging_handler(
-    name: str, plugin_invoker: PluginInvoker, stderr: StreamReader
+    name: str, plugin_invoker: PluginInvoker, stderr: StreamReader, *other_dsts
 ) -> asyncio.Task:
     """Route debug log lines to stderr or an OutputLogger if one is present in our invocation context.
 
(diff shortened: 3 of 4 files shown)


edgarrmondragon (approved):
nice!

pandemicsyn (commented):
@cjohnhanson This *only* prints discovery stderr if the process exited with a non status code right? Any scenario's where you'd still want to enable printing of discovery stderr via a --log-level=debug or the like? 

Like if the plugin is misbehaving and printing log lines but still exiting with a 0 exit code?

cjohnhanson:
> @cjohnhanson This _only_ prints discovery stderr if the process exited with a non status code right? Any scenario's where you'd still want to enable printing of discovery stderr via a --log-level=debug or the like?
> 
> Like if the plugin is misbehaving and printing log lines but still exiting with a 0 exit code?

@pandemicsyn yeah good point, as-is this redirects to a buffer that only ever gets printed as part of a `PluginExceptionError` error message if discovery has a nonzero exit. 

I'm not sure how common it is for Singer taps to print to stderr as part of a successful invocation of discovery, but there's definitely nothing in the Singer spec that disallows it and I think it makes sense to surface stderr in debug logging regardless of whether or not discovery was successful. @tayloramurphy and/or @aaronsteers I'd be interested in your thoughts on this.

So there's a few options::
1. Redirect the plugin's stderr to a buffer iff debug logging is not enabled, else redirect to `PIPE` so it gets logged as normal and then omit stderr from any error message that results from a nonzero exit. This is the lowest effort option but in principle I don't like the idea of logging settings affecting any behavior besides logging.
1. Copy the plugin's stderr stream to _both_ `PIPE` _and_ a buffer iff debug logging is enabled. This means that if discovery fails, the plugin's stderr will be printed in both the logs _and_ the error messages. This will be very noisy but that's to be expected with debug logging. 
1. Leave as-is and assume it's okay to swallow stderr output from a successful invocation even if debug logging is enabled.

Thoughts? I lean towards the second option because I think it adheres best to the Principle of Least Astonishment (debug logging logs everything you'd expect it to and doesn't change what gets included in error messages) but it also means a lot of noisy and redundant output in the case of nonzero exit.

pandemicsyn:
I'm a +1 on option 2 as well.

cjohnhanson:
@pandemicsyn @tayloramurphy [the latest changes](https://github.com/meltano/meltano/pull/6432/files#diff-a6627846eb7707c408d1e7bd87f16d9f100410c7f9569342ac4155fce916033cL40) make it so behavior is in line with option 2 described above.

pandemicsyn (approved):
🚀

PR #6433, opened 2022-07-15
fix: Validating Mapper and Mappings names and alerting if necessary

### LINKED ISSUE #6189: Disallow calling a `mapper` by name in `meltano run` and alert to use a `mapping` name
The document misses a step for mapper configuration. Property `mapping_name` has to be added to the mappers extras, and its value need to match with the name under mappings. It doesn't seem been documented anywhere.

```yaml
plugins:
  mappers:
  - name: transform-field
    variant: transferwise
    pip_url: pipelinewise-transform-field
    executable: transform-field
    mapping_name: transform-field-hash
    mappings:
    - name: transform-field-hash
      config:
        transformations:
          - field_id: Name
            tap_stream_name: dbo-Test
            type: MASK-HIDDEN
```

BuzzCutNorman:
This is an attempt to resolve issue 6189.  I am needing some feed back from the Meltano team on if the validation is occurring were you expected it to ?  I placed a number of loggers so I could find out where I could attempt the validation and wanted to say I am planning on removing them later on.  

Closes #6189


pandemicsyn:
@BuzzCutNorman you've got the right spot! If you wanna clean up the PR and flip it off draft status, we can probably get this one merged pretty quick. 

There are a bunch of example tests you can crib from in https://github.com/meltano/meltano/blob/main/tests/meltano/cli/test_run.py. You can either add a new one explicitly covering this, but https://github.com/meltano/meltano/blob/main/tests/meltano/cli/test_run.py#L199 might be the best spot. Give me a shout if you have any questions about adding test, would be happy to send along an example.

BuzzCutNorman:
@pandemicsyn Thanks for clarifying that this was the proper place to put this check.  I have not completed the second part of the item which was to give the user what mapper name we think they were going to use.  I can always come back to that later we would just need to keep the issue open?  The other thing that is bugging be is I can't get the traceback to not show.  I tried `.with_traceback(None)` and I still got the traceback.

I looked at the test examples and I have a question I see a `tap.name` and `target.name` do I need to add `mapper.name` to the function input variables?  Based on the examples I was thinking I need ` args = ["run", tap.name, mapper.name, target.name]`.  Then again I could be over complicating this and I just need have `pytest.raises` the error message a put in place and have the test check for it.

BuzzCutNorman:
Ok I should clarify I keep getting the traceback showing with I have `--log-level debug` on please let me know if that is the default.  I just didn't see that code being helpful but I guess if you want to know the code that is flagging this it makes sense.

pandemicsyn:
> I looked at the test examples and I have a question I see a `tap.name` and `target.name` do I need to add `mapper.name` to the function input variables? Based on the examples I was thinking I need ` args = ["run", tap.name, mapper.name, target.name]`. Then again I could be over complicating this and I just need have `pytest.raises` the error message a put in place and have the test check for it.

Its a little more than that, those tests actually call `meltano run` in a more integration test like fashion, but in way where taps, targets, mappers are mocked.

So you you replicate what you acutally want to happen. So you'd need to add yourself a mapping that actually has a name collision with the mocker. The easiest way to do is probably just add yourself a mapping in: https://github.com/meltano/meltano/blob/3a6ff7e7e167f51b9a5570eeaee48faed032a9c0/tests/fixtures/core.py#L504-L530

There's already two mapping present, so just adding one called "mapper-mock" (which should be the name of the mapper), would let you trigger the collision you're looking for.  Then over in the test you could test by adding a scenario something like:


```python
        # test mapper/mapping name collision detection
     args = ["run", tap.name, mapper.name, target.name] # this is effectively your meltano run statement, so you wanna include that mapper name here in the invocation because it will also match a mapping name
        with mock.patch.object(SingerTap, "discover_catalog"), mock.patch.object(
            SingerTap, "apply_catalog_rules"
        ), mock.patch(
            "meltano.core.plugin_invoker.asyncio"
        ) as asyncio_mock2, mock.patch(
            "meltano.core.block.parser.ProjectPluginsService",
            return_value=project_plugins_service,
        ):
            asyncio_mock2.create_subprocess_exec = create_subprocess_exec
            with pytest.raises(
                Exception,
                match="YOUR NEW EXCEPTION MESSAGE", 
            ):
                result = cli_runner.invoke(cli, args, catch_exceptions=False)
                assert result.exit_code == 1
```

pandemicsyn:
> The other thing that is bugging me is I can't get the traceback to not show. I tried `.with_traceback(None)` and I still got the traceback.

You're still raising an exception so you'll still get a traceback, but the contents of it should be different? It its not than the other error condition is probably just triggering before you get to this new check.

BuzzCutNorman:
Thanks, for showing me how the tests are done.  I got the test added using (cut and pasting) your example code.  Please review when you get a chance.  

I was not able to come up with any mapper suggestion logic.  Seems like I would need to add a `ProjectPluginService` function to `find_mappers_by_plugin_name` then present say the top 5 at max if present.

tests/meltano/cli/test_run.py
```diff
@@ -1085,6 +1085,24 @@ def test_run_mapper_config(
             assert result.exit_code == 1
             assert "Error: Block not-a-valid-mapping-name not found" in result.stderr
 
+        # test mapper/mapping name collision detection
+        args = ["run", tap.name, "mapper-mock", target.name]
```
pandemicsyn, line 1089: Think you missed adding a mapping name that collides with mapper name. Test was still passing because it would still catch when being invoked with a mapper name, but it wasn't actually being triggered by a collision.

```suggestion
        project_add_service.add(
            PluginType.MAPPERS,
            "mapper-collision",
            inherit_from=mapper.name,
            mappings=[
                {
                    "name": "mapper-collision",
                    "config": {
                        "transformations": [
                            {
                                "field_id": "author_email1",
                                "tap_stream_name": "commits1",
                                "type": "MASK-HIDDEN",
                            }
                        ]
                    },
                }
            ],
        )

        args = ["run", tap.name, "mapper-collision", target.name]
```

tests/meltano/cli/test_run.py
```diff
@@ -1085,6 +1085,24 @@ def test_run_mapper_config(
...
+        ), mock.patch(
+            "meltano.core.plugin_invoker.asyncio"
+        ) as asyncio_mock2, mock.patch(
+            "meltano.core.block.parser.ProjectPluginsService",
+            return_value=project_plugins_service,
+        ):
+            asyncio_mock2.create_subprocess_exec = create_subprocess_exec
+            with pytest.raises(
+                Exception,
+                match="block violates set requirements: Expected unique mappings name not the mapper plugin name: mapper-mock",
```
pandemicsyn, line 1101: To go along with the mapping/mapper name change from above:

```suggestion
                match="block violates set requirements: Expected unique mappings name not the mapper plugin name: mapper-collision",
```

src/meltano/core/block/parser.py
```diff
@@ -290,11 +290,22 @@ def _find_next_elb_set(  # noqa: WPS231, WPS213
...
-                blocks.append(
-                    builder.make_block(
-                        plugin,
+                # Checks to see if the mapper plugin name is the same as the mappings name
+                # If they both match then a validation error is raised because the
+                # meltano run command needs the mappings name to obtain the settings to
+                # pass to the parent mapper plugin.  We also want to fail if the user names them
+                # the same to stop errors due to ambiguous commands.
+                if plugin.name == self._mappings_ref.get(next_block):
+                    self.log.warning("mapper plugin name used")
```
pandemicsyn, line 299:
```suggestion
                    self.log.warning("Found unexpected mapper plugin name." plugin_name=plugin_name)
```

Just adds a little more context (namely what plugin_name was seen)

pandemicsyn (changes requested):
looking great! some minor stuff and a test adjustment suggested inline.

tests/meltano/cli/test_run.py
BuzzCutNorman: Nope, that was intentional since my added parser check is just evaluating if the `PluginType.MAPPERS` block is the same as a mapper plugin name.  Say someone installed the mapper and just tried to run the command `meltano tap-csv transform-field target-postgres` it will fail out and give the error even with no mappings config present in the `meltano.yml`.  That is at least the behavior on my machine.  I will add this as an additional check.

pandemicsyn: 👍 gotcha , two explicit ones makes sense then.

pandemicsyn (approved):

PR #6455, opened 2022-07-19
ci: Automatically cancel stale GitHub workflows

### LINKED ISSUE #6451: Cancel stale workflows
When a PR has multiple pushes in moderately quick succession it ends up running the PR workflows in parallel, but we usually only care about the results from the latest run. Because of that, we should cancel the earlier workflows when a new one is started.

WillDaSilva:
Closes #6451

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -7,6 +7,10 @@ on:
   workflow_dispatch:
     inputs: {}
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 env:
   FOSSA_CLI_VERSION: '3.3.9'
 
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -7,6 +7,10 @@ on:
   workflow_dispatch:
     inputs: {}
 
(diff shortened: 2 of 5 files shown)


WillDaSilva:
The change appears to work. Merging `main` into this branch caused the in-progress workflows for [ff06a07](https://github.com/meltano/meltano/pull/6455/commits/ff06a075e117efefe9e3fc9d86e5cbd1178c1081) to be cancelled.

Annoyingly that counts as a failure rather than having a special status for cancelled, and it causes an email to be sent for cancelled workflow.

.github/workflows/dependency-review.yml
```diff
@@ -5,6 +5,10 @@ on:
   workflow_dispatch:
     inputs: {}
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```
edgarrmondragon, line 9: @WillDaSilva So `head_red` is available in PRs and commits, but not workflow dispatch, hence the fallback?

edgarrmondragon:
docs for `concurrency` in case anyone's curious: https://docs.github.com/en/actions/using-jobs/using-concurrency

edgarrmondragon (approved):

pandemicsyn (approved):

PR #6456, opened 2022-07-20
docs: Update for Windows and for using pipx --python version

seajhawk:
1. Setting env variable in Windows PowerShell. Added:
#For Windows (PowerShell):
$env:MELTANO_ENVIRONMENT="dev"

2. Python version - 
I learned that Meltano must be pipx installed using a version of Python that is compatible with all of the extractors and loaders (see https://meltano.slack.com/x-p526315946325-3801798270198-3808513868754/archives/C01TCRBBJD7/p1657981158809849). I've added instructions for performing a pipx install with a specified version of Python, but I think this issue could use a bit more documentation to help the reader understand when they may need to use that command, especially because line 24 indicates everything _should_ work with Python 3.10, but tap-rest-api-msdk requires Meltano to be installed with 3.9.7 or earlier.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -48,9 +48,11 @@ You will need to be running Linux, macOS, or Windows, and have [Python](https://
 
     ```bash
     pipx install meltano
+    ```
+    If you have multiple versions of Python installed, you can use a specific one with the `--python` arugment:
     
-    #If you have multiple versions of Python installed, you can use a specific one with the --python arugment
-    #  pipx install meltano --python [path to desired Python exe]
+    ```bash
+    pipx install meltano --python [path to desired Python exe]
     ```
 
 1. Optionally, verify that the [`meltano` CLI](/reference/command-line-interface) is now available by viewing the version:
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -48,9 +48,11 @@ You will need to be running Linux, macOS, or Windows, and have [Python](https://
 
     ```bash
     pipx install meltano
+    ```
+    If you have multiple versions of Python installed, you can use a specific one with the --python arugment
     
-    #If you have multiple versions of Python installed, you can use a specific one with the --python arugment
-    #  pipx install meltano --python [path to desired Python exe]
+    ```bash
+    pipx install meltano --python [path to desired Python exe]
     ```
 
 1. Optionally, verify that the [`meltano` CLI](/reference/command-line-interface) is now available by viewing the version:
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -139,10 +139,14 @@ As part of creating your Meltano project, we automatically added your first [env
 1. Activate your environment for your shell session:
 
     ```bash
-    #For Windows (PowerShell): $env:MELTANO_ENVIRONMENT="dev"
-    export MELTANO_ENVIRONMENT=dev
+        export MELTANO_ENVIRONMENT=dev
     ```
-
+    or for Windows PowerShell:
+    
+    ```powershell
+        $env:MELTANO_ENVIRONMENT="dev"
+    ```
+    
     Alternatively you can include the `--environment=dev` argument to each meltano command. You should now see a log message that says `Environment 'dev' is active` each time you run a meltano command.
 
 1. [optional] Add a new environment:
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -144,7 +144,7 @@ As part of creating your Meltano project, we automatically added your first [env
     or for Windows PowerShell:
     
     ```powershell
-        $env:MELTANO_ENVIRONMENT="dev"
+    $env:MELTANO_ENVIRONMENT="dev"
     ```
     
     Alternatively you can include the `--environment=dev` argument to each meltano command. You should now see a log message that says `Environment 'dev' is active` each time you run a meltano command.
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -48,6 +48,9 @@ You will need to be running Linux, macOS, or Windows, and have [Python](https://
 
     ```bash
     pipx install meltano
+    
+    #If you have multiple versions of Python installed, you can use a specific one with the --python arugment
+    #  pipx install meltano --python [path to desired Python exe]
```
WillDaSilva, line 53: The line "If you have multiple [...]" should be moved outside of the codeblock, and then the `pipx install meltano --python [...]` line should be in its own codeblock beneath that.

docs/getting-started.md
```diff
@@ -133,6 +136,7 @@ As part of creating your Meltano project, we automatically added your first [env
 1. Activate your environment for your shell session:
 
     ```bash
```
WillDaSilva, line 138: Instead of having a comment in the `bash` block for Windows, we should use a tabbed code block with a tab for `bash`, and another for `powershell`. Unfortunately I don't believe Kramdown supports this. For now we can just use two code blocks instead, one after the other.

@afolson when we get started on https://github.com/meltano/meltano/issues/5948, we should consider switching to a different markdown processor. I'm not sure how much attachment there is to our use of Jekyll for the docs. If there isn't much, and it wouldn't be too much trouble, we could switch the entire docs site to Mkdocs, and get the more powerful/extensible markdown process for free. Otherwise we can have the generated API docs as a sub-site under the Jekyll docs site, and make an effort towards making their styles match at least enough that switching between them wouldn't be distracting.

WillDaSilva (changes requested):
Thanks for the doc improvements! I've requested a couple small changes so as to avoid having the Windows instructions as comments.

docs/getting-started.md
seajhawk: Regarding use of "tabbed code block". My company does this and they can be extremely helpful in providing clarity. However, we are still working on exactly how/when to use them for the best user experience. Here is some recent advice I gave to my content team that you may want to consider:
 
 - **No tabs where possible** - if the command is the same between bash and powershell, and it's not so long that you need line continuation characters just show one code block. Less cognitive load for the reader, less maintenance.
 
 - **Link the tabs** - if you have multiple tabbed code block examples in a doc, when you switch one to bash or powershell the rest should switch as well. Prevents reader from 

 - **Line continuation characters** - \ for bash and ` for powershell. Use them sparingly as they will force you to have different tabs for bash/powershell.

 - **Variables** - Variables can make for easier to read commands, but they will force you to have different tabs and individual lines that use variables cannot be copied and run on their own.

docs/getting-started.md
seajhawk: I was trying to follow other examples, but I like this a lot better.
Fixed.

docs/getting-started.md
```diff
@@ -50,6 +50,12 @@ You will need to be running Linux, macOS, or Windows, and have [Python](https://
     pipx install meltano
     ```
 
+    If you have multiple versions of Python installed, you can use a specific one with the `--python` arugment:
+    
+    ```bash
+    pipx install meltano --python [path to desired Python exe]
```
WillDaSilva, line 56: - `<...>` instead of `[...]` to indicate that the argument is not optional.
- `executable` instead of `exe` to make it clear that we aren't necessarily referring to a Windows `.exe` file.

```suggestion
    pipx install meltano --python <path to desired Python executable>
```

WillDaSilva (changes requested):

tayloramurphy (approved):

tayloramurphy:
@WillDaSilva are the updates you requested done? Is this ready to go?

docs/getting-started.md
WillDaSilva: @seajhawk Thanks for the advice around tabbed code blocks. Since we can't use them yet, we should split this into 2 code blocks, one after the other.

WillDaSilva:
> are the updates you requested done? Is this ready to go?

@tayloramurphy almost, but there's still a code block that should be split into 2. See the unresolved review comment above.

docs/getting-started.md
seajhawk: @WillDaSilva - split now.

docs/getting-started.md
```diff
@@ -133,9 +139,14 @@ As part of creating your Meltano project, we automatically added your first [env
...
 
     ```bash
-    export MELTANO_ENVIRONMENT=dev
+        export MELTANO_ENVIRONMENT=dev
     ```
-
+    or for Windows PowerShell:
+    
+    ```powershell
+        $env:MELTANO_ENVIRONMENT="dev"
```
WillDaSilva, line 147:
```suggestion
    $env:MELTANO_ENVIRONMENT="dev"
```

docs/getting-started.md
```diff
@@ -133,9 +139,14 @@ As part of creating your Meltano project, we automatically added your first [env
 1. Activate your environment for your shell session:
 
     ```bash
-    export MELTANO_ENVIRONMENT=dev
+        export MELTANO_ENVIRONMENT=dev
```
WillDaSilva, line 142:
```suggestion
    export MELTANO_ENVIRONMENT=dev
```

WillDaSilva (approved):
Thanks again @seajhawk! I've left 2 suggestions regarding the indentation of the code blocks, and other than that this looks good to me!

tayloramurphy:
@WillDaSilva I think this is good to merge. I'm not sure if we need to solve the PR pre-commit check or not so I'll leave that to you 😄

PR #6466, opened 2022-07-22
fix(ui): When feature flag `enable_uvicorn` is true require write lock when accessing Project.meltano attribute 

BuzzCutNorman:
This is an attempt at option 3 proposed in PR #6297 that asked to put the `write_lock()` behind the `enable_uvicorn` feature flag.

### DIFF
--- a/src/meltano/core/project.py
+++ b/src/meltano/core/project.py
@@ -231,9 +231,23 @@ class Project(Versioned):  # noqa: WPS214
             the current meltano config
         """
         from .meltano_file import MeltanoFile
+        from .settings_service import FEATURE_FLAG_PREFIX, FeatureFlags
+        from .yaml import configure_yaml
 
-        with self._meltano_rw_lock.read_lock():
-            return MeltanoFile.parse(self.project_files.load())
+        with open(self.meltanofile) as melt_ff:
+            meltano_ff: dict = configure_yaml().load(melt_ff)
+
+        uvicron_enabled = (
+            meltano_ff.get(f"{FEATURE_FLAG_PREFIX}.{FeatureFlags.ENABLE_UVICORN}")
+            or False
+        )
+
+        if uvicron_enabled:
+            with self._meltano_rw_lock.write_lock():
+                return MeltanoFile.parse(self.project_files.load())
+        else:
+            with self._meltano_rw_lock.read_lock():
+                return MeltanoFile.parse(self.project_files.load())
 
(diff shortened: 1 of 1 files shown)


aaronsteers:
@BuzzCutNorman - Very cool to see this PR created. I don't know if I'm the best approver so I won't jump into the code review quite yet. I do want to call out that this test is a new one and shouldn't have anything to do with the code you've changed. Please ignore:

image

WillDaSilva:
@aaronsteers The dependency review workflow will work for forks once #6462 has been merged.

src/meltano/core/project.py
```diff
@@ -231,9 +231,23 @@ def meltano(self) -> MeltanoFileHint:
...
         from .meltano_file import MeltanoFile
+        from .settings_service import FEATURE_FLAG_PREFIX, FeatureFlags
+        from .yaml import configure_yaml
 
-        with self._meltano_rw_lock.read_lock():
-            return MeltanoFile.parse(self.project_files.load())
+        with open(self.meltanofile) as melt_ff:
+            meltano_ff: dict = configure_yaml().load(melt_ff)
+
+        uvicron_enabled = (
```
cjohnhanson, line 240:
```suggestion
        uvicorn_enabled = (
```

src/meltano/core/project.py
```diff
@@ -231,9 +231,23 @@ def meltano(self) -> MeltanoFileHint:
...
+            meltano_ff.get(f"{FEATURE_FLAG_PREFIX}.{FeatureFlags.ENABLE_UVICORN}")
+            or False
+        )
+
+        if uvicron_enabled:
```
cjohnhanson, line 245:
```suggestion
        if uvicorn_enabled:
```

cjohnhanson (changes requested):
@BuzzCutNorman Thanks for tackling this! I've got a couple nitpicks but other than those, this looks good to go.

src/meltano/core/project.py
BuzzCutNorman: @cjohnhanson thanks for catching that.  I made and committed the suggested changes.

cjohnhanson (approved):
Looks good!

We'll wait on an additional codeowner approval from either @pandemicsyn or @edgarrmondragon. Barring any requested changes there, this should be good to go.

edgarrmondragon (approved):

Issue #6160, opened 2022-06-10
Windows Requires Administrator Access


visch:
If you run on Windows without Administrator Access you currently get at least this error

```
>       self._accessor.symlink(target, self, target_is_directory)
E       OSError: [WinError 1314] A required privilege is not held by the client: 'e:\\code\\meltano\\.venv\\scripts\\meltano' -> 'C:\\Users\\derek\\AppData\\Local\\Temp\\pytest-of-derek\\pytest-7\\meltano_root0\\project_testwindowselt\\.meltano\\run\\bin'

..\..\..\..\Programs\Python\Python38\lib\pathlib.py:1391: OSError
```

Since this one is related to symlinks we could look into using Junctions on windows https://stackoverflow.com/questions/9042542/what-is-the-difference-between-ntfs-junction-points-and-symbolic-links

visch:
This needs some more testing, I setup a project as a user account on windows and didn't hit this error over the weekend didn't dive into why though.

visch:
> This needs some more testing, I setup a project as a user account on windows and didn't hit this error over the weekend didn't dive into why though.

pytest requires administrator, but generally everything works on Windows without Administrator. At least these commands

- meltano init
- meltano run
- meltano add
- meltano state (seemed to work didn't have state to copy around)
- meltano invoke

Pretty much every test fails when running `poetry run pytest` as a user instead of Administrator

Looks like it's due to the `project` fixture, we could probably tweak it to work, and we'd need to figure out how we could run github actions as a non administrator as well

```
E:\code\meltano\src\meltano\core\project_init_service.py:57: in init
    Project.activate(self.project)
e:\code\meltano\.venv\lib\site-packages\fasteners\lock.py:286: in wrapper
    return f(self, *args, **kwargs)
E:\code\meltano\src\meltano\core\project.py:139: in activate
    project.run_dir().joinpath("bin").symlink_to(executable)
[...]

PR #6468, opened 2022-07-22
fix: symbolic link creation for meltano.exe when on windows

### LINKED ISSUE #6467: bug: Meltano Upgrade on Windows stops with error "is not a valid Win32 application"
### Meltano Version

2.3.0

### Python Version

3.9

### Bug scope

### Operating System

Windows Server 2022 Standard

### Description

I have noticed on Windows if the file(symbolic link) `.meltano/run/bin` is present in your project when you run the upgrade via `meltano upgrade` or the `melatno ui` "Upgrade Button" you will get the following `WinError`.

`[WinError 193] %1 is not a valid Win32 application`

if you remove the file and run `meltano upgrade` the upgrade will complete. I am not sure what the long term ramification of deleting the file is.

image

### Code

_No response_

BuzzCutNorman:
This fixes the bug #6467 where on Windows `meltano upgrade` stops because `MeltanoInvoker.invoke()` is calling `subprocess.run()` which requires a executable target.  Since the symbolic link `.meltano\run\bin` is targeting `meltano` not `meltano.exe` a failure occurs.  I updated the `Project.activate()` function to target `meltano.exe` when Windows is detected.

This does not fix any previously created symbolic links that might exist in projects.

Close #6467


visch:
Still need to dive into this a bit more, we should enable the upgrade test on Windows as well. https://github.com/meltano/meltano/blob/main/tests/meltano/cli/test_upgrade.py#L12 is probably the right test to be "reenabled for windows here!)

BuzzCutNorman:
Vish, Would you like me to comment the if statement out save, commit, and publish to see what happens ?

visch:
> Vish, Would you like me to comment the if statement out save, commit, and publish to see what happens ?

Well I'm trying to understand what we're fixing here. `meltano upgrade` actually runs fine for me locally. So it's just `meltano ui` that doesn't. 

A test helps clear up what code path(s) we're taking here that caused this issue. Maybe it's not easy to do because of the UI?

BuzzCutNorman:
Vish, This is an edge case that I have run into.  If you run `dir .\.meltano\run` inside a project folder and you see the file `bin` the upgrade will fail.  This files is created when a project is activated.  If `.\.meltano\run\bin`  is present and you run `meltano upgrade` the you will get the error if it is not present you won't.  

You are correct this is separate from the UI issue you filed.   That is why I created a new bug issue.  Though it could in some cases be a root cause of a UI upgrade failure.

BuzzCutNorman:
Visch, I guess I don't even know what purpose this `bin` symbolic link has or if it is even needed anymore.   Windows fails out at initialization since it can't find a meltano file and hasn't been creating a `bin`.  I wonder if maybe what needs to be done is instead of pointing it to the `meltano.exe`.  I just need to make sure always fails and never creates the file for windows for any enviroment `venv`, `pipx`, `poetry` .

BuzzCutNorman:
If you want to see what the symbolic link `.\.meltano\run\bin` is targeting in Windows run this from a powershell prompt while in a Meltano project folder.

` dir .\.meltano\run\ -recurse -force | ?{$_.LinkType} | select FullName,LinkType,Target`

I think you can do this other ways but I found this one at it has worked well for me.

visch:
@BuzzCutNorman  nice work this works great on Windows (as long as you run as an administrator, but I think tackling that in https://github.com/meltano/meltano/issues/6160 makes sense as we'd probably want to offer junctions and symlinks re https://stackoverflow.com/questions/9042542/what-is-the-difference-between-ntfs-junction-points-and-symbolic-links )

visch:
@edgarrmondragon  looks good to me! Unless we can get rid of `.meltano\run\bin` which sounds like a no go as it's needed for this and airflow (maybe others?)

BuzzCutNorman:
@visch that is wonderful to hear this PR is working great for you as well.  I know it stinks that they require administrator rights to create symlinks in Windows.  I looked into the nfs junctions.  Junction can only point to directories. (hard links are the same style but point to files).  Since the code is expecting the destination to be a file a junction would cause an error.  The hard links which I think you create with os.link() point to files and might work.

- https://docs.microsoft.com/en-us/windows/win32/fileio/hard-links-and-junctions
- https://docs.microsoft.com/en-us/windows/win32/fileio/creating-symbolic-links
- https://docs.python.org/3/library/os.html#os.link
- https://docs.python.org/3/library/os.html#os.symlink

BuzzCutNorman:
@visch just tried to use a hard link (os.link()) and it works if you are an administrator 😄  but it doesn't if you are just a user. 😿

BuzzCutNorman:
I might have been wrong.  😸 In the poetry virtual environment for the user account there is no `meltano.exe` present.  I am going to try and have it create a link to the `meltano.cmd` file that is present.  If a hard link to it is created, then yay we are a little closer to not needing to run as an administrator.

BuzzCutNorman:
@visch Yay!!! it did create a link to the `meltano.cmd` file 🥳 .  I committed the change and just watched the automated GitHub tests for Windows since they are setup to create hard links across drives.  That is a limitation of hard links, you have to stay on the same drive.  😞 .  I can see people keeping python installed on one drive and the data projects on another so I don't think hard links are going to be our answer.

BuzzCutNorman:
@visch  all the tests pass now.  This version uses the `symlink_to()` but I put in place a test to see if the user is an administrator.  Please test this version out when you get a chance.

src/meltano/core/project.py
```diff
@@ -123,13 +123,33 @@ def activate(cls, project: Project):
...
         project.ensure_compatible()
 
         # create a symlink to our current binary
         try:
-            executable = Path(os.path.dirname(sys.executable), "meltano")
-            if executable.is_file():
-                project.run_dir().joinpath("bin").symlink_to(executable)
+            # check if running on Windows
+            if os.name == "nt":
+                executable = Path(os.path.dirname(sys.executable), "meltano.exe")
```
WillDaSilva, line 134:
```suggestion
                executable = Path(sys.executable).parent / "meltano.exe"
```

src/meltano/core/project.py
```diff
@@ -123,13 +123,33 @@ def activate(cls, project: Project):
...
+                # See if the user is part of the at leas the local administrator group
+                is_admin = ctypes.windll.shell32.IsUserAnAdmin() != 0
+                # Only users who have amdin privilages are able to create symlinks
+                if is_admin:
```
WillDaSilva, line 138:
```suggestion
                # Admin privileges are required to create symlinks on Windows
                if ctypes.windll.shell32.IsUserAnAdmin() != 0:
```

src/meltano/core/project.py
```diff
@@ -123,13 +123,33 @@ def activate(cls, project: Project):
...
+                    if executable.is_file():
+                        project.run_dir().joinpath("bin").symlink_to(executable)
+                    else:
+                        logger.debug(
+                            f"Could not create symlink: meltano.exe not present in:\n{str(Path(os.path.dirname(sys.executable)))}"
```
WillDaSilva, line 143:
```suggestion
                            "Could not create symlink: meltano.exe not "
                            f"present in {str(Path(sys.executable).parent)!r}"
```

src/meltano/core/project.py
```diff
@@ -123,13 +123,33 @@ def activate(cls, project: Project):
...
+                        )
+                else:
+                    logger.debug(
+                        "Could not create symlink: The creation of a symlink requires administrator previlages"
+                    )
+            else:
+                executable = Path(os.path.dirname(sys.executable), "meltano")
```
WillDaSilva, line 150:
```suggestion
                executable = Path(sys.executable).parent / "meltano"
```

src/meltano/core/project.py
```diff
@@ -123,13 +123,33 @@ def activate(cls, project: Project):
...
+                if is_admin:
+                    if executable.is_file():
+                        project.run_dir().joinpath("bin").symlink_to(executable)
+                    else:
+                        logger.debug(
+                            f"Could not create symlink: meltano.exe not present in:\n{str(Path(os.path.dirname(sys.executable)))}"
+                        )
+                else:
+                    logger.debug(
+                        "Could not create symlink: The creation of a symlink requires administrator previlages"
```
WillDaSilva, line 147:
```suggestion
                        "Failed to create symlink to 'meltano.exe': "
                        "administrator privilege required"
```

WillDaSilva (approved):
Save for a few style changes and typo corrections, this looks good to go. Thanks @BuzzCutNorman!

src/meltano/core/project.py
```diff
@@ -123,13 +123,33 @@ def activate(cls, project: Project):
...
         # create a symlink to our current binary
         try:
-            executable = Path(os.path.dirname(sys.executable), "meltano")
-            if executable.is_file():
-                project.run_dir().joinpath("bin").symlink_to(executable)
+            # check if running on Windows
+            if os.name == "nt":
+                executable = Path(sys.executable).parent / "meltano.exe"
+                # Admin privileges are required to create symlinks on Windows
+                if ctypes.windll.shell32.IsUserAnAdmin() != 0:
```
WillDaSilva, line 136:
```suggestion
                if ctypes.windll.shell32.IsUserAnAdmin():
```

BuzzCutNorman:
@WillDaSilva Thank You, for clean up my code.  Looks much better now.  I was testing it out and noticed that when I use `--log-level debug` I am not seeing the messages about the file not being found or that you don't have admin privileges.  I used `logger.debug` is there a different logger I should be using now?

WillDaSilva:
> I was testing it out and noticed that when I use --log-level debug I am not seeing the messages about the file not being found or that you don't have admin privileges. I used logger.debug is there a different logger I should be using now?

@BuzzCutNorman That's strange - I don't know why that would be. I'll see if I can reproduce this later today.

WillDaSilva:
@BuzzCutNorman I have reproduced the problem you mentioned, where `logger.debug` does not result in any output after being called, despite the log level being set to `debug`. Nice catch by the way. It does emit a log line if `logger.info` or a higher-level log function is called instead.

After some digging, it seems like the issue is that the the log config is only applied after the project has been activated, and so when `logger.debug` is called it will always be using the default log config. This seems tricky to work around, so I propose for this PR we simply promote the `logger.debug` calls to `logger.warn`, since that seems appropriate for these two messages imo.

This is probably an issue for other `logger.debug` calls that are made before the log config is applied, so I'll open an issue to address that in another PR.

src/meltano/core/project.py
```diff
@@ -123,13 +123,33 @@ def activate(cls, project: Project):
...
-                project.run_dir().joinpath("bin").symlink_to(executable)
+            # check if running on Windows
+            if os.name == "nt":
+                executable = Path(sys.executable).parent / "meltano.exe"
+                # Admin privileges are required to create symlinks on Windows
+                if ctypes.windll.shell32.IsUserAnAdmin():
+                    if executable.is_file():
+                        project.run_dir().joinpath("bin").symlink_to(executable)
+                    else:
+                        logger.debug(
```
WillDaSilva, line 140:
```suggestion
                        logger.warn(
```

src/meltano/core/project.py
```diff
@@ -123,13 +123,33 @@ def activate(cls, project: Project):
...
+                            "Could not create symlink: meltano.exe not "
+                            f"present in {str(Path(sys.executable).parent)!r}"
+                        )
+                else:
+                    logger.debug(
```
WillDaSilva, line 145:
```suggestion
                    logger.warn(
```

BuzzCutNorman:
@WillDaSilva Thank you for taking the time to solving that mystery.  I changed the messages to use `logger.warn` as suggested.

PR #6481, opened 2022-07-26
chore: Address all non-doc lint issues

### LINKED ISSUE #6464: Resolve all non-doc lint issues
As discussed in a recent call, editing an old file only to have it spring dozens of linter issues on you because we currently only lint changed files is always a nasty surprise, particularly for an external or first-time contributor.

While resolving all of the doc/docstring lint issues will be tricky (since it requires good knowledge of every piece of the codebase that lacks docstrings), and is potentially not worth it at this time, resolving all of the other ones should be an easy win.

CC @aaronsteers

WillDaSilva:
Closes #6464


WillDaSilva:
[The failing job](https://github.com/meltano/meltano/runs/7533100888?check_suite_focus=true) is unsurprisingly because of doc-related lint issues. They'll still be a pain for us (and our community contributors) for the foreseeable future, but this PR should reduce both the number and severity of people getting hit by lint issues unrelated to their changes.

With this PR merged in we'll be a bit closer to having no lint issues on `main`, and so a bit closer to being able to use https://pre-commit.ci/ instead of our custom pre-commit workflow. Note that we don't currently use https://pre-commit.ci/ because it provides no way to only lint changed files, and so its check would fail for every PR.

pandemicsyn:
I've got mixed feelings about the doc string changes (e.g. `args` to `parameters`). I guess that's coming from a flake8 check we where skipping in the old ci setup? Is that coming from darqlint ? I'm a little confused because it mentions wanting google style, numpy style, or sphinx style doc strings but the new pattern doesn't seem to conform to any of those three. 

We explicitly picked the google style guide for docs when we settled on a doc string convention last year. I'm not certain but I feel like we already had lint checks covering that on the Gitlab side because we actually did a big batch of manual doc string addition's at one point.

Is that something we actually want to change? /cc @edgarrmondragon @aaronsteers

edgarrmondragon:
> I've got mixed feelings about the doc string changes (e.g. `args` to `parameters`). I guess that's coming from a flake8 check we where skipping in the old ci setup? Is that coming from darqlint ? I'm a little confused because it mentions wanting google style, numpy style, or sphinx style doc strings but the new pattern doesn't seem to conform to any of those three.
> 
> We explicitly picked the google style guide for docs when we settled on a doc string convention last year. I'm not certain but I feel like we already had lint checks covering that on the Gitlab side because we actually did a big batch of manual doc string addition's at one point.

@pandemicsyn @WillDaSilva I agree with Florian and we indeed had those checks in GitLab CI.

I'm also not too comfortable reviewing and merging such a large changeset. What do you guys think about splitting the work in PRs either 1) per module or 2) per styleguide violation (e.g. WPS115).

I remember we agreed tackling typing with strategy 1, so it might make sense to agree on a similar approach.

pandemicsyn:
> I'm also not too comfortable reviewing and merging such a large changeset. What do you guys think about splitting the work in PRs either 1) per module or 2) per styleguide violation (e.g. WPS115).

👍 yea I agree, a lot of these changes appeared like doc string changes - so wasn't sure what the line count would look like with those removed. It definitely makes it hard to pick out manual fixes that could be more prone to errors and warrant some scrutiny. 

Not quite as concerned about changes like that in meltano core where we have pretty good coverage, but the API's testing is iffy so would be easy for small unintentional bugs to slip through.

WillDaSilva:
@edgarrmondragon I'll close this PR and split the changes into 1 PR per module (or other logical group for sparse changes, so as to avoid making 100+ PRs).

@pandemicsyn The `Parameters:` section is valid for Google docstrings, and as explained in the commit where it was done, it is more accurate. It is also the key used by Mkdocstrings to populate the table of parameters (see: https://github.com/meltano/meltano/issues/5948). I'll pull it out into a separate PR.

PR #6487, opened 2022-07-27
chore: Remove unused Dockerfiles

WillDaSilva:


cjohnhanson (approved):
Looks good to me.

Note for perpetuity: the changes are safe because `docker/base` and `docker/main` were previously used in our now-defunct `Makefile` and are now replaced by the single `docker/meltano` for building and publishing to Dockerhub. `docker/docs` is no longer necessary because docs are built and published via `netlify.toml`.

PR #6484, opened 2022-07-26
build: Use `sass` instead of the deprecated `node-sass` library

WillDaSilva:
https://sass-lang.com/blog/libsass-is-deprecated

Closes #6483

PR #6488, opened 2022-07-27
chore: Remove Makefile

WillDaSilva:
Several of the commands in the Makefile do not work anymore, are not necessary, or should not be run locally. The Makefile has a considerable amount of dead code within it, and presents a small but significant burden to contributors. It all adds up: the Makefile does not prevent contributors from having to learn various poetry and yarn invocations to build/test/etc. our software - it just adds more on top of that.

From asking around, and searching through the repository, it seems like there are only 2 commonly used `make` commands: `ui` and `bundle`. They have been replaced by `scripts/bundle_ui.sh`, which can be run from anywhere to achieve the same effect.

Besides that, all of the local tasks that used to be carried out through the Makefile should now instead be done through poetry, yarn, and optionally pre-commit.

Much of what was in the Makefile was for CI, which is now all handled in the workflows and actions under `.github/`. In the unlikely circumstance that we need to run those commands locally, due to the simplicity of how we have our CI set up, we can copy-paste the commands from the workflows/actions manually.

Relates to:
- #6262
- #6487


WillDaSilva:
I intend for the changes in this PR to function as a scream test. If none of us have a good reason to keep the Makefile around, then we should remove it on `main` (by merging this PR) and see if that causes any problems. If not, then we continue on our merry way. Otherwise, we revert this change, and possibly revisit the topic with the knowledge we'll have gained from the problems that arose.

aaronsteers:
@WillDaSilva - Thanks for your work to proactively invest in cleaner and leaner build processes. 

What the process doesn't really give us in this iteration is what actions we'd take if something went wrong or if a file gets committed to the repo which shouldn't have.

Also, for local UI dev and testing it seems likely that we'll accidentally commit files back to the repo which should not have been committed. (Or perhaps they are already git ignored?)

At least to keep functional parity with the make script, can we add something like a `clean.sh` and some readme or contributor doc with a few lines on what does what? One approach is to just document the steps a maintainer would run if the process has to be executed outside of CI, including the cleanup afterwards.

cc @alexmarple and @rwfeather who've been doing the UI dev as of recently and who might have input into the build and cleanup steps. 

Also - the makefile previously insulated us somewhat against having to invoke yarn directly and I'm not sure if that changes with this update and we'd need to become more familiar with yarn to perform general dev/test functions.

pandemicsyn:
@WillDaSilva Double checked the docs - the ui/api dev guides don't seem to mention the make files at all (🥳), I think the `make lint` reference on https://docs.meltano.com/contribute/style **might** be the only spot. 

> the makefile previously insulated us somewhat against having to invoke yarn directly and I'm not sure if that changes with this update and if we'd need to become more familiar with any specific commands in yarn to perform general dev/test/build/cleanup functions.

@aaronsteers That's probably mostly covered by https://docs.meltano.com/contribute/ui or https://github.com/meltano/meltano/blob/main/src/webapp/README.md already, might just be worth having quick skim of those and see what's missing.

WillDaSilva:
We can probably actually get rid of `scripts/bundle_ui.py`, and just call `build.py` instead, which does the same thing, but is automatically run by poetry when you build a wheel. Anyone who wishes to build & bundle the UI manually can simply manually call `build.py`. Plus now we can have it as a Python file that just about everyone here will understand, rather than a bash script that might be intimidating to some.

> I think the make lint reference on https://docs.meltano.com/contribute/style might be the only spot.

Thanks for catching that @pandemicsyn! I'll update it.

> the makefile previously insulated us somewhat against having to invoke yarn directly and I'm not sure if that changes with this update and if we'd need to become more familiar with any specific commands in yarn to perform general dev/test/build/cleanup functions.

@aaronsteers In theory this was true, but in practice it seems like everyone was just calling poetry and yarn anyway.

> What the process doesn't really give us in this iteration is what actions we'd take if something went wrong or if a file gets committed to the repo which shouldn't have.

I'm not sure what the concern is here.

> Also, for local UI dev and testing it seems likely that we'll accidentally commit files back to the repo which should not have been committed. (Or perhaps they are already git ignored?)

The build artifacts are all already git ignored, so this shouldn't be an issue.

> At least to keep functional parity with the make script and to address the above, can we add something like a `clean.sh` [...]

I'll add it as a Python script to provide Windows compatibility, and to have it be in a familiar language for our contributors. Overall I want to minimize the number of scripts we have, but this one seems self-explanatory and useful.

> [can we add] some readme or contributor doc with a few lines on what does what [...]

The docs already generally say what to do using poetry and yarn, rather than the Makefile, but I can do another pass through and see what can be improved on that front.

> [...] especially which steps are needed to build a release and/or building for local tests? One approach is to just document the 
steps a maintainer would run if the process has to be executed outside of CI, including the cleanup afterwards.

I believe having such a document is dangerous. When it comes to something as important and precise as releasing to PyPI, it should only be done in CI, and there should be a single source of truth about what to do, which is the code that runs in CI. If something goes horribly wrong and we need to do a release outside of CI, the steps to take can be summarized as "do exactly what CI does with as few deviations as possible".

tayloramurphy (approved):
Docs approval

WillDaSilva:
Blocked by https://github.com/meltano/meltano/pull/6484, because we now run `yarn && yarn build` when we build the wheel.

afolson (approved):

pyproject.toml
```diff
@@ -14,13 +14,14 @@ classifiers = [
 include = [
   # Explicitly include non-Python files and gitignored files:
   "LICENSE",
-  "*.m5o",
+  "scripts/build.py",
```
aaronsteers, line 17: @WillDaSilva - Do we need this to be bundled/included into the distribution in order for the build to work? If not, can we exclude without impact to the users?

WillDaSilva: It is included by default now that it resides in the root of the project. It must reside in the root of the project due to a Poetry bug.

.github/workflows/build.yml
```diff
@@ -51,12 +51,13 @@ jobs:
...
-      run: |
-        touch src/meltano/core/tracking/.release_marker
+      # The release marker differentiates installations 'in the wild' versus internal dev builds and tests
+      run: touch src/meltano/core/tracking/.release_marker
 
     - name: Build distribution
       run: |
-        make sdist_public
+        poetry run scripts/alembic_freeze.py
+        poetry build
```
aaronsteers, line 60: @WillDaSilva - In 

- #6145

@edgarrmondragon and I found that the `poetry build` process worked for the wheel but did not properly execute `build.py` when running for `sdist`. Discussion thread here: https://github.com/meltano/meltano/pull/6145/files#r897501873

Can we simply run `build.py` manually as in the suggestion below in order to mitigate that discovered issue?

```suggestion
        poetry run scripts/alembic_freeze.py
        poetry run scripts/build.py  # needed for sdist due to https://github.com/python-poetry/poetry/issues/2740#issuecomment-860744196
        poetry build
```

@edgarrmondragon - Any thoughts? I'd rather not remove `sdist` distributions entirely as part of this particular PR, but that's something we also discussed as a future option.

aaronsteers (changes requested):

.github/workflows/build.yml
WillDaSilva: The `sdist` ought to contain the unbuilt source of the repository. It comes with `build.py` within it, and when it is installed, `build.py` is run, as the first step of installing an `sdist` is building a wheel out of it.

If we can accept that anyone who is installing an `sdist` should have `yarn` installed so that they can build the Meltano UI, then the fix here will be to ensure the `sdist` contains the unbuilt Javascript code in `src/webapp` in addition to `src/meltano`.

aaronsteers: Want to add I really like these updates:

> We can probably actually get rid of scripts/bundle_ui.py, and just call build.py instead, which does the same thing, but is automatically run by poetry when you build a wheel. Anyone who wishes to build & bundle the UI manually can simply manually call build.py. Plus now we can have it as a Python file that just about everyone here will understand, rather than a bash script that might be intimidating to some.

The rename to `build.py` addresses some of my concerns around docs, since this is self-documenting according to conventions many would reasonably expect. A paired `clean.py` also would serve a self-documenting function, in that it would remove those files that should not remain in the tree, and be self-documenting of those files without expecting too much inference from developers/contributors.

edgarrmondragon: > If we can accept that anyone who is installing an `sdist` should have `yarn` installed so that they can build the Meltano UI, then the fix here will be to ensure the `sdist` contains the unbuilt Javascript code in `src/webapp` in addition to `src/meltano`.

@WillDaSilva Problem is we've been shipping the `sdist` with the built UI assets, so not including them would be a breaking change for anyone relying on the `sdist`.

WillDaSilva: @edgarrmondragon True, but is it reasonable for anyone to specifically depend on the latest version of an `sdist` not requiring anything special to build it? It's common for `sdist`s to have many system level dependencies (often on C libraries).

The only 3 categories of users I know who depend specifically on an `sdist` are:
- Those who want to build it themselves (e.g. to build it into a Conda package instead of a wheel).
- Those who want to audit it or its build.
- People who haven't installed the `wheel` package.

The first two categories of users who depend on an `sdist` are generally fine dealing with the difficulties around building package with system-level dependencies. The third is rare, and ought to install the `wheel` package.

edgarrmondragon: @WillDaSilva 

> but is it reasonable for anyone to specifically depend on the latest version of an `sdist` not requiring anything special to build it? It's common for `sdist`s to have many system level dependencies (often on C libraries).

Certainly not reasonable, and you're right in that it's how numpy and the like do things.

> The only 3 categories...

Since we publish a "pure Python" wheel, no user installing from PyPI should ever be forced to install from `sdist` when using pip on a niche platform, right?

> Another option to consider is that we could containerize the JS build process. That would introduce a dependency on Docker for any who wish to build the `sdist`, but that's a much simpler dependency (and less likely to change over time) than requiring specific versions of Node, Yarn, and the various C libraries they depend on.

I'm not a fan of that. I think documenting the required versions of node, yarn, etc. is reasonable enough and AFAIK can even be explicit requirements set in `package.json`.

cc @aaronsteers I think there's a good case for publishing a _raw_ `sdist` without built UI assets, and that would simplify our workflows a bit (see https://github.com/meltano/meltano/issues/6566#issuecomment-1205649134)

WillDaSilva: > Since we publish a "pure Python" wheel, no user installing from PyPI should ever be forced to install from sdist when using pip on a niche platform, right?

@edgarrmondragon Correct, so long as they have `wheel` installed, which is generally true. We can ensure it by adding `wheel` to our `build-system` in `pyproject.toml`, which is currently:

```toml
[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

> I think documenting the required versions of node, yarn, etc. is reasonable enough and AFAIK can even be explicit requirements set in package.json.

I'm okay with that, but it's worth noting that this requires the user either change their globally installed version of Node/Yarn, learn how to use a tool like NVM, or create a temporary environment (e.g. within an interactive Docker container or VM) in which to build Meltano. The requirement "have Docker/Podman installed" is much more simple than "have version X of Node and version Y of Yarn installed".

WillDaSilva:
@edgarrmondragon @aaronsteers When Poetry uses `build.py`, it builds a platform-specific wheel, rather than a pure wheel.

The offending code: https://github.com/python-poetry/poetry-core/blob/d5db91af41e0befd413ce54190bfb7fb7fed10f1/src/poetry/core/masonry/builders/wheel.py#L390

Related issues:
- https://github.com/python-poetry/poetry/issues/3594
- https://github.com/python-poetry/poetry/issues/2051

We can't use `build.py` until this has been resolved. I'm going to ask the Poetry core maintainers on Discord about these issues, and see if I can get things moving along. Until we can use `build.py` with Poetry to build a pure wheel, this PR will be a draft.

afolson (approved):

docs/src/_contribute/style.md
```diff
@@ -20,16 +20,17 @@ Python:
...
 - [MyPy](https://mypy.readthedocs.io/en/stable/)
 
-Flake8 is a python tool that glues together `pycodestyle`, `pyflakes`, `mccabe`, and third-party plugins to check the style and quality of python code,
-and `wemake-python-styleguide` is a plugin for Flake8 that offers an extensive set of opinionated rules that encourage clean and correct code.
+Flake8 is a python tool that glues together `pycodestyle`, `pyflakes`, `mccabe`, and third-party plugins to check the style and quality of python code. Notable among these is `wemake-python-styleguide`, which offers an extensive set of opinionated rules that encourage clean and correct code.
+
+To lint your Python code, run `poetry run flake8` from the root of the project. If you only want to lint the files that have changed (and have been staged in git), you can run `pre-commit`. The `pre-commit` check will be run in CI on all PRs. Some lint issues, particularly around documentation, are pre-existing. Over time we are working to address these issues. If your PR is failing the `pre-commit` check because of pre-existing lint issues, feel free to leave them unaddressed.
```
WillDaSilva, line 25:
```suggestion
To lint your Python code, install the project using `poetry install`, then run `poetry run pre-commit --all-files flakeheaven` from the root of the project. The `pre-commit` check will be run in CI on all PRs.
```

afolson (approved):

edgarrmondragon (approved):

aaronsteers (approved):
@WillDaSilva - Thanks very much for tackling this and for pushing for increased clarity and streamlined footprint of legacy code.

Today's discussion was much appreciated. I've logged a new Discussion thread here:

- https://github.com/meltano/meltano/discussions/7153

And having had those discussions, I am happy to approve this to merge as-is.

Thanks again!

PR #6506, opened 2022-07-28
chore: noop re: docstrings

WillDaSilva:
Note [AJ]: I renamed from the prior name so that the commit history will reflect as a noop. Prior title was: `"Args:" -> "Parameters:" in docstrings`

It is more accurate to call them parameters, since that is what is being documented. The arguments are the objects themsleves that are being provided to "fill" the parameters.

Relates to https://github.com/meltano/meltano/pull/6481#issuecomment-1196936867


pandemicsyn:
Happy to be overruled on this one by the majority, but I'm not a huge fan of this. I understand this is a technically correct change since these are parameters and not args.  

However, as far as I could find, this isn't documented in the Google Python Style Guide https://google.github.io/styleguide/pyguide.html#doc-function-args or in the alternate google style guide referenced by the utility mentioned in the O.G. pull request ( https://mkdocstrings.github.io/griffe/docstrings/ -> https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html )

We opt'd to follow the google python style guide for docs in part because its a well known standard/convention that folks are familiar with. If folks preferred we use this pattern, my only ask would be that we update our developer docs so that we have a consistent document to point contributors too. 

/cc @edgarrmondragon @aaronsteers

WillDaSilva:
> However, as far as I could find, this isn't documented in the Google Python Style Guide https://google.github.io/styleguide/pyguide.html#doc-function-args or in the alternate google style guide referenced by the utility mentioned in the O.G. pull request ( https://mkdocstrings.github.io/griffe/docstrings/ -> https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html )

Worth noting that it's used throughout https://mkdocstrings.github.io/griffe/docstrings/, and the example given at https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html is out of date, as it places the type annotations in the docstring instead of in the function signatures.

aaronsteers:
@WillDaSilva - I can definitely see the argument to rename. The Google docs themselves describe the 'Args:' section as listing "each parameter by name".: https://google.github.io/styleguide/pyguide.html#doc-function-args

That said though, I think we have to bias towards consistency and towards IDE and linting support. My understanding is that the linters that do parse and check docstrings using the google docstring style may get hung up on the different header.

- https://gitlab.com/meltano/sdk/-/issues/146

The longer conversations we had on this topic over on the SDK was duplicated for Meltano but didn't get carried into the handbook - very sorry about that. We did evaluate the numpy doc style but declined to use that style approach due to its own limitations and issues there.


Logged this to get something into the handbook:

- https://github.com/meltano/handbook/issues/312

WillDaSilva:
> That said though, I think we have to bias towards consistency and towards 3rd party IDE and linting support. My understanding is that the linters that do parse and check docstrings using the google docstring style may get hung up on the different header.

@aaronsteers Meltano already has a mix of `Args:` and `Parameters:` for the docstring section titles - has anyone run into any issues with this?

We know that `flake8` and its plugins support using `Parameters` as the section header, and the error they raise when it's missing calls for the addition of "parameters" to the docstring (e.g. `DAR101`). I don't believe we want to support any other linters, so even if other linters did take issue with this (and I don't believe that any do), it wouldn't be a problem for us.

As for third-party tools for docstrings, our primary concern should be how they are rendered in the docs (which we are not currently doing at all). That's where `mkdocstrings` would come in, which recommends the use of `Parameters:` over `Args:`. Alternatively we could use `sphinx-autodoc`, though I would recommend against it since that would force us to use restructuredtext instead of markdown. See:
- https://github.com/meltano/meltano/issues/5948

aaronsteers:
> @aaronsteers Meltano already has a mix of `Args:` and `Parameters:` for the docstring section titles - has anyone run into any issues with this?

Meltano has tons of legacy code still and we've only been doing progressive incremental cleanup, so I don't think we have as stringent of requirements or very clear expectations. Would be worth checking over on the SDK repo though - that repo is already doing auto-docs generation.

> As for third-party tools for docstrings, our primary concern should be how they are rendered in the docs (which we are not currently doing at all). That's where `mkdocstrings` would come in, which recommends the use of Parameters: over Args:. Alternatively we could use `sphinx-autodoc`, though I would recommend against it since that would force us to use restructuredtext instead of markdown. 

We are using Sphinx on the SDK. We started out with restructuredtext (which was a pain) and I believe @edgarrmondragon more recently added Markdown support via another plugin.

@edgarrmondragon and @WillDaSilva - Do one of you want to open a new issue (or GitHubn discussion) to discuss the best path forward here? I'd like to keep the same standards and conventions across both repos (and ideally any other Python repos we own as well).

edgarrmondragon:
@WillDaSilva

> Happy to be overruled on this one by the majority, but I'm not a huge fan of this. I understand this is a technically correct change since these are parameters and not args.

I agree with @pandemicsyn here.

> Meltano has tons of legacy code still and we've only been doing progressive incremental cleanup, so I don't think we have as stringent of requirements or very clear expectations. Would be worth checking over on the SDK repo though - that repo is already doing auto-docs generation.

@aaronsteers is correct, and I think we should standardize to the lowest common denominator and  just use `Args` for both. Will say more on https://github.com/meltano/handbook/issues/312

WillDaSilva:
@aaronsteers Per our discussion on Zoom, I have updated this PR to be a no-op.

aaronsteers (approved):
Approving what is now a 'no op' per discussion. This is with the goal of being pragmatic about the other PRs, at the cost of an extra commit on `main`.

PR #6509, opened 2022-07-28
chore: Remove unused file `src/meltano/models.py`

WillDaSilva:
Relates to #6464


pandemicsyn (commented):
Is this actually a file that safe to remove, or did something get lost in the refactor from Job -> State?

pandemicsyn:
> Is this actually a file that safe to remove, or did something get lost in the refactor from Job -> State?

Does seem ancient (nothing since the initial commit in 2019) and unused as far as I can tell.

pandemicsyn (approved):

edgarrmondragon (approved):

PR #6508, opened 2022-07-28
chore: Address all non-doc lint issues under `tests/`

WillDaSilva:
The `CliRunner` invocation `results.output` only has `stdout`, but our logging goes to `stderr` (usually). In another PR to address lint issues, I updated a file to use `log.info` instead of `print`. This resulted in test failures because the expected output was no longer part of `results.output`. For consistency I updated the other places where we were using `results.output`. 

Going forward we should never use `results.output`, and instead favor `results.stdout` or `results.stderr`, since that's unambiguous.

Relates to #6464

PR #6519, opened 2022-07-28
chore: Address all non-doc lint issues in `src/meltano/core/plugin/*.py`

WillDaSilva:
Relates to #6464

The tests for this PR will fail until #6508 has been merged, since it includes the fix to check stderr for the log output, rather than just stdout.


pandemicsyn (approved):

src/meltano/core/plugin/file.py
```diff
@@ -1,3 +1,5 @@
+import structlog
```
edgarrmondragon, line 1: @WillDaSilva I may have missed in other PRs, but I notice `from __future__ import annotations` is not added to all files. Does that mean it's being added manually?

If so, I think we can fix that by adding the import as part of an `isort` fix

```toml
[tool.isort]
add_imports = [
    "from __future__ import annotations",
]
```

Wdyt?

WillDaSilva: That's a cool `isort` feature. It'll prevent us from accidentally re-introducing the old annotation style, and eventually (many years from now) once `from __future__ import annotations` is not needed for any Python version we support, `pyupgrade` will remove it from all of our files automatically.

I've added that line to the `isort` config in:
- https://github.com/meltano/meltano/pull/6546

PR #6523, opened 2022-07-28
chore: Remove unused code & fix lint issues in `src/meltano/core/error.py`

WillDaSilva:
The tests will fail because some elements from this file are imported (but unused) elsewhere. Those imports have been removed in other PRs related to #6464.

Relates to #6464


pandemicsyn (approved):

pandemicsyn:
fyi - i reviewed/approved the cli changes so that should let the integration test go green.

src/meltano/core/error.py
```diff
@@ -83,102 +53,8 @@ async def stderr(self) -> Optional[str]:
...
-        def __init__(self, exceptions: []):
-            self.exceptions = exceptions
-
-        def __str__(self):
-            return "\n".join(str(e) for e in self.exceptions)
-
-    if error_cls != Exception:
-        error_cls.Aggregate = Aggregate
-
-    return error_cls
```
edgarrmondragon, line 110: First time seeing taking a look at this code. Interesting that Meltano devs were trying what's roughly [PEP 654](https://peps.python.org/pep-0654/).

edgarrmondragon (approved):

src/meltano/core/error.py
WillDaSilva: It's a good idea, but difficult to do without library/language support. Even with library/language support, it's difficult to go from the current exception model to this new one.

PR #6530, opened 2022-07-29
fix(core): Update env var tests to be in line with correct reading of spec.

### LINKED ISSUE #5982: Update terminal environment variable precedence to be in line with currently xfail-ing tests

cjohnhanson:
These xfail-ing tests expected _any_ environment variables set at "higher-precedence" config layers such as the plugin-level or active environment-level plugin-level env to override values set at "lower-precedence" config layers such as top-level env. These tests have been changed to reflect the correct behavior, which is for environment variables that are mapped to plugin settings to be parsed as settings and mapped to environment variables in the plugin's runtime environment, basically meaning that terminal-level or active environmnet level values will override plugin-level values, since settings mapped to env vars have the highest precedence order.

This PR also removes the `xfail=True` option from the relevant tests. 

See thread in #5982 for specifics.

Closes #5982


tests/meltano/core/test_environment_variables.py
```diff
@@ -166,72 +166,54 @@ def _meltanofile_update_dict(
     ),
     # Original xfail'ing tests, as per comment above
```
edgarrmondragon, line 167: @cjohnhanson should this comment and the one above be removed?

cjohnhanson: Yep, along with one other comment. I also removed the `xfail` attribute from these test cases since it's no longer  used.

edgarrmondragon (approved):
LGTM!

(there's a failing look but it seems to be transient)

WillDaSilva (approved):
Happy to see all these xfailing tests be cleaned up!

pandemicsyn (approved):
🥳

PR #6533, opened 2022-07-29
fix(core): Timeout Snowplow requests

### LINKED ISSUE #6532: [Bug]: Snowplow tracker does not timeout when endpoint is unavailable
### Meltano Version

2.3.0

### Python Version

NA

### Bug scope

Other

### Operating System

NA

### Description

First reported by @dingobar here: https://meltano.slack.com/archives/C01TCRBBJD7/p1659099832402809

> Hi all, thanks for a great product! Everything is working great locally, and we've now deployed our meltano project on kubernetes where we already had airflow running. However, there is a hiccup: The job hangs after Environment 'dev' is active (see screenshot). 
> image
> 
> After about 15 minutes or so, the actual task completes, but the container keeps hanging telling me that the meltano run process has not yet exited and is blocked somehow. To me it seems like it is either stuck in an infinite loop, or trying to call something over the network with (seemingly) infinite retries/timeout.
> 
> We have a strict firewall policy where we explicitly name the domains to be whitelisted for egress out from our environment (required as we are in the telco business with very strict data protection requirements). My theory is that meltano is trying to call out to some external site. I have disabled telemetry in meltano in general (MELTANO_SEND_ANONYMOUS_USAGE_STATS="False" as stated in the docs).
> 
> As evident in the logs, it hangs for about 6 minutes before getting started:
> image
> 
> Update: So the culprit is found. It is trying to contact "[sp.meltano.com](http://sp.meltano.com/)" with an exponential retry backoff. How can I disable this behavior?
> image

We should:
- Change the default `request_timeout` value for `meltano.core.tracking.Tracker` from `None` to `3.5`.
- Add a test that simulates the endpoint not responding.

### Code

_No response_

WillDaSilva:
Closes #6532

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,7 +102,10 @@ src_paths = ["src/", "tests/"]
 skip_glob = ["*.md", "*.vue", "*.js"]
 
 [tool.coverage.run]
-concurrency = [ "multiprocessing" ]
+concurrency = [
+  "multiprocessing",
+  "thread",
+]
 parallel = true
 branch = true
 
--- a/src/meltano/core/tracking/tracker.py
+++ b/src/meltano/core/tracking/tracker.py
@@ -85,13 +85,16 @@ class Tracker:  # noqa: WPS214 - too many methods 16 > 15
     def __init__(
         self,
         project: Project,
-        request_timeout: float | tuple[float, float] | None = None,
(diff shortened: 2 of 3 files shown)


edgarrmondragon (approved):
🚀

pandemicsyn (approved):

PR #6534, opened 2022-07-29
fix: Lock plugins using the latest Hub definition if plugin has no variant name and no lock file exists

### LINKED ISSUE #6360: Meltano lock fails on migrating to >2.0.0 when variant didnt previously exist
We should update the log message and add a section to the migration docs to help users get variants added for plugins that didnt have a variant previously. For example Airflow locking fails because a variant isnt set for all installations prior to 2.0.0. The error message isnt totally clear in saying this and users will need to manually update their meltano.yml.

Originally discussed in https://github.com/meltano/meltano/issues/6359#issuecomment-1175432513

> Additionally I'm getting Orchestrator 'airflow' variant 'original' is not known to Meltano. Variants: ['apache (default)'] from a default installation of Airflow. I know that its hard because when I originally installed it there wasnt a variant name for airflow but now there is. Is there a way for us to resolve that its the default variant? Maybe that wont work because the default could change, maybe using the discovery.yml when we think were migrating to >2.0.0?
> 
> Or we could log it better and say something like "I notice youre migrating and you dont have a variant, go to the Migration guide and learn how to define a variant" and we write up the steps for locking post migration and adding a variant name:
> 
> remove executable/namespace if defined
> add variant using one available on MeltanoHub. We can list the mapping like airflow -> apache, dbt -> dbt-labs, etc. for plugins that just got variant names
> run meltano lock --all

cc @edgarrmondragon

edgarrmondragon:
Closes #6360

### DIFF
--- a/src/meltano/core/hub/client.py
+++ b/src/meltano/core/hub/client.py
@@ -26,7 +26,7 @@ from .schema import IndexedPlugin, VariantRef
 logger = get_logger(__name__)
 
 
-class HubPluginTypeNotFound(Exception):
+class HubPluginTypeNotFoundError(Exception):
     """Raised when a Hub plugin type is not found."""
 
     def __init__(self, plugin_type: PluginType):
@@ -49,7 +49,7 @@ class HubPluginTypeNotFound(Exception):
         )
 
 
-class HubPluginVariantNotFound(Exception):
+class HubPluginVariantNotFoundError(Exception):
     """Raised when a Hub plugin variant is not found."""
 
     def __init__(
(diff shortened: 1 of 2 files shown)


pandemicsyn (approved):

edgarrmondragon:
@aaronsteers pinging you just to confirm the approach and whether we should still emit some kind of warning when there's no explicit variant in a plugin definition in `meltano.yml`

aaronsteers:
> @aaronsteers pinging you just to confirm the approach and whether we should still emit some kind of warning when there's no explicit variant in a plugin definition in `meltano.yml`

Yes, I think it would be helpful to print something like the below, but not blocking, I think, for this specific PR / fix.

> WARNING: Lock file not found for plugin '<plugin-name>'. Installing latest plugin definition from hub.meltano.com...

PR #6539, opened 2022-08-01
docs: CLI design guidelines 

### LINKED ISSUE #6440: Document CLI design guidelines
Similar to https://docs.meltano.com/contribute/api#api-v2-design-guidelines we should have clearly defined guidelines for creating/updating the meltano `cli`.  How to structure, organize, and even name new CLI additions is a **frequent** source of contention and a stumbling block during iterations and builds.

I think it's worth calling out that it's OK that our CLI's inconsistent right now, and not all `meltano commands` will follow this style guide immediately. However, I think we need a guide in place (even if it's an aspirational one) moving forward so that we can start aiming for a consistent CLI experience for users, and to have a consistent UX style guide in place during feature planning.

This doc should cover:

- Desired verb/command linkage and structure 
- Expected common sub-command's and options
- Expected common output formats
- Use of short and long flags
- Abbreviations
- Global vs argument level flags
- Reusing short flags
- Flag aliases
- Consistent definition of Groups, Commands, Sub-commands, Arguments, and Options
- Expected help and usage
- Doc style guidelines for how to style examples and usage of:
  - Optional arguments/options/flags
  - Required items 
  - Mutually exclusive args
  - Arguments that can repeat
- What belongs the cli module (i.e.`meltano/cli/the_command.py`) and what belongs in meltano.core. 
- Phrasing guidelines for command Deprecation and Preview/Beta commands.

pandemicsyn:
first pass at a CLI design guidelines doc

closes #6440


docs/src/_contribute/cli.md
```diff
@@ -0,0 +1,225 @@
...
+For a technical explanation of how commands and groups work see https://click.palletsprojects.com/en/8.0.x/commands/
+
+### Use of short and long options
+
+The options you expect to be used most frequently should always include both the long AND short variation of the option.
+For example `--help` and `-h`, `--tasks` and `-t`, `--force` and `-f`.
+
+
+### Desired verb/command linkage and structure
+
```
pandemicsyn, line 56: @tayloramurphy @aaronsteers In the past we've had a lot back and forth on how to structure the cli components and it sounds like y'all had differing philosophies on this perhaps. So to get us started, I only briefly described what we did for `job` and what we refactored `schedule` too. 

If this is the structure we want to maintain moving forward I can send along an update with more perscriptive guidance and examples.

docs/src/_contribute/cli.md
```diff
@@ -0,0 +1,225 @@
...
+#### Repeating items
+
+Use three trailing dots and no spaces (`...`) to indicate that the user can specify multiple values for the items:
+
+```
+meltano run BLOCKS ...
+meltano install [] [PLUGIN_NAME] ...
+```
+
+### Phrasing guidelines for command Deprecation and Preview commands
```
pandemicsyn, line 217: @tayloramurphy and specific call outs or considerations for how to handle preview commands or deprecating commands?

docs/src/_contribute/cli.md
tayloramurphy: @pandemicsyn this is a great PR. I love clear guidance docs like this. I'm quite happy with the structure we have for `jobs` and `schedule` and I'm happy to move forward with that. 

The only note I had is to add guidance for when (or when not) to use angle brackets `<>` in the docs and/or help text.

docs/src/_contribute/cli.md
tayloramurphy: @pandemicsyn only thing that comes to mind is recommending that we call out which major version number we expect to promote or deprecate a preview command.

docs/src/_contribute/cli.md
pandemicsyn: @tayloramurphy Yea, i was tempted to explicitly say to never use. Buts its kinda handy and pretty common. The problem is, we use `<stuff>` interchangeably as a placeholder/var substitution for both required AND optional items, so it makes it a bit confusing. What about adding "placeholders" section that clarifies use of angle brackets, as well as when to use generic names like `tap`, `target` vs `tap-gitlab` in examples?

docs/src/_contribute/cli.md
pandemicsyn: 👍 thats a good call out.

docs/src/_contribute/cli.md
tayloramurphy: @pandemicsyn "placeholders" makes sense 👍 I agree we have to have something b/c I'm not sure how we're going to get away from having `<>` in the docs at all 😅

pandemicsyn: 👍 added a section.

tl;dr better for docs than CLI help out. If you use the angle bracket pattern you should still follow the rest of the rules e.g. using `[<MAPPING1> ...]` to indicate whether its required/optional. 

@tayloramurphy also i said to prefer Tap/Target over Extractor/Loader. Is the inverse better?

docs/src/_contribute/cli.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+### Placeholders and angle brackets
+
+When writing help output its often useful to provide examples that utilize place holders. In complex cases where you're trying illustrate how a user might use feature, its often useful to use actual tap/target/mapping names to illustrate intent. But most commonly, especially for simple examples use common terms e.g. `TAP` rather than `extractor` or `tap-something`. 
+
+We also have significant historic use of angle brackets as place holders for user input in our documentation (and also help out). Which often looks like:
+
+```
+meltano elt <tap> <target>
```
tayloramurphy, line 224:
```suggestion
meltano elt <tap_name> <target_name>
```

docs/src/_contribute/cli.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+In the context of something like a more conversational guide or other online docs this does help indicate that we expect the user to supply this input. However, our current use in help output is often ambigous when it comes to indicating whether this is a required item or optional item. So moving forward if you use this form in CLI help out you should still follow the documented conventions for indicating whether something is required:
+
+```
+# TAP and TARGET are upper case since they are required inputs
+meltano run <TAP> <TARGET>
```
tayloramurphy, line 231:
```suggestion
meltano run <TAP_NAME> <TARGET_NAME>
```

Perhaps we should have `<PLUGIN_NAME>` in there as well? I don't know if using `meltano run` is the best example because technically a tap/target pair isn't required.

docs/src/_contribute/cli.md
tayloramurphy: @pandemicsyn tap/target over extractor/loader I think is fine - that's primarily what we're using in the docs currently, though usually scoped to the name (tap_name / target_name). 

Otherwise the addition looks good with the 2 minor comments I had.

docs/src/_contribute/cli.md
```diff
@@ -0,0 +1,247 @@
...
+
+The options you expect to be used most frequently should always include both the long AND short variation of the option.
+For example `--help` and `-h`, `--tasks` and `-t`, `--force` and `-f`.
+
+
+### Desired verb/command linkage and structure
+
+1. Organize in a command group around a feature below the top level where possible.
+2. Coalesce commands that operate on the same resource of a feature as a group of sub-commands.
+3. Use transitive verbs like `delete`, `set`, `list`, `get` as sub-commands to perform work within a group where appropriate.
```
aaronsteers, line 59: I think we are preferring `remove` over `delete` so it might be good to adjust the samples here slightly.

Related: I wonder if it would be helpful to document the common/preferred CRUD-esque verbs: 

- `add` and `remove` to create or remove a construct like "schedules" or "jobs". (Maybe also `set` to override/update after creation, as in `meltano job JOB_NAME set ...`.)
- `get`, `set`, `list`, `clear` for fetching and setting artifacts like "state" and "secrets".
- `list` to enumerate items.
- `describe` to show details about an item.

pandemicsyn: 👍 Swapped in `remove` and add a standardized verbs section.

tayloramurphy (approved):

docs/src/_contribute/cli.md
aaronsteers: Looks great. Thanks!

docs/src/_contribute/cli.md
```diff
@@ -0,0 +1,255 @@
...
+- `set` is a `sub-command` of the `job` command.
+- `JOB_NAME` is an `argument` of the `set` sub-command.
+- `--tasks` is named `option` of the `set` sub-command.
+
+For a technical explanation of how commands and groups work see https://click.palletsprojects.com/en/8.0.x/commands/
+
+### Use of short and long options
+
+The options you expect to be used most frequently should always include both the long AND short variation of the option.
+For example `--help` and `-h`, `--tasks` and `-t`, `--force` and `-f`.
```
aaronsteers, line 52: I might have a less favorable opinion about short options (and also abbreviations) in general, since I find that the benefit in brevity comes at a high cost in readability and having to frequently revisit the docs and/or `--help` output. (Remembering that `-e` means one thing but `-E` means something different, and then `-o` means something whose name _also_ starts with `e` but we had already used `-e` so the short option had to be something completely different.)

I would be happy to remove both of these two topics (short options and abbreviations) from this iteration of the CLI guidelines doc, or else perhaps mitigate using one or more of the below.

1. We can suggest that short options and abbreviations should be discussed and agreed to in the spec ahead of time, prior to development starting. (Prevents stalling a PR later for discussion which otherwise would have been ready to go.)
2. We can suggest that short options and abbreviations be _excluded_ from initial feature releases and instead added after the fact in a follow-on PR. (Similar mitigation as above. We just treat abbreviations as QOL improvements after the fact, and therefor by definition they won't be blocking critical functionality and we won't rush to decisions without proper consideration.)
3. We can just say we don't prefer short options or abbreviations at this time, but that we plan to add them in a considered manner later in the product lifecycle.


Any or all of the above recommendations would prevent us from blocking new features from going out due to internal deliberations of acceptable convenience methods. 

@tayloramurphy and @pandemicsyn - Would it be okay with you both if we take a conservative approach to these for the time being? As the adage goes, "easier to add later than to try to remove after adding".

Also, assuming we keep some iteration of these topics in this doc ("short options" and "abbreviations"), and since they are closely related, what do you think about combining them into a single section, perhaps with subheaders? Then we can offer some type of guidance of when to use them and how to propose new ones as a single guidance for both topics.

Wdyt?

docs/src/_contribute/cli.md
```diff
@@ -0,0 +1,255 @@
...
+
+### Abbreviations
+
+For excessively long options, especially in cases where no short flag is provided or a commonly used abbreviation is exists,
+you should allow the use of the abbreviation as a documented alias:
+
+```
+meltano --env=production
+meltano --environment=production
+```
```
aaronsteers, line 83: Ditto comments from above. Sometime abbreviations can cause confusion and I would hate to (1) slow down an otherwise-ready PR to discuss whether an abbreviation is necessary + sufficiently helpful, or (2) say "yes" too quickly because we want the feature to go out, without thinking through if the abbreviation is needed.

I think we can use the above thread for this discussion but just wanted to call it out here inline as well.

Perhaps not to others, but as a case in point, `--env` brings to mind a specific env variable mapping, and is less obviously a declaration of the environment name. While I realize `--environment` is longer to type, I just don't want to rush through these decisions without thinking through them.

docs/src/_contribute/cli.md
tayloramurphy: @aaronsteers I don't have strong opinions on short flags and am ok with a more conservative approach 👍

pandemicsyn: > Personally, I find that the benefit in brevity comes at a high cost in readability and having to frequently revisit the docs and/or --help output. 

I think that's only true until you're comfortable with the application. And having purely long options doesn't mean you won't visit help as often - and to be frank I think the opposite is true. I have a much better chance of remember that I need `grep -b` than rembering that I need `grep --before-context`. My push back is, day to day - how often do you folks type `ls --long` vs `ls -l` , `grep -c` vs `grep --count`, `grep -b vs grep --before-context`?

> @tayloramurphy and @pandemicsyn - Would it be okay with you both if we take a conservative approach to these for the time being? As the adage goes, "easier to add later than to try to remove after adding".

I think saying "flags should be in the spec and short flags should be vetted" is totally acceptable. But why completely remove the guidance - that just opens us up for more bike shedding, because it means there's not even standard to propose or a standard for a proposed flag to adhere too.

docs/src/_contribute/cli.md
pandemicsyn: Personally, I think having only long options is terrible ux. Everytime i have to type `--environment` I get sad.

docs/src/_contribute/cli.md
tayloramurphy: @tayloramurphy I'll add that it makes intuitive sense to me, and will likely result in a more enjoyable UX for more experienced users, to have the short flags available but I understand the bikeshedding concerns. A considered approach feels reasonable but I would not be in favor a blanket "no short flags".

edgarrmondragon: I agree with @pandemicsyn. For experienced users, it's easier to remember the short form and I can't think of a CLI that I used heavily where I don't prefer the short flag: `nox -rs`, `poetry install -E docs`, etc.

Now, we don't have to add a short form to _all_ options, just the most used ones. We should leave the other open and offer clear guidance for the team and contributors.

Some middle would be to require long flags be used in documented examples, tutorials, etc. so it's more clear what those flags do, and where possible linking to the full command documentation, where short flags are shown alongside their long counterparts.

pandemicsyn: > Now, we don't have to add a short form to all options, just the most used ones. We should leave the other open and offer clear guidance for the team and contributors.

+1 , right now the guidelines say `The options you expect to be used most frequently should always include both the long AND short variation of the option.` 

 Could definitely revise that to say something along the lines `reserve short flag usage only for what you expect the most frequently used options to be, if you feel the short flag can be confused with another, bias towards using the long variant only`

aaronsteers: > > Now, we don't have to add a short form to all options, just the most used ones. We should leave the other open and offer clear guidance for the team and contributors.
> 
> +1 , right now the guidelines say `The options you expect to be used most frequently should always include both the long AND short variation of the option.` 
> 
>  Could definitely revise that to say something along the lines `reserve short flag usage only for what you expect the most frequently used options to be, if you feel the short flag can be confused with another, bias towards using the long variant only`

Yeah, this makes a lot of sense to me. My point in the first comment above is not that we shouldn't have short flags but only that I want to be mindful ahead of time where collisions are going to occur. And figuring them out before the feature is under development, or after the feature had shipped, is preferable to moving too fast or too slow when the PRs are otherwise ready to ship.

aaronsteers: To come to a short flag strategy we can all feel good about, I do think it'd be helpful to take a pass over existing already shipped features. We'd have some good lessons learned there and would likely come out of that exercise with a pretty holistic set of leanings based on existing use cases and usage patterns.

(I really liked, for example, the uppercase -E for environments because it's a global setting and the lowercase -e being available on specific commands.)

pandemicsyn: > I do think it'd be helpful to take a pass over existing already shipped features.

@aaronsteers that definitely makes sense. I was planning on pop'ing a follow-up issue to clean up some of our help output (to conform to our new guidelines), and propose some flag additions. Would you prefer I send these along as part of this PR?

pandemicsyn: @aaronsteers just wanted to give this thread bump, to keep this PR rolling.

pandemicsyn: @aaronsteers @tayloramurphy I updated the wording to more strongly indicate that short flags should be used only when unambiguous, and the long version is excessively long or expected to be frequently used. I also switched the tone to from "you should" to "you may".

I also clarified how they should be presented in docs and that we'd prefer that only long options be used in things like getting start guides and walkthroughs. 

I also completely nuked the abbreviations/aliases section. 

Are these changes enough to keep this rolling?

tayloramurphy: @pandemicsyn love it. this already has my stamp of approval ✅

pandemicsyn:
@aaronsteers bump. Would really love to finally get a resolution on this one way or another.

aaronsteers (approved):

aaronsteers:
@pandemicsyn - I appreciate your incorporating the my feedback from short options and I really like the rewrite. 👍 

Thanks again for taking the time to design these guidelines. I think these will be a big help for us as we move forward.

Merge when ready.  🚀

PR #6541, opened 2022-08-01
Windows workaround for multiline runs while calling exectuables

### LINKED ISSUE #6540: Windows CI when calling remote exe doesn't fail properly for multiline runs
> Why is Install Dependencies not failing

Found the issue, it's still perplexing but atleast I understand more about what is happening

```
      - name: Build tap on Python ${{ matrix.python-version }} 
        run: |
          poetry env use purposelyfail
          poetry debug
```

You would expect to fail, because `purposelyfail` is an invalid environment. I have a test repo setup.  When you run this on these platforms here's what you get

- [Ubuntu](https://github.com/visch/gh_actions_windows_poetry_test/runs/7615264438?check_suite_focus=true) - Fails as expected
- [Windows](https://github.com/visch/gh_actions_windows_poetry_test/runs/7615264340?check_suite_focus=true) - Succeeds 

Even weirder is on Windows if you run  
```
      - name: Build tap on Python ${{ matrix.python-version }} 
        run: |
          poetry env use purposelyfail
```

The action does fail as you would expect. 

Searching around I found a few issues written up but could only find them cached on google it looks like discussions at least in this area did not index properly https://webcache.googleusercontent.com/search?q=cache:c48kgTB1SeMJ:https://github.community/t/multiline-commands-on-windows-do-not-fail-if-individual-commands-fail/16753+&cd=1&hl=en&ct=clnk&gl=us

Other resources:
- https://stackoverflow.com/questions/69615233/githubactions-on-windows-host-powershell-exit-code-of-previous-lines-being-i
- https://stackoverflow.com/questions/9948517/how-to-stop-a-powershell-script-on-the-first-error

>  Error action preference only stops for a failed cmdlet. A failed external command (e.g. npm process) does not stop the script from proceeding.
> An an alternative, please use bash/cmd shell.

Github documentation is here on the `pwsh` shell https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference

From these docs
> We append if ((Test-Path -LiteralPath variable:\LASTEXITCODE)) { exit $LASTEXITCODE } to powershell scripts so action statuses reflect the script's last exit code.

Is problematic for us because `poetry debug` returns exit code 0. I'd think that `$ErrorActionPreference = 'stop'` would do what we need here and fail at the first failure, but this doesn't happen probably due to calling an external application. 

Still thinking but we're much closer to an answer here. There's plenty of ways forward as well now that we know what's happening 

Put in https://github.com/orgs/community/discussions/27923 maybe we'll get some pointers there

_Originally posted by @visch in https://github.com/meltano/meltano/issues/6479#issuecomment-1200554306_

visch:
Closes #6540

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,3 +1,6 @@
+#We use && for all multi-line runs because Windows won't fail in all scenarios
+#See https://github.com/meltano/meltano/issues/6540 for more information
+
 name: Tests
 
 on:
@@ -53,21 +56,21 @@ jobs:
     - name: Install Poetry
       env:
         PIP_CONSTRAINT: .github/workflows/resources/constraints.txt
-      run: |
+      run: >
         pipx install poetry
-        poetry --version
+        && poetry --version
 
     - name: Upgrade pip
       env:
         PIP_CONSTRAINT: .github/workflows/resources/constraints.txt
-      run: |
+      run: >
         pip install pip
-        pip --version
+        && pip --version
 
     - name: Configure Poetry
-      run: |
+      run: >
         poetry env use "${{ matrix.python-version }}"
-        poetry debug
+        && poetry debug
 
     - name: Start Postgres Container
       if: always() && (matrix.backend-db == 'postgresql')
@@ -83,15 +86,14 @@ jobs:
         postgres:11
 
     - name: Check running containers
-      run: |
-        docker ps -a
+      run: docker ps -a
 
     - name: Install Nox
       env:
         PIP_CONSTRAINT: .github/workflows/resources/constraints.txt
-      run: |
+      run: >
         pipx install nox
-        pipx inject nox nox-poetry
+        && pipx inject nox nox-poetry
 
     - name: Run pytest
       env:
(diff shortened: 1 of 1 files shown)


visch:
@edgarrmondragon / @WillDaSilva  

I put this PR up to see if you all like this idea as a workaround? Wanted to prevent an issue like this from cropping up again on Windows. The fix for getting the correct Python version is seperate and much easier, but detecting an issue like this for the future seemed very important to me (maybe I think it's more important than it really is)

.github/workflows/test.yml
```diff
@@ -1,3 +1,6 @@
+#We use && for all multi-line runs because Windows won't fail in all scenarios
```
WillDaSilva, line 1: I don't understand this explanation. Why does using `&&` have an impact on Windows? Previously, if a command failed, that job would immediately exit without executing the lines that followed the failing line. Wouldn't the behaviour be the same with `&&`, regardless of if it's running on Windows?

visch: https://github.com/meltano/meltano/issues/6540 has the information (1 line down!) Not sure how much to get into it all in the comment? Maybe it should be very detailed, posted this mainly for this reason as it's a fix but not an amazing one :/

visch: @WillDaSilva  Dug a little more, there is a fix on Windows!!! Will have to look into how to implement this but if it works as I hope it does we should be able to do this and move on!
https://github.com/PowerShell/PowerShell/pull/15897

visch: I'll just close this PR for now!

PR #6542, opened 2022-08-01
fix: Windows fails silently on Configure Poetry step 

### LINKED ISSUE #6479: [Bug]: Windows fails silently on Configure Poetry step
### Meltano Version

NA

### Python Version

NA

### Bug scope

Other

### Operating System

Windows

### Description

Our tests in CI are running only for Pytho 3.9 on Windows even though our matrix suggests we test on 3.7-3.10. I started https://github.com/meltano/meltano/pull/6478 to validate this and confirmed it after seeing the following debug info in all Python versions on Windows:

```
Poetry
Version: 1.1.14
Python:  3.9.13

Virtualenv
Python:         3.9.13
Implementation: CPython
Path:           C:\Users\runneradmin\AppData\Local\pypoetry\Cache\virtualenvs\meltano-YRQtpV-c-py3.9
Valid:          True

System
Platform: win32
OS:       nt
Python:   C:\hostedtoolcache\windows\Python\3.9.13\x64
Installing dependencies from lock file
```

That means we're not really testing Meltano on Windows for Python 3.7, 3.8 and 3.10. I suspected this while working on https://github.com/meltano/meltano/pull/6395 since that required some refactoring to how Poetry is installed and used, and saw Windows tests passing on Python 3.8 and 3.9, but failing on 3.7 and 3.10.

### Code

_No response_

visch:
Closes https://github.com/meltano/meltano/issues/6479  (when the issue name is updated :D )

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,10 +26,10 @@ jobs:
         - { python-version: "3.10", os: "ubuntu-latest",  backend-db: postgresql }
         # We'd like to run Windows tests for all backend-dbs see https://github.com/meltano/meltano/issues/6281
         # Windows tests for Python 3.7 and 3.10 are temporarily disabled - see https://github.com/meltano/meltano/issues/6479
-        # - { python-version: "3.7",  os: "windows-2022",   backend-db: sqlite }
+        - { python-version: "3.7",  os: "windows-2022",   backend-db: sqlite }
         - { python-version: "3.8",  os: "windows-2022",   backend-db: sqlite }
         - { python-version: "3.9",  os: "windows-2022",   backend-db: sqlite }
-        # - { python-version: "3.10", os: "windows-2022",   backend-db: sqlite }
+        - { python-version: "3.10", os: "windows-2022",   backend-db: sqlite }
       fail-fast: false
 
     name: "Pytest on py${{ matrix.python-version }} (OS: ${{ matrix.os }}, DB: ${{ matrix.backend-db }})"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,13 +50,6 @@ jobs:
         cache: 'pip'
         cache-dependency-path: 'poetry.lock'
 
-    - name: Install Poetry
-      env:
-        PIP_CONSTRAINT: .github/workflows/resources/constraints.txt
-      run: |
-        pipx install poetry
-        poetry --version
-
     - name: Upgrade pip
       env:
         PIP_CONSTRAINT: .github/workflows/resources/constraints.txt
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -50,23 +50,25 @@ jobs:
...
 
-    - name: Upgrade pip
+    - name: Install pipx
       env:
         PIP_CONSTRAINT: .github/workflows/resources/constraints.txt
       run: |
-        pip install pip
-        pip --version
+        python -m pip install pipx
+        pipx --version
```
edgarrmondragon, line 65: AFAICT GitHub runners come with `pipx` out of the box. Would be good to confirm, though

visch: It does come with `pipx` out of the box documented [here](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md#:~:text=22.2%20(python%203.9)-,Pipx%201.1.0,-RubyGems%203.2.33) the issue is that this causes our issue here https://github.com/meltano/meltano/issues/6479 of Windows running the wrong version of Python while doing testing as the `pipx` out of the box is installed with Python 3.9. There's others way to handle this but I think being explicit like this is probably the easiest to understand 🤷

edgarrmondragon: @visch per my comment on https://github.com/meltano/meltano/issues/6479#issuecomment-1201855709  I don't think we need to focus on that since #6395 landed, and we should probably focus on the actual Windows failures that have been surfaced. Wdyt?

visch: > > Going to focus first on why the https://github.com/meltano/meltano/runs/7509637290?check_suite_focus=true "Install Dependencies" task isn't reporting a failure to github actions.
> 
> @visch that's an intriguing issue with GitHub actions in general, but may not be too relevant anymore at least for our Windows case, since #6395 got merged. I'd be in favor of changing the title of this issue to reference the actual Windows failures, or closing it and create new and separate issues for the failures on Python 3.7 and 3.10. The former I think is the trickier one since it seems to relate to `asyncio`.

+1

Posted the [3.7 issue here](https://github.com/meltano/meltano/issues/6544) , [3.10 issue here](https://github.com/meltano/meltano/issues/6545)

I get what you're saying now, `nox` works properly on the different versions of python on Windows. 

I was really focused on
image

I thought that's what all of this was about, but you already made something else work here. Would have been smarter for me to open a seperate issue I think? I was just extremely confused with that picture failing.

We could update the https://github.com/meltano/meltano/issues/6479 issue to be [Bug]: Window s fails silently on Configure Poetry step . 

Maybe that makes the most sense from here (I can't update the issue name :D)

edgarrmondragon: @visch awesome, thanks!

visch:
Closing for now, may reopen after https://github.com/meltano/meltano/issues/6544 and https://github.com/meltano/meltano/issues/6545 are complete

PR #6543, opened 2022-08-01
fix: Register legacy `DiscoveryFile` object in Ruamel YAML instance

### LINKED ISSUE #6128: docs: Update 2.0 migration guide to include confusing/scary warning about discovery.yml not being parsed
As reported in https://meltano.slack.com/archives/C01TCRBBJD7/p1654785151254499:

> I wanted to hold off the upgrade to Meltano 2.0 a bit in case there were some small issues, but now I'm getting `The `discovery.yml` manifest received from https://discovery.meltano.com/discovery.yml could not be parsed.` errors so I guess I have no choice.

It's just a warning, and Meltano will fall back on the cached or bundled copy, but it still makes people seem like something's wrong. We can't go back and make the warning message sound less concerning, of course.

One option would be to roll https://discovery.meltano.com/discovery.yml back to the latest version supported by <2.0 (version 20, iirc), since 2.0 will use the Hub anyway. 2.0 will find an outdated copy at the location, and fall back on its own bundled version, which is OK. It will also show a warning about the remote `discovery.yml`, but there are fewer users on 2.0 than on pre-2.0 affected by that, and we can suppress that in 2.0.x and only use the bundled copy, since the remote discovery.yml won't be updated anymore anyway and the Hub API is the preferred way to look up plugin definitions going forward.

z3z1ma:
As discussed on slack.

Closes https://github.com/meltano/meltano/issues/6128


edgarrmondragon:
@z3z1ma do you think removing this line from the relevant test file

https://github.com/meltano/meltano/blob/a9821667a5f49bd1d30318c4bb5e1d7c41de2b40/tests/meltano/core/test_plugin_discovery_service.py#L23

would confirm that `discovery.yml` is being correctly handled?

z3z1ma:
Hey @edgarrmondragon ! 👋  

Do you mean using the default pyyaml lib instead?
I updated the plugin discovery service to use the same as what was in the test plugin discovery service which is what is supposed to be used as far as I follow

src/meltano/core/discovery_file.py
```diff
@@ -0,0 +1,44 @@
+"""Legacy discovery file object."""
```
edgarrmondragon, line 1: I had to move the `DiscoveryFile` object to a separate module to avoid circular imports.

tests/meltano/core/test_plugin_discovery_service.py
```diff
@@ -69,7 +65,6 @@ def test_discovery_url_mock(self, subject):
     @pytest.fixture
     def discovery_yaml(self, subject):
         """Disable the discovery mock."""
-        yaml.register_class(DiscoveryFile)
```
edgarrmondragon, line 72: This was causing false negatives (i.e. tests were passing but they shouldn't)

edgarrmondragon (approved):

edgarrmondragon:
> Hey @edgarrmondragon ! 👋
> 
> Do you mean using the default pyyaml lib instead? I updated the plugin discovery service to use the same as what was in the test plugin discovery service which is what is supposed to be used as far as I follow

@z3z1ma I meant I had added some code that was causing false negatives (the `DiscoveryFile` object was being registered _only_ when running tests. I've removed that now, so now tests should really tell us if there's an error in parsing `discovery.yml` 😄

@cjohnhanson @kgpayne I've approved, but I'm also partial author so I'd rather get someone else's 👀 on this.

cc @tayloramurphy to groom the PR title

pandemicsyn (approved):

Issue #6300, opened 2022-06-27
Use `pytest-randomly` plugin


WillDaSilva:
https://github.com/pytest-dev/pytest-randomly

This plugin randomizes the order the tests are run in which prevents us from relying on their order to succeed. It also seeds various sources of randomness to improve reproducibility. For instance, if we encounter a test failure in CI due to randomness, we can run the tests locally with the `pytest --randomly-seed=<seed used in the CI test run>` to reproduce.

I've used this plugin before, and it has been helpful numerous times. I've already run into problems with Meltano's tests that are caused by their execution order. Using this plugin should force us to fix those issues, and prevent them from returning.

cjohnhanson:
@WillDaSilva big :+1: to this. CC'ing @visch, I believe that he has done some exploration on this as part of getting local testing working on Windows.

WillDaSilva:
@aaronsteers Using the plugin is causing a large but variable number of tests to fail. This issue will now also encompass fixing all of the bugs in the tests where they make unreliable assumptions about the global state.

aaronsteers:
@WillDaSilva - Thanks for the update. And yes, that sounds right. No surprise from my side; totally understood that we will have some revamp to do in order to get our tests in a randomizeable state.

PR #6547, opened 2022-08-02
chore: Upgrade to `pytest-docker` v1

### LINKED ISSUE #6496: Upgrade to `pytest-docker` v1
https://pypi.org/project/pytest-docker/

`pytest-docker==1.0.0` was released earlier today, and it includes support for Docker compose v2, which is not a Python package. We'll be able to use that to drop our dev dependency on Docker compose v1, which in turn will let us upgrade to the latest major version of `jsonschema`, which is the final Meltano dependency that is considered "out of date" by `libraries.io`/`SourceRank`.

We should wait a week or two to see if any issues come up with the new major release. We're in no rush to make this change, so maybe we'll end up using `pytest-docker==1.0.1`, or something like that instead.

WillDaSilva:
Closes #6496


WillDaSilva:
I updated the Poetry lock file using Poetry `1.1.13`, which had a bug that would set all updated entries under `[metadata.files]` to `[]`. My hope is that updating the lock file with Poetry `1.1.14`, then running the tests again should fix the issue in CI.

edgarrmondragon:
> I updated the Poetry lock file using Poetry `1.1.13`, which had a bug that would set all updated entries under `[metadata.files]` to `[]`. My hope is that updating the lock file with Poetry `1.1.14`, then running the tests again should fix the issue in CI.

@WillDaSilva this might help https://github.com/python-poetry/poetry/issues/5988#issuecomment-1180360559

WillDaSilva:
Keeping `flask-login` at `0.6.1` fixes [the failing tests](https://github.com/meltano/meltano/runs/7634167148).

When the tests are run in isolation, they do not error, regardless of whether I had `flask-login` at `0.6.1` or `0.6.2`, which suggests that resolving #6300 would address the underlying issue here.

Relevant changelog: https://github.com/maxcountryman/flask-login/blob/main/CHANGES.md#version-062

pyproject.toml
```diff
@@ -38,11 +38,12 @@ fasteners = "^0.17.3"
 flask = "^2.1"
 flask-cors = "^3.0.7"
 flask-executor = "^0.10"
+flask-login = "=0.6.1" # https://github.com/meltano/meltano/issues/6548
```
edgarrmondragon, line 41: Is this dep syntax the same as

```suggestion
flask-login = "==0.6.1" # https://github.com/meltano/meltano/issues/6548
```

?

WillDaSilva: [The Poetry docs](https://python-poetry.org/docs/dependency-specification/#exact-requirements) don't have any info about using `=` or `==`. Apparently to specify an exact requirement, you're supposed to use the version number on its own. Seems like `=`, `==`, and the plain version number are all equivalent.

edgarrmondragon: Yeah, Poetry does its own dep parsing, though [PEP 440](https://peps.python.org/pep-0440/#version-matching) prefers `==`. Regardless, I don't have any strong preference

edgarrmondragon (approved):

edgarrmondragon:
@WillDaSilva can you enable the Windows x Python 3.10 tests on this PR to see if this fixes them?

WillDaSilva:
> @WillDaSilva can you enable the Windows x Python 3.10 tests on this PR to see if this fixes them?

I was planning to do that after this PR goes through, to avoid introducing errors to this one if something goes wrong.

edgarrmondragon:
> I was planning to do that after this PR goes through, to avoid introducing errors to this one if something goes wrong.

@WillDaSilva sounds good 👍

PR #6557, opened 2022-08-03
fix: Remove broken unused method `PluginsTrackingContext.append_plugin_context`

WillDaSilva:
- Updated the type hint for the command string to include `None`, as the schema also accepts that, and we explicitly use `None` for that parameter  in `PluginsTrackingContext.from_elt_context`.
- Removed the unused method `PluginsTrackingContext.append_plugin_context`:
  - It was broken because it accessed the `PluginsTrackingContext` as a dict, despite it (and its parent classes) not defining `__getitem__`.
  - It was broken because it tried to append a set containing a single plugin context dict, rather than the dict itself.

### DIFF
--- a/src/meltano/core/tracking/contexts/plugins.py
+++ b/src/meltano/core/tracking/contexts/plugins.py
@@ -16,7 +16,7 @@ from meltano.core.utils import hash_sha256, safe_hasattr
 logger = get_logger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


WillDaSilva:
Codecov doesn't seem to be working (for this PR):

image

edgarrmondragon (approved):

PR #6563, opened 2022-08-03
feat: add support for env var inheritance across config layers

### LINKED ISSUE #5983: Update environment variable inheritance to be in line with currently xfail-ing tests

cjohnhanson:
This PR changes the ~`PluginSettingsService.env_override`~ `PluginInvoker.env()` to expand environment variables at each config layer using the environment provided by the "enclosing" config layer, and then uses those expanded environments in the correct precedence order to finally get behavior in line with the spec mentioned #5983. 

Closes #5983 (finally)


cjohnhanson:
Heads up to reviewers: the failing checks are:
1. From fossa, apparently an unexpected API error. I think this is on their end.
1. flake8 errors for `utils/__init__.py`, which I'm working through but will not have any implications for code review.

src/meltano/core/plugin_invoker.py
```diff
@@ -321,10 +324,63 @@ def env(self):
...
+            # Expand active env plugin env w/ expanded root plugin env
+            expanded_active_env_plugin_env = (
+                expand_env_vars(
+                    self.settings_service.environment_plugin_config.env,
+                    expanded_root_plugin_env,
+                    raise_if_missing=strict_env_var_mode,
+                )
+                if self.settings_service.environment_plugin_config
+                else {}
+            )
```
edgarrmondragon, line 374: Nice! It's very clear what's going on here and which variables are used to expand the next level down 👍

edgarrmondragon (approved):

src/meltano/core/plugin_invoker.py
pandemicsyn: Big +1

pandemicsyn (approved):

PR #6569, opened 2022-08-04
ci: Support pushing images to GitHub registry

### LINKED ISSUE #6493: feature: Publish Docker images to the GitHub registry
As discussed during the call on 2022-07-26 with myself, @aaronsteers, and @edgarrmondragon, we should run our Docker publish GitHub workflow on a weekly basis. This will let us detect security issues that have arisen since the last scan was run.

For testing purposes we should introduce a new setting `registry`, and set it to `ghcr.io`, which is the GitHub registry. The alternative would be to do dry runs - either way we don't want to publish the images from these weekly runs to Docker Hub.

To avoid filling up the GitHub registry (I'm not sure how much space we get for free / on our plan), we should only publish the `:latest` tag for each image. That way they'll be overwritten each week.

Eventually (beyond the scope of this issue) we'll want to run tests against Meltano in these images. We may even want to use the images for our regular Linux pytest runs in CI.

WillDaSilva:
Closes #6493

### DIFF
--- a/.github/actions/docker-build-scan-push/action.yml
+++ b/.github/actions/docker-build-scan-push/action.yml
@@ -1,5 +1,5 @@
 name: 'build-scan-push'
-description: 'Build a Docker image, scan it, then push it to Docker Hub'
+description: 'Build a Docker image, scan it, then push it to a container registry'
 
 inputs:
   tags:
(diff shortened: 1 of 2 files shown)


edgarrmondragon (approved):
🐳 🚀

PR #6577, opened 2022-08-07
ci: Fix Docker scan action

WillDaSilva:
This should've been done originally as part of https://github.com/meltano/meltano/pull/6562. The lines I ran locally to test it included the `jq` step, but I mistakenly didn't update the action with it.


WillDaSilva:
Successful run here: https://github.com/meltano/meltano/actions/runs/2813497937

Apparently we can run workflow dispatch runs for PRs. Maybe that's a new feature? I wasn't able to find any reference for doing that on any of the several previous times I checked. In any case, this feature will make changes to workflow dispatch events far easier to develop.

pandemicsyn (approved):

PR #6578, opened 2022-08-08
feat: Add `--no-install` option to `meltano add`

### LINKED ISSUE #6550: [Feature]: meltano add to project without install
### Feature scope

### Description

**As a user, I need to add to my meltano project without install**

Adding to a meltano project is currently a very slow operation.  Our requirement is to add all the plugins to the project and then later build (i.e. 'meltano install') the whole project.  This means that the initial 'meltano add' is unnecessarily slow as we don't use any of the pip-installed modules from the initial 'meltano add'

Work arounds considered:
- directly add a meltano.yml format file with our plugin definition to a directory with the root meltano.yml includes referencing that directory path & directly save a lock file to the /plugins directory  (essentially avoiding using 'meltano add' at all)

This proposal is to be able to add to the meltano project rapidly without installing all pip dependencies, and we'd ideally like to be able to freeze those dependencies.

`meltano add --skip-install` should
1. download the lock file
2. add to the meltano.yml
3. add all the lock and meltano.yml files for any `requires` plugins (https://github.com/meltano/hub/pull/505 & https://github.com/meltano/meltano/issues/6107)

Related to:
This feature is related to https://github.com/meltano/meltano/issues/6416
**As a user, I need to add to my meltano project without install and create pip freeze**

`meltano add --skip-install --freeze` consider whether this --skip-install works if --freeze is used (assuming https://github.com/meltano/meltano/issues/6416 might implement --freeze)

ReubenFrankel:
Resolves #6550 to skip the install step when running `meltano add`.

---

### Checklist

- [x] Implement feature
- [x]  Add test(s)
- [x]  Update docs

### DIFF
--- a/docs/src/_reference/command-line-interface.md
+++ b/docs/src/_reference/command-line-interface.md
@@ -80,7 +80,7 @@ meltano add <type> <name> --inherit-from <existing-name>
 meltano add extractor tap-ga--client-foo --inherit-from tap-google-analytics
 ```
 
-By default, `meltano add` will attempt to install the plugin after adding it. Use `--no-install` to short-circuit this behaviour:
+By default, `meltano add` will attempt to install the plugin after adding it. Use `--no-install` to skip this behavior:
 
 ```bash
 meltano add <type> <name> --no-install
--- a/docs/src/_reference/command-line-interface.md
+++ b/docs/src/_reference/command-line-interface.md
@@ -20,7 +20,7 @@ Specifically, it will:
 3. Store the plugin definition in the `./plugins` directory (see [Lock Artifacts](/concepts/plugins#lock-artifacts))
 4. Assuming a valid `pip_url` is specified, install the new plugin using [`meltano install <type> <name>`](#install), which will:
    1. Create a dedicated [Python virtual environment](https://docs.python.org/3/glossary.html#term-virtual-environment) for the plugin inside the [`.meltano` directory](/concepts/project#meltano-directory) at `.meltano/<type>s/<name>/venv`, e.g. `.meltano/extractors/tap-gitlab/venv`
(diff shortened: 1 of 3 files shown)


docs/src/_reference/command-line-interface.md
```diff
@@ -80,6 +80,15 @@ meltano add <type> <name> --inherit-from <existing-name>
 meltano add extractor tap-ga--client-foo --inherit-from tap-google-analytics
 ```
 
+By default, `meltano add` will attempt to install the plugin after adding it. Use `--no-install` to short-circuit this behaviour:
```
tayloramurphy, line 83:
```suggestion
By default, `meltano add` will attempt to install the plugin after adding it. Use `--no-install` to skip this behavior:
```

tayloramurphy (approved):
Minor docs change - excited about this!

docs/src/_reference/command-line-interface.md
ReubenFrankel: :us:

afolson (approved):

pandemicsyn:
@ReubenFrankel looking good! Looks like might have some issues with the test though when running as part of the larger suite (the plugins are already present): https://github.com/meltano/meltano/runs/7732169282?check_suite_focus=true

ReubenFrankel:
@pandemicsyn Yeah, I had a look an can see how that's happening, since the test is just a slightly modified version of `test_add`. Any ideas on the best approach here (e.g. reset test fixture, different plugin names)?

edgarrmondragon (approved):

pandemicsyn (approved):
lgtm!

ReubenFrankel:
@pandemicsyn Did you want a changelog entry added?

PR #6581, opened 2022-08-08
change: add new settings for `hub_api_root` and `hub_url_auth`

### LINKED ISSUE #6579: feature: Make `hub_url` setting override the whole url & add a `hub_url_auth` setting to support for custom hubs.
### Feature scope

### Description

The current `hub_url` setting only allows you to change part of the url.

```python
    @property
    def hub_api_url(self):
        """Return the URL of the Hub API.

        Returns:
            The URL of the Hub API.
        """
        hub_url = self.settings_service.get("hub_url")
        return f"{hub_url}/meltano/api/v1"
```

We have been using a custom `discovery_url` with `discovery_url_auth` for a while, and now we are making the move over to Meltano 2.0+ we want to take advantage of the `hub_url`. We would like to use our `hub_url` with our own path and we also need to be able to supply auth with our request.

I was thinking we could make the default value for the `hub_url` to be Meltano's complete url `https://hub.meltano.com/meltano/api/v1` and make `hub_url` replace this completely. (This makes this setting the same as the `discovery_url` setting logic).

Also for supplying the auth with the request to the `hub_url` just copying the same logic from the `discovery_url_auth` setting as a new `hub_url_auth` setting.

The only issue I can think of is that in changing the logic for the `hub_url` this would break anyone with a project already using the `hub_url` in its current form.

DanielPDWalker:
Addresses #6579 to make `hub_url` override the whole url and add an optional `hub_url_auth` setting.

---

## Summary of changes

- Changes the default value of `hub_url` to `https://hub.meltano.com`. Corresponding changes are made to the Hub client.
- Adds `hub_api_root` setting. Default value is `None`, but default _computed_ value is `{hub_url}/meltano/api/v1`.
- Adds `hub_url_auth` setting. Default value is `None`. If provided, it sets the `Authorization` header for requests to the Hub.


tayloramurphy (approved):
docs approval

afolson (approved):

docs/src/_reference/settings.md
```diff
@@ -254,6 +254,28 @@ meltano config meltano set hub_url http://localhost:4000
 export MELTANO_HUB_URL=http://localhost:4000
```
edgarrmondragon, line 254: @DanielPDWalker can you update the env var and `meltano config meltano set hub_url ...` examples to reflect the new expectation?

```suggestion
export MELTANO_HUB_URL=http://localhost:4000/meltano/api/v1
```

edgarrmondragon (changes requested):

docs/src/_reference/settings.md
DanielPDWalker: Sure thing, Ill make the changes tomorrow.

DanielPDWalker:
Also I just wanted to check if you all had any views on how this was implemented? Want to make sure this is inline with how everyone would want it/uses it.

edgarrmondragon:
> Also I just wanted to check if you all had any views on how this was implemented? Want to make sure this is inline with how everyone would want it/uses it.

@DanielPDWalker thanks, there's was some discussion a while ago in the PR that added `hub_url`: https://github.com/meltano/meltano/pull/6002#discussion_r890608473.

@aaronsteers commented:

> I think I'd prefer to rename to `DISCOVERY_API_ROOT`,`MELTANO_DISCOVERY_API_ROOT` or similar, and use the full path including version, but we could later add an `API_ROOT` that overrides the other if/when this becomes an issue.
> 
> On tap-gitlab we ended up having to support _both_ conventions - the site base URL and _also_ the versioned path, which is less than ideal, and I think we can avoid here if we can and just take exactly the root of the API endpoint before additional `path` items are added.
> 
> Either way is fine though, and I don't think we need to block or delay on this point.

so perhaps he or @tayloramurphy have a preference.

aaronsteers (commented):
Dropping a 'hold' marker on this PR while we evaluate the name `hub_url` vs `discovery_api_root` vs an alternate proposal of just exposing again `discovery_yml_url` and whether/if to support some kind of legacy `discovery.yml` endpoint that is a single file.

This PR would just be to allow the API base to be fully qualified, to be clear, but in Slack and office hours the topic of supporting the legacy format came up and I want to make sure we are on the same page about path forward before merging.

aaronsteers (changes requested):
[Not a 'request for change' but a request to hold on resolving above discussions. Thanks.]

aaronsteers:
@edgarrmondragon - To your question above, I think we could avoid the breaking change if we add a `hub_api_root` as defaulting to `${hub_url}/meltano/api/v1`. In that case, overriding `hub_api_root` would override any effect from `hub_url`.

What do you think about this approach and the naming of each?

edgarrmondragon:
> @edgarrmondragon - To your question above, I think we could avoid the breaking change if we add a `hub_api_root` as defaulting to `${hub_url}/meltano/api/v1`. In that case, setting a constant value for `hub_api_root` would override any effect from `hub_url`.
> 
> @edgarrmondragon and @DanielPDWalker - What do you think about this approach and the naming of each?

@aaronsteers that works for me!

aaronsteers:
> To clarify, does it mean
> 
> 1. that users overriding `hub_api_root` would be able to point to a different domain from `hub_url`? e.g. `http://localhost:4000/hub` and `http://localhost:5000/hub-api` or
> 2. changing `hub_url` always affects the computed value of `hub_api_root`.

I think the first, not the second.

Unless we create a new use case for `hub_url` in the future (such as quick link for web browsers?), these would conflct and the second would win:

## Scenario 1: new `hub_api_root` overrides old `hub_url`

1. `hub_url` overridden by the user to custom value `https://doesnt-exist.com`. (Would fail on its own.)
2. `hub_api_root` overridden by the user to custom value `https://mysite.com/my-plugins`. (Makes `hub_url` irrelevant.)

And then, the expected endpoints would be :

- `https://mysite.com/my-plugins/plugins/index`
- `https://mysite.com/my-plugins/plugins/extractors/index`
- etc.

## Scenario 2: same as prior behavior 

1. `hub_url` overridden by the user to custom value `https://mysite.com`
2. `hub_api_root` is not set by the user.

And then, the expected endpoints would be :

- `https://mysite.com/meltano/api/v1/plugins/index`
- `https://mysite.com/meltano/api/v1/plugins/extractors/index`
- etc.

DanielPDWalker:
All sounds good to me, think the new setting sounds like the only way to make a non breaking change. I've added the new `hub_api_root` setting and docs. Reverted the `hub_url` changes.

edgarrmondragon (approved):

edgarrmondragon:
@aaronsteers I've approved this and added a summary of changes to the PR description.

DanielPDWalker:
@edgarrmondragon Think its worth asking for the docs to be checked again since I've added the new setting?

tayloramurphy (approved):

src/meltano/core/hub/client.py
```diff
@@ -108,15 +108,29 @@ def __init__(self, project: Project) -> None:
...
         """Return the URL of the Hub API.
 
         Returns:
             The URL of the Hub API.
         """
+        hub_api_root = self.settings_service.get("hub_api_root")
         hub_url = self.settings_service.get("hub_url")
-        return f"{hub_url}/meltano/api/v1"
+
+        return hub_api_root or f"{hub_url}/meltano/api/v1"
```
edgarrmondragon, line 124: > I'm assuming a logic expecting "one or none but not both". And "if both are passed, `hub_api_root` wins and the other is ignored".

@aaronsteers I think this implementation meets those requirements. Does this have your 👍?

aaronsteers: @edgarrmondragon -  Looks good! Thanks!

aaronsteers (approved):

PR #6582, opened 2022-08-08
fix: Make `default_environment` a top-level setting

### LINKED ISSUE #6115: `default_environment` can only be set in `meltano.yml`, not using `meltano config meltano set`
@tayloramurphy pointed out that `meltano config meltano set default_environment prod` or `meltano config meltano unset default_environment` doesn't do anything.

This is because `default_environment` is not actually a setting, but a core property of `meltano.yml`: https://github.com/meltano/meltano/blob/dfd9650170ff97fd69fdff7ac8dc7dc0841bf663/src/meltano/core/meltano_file.py#L21

https://docs.meltano.com/concepts/environments#default-environments doesn’t suggest that `meltano config meltano` would work here, only setting it directly in `meltano.yml`. But https://docs.meltano.com/getting-started#view-and-activate-your-environments does call it a setting, and someone seeing it in `meltano.yml` could reasonably expect it to work just like the other keys in there like `send_anonymous_usage_stats` and `project_id`.

If we want to make it a proper setting, we should be able to define it in `settings.yml` and replace https://github.com/meltano/meltano/blob/main/src/meltano/cli/cli.py#L70 with `ProjectSettingsService(project).get("default_environment")`.

WillDaSilva:
Closes #6115

### DIFF
--- a/src/meltano/cli/cli.py
+++ b/src/meltano/cli/cli.py
@@ -1,6 +1,8 @@
+"""Definition of the top-level Click group for the Meltano CLI."""
+
 from __future__ import annotations
 
-import logging  # noqa: D100
+import logging
 import sys
 from typing import NoReturn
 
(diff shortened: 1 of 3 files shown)


WillDaSilva:
Draft until #6583 has been resolved.

edgarrmondragon (approved):
👌

PR #6601, opened 2022-08-10
fix: raise error on migration fail

### LINKED ISSUE #6406: bug: `meltano install` exits success even if migrations fail
I would expect `meltano install` to fail and exit with a non-zero exit code if the migrations fail. Instead, it proceeds to install plugins.

To reproduce:

```
$ meltano --version && python --version
meltano, version 2.2.0
Python 3.9.13
$ meltano init example
$ cd example
# this is just an example of a way to make migrations fail
$ sqlite3 .meltano/meltano.db "UPDATE alembic_version SET version_num = 'garbage';"
$ meltano install                                                                   
2022-07-12T16:35:37.771519Z [info     ] Environment 'dev' is active
2022-07-12T16:35:38.860146Z [error    ] No such revision or branch 'garbage' 
  File ".venv/lib/python3.9/site-packages/meltano/core/migration_service.py", line 89, in upgrade
    self.ensure_migration_needed(script, context, head)
  File ".venv/lib/python3.9/site-packages/meltano/core/migration_service.py", line 55, in ensure_migration_needed
    for rev in script.iterate_revisions(current_head, "base"):
  File ".venv/lib/python3.9/site-packages/alembic/script/revision.py", line 799, in iterate_revisions
    revisions, heads = fn(
  File ".venv/lib/python3.9/site-packages/alembic/script/revision.py", line 1426, in _collect_upgrade_revisions
    for rev in self._parse_upgrade_target(
  File ".venv/lib/python3.9/site-packages/alembic/script/revision.py", line 1218, in _parse_upgrade_target
    return self.get_revisions(target)
  File ".venv/lib/python3.9/site-packages/alembic/script/revision.py", line 552, in get_revisions
    return tuple(
  File ".venv/lib/python3.9/site-packages/alembic/script/revision.py", line 553, in <genexpr>
    self._revision_for_ident(rev_id, branch_label)
  File ".venv/lib/python3.9/site-packages/alembic/script/revision.py", line 624, in _revision_for_ident
    raise ResolutionError(
alembic.script.revision.ResolutionError: No such revision or branch 'garbage'
Cannot upgrade the system database. It might be corrupted or was created before database migrations where introduced (v0.34.0)
Installing 0 plugins...
$ echo $?                                                                           
0

### PR DESCRIPTION
**Pull request by kgpayne**
Closes #6406 

- #6406

### DIFF
--- a/src/meltano/core/migration_service.py
+++ b/src/meltano/core/migration_service.py
@@ -105,14 +105,9 @@ class MigrationService:
                 click.secho("System database up-to-date.")
         except Exception as err:
             logging.exception(str(err))
-            click.secho(
-                "Cannot upgrade the system database. It might be corrupted or was created before database migrations where introduced (v0.34.0)",
-                fg="yellow",
-                err=True,
-            )
             raise MigrationError(
                 "Cannot upgrade the system database. It might be corrupted or was created before database migrations where introduced (v0.34.0)"
-            ) from err
+            )
         finally:
             conn.close()
 
--- a/src/meltano/core/migration_service.py
+++ b/src/meltano/core/migration_service.py
@@ -110,6 +110,9 @@ class MigrationService:
                 fg="yellow",
                 err=True,
             )
+            raise MigrationError(
+                "Cannot upgrade the system database. It might be corrupted or was created before database migrations where introduced (v0.34.0)"
+            ) from err
         finally:
             conn.close()
 
--- a/src/meltano/cli/params.py
+++ b/src/meltano/cli/params.py
@@ -73,7 +73,7 @@ class pass_project:  # noqa: N801
                     migration_service.upgrade(silent=True)
                     migration_service.seed(project)
                 except MigrationError as err:
-                    raise CliError from err
+                    raise CliError(str(err))
 
             func(project, *args, **kwargs)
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/meltano/core/migration_service.py [hunk 1]
```diff
@@ -110,6 +110,9 @@ def upgrade(  # noqa: WPS213, WPS231 too many expression and too complex
                 fg="yellow",
                 err=True,
             )
+            raise MigrationError(
+                "Cannot upgrade the system database. It might be corrupted or was created before database migrations where introduced (v0.34.0)"
+            ) from err
```
pandemicsyn, line 115: We'd log the same error message twice back to back with this change right? 

First via the click.secho in the line above, and then again when you raise the CliError from this MigrationError in params.py. Could probably just omit the click.secho call above.

pandemicsyn (commented):
I'm guessing

edgarrmondragon (approved):
This LGTM, save for @pandemicsyn's suggestion.

src/meltano/core/migration_service.py
kgpayne: Thanks @pandemicsyn for checking this out. Using the `raise ... from ...` pattern I wasn't getting the "Cannot upgrade..." message in the stack trace during testing, only the root cause (`alembic.script.revision.ResolutionError: No such revision or branch 'garbage'`), so the `secho` was helpful context. However I have modified the exception raises slightly to produce both the root cause stack trace and a final helpful message via the `CliError` (in stead of relying on `secho`) 🙂

PR #6602, opened 2022-08-10
docs: Update `Docker` definition in `glossary.yml`

WillDaSilva:
Came across our definition of "Docker" on https://docs.meltano.com/reference/glossary#docker, and thought it could use an upgrade.


edgarrmondragon (approved):
So, is it not `a thing`? 😅

pandemicsyn (approved):
😆

Issue #3437, opened 2022-05-24
Create `cron` utility extension


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3522

Originally created by @alxthm on 2022-05-24 14:54:47

---

For simple meltano deployments, it would be great to have a cron-based orchestrator using the `meltano schedule` configuration, without having to set-up an Airflow instance with a db, users, etc.

The meltano UI already makes it possible to manually run meltano schedules, so this would be a nice complement :smile:

pandemicsyn:
> @pandemicsyn would https://github.com/meltano/meltano/issues/5962 resolve this issue? I think it would, since it could essentially do what I suggested here: https://gitlab.com/meltano/meltano/-/issues/3522#note_962685280. Though I'm not sure about having Meltano itself edit the user's crontab, rather than Meltano printing the crontab lines to stdout.

Yep, i think https://github.com/meltano/meltano/issues/5962 would solve this. I'd probably lean toward not using user crontab entries and generating meltano cron file that can be dropped in /etc/cron.d but that seems like a user preferences thing and controllable through a cron export plugin config.

alxthm:
Based on your input, I made a small python script to transform meltano schedules (both job and elt schedule, in json format) into a crontab file:
https://gist.github.com/alxthm/b62eceb1e47f0320e5bd82a67f33e042

I am currently using it this way:
```
meltano schedule list --format=json | python make_crontab.py >crontab.file
crontab crontab.file
```

It relies on a `CRONTAB_MELTANO_EXECUTABLE` environment variable to specify the path to `meltano` (in case it is not in the crontab PATH).

Happy if it can help (and open to feedback if you have any ofc).

aaronsteers:
```
meltano invoke cron-util schedule
meltano invoke cron-util deschedule
```

aaronsteers:
The pattern ideally would also work for:

- AWS Cloudwatch Events (could drive step functions or ECS executions)

longtomjr:
I don't know if it will be out of scope for this feature request, but it might be valuable to have an option to integrate with cron / dead man's switch tools like https://healthchecks.io/.

It might also be valuable for setups based on the proposal at #5962. 

The simplest way I can think of would be to add a way to have an option for an extra command that will be be added to the cron entry.


So instead of generating
```
0 0 * * * (cd /project && meltano run ...) 2>&1 | /usr/bin/logger -t meltano
```
meltano generates the following cron
[...]

PR #6610, opened 2022-08-11
refactor: Replace `StaleJobFailer` class with `fail_stale_jobs` function

WillDaSilva:
Just a small simplification I made while working on #3437 that I figured should be pull out as a separate PR.


pandemicsyn (approved):
lgtm. Also TIL only schedule list fails all stale jobs 🤔

WillDaSilva:
> lgtm. Also TIL only schedule list fails all stale jobs 🤔 

Yeah, that struck me as odd. I may put up a PR later that has it fail all stale jobs the first time a database session is opened on each run, or something like that.

Before I do that though I wanted to get some more informed opinions on it. Is it okay/desirable to only fail stale jobs when a user lists the schedules?

PR #6614, opened 2022-08-12
ci: Generate test workflow summary

### LINKED ISSUE #6454: Generate test workflow summary
As explained [in this GitHub blog post](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/), and detailed [in the GitHub docs](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary), we can create a markdown job summary for a workflow. Creating a test workflow job summary similar to the one shown below would be nice:

image

We could also add the number of xfailed/xpassed tests, the number of warnings, and the coverage. Having improved visibility on this information will be useful. Consider for instance that we currently probably wouldn't notice if one of our tests began xpassing, or emitting a warning. We rely on warnings in tests to indicate if we have any typos in setting names, among other things.

WillDaSilva:
Example test summary: https://github.com/meltano/meltano/actions/runs/2849461047

Looks like a lot of warnings are raised by the mssql tests, and Windows has several test which are xpassing. Glad to see this summary table is already paying off. I'll open issues to address the warnings and xpassing tests, if we don't already have issues open for that.

EDIT: Opened issues:
- https://github.com/meltano/meltano/issues/6629
- https://github.com/meltano/meltano/issues/6630

Closes #6454

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,25 +29,44 @@ jobs:
     strategy:
       matrix:
         include:
-        - { python-version: "3.7",  os: "ubuntu-latest",  backend-db: sqlite }
-        - { python-version: "3.8",  os: "ubuntu-latest",  backend-db: sqlite }
-        - { python-version: "3.9",  os: "ubuntu-latest",  backend-db: sqlite }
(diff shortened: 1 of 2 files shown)


edgarrmondragon (approved):
Nice 👍

Issue #6620, opened 2022-08-12
Remove legacy telemetry events


pnadolny13:
We've had full coverage for our new unstructured telemetry events for some time now, presumably all major bugs have been spotting, and the downstream Squared dbt transformations [don't rely on the legacy events](https://github.com/meltano/squared/issues/340) as a primary source. We can go ahead and remove any extra code for these legacy events and stop double sending a backup structured event (if were still doing that).

cc @WillDaSilva @tayloramurphy

WillDaSilva:
Thanks for ensuring they're no longer necessary @pnadolny13! Removing the legacy telemetry code will be satisfying.

PR #6621, opened 2022-08-12
refactor(core): Make `ProjectInitService.init` generate/set a `project_id`

WillDaSilva:
While addressing #6620 I realized that without the legacy tracker, nothing generates a project ID during initialization. The fact that that responsibility falls on it seems like an oversight.

This PR makes it so that new projects get a default project ID in the form `<project name>-<random UUID>`.

### DIFF
--- a/src/meltano/core/project_init_service.py
+++ b/src/meltano/core/project_init_service.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import os
+import uuid
 
 import click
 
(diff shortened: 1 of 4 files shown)


edgarrmondragon (approved):

pandemicsyn (commented):
Did we previously prefix the project id with the project name? If not - are we ok adding that from a privacy/analytics standpoint.

WillDaSilva:
> Did we previously prefix the project id with the project name? If not - are we ok adding that from a privacy/analytics standpoint.

We did not. This project ID is not the hashed project ID that gets sent to us, so I do not see any privacy/analytics issues. We still hash the project ID before sending it in telemetry data. This project ID is the one that gets written into `meltano.yml`. Previously a newly created project would have no project ID in `meltano.yml`, then one would be generated by the telemetry. Now one gets generated by `meltano init` directly, and the telemetry code only has to set the project ID if it's being restored from `analytics.json`.

pandemicsyn (approved):

PR #6622, opened 2022-08-12
chore: Remove legacy Google analytics telemetry

### LINKED ISSUE #6620: Remove legacy telemetry events

WillDaSilva:
Closes #6620


src/meltano/cli/install.py
```diff
@@ -8,6 +8,7 @@
 from meltano.core.project_plugins_service import ProjectPluginsService
 from meltano.core.tracking import CliEvent, PluginsTrackingContext
 
+from ..core.tracking.tracker import Tracker
```
pandemicsyn, line 11: I know this technically works, just making sure it's intentional (ditto for the similar one below).

pandemicsyn (approved):
🥳 Screen Shot 2022-08-15 at 7 04 09 PM 🎉 

One question inline but generally lgtm - and super glad to have this gone!

src/meltano/cli/install.py
WillDaSilva: Good catch. It must've been added automatically, since I certainly didn't write that. I'll remove it.

pandemicsyn: The auto import hook from my editor used to do this occasionally - so if you come across some pre-existing ones, they're probably my fault 😁

PR #6627, opened 2022-08-14
fix: Raise exception when meltano.yml is empty

rawwar:
When meltano.yml is empty, the following error is raised 
```
Traceback (most recent call last):
  File "/home/kalyan/anaconda3/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/kalyan/anaconda3/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/kalyan/.vscode-server/extensions/ms-python.python-2022.12.0/pythonFiles/lib/python/debugpy/adapter/../../debugpy/launcher/../../debugpy/__main__.py", line 39, in <module>
    cli.main()
  File "/home/kalyan/.vscode-server/extensions/ms-python.python-2022.12.0/pythonFiles/lib/python/debugpy/adapter/../../debugpy/launcher/../../debugpy/../debugpy/server/cli.py", line 430, in main
    run()
  File "/home/kalyan/.vscode-server/extensions/ms-python.python-2022.12.0/pythonFiles/lib/python/debugpy/adapter/../../debugpy/launcher/../../debugpy/../debugpy/server/cli.py", line 284, in run_file
    runpy.run_path(target, run_name="__main__")
  File "/home/kalyan/.vscode-server/extensions/ms-python.python-2022.12.0/pythonFiles/lib/python/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py", line 321, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "/home/kalyan/.vscode-server/extensions/ms-python.python-2022.12.0/pythonFiles/lib/python/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py", line 135, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "/home/kalyan/.vscode-server/extensions/ms-python.python-2022.12.0/pythonFiles/lib/python/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py", line 124, in _run_code
    exec(code, run_globals)
  File "/home/kalyan/work/meltano_project/meltano_debug/meltano/src/tester.py", line 8, in <module>
    cli(["environment",'list'])
  File "/home/kalyan/.cache/pypoetry/virtualenvs/meltano-P8gWqx08-py3.9/lib/python3.9/site-packages/click/core.py", line 1129, in __call__
    return self.main(*args, **kwargs)
[...]

PR #6628, opened 2022-08-14
feat: Make `.meltano`  folder path configurable via new `MELTANO_SYS_DIR_ROOT` env var

### LINKED ISSUE #6619: feature: Make `.meltano` path configurable
### Feature scope

### Description

Making creation of `.meltano`  folder configurable will make it easy when running Meltano projects in read-only file systems. This will be similar to how Meltano accepts `db` path

rawwar:
draft PR for #6619


aaronsteers:
@rawwar - Thanks for this contribution! For my part, the implementation looks pretty clean. I'll let @pandemicsyn take the official review though. I've approved CI tests to run on this PR. Feel free to move out of "Draft" status when you see tests passing. 

@pandemicsyn, @edgarrmondragon, @tayloramurphy - The only things I see to pause on here is the naming of the environment variable - and whether this needs to be a setting or if a env var parse is fine.

Some options regarding possible names/spellings:

- `DOT_MELTANO_ROOT`
- `MELTANO_TEMPDIR`
- `MELTANO_TEMP_DIR`
- `MELTANO_TEMP_DIR_ROOT`
- `MELTANO_SYSTEM_DIR`
- `MELTANO_SYSDIR_ROOT`
- `MELTANO_SYS_DIR_ROOT`

I think I like following the convention of all env vars starting with `MELTANO_`.

tayloramurphy:
@aaronsteers since we have `MELTANO_PROJECT_ROOT` I like keeping "root" in the name. My preference would either be `MELTANO_TEMP_DIR_ROOT` or `MELTANO_SYS_DIR_ROOT`. I'm lean towards the first once since it makes it seems less "stable" and would encourage folks to rely on it less. But I'll leave it to you to make the final call now that my preference is voiced 👍

pandemicsyn:
> @aaronsteers since we have MELTANO_PROJECT_ROOT I like keeping "root" in the name. My preference would either be MELTANO_TEMP_DIR_ROOT or MELTANO_SYS_DIR_ROOT

I'm a no-go on `MELTANO_TEMP_DIR_ROOT` because I think that could lead contributors (either to meltano, or via extensions) to use it as unstructured dumping ground like they would an actual temp dir. That might be ok, but feel like that should be something more intentional, like `$MELTANO_TMP_DIR_ROOT=$MELTANO_PROJECT_ROOT/.meltano/temp`. 

Beyond that I have no prefs on naming.  Sounds like MELTANO_SYS_DIR_ROOT is the defacto choice then ?

src/meltano/core/bundle/settings.yml
```diff
@@ -12,7 +12,7 @@ settings:
   # Replaces negated env_alias from `send_anonymous_usage_stats`
 - name: project_id
 - name: database_uri
-  value: sqlite:///$MELTANO_PROJECT_ROOT/.meltano/meltano.db
+  value: sqlite:///$DOT_MELTANO_ROOT/.meltano/meltano.db
```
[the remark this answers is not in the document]
rawwar: yes. its supposed to be without actual .meltano string

rawwar:
The only reason I raised it as a Draft PR was that I _KNEW_  that variable name was bad :-D . I'll use `MELTANO_SYS_DIR_ROOT` and update the code.  Also, do I include it as a `$MELTANO_SYS_DIR_ROOT/.meltano` or consider `$MELTANO_SYS_DIR_ROOT` as a replacement to `.meltano`?

pandemicsyn:
> The only reason I raised it as a Draft PR was that I KNEW that variable name was bad 😄 . I'll use MELTANO_SYS_DIR_ROOT and update the code. Also, do I include it as a $MELTANO_SYS_DIR_ROOT/.meltano or consider $MELTANO_SYS_DIR_ROOT as a replacement to .meltano?

My vote would be to consider  $MELTANO_SYS_DIR_ROOT as a replacement to .meltano. So that the default is implied to be the equivalent of:

`MELTANO_SYS_DIR_ROOT=$MELTANO_PROJECT_ROOT/.meltano`

but you could override it too:

`MELTANO_SYS_DIR_ROOT=/some/custom/.meltano`
`MELTANO_SYS_DIR_ROOT=/mnt/persistent0`
...

@aaronsteers @tayloramurphy curious though if y'all where thinking about it the same way.

edgarrmondragon (changes requested):

aaronsteers:
> My vote would be to consider $MELTANO_SYS_DIR_ROOT as a replacement to .meltano. So that the default is implied to be the equivalent of:\n\nMELTANO_SYS_DIR_ROOT=$MELTANO_PROJECT_ROOT/.meltano\n\nbut you could override it too:\n\nMELTANO_SYS_DIR_ROOT=/some/custom/.meltano\nMELTANO_SYS_DIR_ROOT=/mnt/persistent0

👍👍

src/meltano/core/bundle/settings.yml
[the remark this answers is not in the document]
rawwar: `$DOT_MELTANO_ROOT` will include `.meltano` folder. made the change

rawwar:
I'm a bit stuck on how to correct `settings.yml`. 
If we want to use `$MELTANO_SYS_DIR_ROOT` as a replacement to `.meltano`,  what do I update the following to? 
```
- name: database_uri
  value: sqlite:///$MELTANO_PROJECT_ROOT/.meltano/meltano.db
  env_specific: true
```
If `$MELTANO_SYS_DIR_ROOT` is defined, we no longer need `.meltano` and if it is not defined, `.meltano` needs to be included. I feel that I can update `db.py` to decide what to use based on the availability of the `$MELTANO_SYS_DIR_ROOT` env variable. If env variable doesn't exist, i go by the default value, else, i use the `$MELTANO_SYS_DIR_ROOT/.meltano.db`. But, We also have `$DATABASE_URI`, which will just override the default settings

src/meltano/core/bundle/settings.yml
edgarrmondragon, line 15:
```suggestion
  value: sqlite:///$DOT_MELTANO_ROOT/meltano.db
```

src/meltano/core/project.py
```diff
@@ -89,6 +90,7 @@ def __init__(self, root: Path | str):
             root: the root directory for the project
         """
         self.root = Path(root).resolve()
+        self.dot_meltano_root = Path(os.getenv(DOT_MELTANO_ROOT_ENV, root)).resolve()
```
edgarrmondragon, line 93:
```suggestion
        self.dot_meltano_root = Path(os.getenv(DOT_MELTANO_ROOT_ENV, root.joinpath(".meltano"))).resolve()
```

edgarrmondragon (changes requested):

edgarrmondragon:
> I'm a bit stuck on how to correct `settings.yml`. If we want to use `$MELTANO_SYS_DIR_ROOT` as a replacement to `.meltano`, what do I update the following to?
> 
> ```
> - name: database_uri
>   value: sqlite:///$MELTANO_PROJECT_ROOT/.meltano/meltano.db
>   env_specific: true
> ```
> 
> If `$MELTANO_SYS_DIR_ROOT` is defined, we no longer need `.meltano` and if it is not defined, `.meltano` needs to be included. I feel that I can update `db.py` to decide what to use based on the availability of the `$MELTANO_SYS_DIR_ROOT` env variable. If env variable doesn't exist, i go by the default value, else, i use the `$MELTANO_SYS_DIR_ROOT/.meltano.db`. But, We also have `$DATABASE_URI`, which will just override the default settings

@rawwar with the changes I suggested, the default dot meltano dir would live in `<project root>/.meltano`, so the default sqlite db would still live in `sqlite:///<project root>/.meltano/meltano.db`

rawwar:
> @rawwar with the changes I suggested, the default dot meltano dir would live in `<project root>/.meltano`, so the default sqlite db would still live in `sqlite:///<project root>/.meltano/meltano.db`

But, From the above discussion, we want to consider `DOT_MELTANO_ROOT`(AKA `$MELTANO_SYS_DIR_ROOT`) will be representing `.meltano` folder itself. So,  If I do keep things as it is in `settings.yml`, I need to update in db.py to consider `$MELTANO_SYS_DIR_ROOT/meltano.db` as the db uri.

pandemicsyn:
> > @rawwar with the changes I suggested, the default dot meltano dir would live in `<project root>/.meltano`, so the default sqlite db would still live in `sqlite:///<project root>/.meltano/meltano.db`
> 
> But, From the above discussion, we want to consider `DOT_MELTANO_ROOT`(AKA `$MELTANO_SYS_DIR_ROOT`) will be representing `.meltano` folder itself. So, If I do keep things as it is in `settings.yml`, I need to update in db.py to consider `$MELTANO_SYS_DIR_ROOT/meltano.db` as the db uri.

@rawwar @edgarrmondragon yea wouldn't the default entry just become:

```
- name: database_uri
  value: sqlite:///$MELTANO_SYS_DIR_ROOT/meltano.db
  env_specific: true
```

Which by default then resolves $MELTANO_SYS_DIR_ROOT to $MELTANO_PROJECT_ROOT/.meltano ?

rawwar:
Correct me If I am wrong. 
so, if we keep default as the following 
```
- name: database_uri
  value: sqlite:///$MELTANO_SYS_DIR_ROOT/meltano.db
  env_specific: true
```
when user sets `$MELTANO_PROJECT_ROOT` to `/home/my_project` and does not set `$MELTANO_SYS_DIR_ROOT`. default value is evaluated to `sqlite:////meltano.db` . This would mean that meltano.db is created in the folder where user is running the command. 
One way I can think of is updating `env` property of `Project` class to include `MELTANO_SYS_DIR_ROOT`. So, i updated it as follows.

```
@property
    def env(self):
        """Get environment variables for this project.

        Returns:
            dict of environment variables and values for this project.
        """
        environment_name = (
            self.active_environment.name if self.active_environment else ""
        )
        return {
            PROJECT_ROOT_ENV: str(self.root),
            PROJECT_ENVIRONMENT_ENV: environment_name,
            PROJECT_SYS_DIR_ROOT: str(self.sys_dir_root)
        }
```

edgarrmondragon:
> One way I can think of is updating `env` property of `Project` class to include `MELTANO_SYS_DIR_ROOT`. So, i updated it as follows.

@rawwar Yes, you're right! I think that's the correct approach.

rawwar:
I just realized while understanding how to write tests for #6627 that I need to write tests for this as well. I might take a while for writing tests.

edgarrmondragon (approved):
Changes look good 👌. I'll defer to @tayloramurphy if he'd like some docs for the new feature to be added.

tayloramurphy:
@edgarrmondragon @rawwar  it looks like `MELTANO_PROJECT_ROOT` is mentioned in a few places. Would we want to add this new env var to https://docs.meltano.com/concepts/environments#the-env-mapping as well as updating https://docs.meltano.com/reference/settings#database_uri as appropriate.

edgarrmondragon:
> Would we want to add this new env var to https://docs.meltano.com/concepts/environments#the-env-mapping as well as updating https://docs.meltano.com/reference/settings#database_uri as appropriate.

@tayloramurphy @rawwar yeah, I'd agree with those two and add https://docs.meltano.com/concepts/project#meltano-directory.

tayloramurphy (changes requested):
Requesting additions for the docs.

rawwar:
> Requesting additions for the docs.

Sure. I'll update the documentation today.

rawwar:
> https://docs.meltano.com/concepts/project#meltano-directory.
@edgarrmondragon , wondering if I should update all references to `.meltano` directory to `$MELTANO_SYS_DIR_ROOT` or just mention in about `.meltano` path can be overridden using `$MELTANO_SYS_DIR_ROOT`

rawwar:
While updating documentation,  I asked myself few things.  
 if user selects a temporary path(probably non-persistent) for `$MELTANO_SYS_DIR_ROOT` , he can no longer user the project properly as `meltano.db` will unavailable. Worse, if user forgets which directory he used for a specific project.  Or any of the possible scenario where all the data in `$MELTANO_SYS_DIR_ROOT` is not available after a while.
I tried thinking and I could only think of possibly having/updating `$MELTANO_SYS_DIR_ROOT` as part of `meltano.yml` itself.  I am still trying to think if this is a valid concern. So, mentioning it here.

docs/src/_concepts/project.md
```diff
@@ -324,6 +324,8 @@ In a newly initialized project, this file will be included in [`.gitignore`](#gi
 
 Meltano stores various files for internal use inside a `.meltano` directory inside your project.
 
+**Note**: `$MELTANO_SYS_DIR_ROOT` can be used as a replacement to `.meltano` directory.
```
tayloramurphy, line 327:
```suggestion
**Note**: `$MELTANO_SYS_DIR_ROOT` can be used as a replacement to `$MELTANO_PROJECT_ROOT/.meltano` directory.
```

Making it clear that's it's not _just_ the `/.meltano` directory.

tayloramurphy (approved):

edgarrmondragon:
> hile updating documentation, I asked myself few things.
> if user selects a temporary path(probably non-persistent) for `$MELTANO_SYS_DIR_ROOT` , he can no longer user the project properly as `meltano.db` will unavailable. Worse, if user forgets which directory he used for a specific project. Or any of the possible scenario where all the data in `$MELTANO_SYS_DIR_ROOT` is not available after a while.
> I tried thinking and I could only think of possibly having/updating `$MELTANO_SYS_DIR_ROOT` as part of `meltano.yml` itself. I am still trying to think if this is a valid concern. So, mentioning it here.

@rawwar afaik there's no good way to check if a directory is temporary, so the risk of users creating resources that will be nuked by the OS or another process is always present 🤷.

You're right that it's probably a good idea to make the path string persistent somewhere. There's a number of ways to make Meltano source the value from `meltano.yml`:

1. Make it a project setting, i.e. adding it to `src/meltano/core/bundle/settings.yml`. This makes the setting configurable with the CLI:

   ```console
   $ meltano config meltano sys_dir_root /path/to/meltano/sys
   ```

2. Adding it as an attribute to `meltano.core.meltano_file.MeltanoFile`. This is the simplest approach and probably good enough.

Wdyt?

rawwar:
>  Adding it as an attribute to `meltano.core.meltano_file.MeltanoFile`. This is the simplest approach and probably good enough.
I'll go with this implementation. Thank you

docs/src/_concepts/project.md
```diff
@@ -336,6 +338,7 @@ While you would usually not want to modify files in this directory directly, kno
 - `.meltano/run/<plugin name>/`, e.g. `.meltano/run/tap-gitlab/`: Directory used by [`meltano invoke`](/reference/command-line-interface#invoke) to store generated plugin config files.
 - `.meltano/<plugin type>/<plugin name>/venv/`, e.g. `.meltano/extractors/tap-gitlab/venv/`: [Python virtual environment](https://docs.python.org/3/glossary.html#term-virtual-environment) directory that a plugin's [pip package](https://pip.pypa.io/en/stable/) was installed into by [`meltano add`](/reference/command-line-interface#add) or [`meltano install`](/reference/command-line-interface#install).
 
+if `$MELTANO_SYS_DIR_ROOT` is set, all the above mentioned paths `.meltano/*` will point to `$MELTANO_SYS_DIR_ROOT/*`.
```
tayloramurphy, line 341:
```suggestion
If `$MELTANO_SYS_DIR_ROOT` is set, all the above mentioned paths `.meltano/*` will point to `$MELTANO_SYS_DIR_ROOT/*`.

```

tayloramurphy (approved):
One suggested change, but otherwise approving from a docs perspective.

aaronsteers (approved):
✅ 👍

edgarrmondragon:
> This would also mean, we need to decide which takes precedence over others.

@rawwar I think it makes sense for the `$MELTANO_SYS_DIR_ROOT ` to take precedence if it's present in the environment. Wdyt?

rawwar:
> @rawwar I think it makes sense for the `$MELTANO_SYS_DIR_ROOT ` to take precedence if it's present in the environment. Wdyt?

I have been trying to implement this. Was having trouble doing it properly. 
Issue is that, my current implementation is  failing for project `init` and working for other commands. Since, during init, there won't be `meltano.yml`, but, my code was trying to read it .. is causing issues. trying to handle this case

aaronsteers:
> > This would also mean, we need to decide which takes precedence over others.
> 
> @rawwar I think it makes sense for the `$MELTANO_SYS_DIR_ROOT ` to take precedence if it's present in the environment. Wdyt?

As of now, I don't think we have a formal dev standard for this. But the above is consistent with what we might adopt as a standard if we take this as reference:

From https://clig.dev/#configuration:

> ### Apply configuration parameters in order of precedence. Here is the precedence for config parameters, from highest to lowest:
> 
> - Flags
> - The running shell’s environment variables
> - Project-level configuration (eg. `.env`)
> - User-level configuration
> - System wide configuration

cc @pandemicsyn, who is taking lead on CLI dev standards.

pandemicsyn:
> As of now, I don't think we have a formal dev standard for this. But the above is consistent with what we might adopt as a standard if we take this as reference:

+1 we've not codified anything as far a control layers but pretty sure when it comes to a hierarchy what @aaronsteers referenced will almost certainly be our default stance. 

> Since, during init, there won't be meltano.yml, but, my code was trying to read it .. is causing issues. trying to handle this case.

@rawwar did you manage to work around this, or is this still a sticking point for ya?

rawwar:
> > As of now, I don't think we have a formal dev standard for this. But the above is consistent with what we might adopt as a standard if we take this as reference:
> 
> +1 we've not codified anything as far a control layers but pretty sure when it comes to a hierarchy what @aaronsteers referenced will almost certainly be our default stance.
> 
> > Since, during init, there won't be meltano.yml, but, my code was trying to read it .. is causing issues. trying to handle this case.
> 
> @rawwar did you manage to work around this, or is this still a sticking point for ya?

I did try a week ago. But, I could not. I was stuck at it. I'll be looking into it today. I'll give an update here once I'm done.

sbalnojan:
Hey @rawwar, do you need need help to push this over the finish line?

rawwar:
> Hey @rawwar, do you need need help to push this over the finish line?

I tried to resolve the issue since weekend and I am unable to. Would appreciate if I can get some help with this

edgarrmondragon:
@rawwar I'll take a look

edgarrmondragon (approved):
@rawwar ok, I got the merge conflicts solved after a few attempts 😅.

This is looking good. Pinging @aaronsteers in case I miss anything.

src/meltano/core/bundle/settings.yml
```diff
@@ -12,7 +12,7 @@ settings:
   # Replaces negated env_alias from `send_anonymous_usage_stats`
 - name: project_id
 - name: database_uri
-  value: sqlite:///$MELTANO_PROJECT_ROOT/.meltano/meltano.db
+  value: sqlite:///$MELTANO_SYS_DIR_ROOT/meltano.db
```
aaronsteers, line 15: Can someone confirm this works in manual testing? 

If so, would lean towards disproving the issue described here:

- #6920

aaronsteers (approved):
@edgarrmondragon - Looks good from my side. ~~Just one question here: https://github.com/meltano/meltano/pull/6628/files#r1000023143~~ [Resolved!]

If all other issues/questions resolved, do you want to do the honors with a merge to `main`?

src/meltano/core/bundle/settings.yml
edgarrmondragon: @aaronsteers Yeah, I can confirm that the interpolation works fine in this use case:

```console
$ export MELTANO_SYS_DIR_ROOT=$HOME/.meltano-data

$ meltano init new-project
...

$ ls -a | grep "^\."

.
..
.gitignore

$ tree $MELTANO_SYS_DIR_ROOT
/Users/edgarramirez/.meltano-data
├── analytics.json
├── meltano.db
├── meltano.db-shm
├── meltano.db-wal
└── run
    ├── bin -> /Users/edgarramirez/Library/Caches/pypoetry/virtualenvs/meltano-SY7IjDqw-py3.10/bin/meltano
    └── meltano.yml.lock

$ meltano config meltano list | grep "DATABASE_URI"
database_uri [env: MELTANO_DATABASE_URI] current value: 'sqlite:////Users/edgarramirez/.meltano-data/meltano.db' (default: 'sqlite:///$MELTANO_SYS_DIR_ROOT/meltano.db')
```

edgarrmondragon: @aaronsteers Is there any other check that comes to mind?

aaronsteers: @edgarrmondragon - No, that's great. Thanks for confirming. 👍

rawwar:
@edgarrmondragon .. Thank you so much!

PR #6644, opened 2022-08-18
perf: Improve performance when handling large yaml files

### LINKED ISSUE #6597: bug: Meltano is slow to start
There appears to be a bug causing some users to have a slow startup, up to 45 seconds in the provided logs

Pre 2.x:

```
(.venv) ➜  dutchie-dagster-meltano git:(NoJIRA-upgrade-dagster-meltano) ✗ meltano --version
meltano, version 1.100.0
(.venv) ➜  dutchie-dagster-meltano git:(NoJIRA-upgrade-dagster-meltano) ✗ time meltano --log-level debug config tap-exchange-rates-api list
2022-08-08T22:33:35.566057Z [debug    ] Creating engine <meltano.core.project.Project object at 0x10e248970>@sqlite:////Users/dwall/repos/dagster/dutchie-dagster-meltano/.meltano/meltano.db
2022-08-08T22:33:35.815951Z [debug    ] Using selector: KqueueSelector
2022-08-08T22:33:35.848522Z [debug    ] Created configuration at /Users/dwall/repos/dagster/dutchie-dagster-meltano/.meltano/run/tap-exchange-rates-api/tap.e8b67d7e-f096-4197-b805-e51bcf5971a8.config.json
2022-08-08T22:33:35.848773Z [debug    ] Could not find tap.properties.json in /Users/dwall/repos/dagster/dutchie-dagster-meltano/.meltano/extractors/tap-exchange-rates-api/tap.properties.json, skipping.
2022-08-08T22:33:35.848938Z [debug    ] Could not find tap.properties.cache_key in /Users/dwall/repos/dagster/dutchie-dagster-meltano/.meltano/extractors/tap-exchange-rates-api/tap.properties.cache_key, skipping.
2022-08-08T22:33:35.849094Z [debug    ] Could not find state.json in /Users/dwall/repos/dagster/dutchie-dagster-meltano/.meltano/extractors/tap-exchange-rates-api/state.json, skipping.

Custom, possibly unsupported by the plugin:
base [env: TAP_EXCHANGE_RATES_API_BASE] current value: 'USD' (from `meltano.yml`)
start_date [env: TAP_EXCHANGE_RATES_API_START_DATE] current value: '2018-01-01' (from `meltano.yml`)
symbols [env: TAP_EXCHANGE_RATES_API_SYMBOLS] current value: ['CAD', 'MXN', 'AUD', 'EUR', 'USD'] (from `meltano.yml`)
meltano --log-level debug config tap-exchange-rates-api list  1.64s user 0.20s system 99% cpu 1.849 total
```

Vs in 2.4.0:

```
(.venv) ➜  dutchie-dagster-meltano git:(NoJIRA-upgrade-dagster-meltano) ✗ meltano --version
meltano, version 2.4.0
(.venv) ➜  dutchie-dagster-meltano git:(NoJIRA-upgrade-dagster-meltano) ✗ date && time meltano --log-level debug config tap-exchange-rates-api list && date
Mon Aug  8 15:46:01 PDT 2022
2022-08-08T22:46:45.104438Z [debug    ] Creating engine <meltano.core.project.Project object at 0x10c521760>@sqlite:////Users/dwall/repos/dagster/dutchie-dagster-meltano/.meltano/meltano.db
2022-08-08T22:46:45.406143Z [debug    ] loaded lazy attr 'SafeConfigParser': <class 'configparser.ConfigParser'>
2022-08-08T22:46:45.406251Z [debug    ] loaded lazy attr 'NativeStringIO': <class '_io.StringIO'>
2022-08-08T22:46:45.406330Z [debug    ] loaded lazy attr 'BytesIO': <class '_io.BytesIO'>

Custom, possibly unsupported by the plugin:
base [env: TAP_EXCHANGE_RATES_API_BASE] current value: 'USD' (from `meltano.yml`)
start_date [env: TAP_EXCHANGE_RATES_API_START_DATE] current value: '2018-01-01' (from `meltano.yml`)
symbols [env: TAP_EXCHANGE_RATES_API_SYMBOLS] current value: ['CAD', 'MXN', 'AUD', 'EUR', 'USD'] (from `meltano.yml`)
meltano --log-level debug config tap-exchange-rates-api list  76.80s user 0.50s system 97% cpu 1:18.94 total
Mon Aug  8 15:47:20 PDT 2022
```

More detail in Slack thread: https://meltano.slack.com/archives/C01TCRBBJD7/p1659998566724249?thread_ts=1659985256.778899&cid=C01TCRBBJD7

WillDaSilva:
This PR pragmatically addresses #6597 by implementing caching at 3 levels, all in an effort to reduce the number of times ruamel collections (e.g. `CommentedMap`, `CommentedSeq`, etc.) are deep-copied.

This PR does not entirely avoid the performance penalty of deep-copying large ruamel collections, and so even with these changes Meltano cannot handle very large YAML config files. Fortunately it seems like few users use very large config files - if there were more, we would have noticed the average execution time increase significantly, and we would have received additional bug reports.

The caching provided by this PR is likely sufficient for @dwallace0723 and @dehume, who brought this issue to our attention, and helped us fix it.

The 3 levels of caching introduced by this PR are:
- `yaml.load`, which uses an unbound cache keyed by the file path. On every call it will open and read the specified file, but it will only load it with ruamel if the content hash of the file differs from what is cached.
- `ProjectFiles.load`, which caches the loaded merged project files on the first call. On future calls, the cached value is used (and the expensive merge operation is skipped) if the `yaml.load` call for each of the files that would get merged together returns the same unchanged object.
- `Canonical.parse`, which uses a bound least-recently-used cache with 4096 slots keyed by the Python object ID. That is to say, if the exact same Python object is provided to `Canonical.parse`, it returns the cached result.

No one or two levels of caching here yields sufficient performance improvements. All 3 are required to work together: The caching for `Canonical.parse` relies on the object returned from `ProjectFiles.load` being identical to previous calls, and the caching for `ProjectFiles.load` relies on the object returned from `yaml.load` being identical to previous calls. Because of this dependency chain, if any of the underlying YAML files are changed, all 3 caches are invalidated automatically. Manual cache invalidation can also be performed.

Closes #6597


WillDaSilva:
The current performance gains are significant, but we still haven't escaped from the terrible time complexity of `CommentedMap.__deepcopy__`. Whereas before some large yaml files (e.g. 1000 lines long) would result in Meltano taking several minutes to perform a simple `config <plugin name> list` command, now it only takes a couple seconds. Unfortunately with a Meltano file that is not much larger (e.g. 1500 lines long) we find ourselves back with an unacceptable performance in the minutes.

Caching `meltano.yml` and the config files it includes is nice and fairly easy, but we'll have to tackle this at its source to really fix it. That means avoiding or alerting the behaviour of `CommentedMap.__deepcopy__`. Maybe an easy approach will be to downcast most of our `CommentedMap` objects to `ordereddict`, which is one of its base classes. My guess is that its `deepcopy` performance is significantly faster - no worse than a `dict`.

WillDaSilva:
Without the changes in this PR, `meltano config tap-exchange-rates-api list` takes ~55 seconds on my computer using the `meltano.yml` file provided to me by Dennis Hume. With the changes, it takes ~1.7 seconds.

I think this provides us with sufficient reason to incorporate these changes, even though they don't fully address the problem. As mentioned previously, to fully address the problem (i.e. have reasonable performance even when the YAML files being handled are a couple thousand lines long) we need to avoid `CommentedMap.__deepcopy__`. I tried patching it to use `copy` instead of `deepcopy`, and that made the performance acceptable, but obviously breaks the functionality. A better approach I'm currently testing is to convert the "commented" collections (i.e. `CommentedMap`, `CommentedSeq`, and `CommentedSet`) to stdlib collections (i.e. `dict`, `list`, and `set` respectively).

WillDaSilva:
> A better approach I'm currently testing is to convert the "commented" collections (i.e. `CommentedMap`, `CommentedSeq`, and `CommentedSet`) to stdlib collections (i.e. `dict`, `list`, and `set` respectively).

This won't work. It totally solves the performance regression, but we can't retain the comment info at depth.

WillDaSilva:
One approach @edgarrmondragon and I considered was making `Project.meltano` a cached property, and only writing updates to the files at the end of the execution of the command. Making `Project.meltano` a cached property helps for the mid-sized YAML file case, but doesn't make dealing with large (e.g. 2k+ lines) YAML files.

The caching I have added in this PR already addresses the mid-sized YAML case, and is almost passing all tests. By contract, having `Project.meltano` as a cached property is causing a very large number of test failures. It'll probably be easier, at least in the short term, to just use the caching. Long-term we'll want to consider what kind of impact these decisions will have on running the Meltano core logic for the managed service.

I'm beginning to think we can't solve the terrible performance around large YAML files. It's something we'll have to accept if we want to preserve comments using ruamel. Fortunately such incredibly large YAML file are rare - the original issue is concerned with a mid-sized YAML file that caching can make fast enough.

src/meltano/core/plugin/command.py
```diff
@@ -80,14 +80,6 @@ def expanded_args(self, name, env):
...
         return expanded
 
-    def canonical(self):
-        """Serialize the command.
-
-        Returns:
-            Python object.
-        """
-        return Command.as_canonical(self)
-
```
WillDaSilva, line 90: This is no longer necessary because the base `Canonical` class now references class methods as `type(self).method` instead of `Canonical.method`.

src/meltano/core/yaml.py
```diff
@@ -2,26 +2,62 @@
...
 from meltano.core.setting_definition import SettingKind
+from meltano.core.utils import hash_sha256
+
+yaml = YAML()
+yaml.default_flow_style = False
+
+yaml.register_class(Canonical)
+yaml.register_class(PluginType)
+yaml.register_class(SettingKind)
+yaml.register_class(DiscoveryFile)
```
edgarrmondragon, line 23: Nice improvement 👍

src/meltano/core/project.py
```diff
@@ -210,44 +203,41 @@ def find(cls, project_root: Path | str = None, activate=True):
...
         """
-        if self._project_files is None:
-            self._project_files = ProjectFiles(
-                root=self.root, meltano_file_path=self.meltanofile
-            )
-        return self._project_files
+        return ProjectFiles(root=self.root, meltano_file_path=self.meltanofile)
+
+    def clear_cache(self) -> None:
+        """Clear cached project files (e.g. `meltano.yml`)."""
```
edgarrmondragon, line 216: Can you add some guidance in the docstring about in which circumstances or stages it's recommended to clear the project file cache?

edgarrmondragon (changes requested):
@WillDaSilva This was a tricky one. Only requests are to document when clearing the project file cache is advised 🙏

src/meltano/core/project.py
WillDaSilva: Good call. I added the following:

> This can be useful if the cached `ProjectFiles` object has been modified in-place, but not updated on-disk, and you need the on-disk version.

It's a pretty niche situation, but we have some tests that explicitly check it.

edgarrmondragon (approved):

PR #6646, opened 2022-08-18
Remove Meltano 2.0 banner on docs

afolson:


WillDaSilva (approved):
LGTM :+1:

afolson:
@DouweM @emilylucie I just hid the banner on the site for now since I don't think we have another CTA at the moment. I'd like to avoid having it there all the time so that we don't train people to ignore it, but let me know if there's something else you think we should put there.

tayloramurphy (approved):

PR #6647, opened 2022-08-18
ci: Use `actions/dependency-review-action` instead of fork

WillDaSilva:
We are currently using my personal fork of `actions/dependency-review-action` because it has the ability to run on `pull_request_target` events (and use custom base/head refs).

Now that those changed have been merged into the upstream repository, and released under the tag `v2.1.0`, we should use it instead of my fork.


edgarrmondragon (approved):

WillDaSilva:
Thanks for the title update @edgarrmondragon

Issue #6485, opened 2022-07-27
bug:  Regression in `meltano remove`


edgarrmondragon:
### Meltano Version

2.3.0

### Python Version

NA

### Bug scope

CLI (options, error messages, logging, etc.)

### Operating System

NA

### Description

Meltano `2.3.0` introduced a regression on `meltano remove` where it complains that the plugin is missing the `settings_with_extras` attribute.

The regression was most likely introduced in https://github.com/meltano/meltano/pull/6267 somewhere in the use of a feature flag:

https://github.com/meltano/meltano/blob/23a9099cefe148cb560d25e7ae335447858948df/src/meltano/core/plugin/settings_service.py#L63-L65

I still haven't figured out the root cause but might be related with trying to check for feature flags in the `PluginSettingsService`.

Other commands may be similarly affected, but at least `invoke` and `run` seem to be working fine.

### Code

<details><summary>Traceback</summary>

```
Traceback (most recent call last):
  File "/Users/edgarramirez/meltano/meltano/src/meltano/core/behavior/canonical.py", line 138, in __getattr__
    value = self._dict[attr]
KeyError: 'settings_with_extras'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/edgarramirez/meltano/meltano/src/meltano/cli/__init__.py", line 66, in _run_cli
    cli(obj={"project": None})
  File "/Users/edgarramirez/Library/Caches/pypoetry/virtualenvs/meltano-SY7IjDqw-py3.9/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/edgarramirez/meltano/meltano/src/meltano/cli/cli.py", line 35, in main
    return super().main(*args, windows_expand_args=False, **kwargs)
  File "/Users/edgarramirez/Library/Caches/pypoetry/virtualenvs/meltano-SY7IjDqw-py3.9/lib/python3.9/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/edgarramirez/meltano/meltano/src/meltano/cli/utils.py", line 540, in invoke
    super().invoke(ctx)  # noqa: WPS608
  File "/Users/edgarramirez/Library/Caches/pypoetry/virtualenvs/meltano-SY7IjDqw-py3.9/lib/python3.9/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
[...]

PR #6656, opened 2022-08-22
fix: Correct STRICT_ENV_VAR_MODE behavior

### LINKED ISSUE #6604: Bug: Env var strict mode not working as expected
While [testing](https://github.com/meltano/squared/pull/312#issuecomment-1191828551) within the `squared` project, and also confirmed by @visch [in Slack](https://meltano.slack.com/archives/CFG3C3C66/p1660078390407119?thread_ts=1659968745.750009&cid=CFG3C3C66), we've not been seeing the strict mode flag working as expected to raise errors.

It could be the case, in the multiple recent changes to this part of the codebase (`env` resolution, specifically) that we may have had a regression.

cjohnhanson:
The `strict_env_var_mode` feature flag hasn't been behaving correctly since the 2.3.0 release. This 
PR fixes the behavior and adds new functional tests in addition to the existing unit tests to catch 
such regressions moving 
forward.

Closes #6604


src/meltano/core/plugin_invoker.py
```diff
@@ -327,7 +327,7 @@ def env(self):
         project_settings_service = ProjectSettingsService(
             self.project, config_service=self.plugins_service.config_service
         )
-        with self.settings_service.feature_flag(
+        with project_settings_service.feature_flag(
```
edgarrmondragon, line 330: I think this might fix #6485 too 🤞

edgarrmondragon (approved):
LGTM!

pandemicsyn (approved):
🚀

src/meltano/core/plugin_invoker.py
cjohnhanson: 🤞

PR #6665, opened 2022-08-24
ci: Use `meltano/commitizen-version-bump`

WillDaSilva:
To customize how Commitizen performs our version bump, Meltano and the SDK will now use https://github.com/meltano/commitizen-version-bump. It will be installed and used by `commitizen-action` in the `version_bump` GitHub workflow.

Relates to https://github.com/meltano/gitops/issues/4

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,6 +77,7 @@ bumpversion = "^0.6.0"
 changelog-cli = "^0.6.2"
 colorama = "^0.4.4"
 coverage = {extras = ["toml"], version = "^6.4.1"}
+commitizen-version-bump = { git = "https://github.com/meltano/commitizen-version-bump.git", branch = "main" }
 freezegun = "^0.3.12"
 isort = "^5.10.1"
 mock = "^4.0.3"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -141,7 +141,7 @@ exclude_lines = [
 ]
 
 [tool.commitizen]
-name = "cz_customize"
+name = "cz_version_bump"
 version = "2.4.0"
 tag_format = "v$major.$minor.$patch$prerelease"
 version_files = [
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -152,25 +152,6 @@ version_files = [
...
-]
-
-[tool.commitizen.customize.change_type_map]
-break = "BREAKING CHANGES"
-feat = "✨ New"
-fix = "🐛 Fixes"
-refactor = "⚙️ Under the Hood"
-docs = "📚 Documentation Improvements"
-perf = "⚡ Performance Improvements"
-
```
aaronsteers, line 173: @WillDaSilva - For my info, how does this play with the changelog automation? Is this mapping moving to another file, for instance?

WillDaSilva: It has been moved to https://github.com/meltano/commitizen-version-bump/blob/main/cz_version_bump/__init__.py

WillDaSilva: Because of this, it applies to both Meltano and the SDK, and any other repository we want to version in the future (e.g. the EDK).

aaronsteers: @WillDaSilva - Thanks! Sorry, I totally missed that this was a new custom repo with our logic baked in. Very cool!

aaronsteers (commented):
Love the refactoring of putting the logic into a new repo that can be shared across other projects. 👍 

If @edgarrmondragon and/or @pandemicsyn can give thumbs up, I think this is clear to go - with follow-ups to be addressed within the https://github.com/meltano/commitizen-version-bump repo.

pandemicsyn (approved):
Considering i'm about to have to set up commitizen on 1-3 new repos, this is very timely and handy 😁 

lgtm.

aaronsteers (approved):

PR #6672, opened 2022-08-25
docs: Add tutorial on how to debug a custom extractor

jared-rimmer:
Adds a guide on how to setup debugging for a custom extractor / tap. 

I've just been through the process and had to look through the Slack thread and Stack Overflow so thought putting a tutorial together might be useful.

At the moment it is VSCode specific but it would be great to add some other IDE specific set ups.


docs/src/_tutorials/debugging-custom-extractor.md
```diff
@@ -0,0 +1,60 @@
...
+description: Learn how to debug a custom data extractor.
+layout: doc
+---
+
+## Add a main block in tap.py of your Custom Extractor
+
+```python
+if __name__ == "__main__":
+  # TapCustomExtractor is the class name of your tap in tap.py
+  TapCustomExtractor.cli()
```
edgarrmondragon, line 12:
```suggestion
    # TapCustomExtractor is the class name of your tap in tap.py
    TapCustomExtractor.cli()
```

docs/src/_tutorials/debugging-custom-extractor.md
```diff
@@ -0,0 +1,60 @@
...
+
+If you're using Poetry you need to create a venv in your custom tap directory so that your IDE can pick it up when debugging.
+
+If you have already installed poetry this will involve removing the existing poetry virtual environment. Running the command below and finally reinstalling poetry.
+
+```bash
+poetry config virtualenv.in-project true
+```
+
+## What to put in VSCode launch.json
```
edgarrmondragon, line 25:
```suggestion
## What to put in VSCode launch.json

Add the following [launch configuration](https://code.visualstudio.com/docs/editor/debugging#_launch-configurations) to your project:
```

edgarrmondragon (changes requested):
this is great @jared-rimmer! Do you think this might make more sense as a "how-to" guide? i.e. put in under `docs/src/_guide/` instead

jared-rimmer:
Thanks for the feedback @edgarrmondragon. I've committed your suggested changes and moved the location from tutorial to guides.

docs/src/_guide/debugging-custom-extractor.md
```diff
@@ -0,0 +1,62 @@
+---
+title: Debug a Custom Extractor
+description: Learn how to debug a custom data extractor.
+layout: doc
```
edgarrmondragon, line 4: @jared-rimmer let's add a weight so it renders after `Logging and Monitoring` in the left nav.

```suggestion
layout: doc
weight: 25
```

edgarrmondragon (changes requested):

edgarrmondragon (approved):

PR #6679, opened 2022-08-27
Delete ELT.md

lauraperezco:
Per https://github.com/meltano/internal-marketing/issues/494 we're deprecating the 'ELT' page in docs since it's competing with our "ELT vs ETL" blog and ELT pages (where we would rather have that traffic, instead of the docs).


tayloramurphy (changes requested):
@lauraperezco @afolson we should setup a redirect so that that folks finding this page will be sent to the blog post. I believe the way to do that is to keep the page but just have the following in the front matter:

```yaml
title: "ELT vs ETL"
redirect_to: https://meltano.com/blog/the-benefits-of-elt-vs-etl-what-you-need-to-know/
```

Based on https://github.com/jekyll/jekyll-redirect-from#redirect-to

lauraperezco:
> @lauraperezco @afolson we should setup a redirect so that that folks finding this page will be sent to the blog post. I believe the way to do that is to keep the page but just have the following in the front matter:
> 
> ```yaml
> title: "ELT vs ETL"
> redirect_to: https://meltano.com/blog/the-benefits-of-elt-vs-etl-what-you-need-to-know/
> ```
> 
> Based on https://github.com/jekyll/jekyll-redirect-from#redirect-to

@tayloramurphy From the link you sent it says that's a plugin. Is that something we have already installed or is it to be installed with this change? and also, sorry Im not sure I got where to paste the redirect from your instructions - where's the "front matter"?🥹

tayloramurphy:
@lauraperezco sorry, it's already installed and we use it on several pages already. 

The front matter is the stuff between the `---` at the top of the markdown file https://github.com/meltano/meltano/blob/b4fa94819e0b4e8dae94b4ba03538e1342c36789/docs/src/_concepts/elt.md?plain=1#L2-L5

If you replace all of that (but keep both of the `---` sections) I think it should work. You can also delete the rest of the content in the markdown file, but not the page, if that makes sense.

PR #6681, opened 2022-08-29
Redirect ELT to blog

lauraperezco:
- Following Taylor's instructions from #6679 to make the ELT page in docs redirect to our blog post, instead of deleting it.
- Why we're doing this (SEO updates): https://github.com/meltano/internal-marketing/issues/494


WillDaSilva (approved):

docs/src/_concepts/elt.md
```diff
@@ -1,78 +1,6 @@
 ---
 title: "ELT vs ETL"
-description: "ELT Meaning, Advantages & More: All You Need To Know To Get Started"
-layout: doc
-weight: 11
+redirect_to: https://meltano.com/blog/etl-vs-elt-which-is-better/
```
tayloramurphy, line 3:
```suggestion
redirect_to: https://meltano.com/blog/etl-vs-elt-which-is-better/
hidden: true
```

tayloramurphy (approved):
I forgot that we needed to add `hidden: true`. I went ahead and did that and the redirect works!

PR #6692, opened 2022-08-31
chore: Update `poetry.lock` to Poetry 1.2 format

WillDaSilva:
Relates to https://github.com/meltano/meltano/pull/6691

From now on we'll all need to use Poetry 1.2+ if we want to push an update for the lock file.

The new lockfile format is backwards compatible. We just don't want to produce a bunch of messy diffs.


edgarrmondragon (approved):
🏎️

Issue #6157, opened 2022-06-10
Run Pytest tests in parallel


visch:
https://meltano.slack.com/archives/CKHP6G5V4/p1653508630941889

Looks like pytest doesn't take 45 minutes, not sure why I though it was taking that long before? Maybe on gitlab CI updated the title to reflect that it takes ~5-10 minutes to run pytest today. 

1. Be sure https://github.com/meltano/meltano/pull/6155/files#diff-65c008e1065903299dfa0da25e1f18d0f87f387867ccbbf2b0888eb138bdb255L242 is implemented (This uses a seperate folder for each Test in your temp directory instead of the same folder over and over again)
1. `poetry add -D pytest-xdist==1.*`
1. `poetry run pytest -n 20` 
1. Watch tests run extremely quickly

There's still a large number of failures here as some tests depend on the other I believe? Not exactly sure what causes the failures, but it's clear they weren't all designed with parallel testing in mind.  The gain is big enough it may be worth the time to figure out what causes the failures.

edgarrmondragon:
@visch some tests seem to rely on running after (or not!) other tests. See https://github.com/meltano/meltano/pull/6135#issue-1266509859. It's a bit crazy 😅

WillDaSilva:
This should be possible once #6300 is complete. Because #6300 introduces test ordering at the module level via `pytest-order`, we'll need to use the `pytest-xdist` option `--dist=loadfile` as explained here: https://pytest-dev.github.io/pytest-order/stable/other_plugins.html#usage-with-pytest-xdist

WillDaSilva:
We may want to upgrade our CI runners. They currently have 2 cores, but with more our CI pipelines would be far faster once we're running the Python tests in parallel. GitHub charges linearly per-core for their CI runners, so as an example, if we want to use runners with 16 cores, it would cost us 8 times as much to execute the Python tests per CI minute, but my guess is that the tests would probably execute ~6 times faster. Seems like a worthwhile tradeoff.

https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners

WillDaSilva:
Good timing on the public beta for large GitHub runners: https://github.blog/changelog/2022-09-01-github-actions-larger-runners-are-now-in-public-beta/

I've added Meltano to the waitlist to be permitted to use them.

[...]

PR #6695, opened 2022-08-31
test: Randomize Pytest runs

### LINKED ISSUE #6300: Use `pytest-randomly` plugin

WillDaSilva:
This PR adds a dependency on `pytest-randomly`, which seeds randomness in our test runs, and randomizes the order the tests are run in.

The randomness introduced by this plugin can be replicated using the `--randomly-seed` argument that is printed out for each Pytest run.

Running the tests in a random order helps to ensure they do not depend on tests that were run previously. This generally makes the test suite more robust, and also will allow us to run the tests in parallel using `pytest-xdist` - see https://github.com/meltano/meltano/issues/6157

Most of the pre-existing implicit dependencies between tests have been addressed by making them explicit with `order` marks provided by the `pytest-order` plugin. This is not the cleanest approach (it would be better if there were no dependencies between our tests), but it should be sufficient to realize most of the benefits, and is much faster than rewriting the tests to not have dependencies on other tests.

<details>
  <summary>Known issues: (EDIT: none remaining)</summary>

- [x] `tests/meltano/cli/test_config.py::TestCliConfig::test_config_meltano` occasionally fails on Windows.
- [x] `tests/meltano/cli/test_state.py::TestCliState::test_list` and `tests/meltano/cli/test_state.py::TestCliState::test_list_pattern` fail occasionally on all platforms. This occurs because other tests that may be run before will create unexpected state IDs. These state IDs cannot be cleared through the CLI because the database is locked because of the `state_service` fixture.
- [x] `psycopg2.OperationalError: connection to server at "localhost" (::1), port 5432 failed: FATAL:  sorry, too many clients already` occurs occasionally when using the Postgres test backend.
  - Increasing the SQLAlchemy pool size in the `create_engine` call from 20 to 512 did not fix the problem. 
  - I have ~~not yet~~ been able to reproduce this failure locally. Using the same `pytest-randomly` seed as the failing tests in CI does not seem to have had an effect. Using `taskset -c 0` to limit Pytest to a single processor core enabled the reproduction of the issue inconsistently.
  - Looks like we've got a Heisenbug. It does not occur when running Pytest with the `--pdb` flag.
  - It occurs during fixture teardown, but I don't know which fixture. Wrapping the teardown of each suspicious fixture with a try-except block that drops into the debugger or prints the current stack might help.
  - Looks like this problem stems from using connection pooling with Postgres, which can be avoided by using the `NullPool`. If we actually need connection pooling then this is obviously unacceptable, but it doesn't seem like we need it.
- [x] `tests/meltano/api/test_workers.py::TestUIAvailableWorker::test_open_browser` fails intermittently. Example: https://github.com/meltano/meltano/runs/8146232589

</details>

Closes #6300


src/meltano/core/db.py
```diff
@@ -41,8 +42,9 @@ def project_engine(
     settings = ProjectSettingsService(project)
 
     engine_uri = settings.get("database_uri")
-    logging.debug(f"Creating engine {project}@{engine_uri}")
-    engine = create_engine(engine_uri, pool_pre_ping=True)
+    logging.debug(f"Creating engine '{project}@{engine_uri}'")
+
+    engine = create_engine(engine_uri, poolclass=NullPool)
```
WillDaSilva, line 47: I want to call attention to this change. Using a `NullPool` seems fine for Meltano, since we don't need connection pooling to the best of my knowledge.

Using connection pooling was causing some tricky issues with the tests. It was introduced when we added support for Postgres as the DB backend. If we don't have a good reason for using connection pooling, I'm inclined to not use it (i.e. use `NullPool`) to simplify it, and in the process, resolve this Heisenbug that I've been dealing with for the past 48 hours

kgpayne: I think this is ok as i) connection pooling is an optimisation for applications with 100's or 1000's of users spawning connections, which our app isn't and ii) because we aren't going to be using Postgres (or any RDBMS) for much longer 🙂

kgpayne (approved):
LGTM 🚢 I am not surprised we have so many tests that require ordering. Looks like 17 was the longest chain 😅

tests/meltano/cli/test_cli.py
```diff
@@ -65,6 +68,7 @@ def test_activate_project_readonly_env(
 
         assert Project._default.readonly
 
+    @pytest.mark.order(2)
```
edgarrmondragon, line 71: Should this be `3`?

```suggestion
    @pytest.mark.order(3)
```

or does it mean the two tests with `2` can run in any order?

tests/meltano/core/plugin/test_plugin_settings.py
```diff
@@ -919,12 +924,14 @@ def test_find_setting_aliases(self, tap, plugin_settings_service_factory):
...
         assert subject.get("aliased") == "value_3"
 
+    @pytest.mark.order(-1)
     def test_strict_env_var_mode_on_raises_error(self, subject):
         subject.set([FEATURE_FLAG_PREFIX, str(FeatureFlags.STRICT_ENV_VAR_MODE)], True)
         subject.set("stacked_env_var", "${NONEXISTENT_ENV_VAR}")
         with pytest.raises(EnvironmentVariableNotSetError):
             subject.get("stacked_env_var")
 
+    @pytest.mark.order(-1)
```
edgarrmondragon, line 934: Ditto duplicated order indexes

tests/meltano/cli/test_cli.py
WillDaSilva: Tests with a particular number can run in any order relative to other tests with that same number. The orders are scoped per-module.

tests/meltano/core/plugin/test_plugin_settings.py
WillDaSilva: Same as with the other duplicate, it's intentional.

edgarrmondragon (approved):

WillDaSilva:
Extra test run for good measure: https://github.com/meltano/meltano/actions/runs/2982188929

PR #6707, opened 2022-09-02
ci: Enable static type checking with mypy

### LINKED ISSUE #6491: MyPy is not being run in CI
The style guide in the contributor docs says:

> MyPy is currently only executed as part of the build pipeline in order to avoid overwhelming developers with the complete list of violations. This allows for incremental and iterative improvement without requiring a concerted effort to fix all errors at once.

We are not currently running MyPy at all. We should immediately update the contributor doc to reflect that. Later we should update our GitHub CI to run MyPy, and work towards getting MyPy passing.

edgarrmondragon:
Relates to https://github.com/meltano/meltano/issues/6491

Example:

- https://github.com/meltano/meltano/pull/6708


pyproject.toml
```diff
@@ -155,6 +155,84 @@ version_files = [
...
+  "meltano.core.settings_store",
+  "meltano.core.sqlalchemy",
+  "meltano.core.state_service",
+  "meltano.core.task_sets",
+  "meltano.core.task_sets_service",
+  "meltano.core.transform_add_service",
+  "meltano.core.upgrade_service",
+  "meltano.core.validation_service",
+  "meltano.core.venv_service",
+  "meltano.core.yaml",
```
edgarrmondragon, line 216: This probably calls for some refactoring to put some of these in dedicated submodules.

WillDaSilva (approved):

PR #6709, opened 2022-09-06
docs: Add DataHub HowTo

sbalnojan:
Added a how for the data hub utility.


tayloramurphy:
@sbalnojan I like the content of this, but I'm not sure it's right for the Meltano Docs. I think this either needs to be a blog post or a Tutorial. Having it in the Guide section puts an overemphasis on DataHub in particular when compared to the ELT functionality. 

Some of the could also live in https://hub.meltano.com/utilities/datahub but likely we should link from the Hub to where this doc winds up living. 

@afolson your thoughts on this?

sbalnojan:
@tayloramurphy mh you think? I mean I agree, we shouldn't overemphasize one plugin, but that just means, we definitely need way more How Tos and not just this one.

DataHub is a great complement to ELT (as is Great Expectations), and I think it would be really helpful to have all the information in one place (the docs). 

Do you really want to redirect someone to some other place to answer the simple "how do I use DataHub with Meltano" question?


But, I would like to propose a restructuring anyways, putting this and 1-2 more pieces into a dedicated "How To" section (specific answers to specific problems while I'm working it). And restructure the "Guides" section into a Tutorial + Getting Started Section. I just thought that belongs into a separate PR.

tayloramurphy:
@sbalnojan fair points. I do think the overall information architecture of the docs could take another iteration. My main point is that I don't think it's right for the _current_ guide section and it seems like "Tutorials" would be better. 

> But, I would like to propose a restructuring anyways, putting this and 1-2 more pieces into a dedicated "How To" section (specific answers to specific problems while I'm working it). And restructure the "Guides" section into a Tutorial + Getting Started Section. I just thought that belongs into a separate PR.

I like that and I wouldn't be opposed to having that PR come in first, but that's primarily because of the pain of having to redirect if/when the URL changes. 

My recommendation would either to update this PR to move the  new content to Tutorials and then have the new PR update the structure or vice versa and wait to get this merged.

sbalnojan:
@tayloramurphy I forgot to tag you @tayloramurphy and I feel like all conversations have been resolved.

tayloramurphy (approved):
Great iteration - let's ship it and add more! 😄

PR #6710, opened 2022-09-06
test: Run Pytest tests in parallel

### LINKED ISSUE #6157: Run Pytest tests in parallel

WillDaSilva:
Comparing https://github.com/meltano/meltano/actions/runs/3001096204 and https://github.com/meltano/meltano/actions/runs/2998845253, we see an average time saving of ~476 seconds (~8 minutes) for each test run.

This time saving should become even greater once we gain access to the large GitHub CI runners:
- https://github.com/meltano/meltano/issues/6157#issuecomment-1234665437
- https://github.com/meltano/meltano/issues/6157#issuecomment-1236048249

I'm not sure how long it'll take for GitHub to give us access. All I know is that we're on the public beta waitlist for them, and they'll be emailing `hello@meltano.com` once we've been given access.

Closes #6157

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -175,7 +175,7 @@ jobs:
         $( pattern='in ([0-9]+)\.[0-9]+s' && [[ ${{ env.pytest_results }} =~ $pattern ]] && echo "${BASH_REMATCH[1]}" )s |"
 
     - name: Upload coverage data
-      if: always() && (matrix.python-version == '3.9')
+      if: always()
       uses: actions/upload-artifact@v3.1.0
       with:
         name: coverage-data
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -132,8 +132,9 @@ jobs:
 
     - name: Run pytest
       env:
+        COLUMNS: 160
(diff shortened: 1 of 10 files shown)


edgarrmondragon (approved):
nice improvement!

pandemicsyn (approved):
> we see an average time saving of ~476 seconds

😮 nice!

PR #6740, opened 2022-09-12
chore: update codeowners

kgpayne:
Updating web app owners.


.github/CODEOWNERS
```diff
@@ -1,36 +1,45 @@
...
 .github/actions/ @WillDaSilva @meltano/engineering
 .github/workflows/ @WillDaSilva @meltano/engineering
 
 # Discovery (End-of-Life)
+
 /src/meltano/core/bundle/discovery.yml @aaronsteers
 
 # Web App and API owners
-/src/webapp/ @alexmarple @pandemicsyn
-/src/meltano/api/ @pandemicsyn @alexmarple
```
pandemicsyn, line 18: We probably still want secondary reviewers here. Maybe @cjohnhanson / @rwfeather since they have some FE foo as well?

/cc @aaronsteers

pandemicsyn:
@kgpayne mind update the PR message to include something a bit descriptive? Just something as simple as "update webapp owners" would be fine. Had to look pretty closely to pick up what was **actually** changing 😁

.github/CODEOWNERS
```diff
@@ -1,36 +1,45 @@
...
+
 /CHANGELOG.md @meltano/engineering
 /docs/package.json @meltano/engineering
-/package.json  @meltano/engineering
-/pyproject.toml  @meltano/engineering
-/src/meltano/__init__.py  @meltano/engineering
-/src/webapp/package.json  @meltano/engineering
+/package.json @meltano/engineering
+/pyproject.toml @meltano/engineering
+/src/meltano/**init**.py @meltano/engineering
```
pandemicsyn, line 44: is that an intentional change ?

pandemicsyn (approved):

Issue #5981, opened 2022-06-02
feat: Add support for non-DB state backends (`s3`, `dynamodb`, etc.)


aaronsteers:
Spec discussion: 

- https://github.com/meltano/meltano/discussions/6270

I think we could introduce state backends such as `s3`, `dynamodb`, and other backends that have better reliability than an RDBMS and near-zero always on cost.

In the future, a Meltano-managed state offering, similar to Pulumi's default experience.

_Originally posted by @aaronsteers in https://github.com/meltano/meltano/issues/2520#issuecomment-1145062214_

## Additional context:

1. Currently a 'current' STATE is a composite on-demand scan through history records. This is not ideal in general - and we have #3340 logged to refactor this so a single table row would be the "backend" to read and write from.
1. Moving to a generic backend store would likely require _also_ solving for #3340 - or at least the same refactoring would (likely) be needed in both cases to eliminate the need for scanning history logs.

## Workarounds:

#2520 talks about potential workaround, but basically the current workaround is to:

1.  `meltano state get...` to pull the latest state into a file.
2. upload that file to S3 before the container is deleted.
3. when next run, download the file from S3 and load to the systemdb with `meltano state set ...`

This works with our without a postgres or other long-lived rdmbs, since the built-in sqlite implementation is created on the fly if no postgresdb is specified, and the process above essentially just removes the long-term state storage requirement from the sqlite backend.

tayloramurphy:
The plugin approach discussed in https://github.com/meltano/meltano/discussions/6270#discussion-4166066 I think does make sense to explore more. Particularly in light of efforts with #6130

rickiesmooth:
I'm running Meltano in a container aswell, but I'm having a hard time figuring out what the best way is to _upload that file to S3 before the container is deleted_. There's no on-run-end hook in Meltano right?

tayloramurphy:
@rickiesmooth which file are asking about? There's a few ways I could imagine doing this but there's currently no `on-run-end` hook exactly like there is in dbt.

rickiesmooth:
Sorry I was referencing the workaround and how it uploads the state file.

rickiesmooth:
ah that would be nice, for now I just do:
```bash
meltano state get dev:tap-google-search-console-to-target-bigquery > meltano_state/gsc_state.json
meltano state get dev:tap-google-analytics-to-target-bigquery > meltano_state/ga_state.json

[...]

PR #6742, opened 2022-09-13
chore: Support "state backends" beginning with systemdb

### LINKED ISSUE #3340: Create a new `state` table in Meltano systemdb
Related to:

- https://github.com/meltano/meltano/issues/5981

Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3418

Originally created by @tayloramurphy on 2022-04-19 14:49:44

---

After [#3290](https://gitlab.com/meltano/meltano/-/issues/3290) and [#2924](https://gitlab.com/meltano/meltano/-/issues/2924) the backend for state should be updated (from https://gitlab.com/meltano/meltano/-/merge_requests/2606#note_916073080)

Overview:

- Create a `state` table that just has one row per `job-id` (which will be called `state-id` in the future).
- This is makes us compatible with a non-rdbms approach for state backend, since the history tables can continue doing what they are doing and the `state` backend (rdbms or otherwise) has a single artifact (per `state-id`) to read and write from directly.

Spec draft:

1. During an EL run: Meltano inserts a row to the job run history table *and* updates or upserts the `state` table.
   * Probably we need 2 columns: `completed_state` and `partial_state`.
     * `completed_state` would be the last full EL execution.
     * `partial_state` could be updated on an ongoing basis as the job is running.
   * After an EL operation completes, the `partial_state` is moved to the `completed_state` column and `partial_state` is cleared.:
     * `UPDATE TABLE state SET completed_state = partial_state, partial_state = NULL WHERE state_id = ?`
   * Before an EL operation states, the tap is sent the result of merging the `partial_state` column against the `completed_state` column. (Only one row needs to be read.)
1. During `meltano set` and `meltano clear`: we can insert a row just as we're doing here (useful for audit purposes) and *also* update the `state` table.
1. During `meltano get`, the result is calculated by merging the `partial_state` column against the `completed_state` column. (Just as in the start of an EL operation, only one row needs to be read.)
1. Never in the new process would we need to perform a full table scan or retrieve all history records for a job. (Unless we decide to accept something like `--history` to get the historic job execution records - in which case we are going to get this from the historic logs _instead of_ the "main" state table.)

cjohnhanson:
**This is a significant change to how state is handled.** 

This PR refactors our state management to use a new `StateStoreManager` ABC similar to how we manage settings. The first `StateStoreManager` is `DBStateStoreManager` which just uses the system db to persist state. The primary difference in behavior is that state is now a dedicated table and we no longer have to query and parse through job run history in order to assemble current state for a job. 

The refactor also gets us 90% of the way towards rolling out a new state backend that can use blob storage such as AWS S3 (see #5981). It also lays the groundwork for us to roll out _any_ new state backend without a ton of effort. With some small amount of additional refactoring, it also lays the groundwork for us to support arbitrary backends in the form of plugins once we're able to support that in the SDK (or maybe EDK).

Currently labeled "chore" rather than "feat" because this is an invisible backend refactor without user impact. The follow-up PR to add remote blob storage backends will be the "release" PR that should be included in release changelog and demos.

To test the migration (since it's a bit more complex than most migrations and includes job run history to populate the new table) I had two venvs: one with the latest meltano version installed and one with my local meltano branch installed. I started a new project using the latest version, ran a tap-github to target-jsonl pipeline to populate minimal job history and state then checked the output of `meltano state list` and `meltano state get`. Then I switched to the local branch, ran `meltano upgrade database` and confirmed that the outputs for state commands matched. **This was a pretty limited manual test and I would love suggestions for how to include some automated integrations tests or otherwise test the migration more thoroughly.**

Closes #3340


src/meltano/core/state_store.py
```diff
@@ -0,0 +1,202 @@
...
+from abc import ABC, abstractmethod
+
+from sqlalchemy import select
+from sqlalchemy.orm import Session
+
+from meltano.core.job_state import JobState
+from meltano.core.utils import merge
+
+
+class StateStoreManager(ABC):
```
edgarrmondragon, line 14: Nice interface 😄

edgarrmondragon (approved):

src/meltano/migrations/versions/f4c225a9492f_create_dedicated_job_state_table.py
```diff
@@ -0,0 +1,404 @@
...
+    Includes State.STATE_EDIT rows which represent CLI invocations of the
+    `meltano state` command which wrote state to the db. Queries that are
+    meant to return only actual job runs should filter out records with
+    state == State.STATE_EDIT.
+    """
+
+    __tablename__ = "runs"
+
+    id = Column(types.Integer, primary_key=True)
+    job_name = Column(types.String)
```
pandemicsyn, line 319: qq - do we need to throw some max lengths on these String and JSONEncodedDict columns ? To maintain compatability with MSSQL (and MySQL later).

basically thinking of:

- #6167

pandemicsyn:
> This was a pretty limited manual test and I would love suggestions for how to include some automated integrations tests or otherwise test the migration more thoroughly.

@cjohnhanson I think it would be pretty easy to set up an integration test to exercise that the new backend works in a more e2e fashion, since the validation step in the integration test can run an arbitrary bash script (https://github.com/meltano/meltano/blob/main/integration/meltano-run/validate.sh), so we could sqlite/psql all the things! Or even just do a couple of runs and then exercise it with the `meltano state` command and make sure `meltano state` works, returns things as expected. 

If you really wanted to test the migration, we could probably also rig it up to do something like import an included sqlite dump as the first step of the integration test.

src/meltano/core/job_state.py
```diff
@@ -0,0 +1,74 @@
...
+        partial_state: dict[Any, Any] = {}
+        incomplete_since = None
+        finder = JobFinder(job_name)
+
+        # Get the state for the most recent completed job.
+        # Do not consider dummy jobs create via add_state.
+        state_job = finder.latest_with_payload(session, flags=Payload.STATE)
+        if state_job:
+            incomplete_since = state_job.ended_at
+            if "singer_state" in state_job.payload:
```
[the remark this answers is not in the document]
cjohnhanson: Good call, I've added that within `job_state.py`

src/meltano/migrations/versions/f4c225a9492f_create_dedicated_job_state_table.py
cjohnhanson: Yeah we definitely do need max length down in the `JobState` table. I've added that for the `JSONEncodedDict`. Turns out MSSQL also requires primary key columns to be <=900 bytes as well, so I've added that constraint there to.

pandemicsyn (approved):
:shipit:

integration/meltano-run/validate.sh
```diff
@@ -14,3 +14,7 @@ grep "Done. PASS=1 WARN=0 ERROR=0 SKIP=0 TOTAL=1 cmd_type=command name=dbt-postg
 # we could also run psql statements
 # check for the existance of files
 # etc
```
cjohnhanson, line 16: @pandemicsyn you mind giving a quick once-over to this new integration test?

pandemicsyn: 👍 lgtm - seems like a good safety check (makes sure `meltano run` is actually tracking state as expected, and that the entry looks roughly legit).

WillDaSilva:
> Currently labeled "chore" rather than "feat" because this is an invisible backend refactor without user impact. The follow-up PR to add remote blob storage backends will be the "release" PR that should be included in release changelog and demos.

@cjohnhanson I'm chiming in a bit too late, but there's a `refactor` semantic type that can be used for backend changes without user impact. Things labeled "chore" are not included in the release notes, but refactors are.

cjohnhanson:
@WillDaSilva thanks for the info, that's definitely what I should have used here 😅

PR #6743, opened 2022-09-13
docs: Describe how `project_id` is hashed

WillDaSilva:
Closes #6729

Issue #6729, opened 2022-09-07
docs: update telemetry docs to accurately describe hashing


tayloramurphy:
* https://github.com/meltano/meltano/pull/6621#issuecomment-1215686833 indicates the follwoing

> We still hash the project ID before sending it in telemetry data

but a reading of https://docs.meltano.com/reference/settings#send_anonymous_usage_stats doesn't make that clear. That should be updated to indicate Snowplow / not GA and that the project_id is hashed.

PR #6744, opened 2022-09-13
chore: Remove deprecated Google Analytics tracking ID settings

WillDaSilva:
Came across these while addressing #6729 (PR: https://github.com/meltano/meltano/pull/6743). Since Google Analytics has already been removed (#6622), these should probably be removed too.


tayloramurphy (changes requested):
@WillDaSilva looks like on that settings page where the anonymous usage stats are (around line 120) there are still a few references to Google Analytics. If you could remove those as a part of this PR that'd be perfect. Thanks!

tayloramurphy (approved):
Nevermind, I see that's in the other PR! Approving!

pandemicsyn (approved):

PR #6745, opened 2022-09-13
fix: Use random UUID as default `project_id`

### LINKED ISSUE #6733: revert: project-name in project_id
In https://github.com/meltano/meltano/pull/6621 we added the `project name` to the project_id on `init`. Even though we hash the id prior to sending, its addition was unnecessary and creates a few problems:

* having the project name in the id confuses the purpose of the id. We don't need to know or include the directory name in the project id and its inclusion sends the wrong signal that we do care about it somehow
* while it's true that `project_id` can be any string, we generally don't want to encourage folks to change it unless they have a strong reason to
* when trying to correlate local testing to events we're seeing through our production snowplow pipeline, having to hash the id is unnecessarily burdensome when we'd rather just look for the GUID itself

I would support a documentation update to make it clear that `project_id` can be any string but that it's generally not recommended to update if you don't need to.

Also, I did notice that as part of this update it seems like `project_id` gets put at the end of the file - is there anyway to bring that back up above the environments?

@WillDaSilva @aaronsteers

WillDaSilva:
Closes #6733

### DIFF
--- a/src/meltano/core/project_init_service.py
+++ b/src/meltano/core/project_init_service.py
@@ -66,7 +66,7 @@ class ProjectInitService:
         self.settings_service = ProjectSettingsService(self.project)
         self.settings_service.set(
             "project_id",
-            f"{self.project_name}-{uuid.uuid4()}",
+            str(uuid.uuid4()),
             store=SettingValueStore.MELTANO_YML,
         )
         self.set_send_anonymous_usage_stats()


pandemicsyn (approved):
lgtm! but definitely hold for a +1 from @tayloramurphy on this one 😁

edgarrmondragon (approved):

tayloramurphy (approved):
LGTM :shipit:

PR #6748, opened 2022-09-13
refactor: Define and recommend `MeltanoError` exception class for improved CLI error messages

### LINKED ISSUE #6645: Give `http://melta.no` troubleshooting help link when plugin fails, with CTA to Slack in web copy
Spec (edited by AJ):

When Meltano fails, we'll print a standard error message:

> [Error description text.]
>
> To contact the Meltano team and access troubleshooting help and resources, visit http://melta.no

<details>
  <summary>Orig. issue description (Douwe)</summary>
  
When `meltano invoke` or `meltano run` fails, a user is more likely to manage to solve the issue and stick with Meltano if they join Slack than if they don't, as evidenced by the number of people who responded to the survey saying the docs weren't always helpful, but the Slack community was. I imagine a good amount of people that try us out and churn would not have if they had known about Slack. That also lets us reach out to them to learn more about their use case and experience, even if they do end up churning.

So I think we should recommend joining Slack in the product wherever that makes sense.

@tayloramurphy Wdyt? Seems like an easy change with potentially large impact.

</details>

### LINKED ISSUE #6684: feat: Improve error log for cron validator and env var strict mode
Discussion:

- https://github.com/meltano/meltano/discussions/6700

Cron validation launched last week. As of now, the invalid cron expression prints in red text but without a clear error message.

Similarly for the env var strict mode, we could expand the error message to be more descriptive and prescriptive.

We can improve usability if we add an explicit error message for this case.

edgarrmondragon:
Related:

- Closes https://github.com/meltano/meltano/issues/6645
- Closes https://github.com/meltano/meltano/issues/6684
- https://github.com/meltano/meltano/discussions/6700

### DIFF
--- a/docs/src/_contribute/errors.md
+++ b/docs/src/_contribute/errors.md
@@ -25,11 +25,9 @@ class ScheduleDoesNotExistError(MeltanoError):
 
         Args:
             name: The name of the schedule that does not exist.
-
         """
-        self.name = name
-
-        reason = f"Schedule '{self.name}' does not exist"
-        instruction = "Use `meltano schedule add` to add a schedule"
-        super().__init__(reason, instruction)
+        super().__init__(
+            reason=f"Schedule '{name}' does not exist",
+            instruction="Use `meltano schedule add` to add a schedule",
+        )
 ```
--- a/docs/src/_contribute/errors.md
+++ b/docs/src/_contribute/errors.md
@@ -9,9 +9,9 @@ This section describes how to create custom exception classes for Meltano.
 
 ## Creating a custom exception
 
-_All_ new custom exceptions must inherit from `meltano.core.error.MeltanoError` or one of its subclasses. The subclass declaration takes the optional
+_All_ new custom exceptions must inherit from `meltano.core.error.MeltanoError` or one of its subclasses.
 
-The subclassed exception has to be instantiated with a required `reason` argument, which is a string describing the error, as well as the optional `instruction` argument that can be used to provide additional next troubleshooting steps to the user.
+The `meltano.core.error.MeltanoError` initializer takes in a `reason` argument, which is a string describing the error, as well as the optional `instruction` argument that can be used to provide additional troubleshooting steps for the user.
 
 ```python
 from meltano.core.error import MeltanoError
--- a/docs/src/_contribute/errors.md
+++ b/docs/src/_contribute/errors.md
@@ -1,0 +1,35 @@
+---
+title: Custom errors
+description: Creating custom exceptions and errors for the Meltano codebase.
+layout: doc
+weight: 10
+---
+
+This section describes how to create custom exception classes for Meltano.
+
+## Creating a custom exception
+
+_All_ new custom exceptions must inherit from `meltano.core.error.MeltanoError` or one of its subclasses. The subclass declaration takes the optional
+
+The subclassed exception has to be instantiated with a required `reason` argument, which is a string describing the error, as well as the optional `instruction` argument that can be used to provide additional next troubleshooting steps to the user.
+
+```python
+from meltano.core.error import MeltanoError
+
+
+class ScheduleDoesNotExistError(MeltanoError):
+    """Occurs when a schedule does not exist."""
+
+    def __init__(self, name: str):
+        """Initialize the exception.
+
+        Args:
+            name: The name of the schedule that does not exist.
+
+        """
+        self.name = name
+
+        reason = f"Schedule '{self.name}' does not exist"
+        instruction = "Use `meltano schedule add` to add a schedule"
+        super().__init__(reason, instruction)
+```
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -182,7 +182,6 @@ module = [
   "meltano.core.elt_context",
   "meltano.core.environment",
   "meltano.core.environment_service",
-  "meltano.core.error",
   "meltano.core.extract_utils",
   "meltano.core.extractor",
   "meltano.core.meltano_file",
--- a/src/meltano/cli/__init__.py
+++ b/src/meltano/cli/__init__.py
@@ -5,7 +5,7 @@ import logging
 import os
 import sys
 from textwrap import dedent
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, NoReturn
 
 from meltano.core.error import MeltanoError
 from meltano.core.logging import setup_logging
@@ -62,12 +62,12 @@ setup_logging()
 logger = logging.getLogger(__name__)
 
 troubleshooting_message = """\
-Need help fixing this problem? Visit http://melta.no/ for troubleshooting steps or to
-join our friendly slack community.
+Need help fixing this problem? Visit http://melta.no/ for troubleshooting steps, or to
+join our friendly Slack community.
 """
 
 
-def handle_meltano_error(error: MeltanoError):
+def handle_meltano_error(error: MeltanoError) -> NoReturn:
     """Handle a MeltanoError.
 
     Args:
@@ -76,12 +76,11 @@ def handle_meltano_error(error: MeltanoError):
     Raises:
         CliError: always.
     """
-    if error.instruction:
-        message = f"{error.reason}. {error.instruction}."
-    else:
-        message = error.reason
-
-    raise CliError(message) from error
+    raise CliError(
+        f"{error.reason}. {error.instruction}."
+        if error.instruction
+        else f"{error.reason}."
+    ) from error
 
 
 def _run_cli():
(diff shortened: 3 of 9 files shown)


WillDaSilva (commented):
https://melta.no/ gives me:
image

http://melta.no works fine

edgarrmondragon:
> https://melta.no/

I can confirm this in Chrome, and both fail with `SSL_ERROR_UNRECOGNIZED_NAME_ALERT` in Firefox for me.

cc @tayloramurphy

tayloramurphy:
@WillDaSilva @edgarrmondragon I've emailed the domain provider to see about getting an SSL Certificate. http://melta.no does work and forwards appropriately from what I see. FYI @therebbie I put an inquiry through the Marcaria contact form to go to the hello meltano address.

pyproject.toml
```diff
@@ -182,7 +182,6 @@ module = [
   "meltano.core.elt_context",
   "meltano.core.environment",
   "meltano.core.environment_service",
-  "meltano.core.error",
```
WillDaSilva, line 185: Nice!

docs/src/_contribute/errors.md
```diff
@@ -0,0 +1,35 @@
...
+
+        Args:
+            name: The name of the schedule that does not exist.
+
+        """
+        self.name = name
+
+        reason = f"Schedule '{self.name}' does not exist"
+        instruction = "Use `meltano schedule add` to add a schedule"
+        super().__init__(reason, instruction)
```
WillDaSilva, line 34:
```suggestion
            name: The name of the schedule that does not exist.
        """
        super().__init__(
            reason=f"Schedule '{name}' does not exist",
            instruction="Use `meltano schedule add` to add a schedule",
        )
```

src/meltano/cli/__init__.py
```diff
@@ -59,6 +61,28 @@
 
 logger = logging.getLogger(__name__)
 
+troubleshooting_message = """\
+Need help fixing this problem? Visit http://melta.no/ for troubleshooting steps or to
+join our friendly slack community.
```
WillDaSilva, line 66:
```suggestion
Need help fixing this problem? Visit http://melta.no/ for troubleshooting steps, or to
join our friendly Slack community.
```

src/meltano/cli/__init__.py
```diff
@@ -59,6 +61,28 @@
...
+
+    Raises:
+        CliError: always.
+    """
+    if error.instruction:
+        message = f"{error.reason}. {error.instruction}."
+    else:
+        message = error.reason
+
+    raise CliError(message) from error
```
WillDaSilva, line 84: This suggestion avoids inconsistent periods.

```suggestion
    raise CliError(
        f"{error.reason}. {error.instruction}."
        if error.instruction
        else f"{error.reason}."
    ) from error
```

src/meltano/cli/__init__.py
```diff
@@ -4,8 +4,10 @@
 import logging
 import os
 import sys
+from textwrap import dedent
 from typing import TYPE_CHECKING
```
WillDaSilva, line 8: This suggestion pairs with the one that adds a type signature to `handle_meltano_error` below.

```suggestion
from typing import TYPE_CHECKING, NoReturn
```

src/meltano/cli/__init__.py
```diff
@@ -59,6 +61,28 @@
 
 logger = logging.getLogger(__name__)
 
+troubleshooting_message = """\
+Need help fixing this problem? Visit http://melta.no/ for troubleshooting steps or to
+join our friendly slack community.
+"""
+
+
+def handle_meltano_error(error: MeltanoError):
```
WillDaSilva, line 70: This suggestion pairs with the one that imports `NoReturn` above. See https://docs.python.org/3/library/typing.html#typing.NoReturn

```suggestion
def handle_meltano_error(error: MeltanoError) -> NoReturn:
```

src/meltano/cli/__init__.py
```diff
@@ -67,16 +91,19 @@ def _run_cli():
...
                 f"The requested action could not be completed: {err}"
             ) from err
         except KeyboardInterrupt:  # noqa: WPS329
             raise
+        except MeltanoError as err:
+            handle_meltano_error(err)
         except Exception as err:
-            raise CliError(str(err)) from err
+            message = f"{troubleshooting_message}\n{err}"
+            raise CliError(message) from err
```
WillDaSilva, line 106: A nitpick - why assign to a variable when it can cleanly be avoided:

```suggestion
            raise CliError(f"{troubleshooting_message}\n{err}") from err
```

WillDaSilva, line 106: This isn't introduced by this PR, but this approach of turning `err` into a string loses the traceback, which can lead to some unhelpful error messages. The traceback is still kept in the `CliError` though, so as long as we log it somewhere it's fine. That said, I believe it is being logged.

src/meltano/core/error.py
```diff
@@ -13,6 +14,21 @@ class ExitCode(int, Enum):  # noqa: D101
...
+        """Initialize a MeltanoError.
+
+        Args:
+            reason: A short explanation of the error.
+            instruction: A short instruction on how to fix the error.
+            args: Additional arguments to pass to the base exception class.
+        """
+        self.reason = reason
+        self.instruction = instruction
+
```
WillDaSilva, line 30:
```suggestion
        super().__init__(*args)

```

src/meltano/core/error.py
```diff
@@ -13,6 +14,21 @@ class ExitCode(int, Enum):  # noqa: D101
     NO_RETRY = 2
 
 
+class MeltanoError(Exception):
+    """Base class for all user-facing errors."""
+
+    def __init__(self, reason: str, instruction: str | None = None, *args: Any) -> None:
```
WillDaSilva, line 20: Why not also accept `**kwargs`?

src/meltano/core/schedule_service.py
```diff
@@ -32,8 +33,11 @@ def __init__(self, schedule: Schedule):
         """
         self.schedule = schedule
 
+        reason = f"Schedule '{self.schedule.name}' already exists"
+        super().__init__(reason)
```
WillDaSilva, line 37:
```suggestion
        super().__init__(reason=f"Schedule '{self.schedule.name}' already exists")
```

src/meltano/core/schedule_service.py
```diff
@@ -45,8 +49,12 @@ def __init__(self, name: str):
         """
         self.name = name
 
+        reason = f"Schedule '{self.name}' does not exist"
+        instruction = "Use `meltano schedule add` to add a schedule"
+        super().__init__(reason, instruction)
```
WillDaSilva, line 54: Why store `name` on the attribute?

I think it would be nicer if `reason` and `instruction` were passed in as keyword arguments.

This same suggestion applies to a few other decedents of `MeltanoError`.

tests/meltano/cli/test_cli.py
```diff
@@ -160,6 +161,11 @@ def test_no_environment_and_null_environment_overrides_default(  # noqa: WPS118
         )
         assert Project._default.active_environment is None
 
+    def test_handle_meltano_error(self):
+        exception = MeltanoError(reason="This failed", instruction="Try again")
+        with pytest.raises(CliError):
```
WillDaSilva, line 166:
```suggestion
        with pytest.raises(CliError, match=r"This failed. Try again."):
```

WillDaSilva (commented):
I've got no major criticisms. I'll be happy to see this merged in.

src/meltano/core/error.py
```diff
@@ -13,6 +14,21 @@ class ExitCode(int, Enum):  # noqa: D101
...
+        """Initialize a MeltanoError.
+
+        Args:
+            reason: A short explanation of the error.
+            instruction: A short instruction on how to fix the error.
+            args: Additional arguments to pass to the base exception class.
+        """
```
edgarrmondragon, line 27:
```suggestion
    def __init__(
        self,
        reason: str,
        instruction: str | None = None,
        *args: Any,
        *kwargs: Any,
    ) -> None:
        """Initialize a MeltanoError.

        Args:
            reason: A short explanation of the error.
            instruction: A short instruction on how to fix the error.
            args: Additional arguments to pass to the base exception class.
            kwargs: Keyword arguments to pass to the base exception class.
        """
```

src/meltano/core/schedule_service.py
edgarrmondragon: @WillDaSilva problem I thought of that it would require the dev to type a reason and instruction every time the exception is instantiated (if the exception may be raised in different places)

WillDaSilva: @edgarrmondragon Sorry for not being clear. When I said

> I think it would be nicer if `reason` and `instruction` were passed in as keyword arguments.

I didn't mean to suggest changing the function signature. I meant calling `super().__init__` like this:

```python
super().__init__(
    reason="...",
    instruction="...",
)
```

It was just a minor style comment, since that way avoids the use of additional variables, and would break (as we desire) if the signature is changed in the future.

edgarrmondragon: @WillDaSilva Ah, that makes sense. Sorry for my confusion 😅. Just pushed the change.

WillDaSilva (approved):

Issue #6750, opened 2022-09-14
bug: `api/v1/version?include_latest=true` fails behind firewall


dingobar:
### Meltano Version

2.6.0

### Python Version

3.9

### Bug scope

API

### Operating System

Linux - Ubuntu 22.04

### Description

Not sure if everyone would agree but I file this as a bug. I find this reasonable because meltano should not depend on external metadata to do its job.

When running behind a restrictive firewall, `meltano ui start` fails due to calling `api/v1/version?include_latest=true`, which calls out to pypi.org to get the latest version of meltano from there. For environments with strict firewall policies (typically environments with a large amount of critical infrastructure like in our case the telco industry), we cannot open for egress to external domains that are not our own. In these cases, I would like meltano to fail gracefully. In general, it could be worth considering how we will deal with these types of external metadata dependencies going forwards - crash? succeed but fail external calls silently? succeed but log warnings of failed calls? global switch to disable egress requests?

[Here](https://github.com/meltano/meltano/blob/b56af55138b1898227ec6e5e1e1429c3123edb5f/src/meltano/api/controllers/root.py#L74:L83) is the controller code for the endpoint.

[Here](https://github.com/meltano/meltano/blob/dfd9650170ff97fd69fdff7ac8dc7dc0841bf663/src/webapp/src/store/modules/system.js#L27) is where meltano ui calls the endpoint with the `latest_version=true` parameter.

### Suggested fix

I suggest we simply add a try-except on that call to the endpoint, and return `latest_version: null` if we cannot retrieve the latest version.

**I don't mind implementing that.** I have some time on my hands the coming days too.

tayloramurphy:
Related to https://github.com/meltano/meltano/issues/6754

@aaronsteers this seems like an uncontroversial change. Any objections?

aaronsteers:
@dingobar 

> ### Suggested fix
> I suggest we simply add a try-except on that call to the endpoint, and return `latest_version: null` if we cannot retrieve the latest version.
> 
> **I don't mind implementing that.** I have some time on my hands the coming days too.

Yes, I agree with this approach. Totally makes sense.

Also, and not mutually exclusive with the above, we could add a UI setting called `ui.update_check` ([here](https://docs.meltano.com/reference/settings#meltano-ui-features) as sibling to `ui.read_only`) which defaults to `true` but can be set to `false` to completely disable the check.

What do you think?

tayloramurphy:
@aaronsteers I like it 👍 

[...]

PR #6751, opened 2022-09-14
fix: catch ConnectionError in /version endpoint

martimors:
My stab at https://github.com/meltano/meltano/issues/6750

I tried to write a test for it but I'm not sure how you usually mock this type of external calls. Also, I don't see a test for the `/version` endpoint from the start. For now I've verified that this works and does not break the UI.

Feedback would be greatly appreciated!


dingobar:
Seems pre-commit is failing with a bunch of missing docstrings. I guess that pre-commit was added after the controller was written?

I also see this
```
 src/meltano/api/controllers/root.py:85:17: WPS323 Found `%` string formatting
                "%s failed with error %s getting latest_version from pypi.org",
```
which I would like to ignore because this is indeed the correct way to interpolate strings when logging [according to pylint](https://pylint.pycqa.org/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html).

aaronsteers:
@dingobar - Re:

> For now I've verified that this works and does not break the UI.

Thanks for confirming the manual test on your side. For my part, this looks good. It's helpful to have the confirmation that returning `None` did not otherwise break the UI.

I'll leave this to @pandemicsyn for review and merge.

pandemicsyn (approved):
lgtm - also sorry for the delay, missed the notification for this one some how.

Issue #6253, opened 2022-06-21
Consolidate environment variable handling on an `EnvironmentVariableService` class


kgpayne:
The logic for collating (including resolving inheritance) and expanding environment variables is currently spread across several classes in Meltano. This already makes maintaining, debugging and reasoning about environment variable origin and precedence difficult, and increases the development cost of improvements such as enforcing/ensuring a coherent resolution hierarchy (#5982 and this miro board) and the 'stacking' feature (#5983 whereby an env var can refer to another env var at or above its level in the resolution hierarchy).

To resolve this we could consolidate the helper functions and class attributes that currently make up or environment variable features into a `EnvironmentVariableService`, responsible for i) collating environment variable definitions (via the `Project` and `ConfigService` classes, ii) managing the resolution of environment variable precedence for a given context and iii) handling the expansion of environment variables, including 'stacking'. As the environment definition does not change during a given invocation, the environment can largely be built (collated and expanded) once and used for fast-lookups from the various `SettingsService*` instances and during invocation thereafter.

## Proposed Spec

TBD

pandemicsyn:
+1 - I think this is a great idea, would get us (folks hacking on meltano) close to having a single source truth and one easier to reason about, test against or mock, and just generally interact with. It's a bit of a big lift, but feels a bit overdue, to be honest.

kgpayne:
As per 1:1 conversation with @aaronsteers we should still do this at some point, but it is less of a priority now as we have added the priority features to the existing implementation:

> Between my changes and Cody's, we haven't needed to do the refactor, but we also haven't reduced the complexity of maintaining the env var sourcing and resolution code. It is still spread across several classes - we just added new logic to those places. So I still think the refactor to consolidate the env var logic on a new service class should be done at some point, but like all tech debt it can be postponed until the next time we need to change this code. I think Florian's comment leans in that direction too.

tayloramurphy:
This was previously weighted at an 8. I'm pushing this to the backlog but we can pull forward if deemed more urget.

PR #6758, opened 2022-09-15
feat: Support custom state IDs in `meltano run` with `--state-id-suffix`

### LINKED ISSUE #6164: Feature proposal: dynamic state IDs based on an input variable
This was called out as "nice to have" but not delivered in the initial iteration of auto-generated state IDs for `meltano run`. 

For scenarios where the 'environment' actually builds to multiple deploy targets, such as one per customer or one per CI/CD build number or one per CI/CD branch name, it would be great to have a configurable and dynamic `state ID suffix` that could take one or more inputs, and use those inputs to ensure a unique set of STATE ID refs when running via `meltano run`.

Today state IDs are generated like:

```
${environment_name}:${tap-name}-to-${target-name}
```

But we could add some per-environment setting such as `state_id_suffix` or `deploy_target_key` which would concatenate with the base state ID calc:

```
${environment_name}:${tap-name}-to-${target-name}:${suffix_str}
```

Taking the branch names example, and an environment name called `cicd`, this could look like the following:

```
cicd:tap-github-to-target-snowflake:main
cicd:tap-github-to-target-snowflake:develop
cicd:tap-github-to-target-snowflake:staging
cicd:tap-github-to-target-snowflake:feat/my-cool-feature
cicd:tap-github-to-target-snowflake:fix/my-super-important-fix
```

Where each of these variations are basically populated with some input like `CI_BRANCH_REF`.

### LINKED ISSUE #6370: Add ability to provide custom `--state-id-suffix` to `meltano run`
One thing that was cut late in development when adding incremental job support to `meltano run` was the ability to add a custom prefix to the auto generated job (now `state`) names: https://gitlab.com/meltano/meltano/-/merge_requests/2544#note_855957749

> I think we've agreed on this already but just want to reiterate given that we've just logged [ADR: Refactor/rename "job_id" as "state-id" (#3290 - closed)](https://github.com/meltano/meltano/-/issues/3290), it is probably best to try to avoid user config/override for the state_id_prefix or job_id_prefix. Just for the reason that if we allow them to configure the naming convention, and then we update what it is called, that will create a breaking change which I think it's best to avoid.

Now that the `job` -> `state` transition work has basically wrapped up we could revisit adding this.

@tayloramurphy fwiw this is a pretty easy-peasy change on the dev side. If only configurable as a cli arg weights probably as low as a 2, but if we want to also support this as a per env meltano.yml option than most likely more realistically a weight of 4.

ReubenFrankel:
Resolves #6370, resolves #6164

### Checklist
- [x] `--state-id-suffix` option for `meltano run`
- [x] `state_id_suffix` option for environments in `meltano.yml`
- [x] Support expansion of environment variables for `state_id_suffix` value
- [x] Add/update tests
- [x] Update docs


ReubenFrankel:
I would like to add an end-to-end test for `--state-id-suffix` in [`test_run.py`](https://github.com/meltano/meltano/blob/fba29109eb36cde61d652e17960a241ed5d8edca/tests/meltano/cli/test_run.py) to check a state ID is generated with the given suffix, but the existing tests are quite complex so not sure where to start. Any pointers?

aaronsteers:
> I would like to add an end-to-end test for `--state-id-suffix` in [`test_run.py`](https://github.com/meltano/meltano/blob/fba29109eb36cde61d652e17960a241ed5d8edca/tests/meltano/cli/test_run.py) to check a state ID is generated with the given suffix, but the existing tests are quite complex so not sure where to start. Any pointers?

@pandemicsyn, @edgarrmondragon - Any ideas on the above?

@ReubenFrankel - I just noticed you'd kicked off this PR - very exciting! 🚀

edgarrmondragon:
> > I would like to add an end-to-end test for `--state-id-suffix` in [`test_run.py`](https://github.com/meltano/meltano/blob/fba29109eb36cde61d652e17960a241ed5d8edca/tests/meltano/cli/test_run.py) to check a state ID is generated with the given suffix, but the existing tests are quite complex so not sure where to start. Any pointers?
> 
> @pandemicsyn, @edgarrmondragon - Any ideas on the above?
> 
> @ReubenFrankel - I just noticed you'd kicked off this PR - very exciting! 🚀

Probably copying and modifying the first bit of `TestCliRunScratchpadOne.test_run_basic_invocations`: https://github.com/meltano/meltano/blob/e4d698495e50ee9d0a81d4885dced8f79f519982/tests/meltano/cli/test_run.py#L295-L340

and using the `job_logging_service` to check that the right state ID was used to create the job log

https://github.com/meltano/meltano/blob/e4d698495e50ee9d0a81d4885dced8f79f519982/src/meltano/core/logging/job_logging_service.py#L55-L56

src/meltano/core/plugin/settings_service.py
```diff
@@ -73,6 +73,17 @@ def __init__(
...
+
+                # expand state_id_suffix
+                self.project.active_environment.state_id_suffix = expand_env_vars(
+                    self.project.active_environment.state_id_suffix,
+                    {
+                        **self.env_override,
+                        **self.project.dotenv_env,
+                    },
+                    raise_if_missing=strict_env_var_mode,
+                )
```
ReubenFrankel, line 85: Implementation related to #6770.

ReubenFrankel:
Unrelated test failure: https://github.com/meltano/meltano/actions/runs/3087082213/jobs/4992077395#step:11:3279

src/meltano/core/plugin/settings_service.py
aaronsteers: @kgpayne - You and @cjohnhanson are our experts on config and env var parsing as of now. 😄 Can you take a look at the above and advise?

pandemicsyn:
> Unrelated test failure: https://github.com/meltano/meltano/actions/runs/3087082213/jobs/4992077395#step:11:3279

Looks like it cleared on subsequent updates.  Based on the conn resets, i'm guessing it was just transient and the host/vm had a networking issues.

tests/meltano/core/block/test_extract_load.py
```diff
@@ -621,19 +622,51 @@ async def test_elb_with_job_context(
...
+
+        assert (
+            generate_state_id(project, "suffix", block1, block2)
+            == "test:block1-to-block2:suffix"
+        )
+
+        assert (
+            generate_state_id(project, "suffix", block1, block2)
+            == "test:block1-to-block2:suffix"
+        )
```
pandemicsyn, line 644:
```suggestion
        )
        
        assert (
            generate_state_id(project, None, block1, block2)
            == "test:block1-to-block2"
        )

```

Worth testing no suffix just incase ?

tests/meltano/cli/test_run.py
```diff
@@ -368,6 +370,128 @@ def test_run_basic_invocations(
...
+                mp.setenv(*env)
+
+            result = cli_runner.invoke(cli, args, catch_exceptions=True)
+            assert result.exit_code == 0
+
+            matcher = EventMatcher(result.stderr)
+            assert matcher.find_by_event("Block run completed.")[0].get("success")
+
+            job_logging_service.get_latest_log(
+                f"dev:{tap.name}-to-{target.name}:${expected_suffix}"
```
pandemicsyn, line 492: nice 🤘 IMHO this should cover us for the change. 

@aaronsteers would be nice to do a integration test to show case/cover some of the state manipulation and behaviors that we support now (pop'd https://github.com/meltano/meltano/issues/6773 for that for us to follow up on).

src/meltano/core/state_service.py
```diff
@@ -16,6 +16,8 @@
 from meltano.core.job import Job, JobFinder, Payload, State
 from meltano.core.utils import merge
 
+STATE_ID_COMPONENT_DELIMITER = ":"
```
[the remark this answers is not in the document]
pandemicsyn: Guessing pulling STATE_ID_COMPONENT_DELIMETER and `meltano.core.block.extract_load.generate_state_id` into meltano.core.utils or the like probably doesn't help ?

pandemicsyn:
@ReubenFrankel looking great! IMHO you're safe to include the doc update whenever you're ready.

tests/meltano/core/block/test_extract_load.py
ReubenFrankel: Oops, that was my original intention... :sweat_smile: https://github.com/meltano/meltano/pull/6758/commits/fc2af322ff9d9aa1ef37b906af609b265ebf8b3d

src/meltano/core/state_service.py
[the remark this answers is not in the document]
ReubenFrankel: I'll have a look to see if that works. Would that be preferable?

docs/src/_concepts/environments.md
```diff
@@ -69,6 +70,12 @@ That is, `MELTANO_PROJECT_ROOT` and `MELTANO_ENVIRONMENT`.
 In the below example, the `$MELTANO_PROJECT_ROOT/path/to/a/file.json` value will properly read the `$MELTANO_PROJECT_ROOT`
 environment variable and inject the full value as `$MY_ENV_VAR` into the environment.
 
+## `state_id_suffix`
```
tayloramurphy, line 73:
```suggestion
## State ID Suffix
```

docs/src/_concepts/environments.md
```diff
@@ -69,6 +70,12 @@ That is, `MELTANO_PROJECT_ROOT` and `MELTANO_ENVIRONMENT`.
...
+
+Environments can also define a `state_id_suffix` - a custom suffix used in the generation of a state ID for each extractor/loader pair passed to [`meltano run`]((/reference/command-line-interface#run)).
```
tayloramurphy, line 75:
```suggestion
Environments can also define a `state_id_suffix` - a custom suffix used in the generation of a state ID for each extractor/loader pair passed to [`meltano run`]((/reference/command-line-interface#run)).
The suffix is appended to the generated ID with a colon prefix: `:<state_id_suffix>`. 

The full ID when a suffix is present is `<environment_name>:<tap_name>-to-<target_name>:<state_id_suffix>`.
```

docs/src/_concepts/environments.md
```diff
@@ -69,6 +70,12 @@ That is, `MELTANO_PROJECT_ROOT` and `MELTANO_ENVIRONMENT`.
...
+
+`state_id_suffix` supports interpolation of environment variables to allow for dynamic state IDs (e.g. unique state for multiple `meltano run` invocations using the same environment and EL pair(s)).
+
 ```yaml
```
tayloramurphy, line 79: This section goes with the "The env mapping" section, not with the state id suffix. It needs to be moved back up. Check the preview https://deploy-preview-6758--meltano.netlify.app/concepts/environments#state_id_suffix

tayloramurphy (changes requested):

src/meltano/core/state_service.py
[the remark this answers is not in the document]
pandemicsyn: Thinking about , I bet you'd still end up with a circular import. The annoy bit is that generate_state_id wants a consumer/producer IOBlock , but *only* uses the string_id's of those. One way to clean this up then would be to update `generate_state_id` to just expect a string for consumer/producer and just have callers start passing the string_id directly rather than passing the whole IOBlock.

That's a bit more work/refactoring though - I think you could keep the pr as-is and then we can log a follow up issue to clean up/refactor.

pandemicsyn (approved):
lgtm - pending the doc changes @tayloramurphy flagged

docs/src/_concepts/environments.md
ReubenFrankel: My bad, fixing now :+1:

docs/src/_concepts/environments.md
ReubenFrankel: See https://github.com/meltano/meltano/pull/6758/commits/fdca026b0c5376d23e7d147e19693e3b13003a3d

docs/src/_concepts/environments.md
ReubenFrankel: See See https://github.com/meltano/meltano/pull/6758/commits/fdca026b0c5376d23e7d147e19693e3b13003a3d

docs/src/_concepts/environments.md
ReubenFrankel: See https://github.com/meltano/meltano/pull/6758/commits/fdca026b0c5376d23e7d147e19693e3b13003a3d

src/meltano/core/state_service.py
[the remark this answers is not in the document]
ReubenFrankel: Sure :+1:

ReubenFrankel:
@pandemicsyn I've resolved all conversations I think are complete, just waiting on the others.

tayloramurphy (approved):
Docs approval

src/meltano/core/plugin/settings_service.py
kgpayne: Thanks for this @ReubenFrankel 🙌 I'm not surprised that env var expansion was a hurdle - we have been talking for a while about centralising env var handling onto its own service (#6253). Your approach is inline with the current state of env var expansion 👍 I was surprised that the `self.project.dotenv_env` was necessary, as I'd have expected this to be filled in already at the `project_settings_service` layer. I see it comes in stead from `get_with_metadata` when accessing 'normal' settings, so is needed here. Will put more thoughts along that line in its own comment...

kgpayne:
Following this comment on [env var expansion](https://github.com/meltano/meltano/pull/6758#discussion_r977547766), it strikes me that manual expansion is needed in this case because `state_id_suffix` is implemented as a new attribute of the `Environment`, and so bypasses the `get`/`get_with_metadata` method on the `SettingsService` (and descendants), which handles env var expansion for 'normal' settings. I wouldn't block on it for now, but in future we might choose to take the same approach as with plugins; treating extra `Environment` attributes as [extras](https://docs.meltano.com/guide/configuration#plugin-extras), accessible via the regular `get_with_metadata` and `set_with_metadata` calls (which handle expansion). Apart from simplifying expansion, this would also make it much easier to add new "Environment setting extras" like this in future, accessible via `project_setting_service.get_with_metadata("_extra.state_id_suffix")` 🙂

I think this would be as simple as updating the `MeltanoEnvStoreManager` to return any `Environment` attrs other than `['name', 'config' 'env']` as `_extras` in its `flat_config` property, but I haven't looked too closely. Configuration and overrides would happen in the usual settings hierarchy (env vars, yml etc.), and the cli can be kept in addition (if required) by passing the given value through as `config_override` to the `ProjectSettingsService` constructor.

Interested in your thoughts @edgarrmondragon, and if you think is worth capturing this in a new issue 👍 FYI @aaronsteers

kgpayne (changes requested):
Ah, actually I think these two need swapping, in line with how we interpolate in `get_with_metadata` [here](https://github.com/meltano/meltano/blob/main/src/meltano/core/settings_service.py#L333) 😅

src/meltano/core/plugin/settings_service.py
```diff
@@ -73,6 +73,17 @@ def __init__(
...
                     for var, value in self.project.active_environment.env.items()
                 }
+
+                # expand state_id_suffix
+                self.project.active_environment.state_id_suffix = expand_env_vars(
+                    self.project.active_environment.state_id_suffix,
+                    {
+                        **self.env_override,
+                        **self.project.dotenv_env,
+                    },
```
kgpayne, line 83:
```suggestion
                    {
                        **self.project.dotenv_env,
                        **self.env_override,
                    },
```

kgpayne (dismissed):

tests/meltano/cli/test_run.py
aaronsteers: Thanks, @pandemicsyn  👍 

Much appreciated.

aaronsteers (approved):
Looks great! Thanks, @ReubenFrankel for contributing.

PR #6759, opened 2022-09-15
docs: Fix the meltano ui command for the ui dev guide

martimors:
Took me several hours to figure this one out. Adding this to hopefully save others from that.

### DIFF
--- a/docs/src/_contribute/ui.md
+++ b/docs/src/_contribute/ui.md
@@ -22,6 +22,7 @@ meltano init $PROJECT_NAME
 cd $PROJECT_NAME
 
 # Start the Meltano API and a production build of Meltano UI that you can ignore
+export FLASK_ENV=development
 meltano ui
 
 # Open a new terminal tab and go to the directory you cloned meltano into
@@ -37,7 +38,6 @@ yarn serve
 The development build of the Meltano UI will now be available at <http://localhost:8080/>.
 
 A production build of the API will be available at <http://localhost:5000/> to support the UI, but you will not need to interact with this directly. However, as mentioned in the [API Development section](/contribute/api) above, users on MacOS may need to specify an alternate [bind to port](/reference/settings#ui-bind-port) to prevent a port conflict with a MacOS system service also running on port 5000.
-
 
 <div class="notification is-warning">
   <p><strong>If you're developing for the _Embed_ app (embeddable <code>iframe</code> for reports) you can toggle <code>MELTANO_EMBED</code>:</strong></p>
--- a/docs/src/_contribute/api.md
+++ b/docs/src/_contribute/api.md
@@ -24,7 +24,8 @@ meltano init $PROJECT_NAME
 cd $PROJECT_NAME
 
 # Start a development build of the Meltano API and a production build of Meltano UI
-FLASK_ENV=development meltano ui
+export FLASK_ENV=development
+meltano ui
 ```
 
 The development build of the Meltano API and a production build of the UI will now be available at <http://localhost:5000/>.
@@ -47,7 +48,7 @@ For V2 of the Meltano API, we generally aim to implement and adhere to our varia
 > The methods can be either the standard methods or custom methods.
 >
 > Where API functionality naturally maps to one of the standard methods, that method **should** be used in the API design.
-> For functionality that does not naturally map to one of the standard methods,*custom methods* **may** be used.
+> For functionality that does not naturally map to one of the standard methods,_custom methods_ **may** be used.
 > Custom methods offer the same design freedom as traditional RPC APIs, which can be used to implement common programming patterns, such as database transactions or data analysis.
 
 ### Concrete methods/verbs
@@ -79,8 +80,8 @@ Where a custom verb is used, the verb will be appended to the resource as a new 
 
 API service umbrella and namespace:
 
- - `meltano/core/v2beta` (during development)
- - `meltano/core/v2`
+- `meltano/core/v2beta` (during development)
+- `meltano/core/v2`
 
 Our top-level collection is `projects/*` with an intermediate of `/envs/*` as we've opted to include support for operating on multiple projects.
 The V2 API spec is in flux, we don't yet actually have strong ties between a lot of the topics beyond this level, but as we start specing v2 we expect a grouping around the below features to evolve.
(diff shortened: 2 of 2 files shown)


docs/src/_contribute/ui.md
```diff
@@ -22,7 +22,7 @@ meltano init $PROJECT_NAME
 cd $PROJECT_NAME
 
 # Start the Meltano API and a production build of Meltano UI that you can ignore
-meltano ui
+FLASK_ENV=development meltano ui
```
WillDaSilva, line 25:
```suggestion
FLASK_DEBUG=True meltano ui
```

The `FLASK_ENV` setting has been [deprecated since Flask 2.2](
https://flask.palletsprojects.com/en/2.2.x/config/#ENV). We are using ^2.1, so `FLASK_DEBUG` should still be available, and setting it to true should accomplish what you want.

WillDaSilva (changes requested):
Thanks for the contribution @dingobar

What issue were you encountering that this change helped with? I can see why we might want to call out Flask's debug mode as a development/debug tip, but I'm not sure that we want to tell contributors to always enable it.

docs/src/_contribute/ui.md
dingobar: If I don't set this flag (or `FLASK_DEBUG` as suggested, works as well), all requests from the dev UI are blocked because of a missing CORS header. So this is, afaict, a prerequisite to getting the dev UI up and running and as such I guess all contributors should use it.

dingobar: Seems it sets this header 
image

WillDaSilva (approved):
> If I don't set this flag (or FLASK_DEBUG as suggested, works as well), all requests from the dev UI are blocked because of a missing CORS header. So this is, afaict, a prerequisite to getting the dev UI up and running and as such I guess all contributors should use it.

> Seems it sets this header
> image

I have not yet been able to reproduce this issue, but this sounds like more than just an issue for contributors engaging in UI development. I'm approving this PR because for development, setting `FLASK_DEBUG=True` makes sense, but there may be a larger issue here.

pandemicsyn:
>  I'm approving this PR because for development, setting FLASK_DEBUG=True makes sense, but there may be a larger issue here.

👍 I think that makes sense, and also agree that there might have been another issue at play. I've never run into CORS issues yet either. 

@dingobar did you by chance set any other config opts like ui.server_name?

docs/src/_contribute/ui.md
```diff
@@ -22,7 +22,7 @@ meltano init $PROJECT_NAME
 cd $PROJECT_NAME
 
 # Start the Meltano API and a production build of Meltano UI that you can ignore
-meltano ui
+FLASK_DEBUG=True meltano ui
```
tayloramurphy, line 25: @WillDaSilva @pandemicsyn should we reccommend that this be done on a separate line?

```suggestion
export FLASK_DEBUG=True 

meltano ui
```

WillDaSilva: Sure. That plus a short comment explaining explanation why would be ideal.

dingobar: It's on two separate lines now. I also fixed in the API dev documentation which had the same command on a single line (which is why I added it to a single line in the first place). Keep in mind we forfeit windows support by adding the `export` keyword, the rest of the script is windows compatible.

pandemicsyn (approved):

dingobar:
> > I'm approving this PR because for development, setting FLASK_DEBUG=True makes sense, but there may be a larger issue here.
> 
> +1 I think that makes sense, and also agree that there might have been another issue at play. I've never run into CORS issues yet either (on a vanilla install).
> 
> @dingobar did you by chance set any other config opts like ui.server_name?

Ok so I tried to reproduce the error, and it seems like setting `FLASK_ENV=development` is the only thing that really helps.

`FLASK_DEBUG=True` has no effect,
image

However, by running with `FLASK_ENV=development` it works. I don't know why I thought it worked on Friday, guess I was tired after a long week and my shell probably had the `FLASK_ENV` variable set already. Sorry for that confusion!

npm --version
8.18.0

node --version
v18.8.0

yarn --version                                             
1.22.15

**Could someone try to verify following these steps** 
1. Create an empty 2.6.0 project using `meltano init testproject` and pipx installed meltano 2.6.0
3. `cd testproject` then `poetry run meltano ui`
4. git clone the meltano repo
5. In another terminal, `cd meltano` and then `yarn build && yarn serve`
6. Open localhost:8080 in FireFox or Chrome and see the CORS errors in the browser console

pandemicsyn:
>Could someone try to verify following these steps

@dingobar at some point I stuck`FLASK_ENV=development` in my `.zshrc` and promptly forgot about it - so initially had some issues reproducing this 😬  But can confirm `FLASK_ENV=development` is required (ran through your steps to be sure).

dingobar:
> > Could someone try to verify following these steps
> 
> @dingobar at some point I stuck`FLASK_ENV=development` in my `.zshrc` and promptly forgot about it - so initially had some issues reproducing this grimacing But can confirm `FLASK_ENV=development` is required (ran through your steps to be sure).

I've fixed the docs accordingly! `FLASK_DEBUG=True` not working seems to be a bug?

pandemicsyn:
> FLASK_DEBUG=True not working seems to be a bug?

I'm not certain, actually.  I don't know if FLASK_DEBUG=True should mean that cors restrictions/host checks are loosened automatically, feels like that might be an unintended behavior change (i.e. if you're trying to debug your production deployment or the like).

@meltano/engineering curious how y'all would expect FLASK_DEBUG=True to behave.

WillDaSilva:
> curious how y'all would expect FLASK_DEBUG=True to behave.

@pandemicsyn 

https://flask.palletsprojects.com/en/2.2.x/config/

image

https://flask.palletsprojects.com/en/2.1.x/config/

image

Based off of these docs, I would expect `FLASK_DEBUG=True` to do whatever `FLASK_ENV=development` does, since the docs say to use `--debug` as a replacement for env.

---

@dingobar What version of Python did you use for the reproduction you provided in https://github.com/meltano/meltano/pull/6759#issuecomment-1250862353?

I tried to reproduce the issue in a clean Python 3.8 container following your instructions, but was unable to. I ran into numerous build issues with step 4:
> In another terminal, cd meltano and then `yarn build && yarn serve`

Also the steps confuse me - are you serving the UI with `meltano ui` or `yarn serve`? Or some mix thereof?

As far as I know, serving the UI with `yarn serve` is not supported.

If you want to build and serve the UI for development, you need to cd into `src/webapp/`, and run `yarn && yarn build` from there, then cd back to the project root, copy the built UI into the proper locations, install the project into your Python environment, then run `meltano ui`.

This is currently encapsulated by `make bundle`, which handles everything before installing the Python package, and running `meltano ui`. (In the future this will be automatically performed as part of installing the Python package, see #6488 and #6262.)

Seems to me like [the UI dev docs](https://github.com/dingobar/meltano/blob/bf76d826df2252b1dba819a69b7c0a974be835b0/docs/src/_contribute/ui.md) are out of date.

As for the CORS issue, @dingobar does it occur if you use the `meltano ui` command exclusively with `FLASK_ENV` and `FLASK_DEBUG` unset?

pandemicsyn:
> Also the steps confuse me - are you serving the UI with meltano ui or yarn serve? Or some mix thereof?

@WillDaSilva its both  


> As far as I know, serving the UI with yarn serve is not supported.

> If you want to build and serve the UI for development, you need to cd into src/webapp/, and run yarn && yarn build from there, then cd back to the project root, copy the built UI into the proper locations, install the project into your Python environment, then run meltano ui.

That's not required (fwiw I've never had to do that and always spin up the dev instance via yarn). The UI Dev guide as additional details - https://docs.meltano.com/contribute/ui#ui-development

Issue #6760, opened 2022-09-15
feature: allow disabling update checks in UI


dingobar:
### Feature scope

Web UI

### Description

See https://github.com/meltano/meltano/issues/6750#issuecomment-1247116799

Figured I'd file a separate issue for it.

Disabling the update check in the UI while still displaying the current version.

PR to follow shortly.

WillDaSilva:
Thank you for the issue @dingobar. As you've hopefully heard by now, the Meltano UI has been deprecated, and is scheduled for removal in Meltano 3.0. Since no additional work/improvements for the UI are planned leading up to its removal, I'll be closing this issue.

PR #6761, opened 2022-09-15
feat: add `ui.update_check` configuration to disable PyPi-based version check in UI

martimors:
Took a stab at solving this [https://github.com/meltano/meltano/issues/6750#issuecomment-1247116799 .](https://github.com/meltano/meltano/issues/6760).

First time messing with Vue.js so bear with me here :pray:

I guess the UI doesn't have a solidly defined concept of an update check in the first place, so the implementation here feels a bit hacky. Basically, the switch just stops the UI from requesting the latest version from the backend when the flag is set to false, and sets the latest version to `null`. The consequence of this is already handled in the `NavBar` component with a null check.

I've tested it manually and it seems to work, however I wonder how to test this with the proper flask-based context manager?

**I'm missing the docs here** but figured it's best to get a review on the implementation first.

### DIFF
--- a/docs/src/_reference/settings.md
+++ b/docs/src/_reference/settings.md
@@ -464,14 +464,20 @@ The host to bind to.
 Together with the [`ui.bind_port` setting](#ui-bind-port), this setting corresponds to
 [Gunicorn's `bind` setting](https://docs.gunicorn.org/en/stable/settings.html#bind).
 
+### <a name="ui-update-check"></a>`ui.update_check`
+
+- [Environment variable](/guide/configuration#configuring-settings): `MELTANO_UI_UPDATE_CHECK`
+- Type: `boolean`
+- Default: `true`
+
+Whether to check for newer Meltano versions and display update notifications in the UI.
+
 #### How to use
 
 ```bash
-meltano config meltano set ui bind_host 127.0.0.1
+meltano config meltano set ui update_check false
 
-export MELTANO_UI_BIND_HOST=127.0.0.1
-
-meltano ui --bind=127.0.0.1
+export MELTANO_UI_UPDATE_CHECK=false
 ```
 
 ### <a name="ui-bind-port"></a>`ui.bind_port`
@@ -492,7 +498,6 @@ meltano config meltano set ui bind_port 80
 
 export MELTANO_UI_BIND_PORT=80
 
-meltano ui --bind-port=80
 ```
 
 ### <a name="ui-server-name"></a>`ui.server_name`
--- a/.gitignore
+++ b/.gitignore
@@ -125,6 +125,7 @@ ENV/
 env.bak/
 venv.bak/
 .env.dev
+.nvmrc
 
 # Spyder project settings
 .spyderproject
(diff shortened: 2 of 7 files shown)


pandemicsyn (approved):
lgtm! 

> I've tested it manually and it seems to work, however I wonder how to test this with the proper flask-based context manager?

I'm not exactly sure what you mean by "test this with the proper flask-based context manager" but to be honest, tests are pretty sparse on the webapp and non-existent on the UI (and UI seem's like where you'd really wanna exercise this). 

/cc @cjohnhanson mind double checking me and giving this a quick once over ?

aaronsteers:
@dingobar - I took a quick stab at adding docs and added as a commit here in the PR.

@cjohnhanson and @tayloramurphy - If you don't have any further changes or requests, this appears ready to go.

aaronsteers (approved):

docs/src/_reference/settings.md
```diff
@@ -464,14 +464,20 @@ The host to bind to.
...
+- [Environment variable](/guide/configuration#configuring-settings): `MELTANO_UI_UPDATE_CHECK`
+- Type: `boolean`
+- Default: `true`
+
+Whether to check for newer Meltano versions and display update notifications in the UI.
+
 #### How to use
 
 ```bash
-meltano config meltano set ui bind_host 127.0.0.1
```
tayloramurphy, line 470: Are we intending to delete these? https://docs.meltano.com/reference/settings#meltano-ui-server

docs/src/_reference/settings.md
```diff
@@ -464,14 +464,20 @@ The host to bind to.
 Together with the [`ui.bind_port` setting](#ui-bind-port), this setting corresponds to
 [Gunicorn's `bind` setting](https://docs.gunicorn.org/en/stable/settings.html#bind).
 
+### <a name="ui-update-check"></a>`ui.update_check`
```
tayloramurphy, line 467: In a separate PR we should probably alphabetize these

tayloramurphy (changes requested):
Some docs updates don't seem right.

cjohnhanson (approved):
@dingobar Looks good! Thanks for taking the time to put this together.

Agreed with @pandemicsyn, we don't currently test this route on the backend and since this is really just a change to the _frontend_ logic where we're no longer requesting for the `/version` endpoint to include the latest version, I don't think there are any new tests required. 

Side note: it's interesting that this endpoint already supported the `include_latest` param but it was always just sent as `true` from the frontend. 

Another side note/follow-up for me: in reviewing this I noticed that the `/version` endpoint is hardcoded to point at PyPI when checking for the latest available version, but I think (not 100% sure) the `/upgrade`  endpoint will use whatever `pip` is pointed towards locally. Not a big deal, but I think it's somewhat common for even mid-sized organizations to host their own private package repositories for security reasons, which could cause issues here. I'll make a follow-on issue for us.

docs/src/_reference/settings.md
dingobar: Hi, yeah, I'll see if I can finish it up this week. Having a lot on my plate at the moment!

dingobar:
I'm closing this for now because there are merge conflicts and it's been hanging for a while. Anyone else feel free to pick it up :)

PR #6764, opened 2022-09-16
docs: Move installation guide to "Getting Started" add Part 1 tutorial

sbalnojan:
Starting to remodel the docs and filling up the getting started section with a long, broken down tutorial. Still want to keep the "getting started guide" for a while, because the tuts are much more reduced and I need way more time to move that content into the proper place (tuts and how tos).


docs/src/_getting-started/part1.md
```diff
@@ -0,0 +1,217 @@
...
+In Part 2, we will then place this data into a PostgreSQL database.
+
+We'll assume you have [Meltano installed](/getting-started/installation) already. You can tell Meltano is installed and which version by running ``` meltano version```
+
+![Meltano Version](images/part1/gif_meltano_version.gif)
+
+
+This tutorial is written using meltano >= v2.0.0.
+
+If you don't have a GitLab account to follow along, you could either exchange the commands for a differe tap, like GitHub or PostgreSQL, or you create a free GitLab account. You will also need a [personal access token to your GitLab account](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) which you can create under "Edit Profile > "Access Tokens" > .
```
tayloramurphy, line 22:
```suggestion
If you don't have a GitLab account to follow along, you could either exchange the commands for a different tap, like GitHub or PostgreSQL, or you create a free GitLab account. You will also need a [personal access token to your GitLab account](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) which you can create under "Edit Profile > "Access Tokens" > .
```

docs/src/_getting-started/part1.md
```diff
@@ -0,0 +1,217 @@
...
+   ```
+
+![Meltano Init](images/part1/gif_meltano_init.gif)
+
+   This action will create a new directory with, among other things, your [`meltano.yml` project file](/concepts/project#meltano-yml-project-file). Your file will look something like this:
+
+   ```yml
+   version: 1
+   default_environment: dev
+   project_id: u3u5u4ueudufhfh4h4h5ddkjh
```
tayloramurphy, line 48:
```suggestion
   project_id: <unique-GUID>
```

It'll default to a true unique GUID so I'd recommend either having a guid there are just indicating that it can be a guid.

docs/src/_getting-started/part1.md
```diff
@@ -0,0 +1,217 @@
...
+title: Tutorial Part 1 - Extract Data
+description: Part 1 - If you're ready to get started with Meltano and run an EL[T] pipeline with a data source and destination of your choosing, you've come to the right place!
+layout: doc
+weight: 3
+---
+Let’s learn by example.
+
+Throughout this tutorial, we’ll walk you through the creation of a end-to-end modern E(t)LT stack.
+
+We're going to take data from a "source", namely GitLab, and extract a list of projects, the authors, messages and so on.
```
tayloramurphy, line 11: Couple of questions on this:

* do you think it's a problem to have a source that a user will need to have an account on? I think GitLab may actually require a credit card now to signup (I could be wrong though). We could have a source that doesnt' require authtentication
* Should we prefer GitHub to GitLab? The bet is that folks will be more likely to complete the tutorial if they have an existing GH account.

docs/src/_getting-started/part1.md
```diff
@@ -0,0 +1,217 @@
...
+
+1. Navigate to the newly created project directory:
+
+   ```bash
+   cd my-meltano-project
+   ```
+
+## Add an Extractor to Pull Data from a Source
+
+Now that you have your very own Meltano project, it's time to add [plugins](/concepts/plugins) to it. We're going to add an extrator for GitLab to get our data. An [extractor](/concepts/plugins#extractors) is responsible for pulling data out of any data source.
```
tayloramurphy, line 63: Not blocking for this PR, but we are planning to rename plugins to components. Just wanted to give you a heads up https://github.com/meltano/meltano/discussions/6561 no exact timeline on this, but likely 3.0

docs/src/_getting-started/part1.md
```diff
@@ -0,0 +1,217 @@
...
+   $ meltano config tap-gitlab set --interactive
+   ```
+2. Follow the prompts to step through all available settings, the ones you'll need to fill out are projects, start_date and your private_token.
+
+![Meltano Version](images/part1/gif_meltano_config_interact.gif)
+
+  This will add the non-sensitive configuration to your [`meltano.yml` project file](/concepts/project#plugin-configuration):
+
+  ```yml
+  environments:
```
tayloramurphy, line 108: By default now, `meltano config` does _not_ use the default environment. The config will just be under the top level `plugins` key. Documented in https://docs.meltano.com/concepts/environments#default-environments

docs/src/_getting-started/part1.md
```diff
@@ -0,0 +1,217 @@
...
+
+1. Find out what entities and attributes are available, using [`meltano select --list --all`](/reference/command-line-interface#select):
+
+   ```bash
+   meltano select tap-gitlab --list --all
+   ```
+  ![Meltano Version](images/part1/gif_meltano_select_all.gif)
+
+
+1. select the entities and attributes for extraction using [`meltano select`](/reference/command-line-interface#select):
```
tayloramurphy, line 143:
```suggestion
1. Select the entities and attributes for extraction using [`meltano select`](/reference/command-line-interface#select):
```

docs/src/_getting-started/part1.md
```diff
@@ -0,0 +1,217 @@
...
+Now that the extractor has been configured, it'll know where and how to find your data, but won't yet know which specific entities and attributes (tables and columns) you're interested in.
+
+By default, Meltano will instruct extractors to extract all supported entities and attributes, but we're going to [select specific ones for this tutorial](/guide/integration#selecting-entities-and-attributes-for-extraction). 
+
+1. Find out what entities and attributes are available, using [`meltano select --list --all`](/reference/command-line-interface#select):
+
+   ```bash
+   meltano select tap-gitlab --list --all
+   ```
+  ![Meltano Version](images/part1/gif_meltano_select_all.gif)
```
tayloramurphy, line 140:
```suggestion

  ![Meltano Version](images/part1/gif_meltano_select_all.gif)
```

Adding the new line may create a bit more breathing room between the elements.

docs/src/_getting-started/part1.md
```diff
@@ -0,0 +1,217 @@
...
+
+   Note that exclusion takes precedence over inclusion, always.
+
+1. Run [`meltano select --list`](/reference/command-line-interface#select) to double-check your selection:
+
+   ```bash
+   meltano select tap-gitlab --list
+   ```
+
+## Add a dummy loader to dump the data into JSON
```
tayloramurphy, line 193: Instead of a dummy loader, I think there's (2) simpler paths:

* run `meltano invoke tap-gitlab` and see if data is output on the command line
* run `meltano config tap-gitlab test`. That may not be supported if it's not an SDK-based tap. https://docs.meltano.com/reference/command-line-interface#how-to-use-1

I like not introducing a loader at this stage because it gets folks thinking about testing their extractor in isolation. What do you think?

tayloramurphy: I do see that it would add an extra step between getting them to a `meltano run` command, so perhaps that's an argument against.

docs/src/_getting-started/part1.md
```diff
@@ -0,0 +1,217 @@
...
+
+There's just one step here: run your newly added extractor and jsonl loader in a pipeline using [`meltano run`](/reference/command-line-interface#run):
+
+```bash
+$ meltano run tap-gitlab target-jsonl
+```
+![Meltano Version](images/part1/gif_meltano_run.gif)
+
+You should see data flowing from your source into the jsonl file.
+You can verify that it worked by looking inside the newly created file called ```output/commits.jsonl```.
```
tayloramurphy, line 214: Perhaps here we can keep them on the command line and have them run `cat output/commits.jsonl`?

docs/src/_getting-started/part1.md
```diff
@@ -0,0 +1,217 @@
+---
+title: Tutorial Part 1 - Extract Data
```
tayloramurphy, line 2: Meta question - since this isn't in the tutorial section, do we want it to be a walkthrough? How are you thinking about that structure?

tayloramurphy (changes requested):
@sbalnojan love where this is heading. A few meta questions and one bigger bug in here, but it's looking good.

docs/src/_getting-started/part1.md
sbalnojan: Mh I just signed up for GitLab, so it doesn't. But yes I do think it's a problem, but I couldn't think of a better alternative at the moment. GitHub could be a better option. I definitely think it should be an external source, not something local, otherwise it's hard to convince people of the feeling of doing actual data integration. So yes we could switch over.

docs/src/_getting-started/part1.md
sbalnojan: Which part are you commenting on? I get that I need to change something, but not sure what :-D I do want to keep out the environment config part in part 1 of the tut if possible.

docs/src/_getting-started/part1.md
sbalnojan: agree.

docs/src/_getting-started/part1.md
sbalnojan: Mh, I agree the loader is a (light) additional weight. Ideally I would like taps to dump to file with some optional flag, without me needing to add this. Or just printing the data straight to stdout (just the data, nothing else). I added this to just get the data, and nothing else, the data as it will flow into the next step. 

Not sure either here.

docs/src/_getting-started/part1.md
sbalnojan: It's still a tutorial. The way I think about it is:
- "Getting Started" is a section which hosts just one tutorial. "The Meltano Tutorial".
- Then we add an "Advanced Tutorials" section.

Fwiw, I don't think the naming is that important, but rather the intent. I'd aim for having both Getting Started & Advanced Tutorials as tutorial sections (in the diataxis framework).

docs/src/_getting-started/part1.md
sbalnojan: found it.

docs/src/_getting-started/part1.md
tayloramurphy: @sbalnojan makes sense! 

Also, don't forget to tag me in basically every comment, otherwise I won't see it. I only saw this because you re-requested a review, so it worked out 😄

tayloramurphy (approved):
@sbalnojan love the updates on this. 

Related to our recent conversation, this makes me think of two ways we could get folks started faster: importing a template and interactive select. We have an issue for the former and I think AJ was going to open one on the latter.

edgarrmondragon (commented):
Left a few suggestions for whitespace issues

docs/src/_getting-started/part1.md
sbalnojan: @tayloramurphy will do.

PR #6765, opened 2022-09-16
chore: Use pyyaml v6

WillDaSilva:
The EDK and extensions created with it required pyyaml `^6.0.0`. As a result is no version of Meltano which can be installed alongside them. In a test environment, I'd want to be able to install Meltano and the extension.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -155,21 +155,21 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
 
 [[package]]
 name = "attrs"
-version = "21.4.0"
+version = "22.1.0"
 description = "Classes Without Boilerplate"
 category = "main"
 optional = false
-python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
+python-versions = ">=3.5"
 
 [package.extras]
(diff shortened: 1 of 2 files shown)


pandemicsyn:
Is it better to be a little loser on the requirements on the EDK side and see if we can get away with 5.3.1 here? I just defaulted to latest pyyaml on the EDK because I didn't think to explicitly asked for a version. Wondering if bumping this on the meltano side will be problematic for other folks.

On the flip side - keeping packages current is nice...

WillDaSilva:
> Is it better to be a little loser on the requirements on the EDK side and see if we can get away with 5.3.1 here? I just defaulted to latest pyyaml on the EDK because I didn't think to explicitly ask for a version. Wondering if bumping this on the meltano side will be problematic for other folks.
> 
> On the flip side - keeping packages current is nice...

@pandemicsyn Either way introduces incompatibilities. Us updating Meltano to use the newer version doesn't prevent anyone from continuing to use an older version of Meltano with their older deps, so I'd lean towards keeping packages current.

pandemicsyn (approved):
lgmt. 

But definitely worth seeking another +1 from @edgarrmondragon or @aaronsteers to be safe.

edgarrmondragon (approved):

edgarrmondragon:
> lgmt.
> 
> But definitely worth seeking another +1 from @edgarrmondragon or @aaronsteers to be safe.

I'd even prefer removing any upper limits on dependencies, but we've been bitten by that before 🤷

PR #6781, opened 2022-09-22
docs: add parts 2-4 to the tutorial

sbalnojan:
- add parts 2-4 of the tutorial
- fix empty lines etc.


docs/src/_getting-started/part4.md
```diff
@@ -0,0 +1,194 @@
...
+In the previous parts, we set up a E(t)L pipeline, extracting data from GitHub, transforming it inline, and loading it into a local PostgreSQL database.
+
+In this part, we're going to unleash the data-built-tool [dbt](https://www.getdbt.com/) onto our data to transform it into meaningful information.
+
+<div class="notification is-success">
+    <p>If you're having trouble throughout this tutorial, you can always head over to the <a href="https://meltano.com/slack">Slack channel</a> to get help.</p>
+</div>
+
+## Install and configure the postgres specific dbt transformer
+Dbt transformers come in flavors. Our flavor is `dbt-postgres`. As usual, you can use the `meltano add` command to add it to your project.
```
tayloramurphy, line 22:
```suggestion
Dbt uses different [adapters](https://docs.getdbt.com/docs/supported-data-platforms) depending on the database/warehouse/platform you use. Meltano transformers match this pattern; in this case our transformer is `dbt-postgres`. As usual, you can use the `meltano add` command to add it to your project.
```

docs/src/_getting-started/part4.md
```diff
@@ -0,0 +1,194 @@
...
+  config(
+    materialized='table'
+  )
+}}
+
+
+with base as (select *
+from {{ source('tap_github', 'commits') }}) {% endraw %}
+
+select distinct (commit -> 'author' -> 'name') as authors from base
```
tayloramurphy, line 117:
```suggestion
WITH base AS (

  SELECT *
  FROM {{ source('tap_github', 'commits') }}
  
) {% endraw %}

SELECT 
  DISTINCT (commit -> 'author' -> 'name') AS authors 
FROM base
```

I'm going to be annoying about structuring SQL https://handbook.meltano.com/data-team/sql-style-guide 😈

docs/src/_getting-started/part4.md
```diff
@@ -0,0 +1,194 @@
...
+#
+```
+
+</div>
+
+You can check the data inside the database. There should now be a table inside the `analytics` schema called `authors` populated with data.
+
+## Run the complete pipeline
+
+To check that everything works together as a pipeline, we clean out once more and run the whole E(t)LT pipeline. Drop the tap_github.commits and the analytics.authors tables, clear the state using `meltano state clear dev:tap-github-to-target-postgres` and run the final pipeline alltogether:
```
tayloramurphy, line 153: Similar comment as above. It may be reasonable to have a utility which spins up a local postgres database and has some of these commands ready to go 🤔 but that can be in another iteration perhaps.

docs/src/_getting-started/part4.md
```diff
@@ -0,0 +1,194 @@
...
+[info     ] 12:43:21                       cmd_type=command name=dbt-postgres stdio=stderr
+[info     ] 12:43:21  Completed successfully cmd_type=command name=dbt-postgres stdio=stderr
+info     ] 12:43:21                       cmd_type=command name=dbt-postgres stdio=stderr
+[info     ] 12:43:21  Done. PASS=1 WARN=0 ERROR=0 SKIP=0 TOTAL=1 cmd_type=command name=dbt-postgres stdio=stderr
+[info     ] Block run completed.           block_type=InvokerCommand err=None set_number=1 success=True
+```
+
+</div>
+
+There we have it, a complete E(t)LT pipeline.
```
tayloramurphy, line 187:
```suggestion
There we have it, a complete E(t)LT pipeline! Congratulations!
```

Thoughts on having a fun picture of Melty or something like that to celebrate?

Also, you can have them run `meltano dragon` to generate a dragon and celebrate that way 😄 Just thinking of a way to make it a bit more playful.

docs/src/_getting-started/part4.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Next Steps
+
+Next, head over to [Part 5, inside the Getting Started Guide](/getting-started/#add-a-loader-to-send-data-to-a-destination).
```
tayloramurphy, line 191:
```suggestion
Next, head over to [Part 5, inside the Getting Started Guide](/getting-started/#next-steps).
```

tayloramurphy (changes requested):

aaronsteers:
@sbalnojan and @tayloramurphy - What do you think about using `meltano config target-postgres --interactive`? This is a more guided experience which could replace:

- `meltano config target-postgres list`
- the series of `meltano config target-postgres <setting>` commands
- the `.env` edits

docs/src/_getting-started/part3.md
```diff
@@ -0,0 +1,176 @@
+---
+title: Part 3 - Inline Transformations, E(t)L
```
aaronsteers, line 2: @tayloramurphy - We've been very careful to avoid 'transforms' and the EtLT because it doesn't translate well and _appears_ on the surface to conflict with our core best practices. I don't want to use the word "transform" or "transformation" anywhere in the header here. I definitely would want to avoid the "little 't'" terminology because it then makes subsequent conversations about "the transformation layer" subject to the constant need to clarify: "which one?".

Even though mappers are very cool, my preference would be to keep mappers out of this initial tutorial and maybe move to an 'advanced topics' section. The other factor here is that we already have a large-enough challenge in getting users (from a variety of backgrounds) onboarded onto "what is ELT" and "how do I use Meltano for ELT". I think as an advanced topic, we'd have more time to let this breath.


Happy to chat more about this in a sidebar if helpful.

aaronsteers (dismissed):

tayloramurphy:
@aaronsteers we use interactive on https://docs.meltano.com/getting-started/part1#configure-the-extractor already, so I'd be in favor of adding it. @sbalnojan ?

docs/src/_getting-started/part4.md
afolson: Inline :melty-bouncy: maybe?

docs/src/_getting-started/part3.md
```diff
@@ -0,0 +1,176 @@
...
+    pip_url: pipelinewise-transform-field
+    executable: transform-field
+   mappings:
+    - name: hide-github-mails
+      config:
+        transformations:
+          - field_id: "commit"
+            tap_stream_name: "commits"
+            field_paths: ["author/email", "committer/email"]
+            type: "MASK-HIDDEN"
```
aaronsteers, line 98: @sbalnojan and @tayloramurphy - Separate from my prior comment proposing we add mappings _after_ the "T" in "ELT":

Whereever we put the content here (and I agree it is valuable), I think it's probably best if we promote the `meltano-map-transfomer` plugin, which we own - and not point people in the direction of `transform-field`. The rationale would be: (1) our plugin has slightly better features, (2) for SDK-based taps and targets, the config will basically be drop-in compatible across taps, targets, and the dedicated mapper, and (3) we don't have ability to improve the `field-transform` plugin when people run into issues, since we aren't maintainers on it.

docs/src/_getting-started/part3.md
aaronsteers: On thinking about this a bit more...

The primary use case in the industry for `mappers` and `mappings` is data obfuscation and data protection - meaning: we actively want the data and don't want the data at the same time. Replicating the data in clear-text is a no-go for security reasons (or at least prohibitively expensive in terms of our liability), but not having the data at all is debilitating to our ability to get insights.

With this framing, if we simply swap the order of `part-3` and `part-4`, one path we could take is to say:

> In step 1 we intentionally did not import email addresses due to privacy considerations. Now we will use the "stream maps" advance feature to import only an anonymized one-way hash of the email address. This will allow downstream analytics applications to calculate "how many unique emails are there?" without ever importing those emails into our database.

Then the section header might be focused on "advanced data protection" or similar, as a specific use case for stream maps.

tayloramurphy: @aaronsteers I'm in favor of the following:

* Swapping parts 3 and 4. By getting to the true dbt transformation first you run the full pipeline and complete the ELT
* Keeping Part 4 but not labeling it advanced. Stream maps are a key feature that separates us from other tools and should be highlighted and promoted
* Renaming the title to something akin to "Inline Data Mapping". This moves us away from the transformation language in the header. IN the body we talk about this type of transformation and how it's distinguished from dbt-style transformation
* Data protection is arguably the main use for mappers but they're powerful and can be used for other things as well. The framing I like is that it gives the user a lot of control over how the data lands. Data protection is one application of that
* We should figure out how to test and talk about E(t)LT. as much as I like EMLT or E(m)LT, nobody knows what we're talking about and at least E(t)LT gets people to understand that it's similar to ETL but clearly something is different!

docs/src/_getting-started/part3.md
```diff
@@ -0,0 +1,176 @@
...
+"sbalnojan/meltano-example-el"}, "time_extracted": "2022-09-22T07:37:06.289545Z"}
+
+...[many more records]...
+
+{"type": "STATE", "value": {"bookmarks": {"sbalnojan/meltano-example-el": {"commits": {"since": "2022-09-22T07:37:06.289545Z"}}}}}
+´´´
+</div>
+
+## Installing the transform-field mapper
+To add inline transformations, we need a new plugin. We're going to use the mapper "transform-field". To add this plugin, use the `meltano add mapper` command:
```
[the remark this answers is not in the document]
sbalnojan: Sure, I just wouldn't put it into the tut, but into the more detailed mapper explanation (which kind of isn't there yet...). That would probably confuse more than add value here esp. if someone then tries to find the docs on doing it differently.

sbalnojan:
> @sbalnojan and @tayloramurphy - What do you think about using `meltano config target-postgres --interactive`? This is a more guided experience which could replace:
> 
> * `meltano config target-postgres list`
> * the series of `meltano config target-postgres <setting>` commands
> * the `.env` edits
> 
> Related to steps here: https://deploy-preview-6781--meltano.netlify.app/getting-started/part2#configure-the-target-postgres-loader
> 
> I don't know how well it would render in the termy visualizer, but we could list the config values the user needs to set and then instruct them to step through `--interactive` and enter each value in the respective spot.

I'm all for it. I just forgot to do that. Btw. everything renders well in termy if you take the time to make it render well. The hard part is just getting the ANSI text converted to HTML and then doing a bit of editing to match the size (which I haven't figured out yet. ansi2html doesn't do it for me).

docs/src/_getting-started/part3.md
sbalnojan: That was my plan, but quite frankly, I wasn't able to get it to work after a full day, so I had to opt for what works out of the box (mostly), even this mapper is IMHO way below average in terms of usability. 

So I propose:
- we push it out as it is
- someone adds docs to the meltano-map-transformer (I already highlighted that on a separate issue)
- we swap it out afterwards.

docs/src/_getting-started/part3.md
sbalnojan: @aaronsteers @tayloramurphy 

- I definitely favor keeping it in the tutorial, have no strong opinion on whether 3 or 4 (fwiw, I chose 3 because it seemed to be an easier progression in terms of difficulty of the tutorial, because it's still inside the meltano context mostly, unlike dbt which leaves that context and requires the user to potentially also learn about dbt).
- I think this is way more than a nice to have. The core to solving data integration problems are the connectors. Thus they come in a generic form, which means we will always need to have the possibility to manipulate the output afterwards in some slight way. Same goes for all the big APIs of SaaS solutions. They are never tailored to our specific request, so having mappers is a pretty powerful tool here. 
- I would also favor the word "Inline Data Mapping" over the (one) use case of "Data protection", but thats mostly because in the customer interviews someone provided a good use case which simply was different (I think they added an identifier to the data). And quite a few others (related to the above point) might also go into a different direction.

I like the E(t)LT acronym because it's real easy to explain: We went from ETL to ELT because ELT gives us 10x more benefits. Turns out, small transformations, we call them more appropriately "inline data mappings" are very useful, so we added them back in (in a way that encourages small transformations, and discourages large ones).

docs/src/_getting-started/part4.md
sbalnojan: Probably, but it's a tutorial, it's about teaching. I felt it's ok to teach people about state a bit in part 4, though I agree that it would probably also fine to wrap it away, might make adoption even higher.

docs/src/_getting-started/part3.md
```diff
@@ -0,0 +1,176 @@
...
+```
+These lines define one transformation. We instruct to target the stream "commits", and therein the field "commit". We then use the field paths to navigate to the two emails we know are contained within this message and set the type to "MASK-HIDDEN" to hide the values.
+
+## Run the data integration (E(t)L) pipeline
+Now we're ready to run the data integration process with these modifications again. To do so, we'll need to clean up first, since we already ran the EL process in part 1. The primary key is still the same and as such the ingestion would fail.
+
+Drop the table inside your local postgres by running
+
+```sql
+DROP TABLE tap_github.commits
```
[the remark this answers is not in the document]
sbalnojan: I don't know, had the same reaction, different schema isn't cool because we want to do it again in part 4. I could wrap it into a docker command, after all they need to have docker anyways.

docs/src/_getting-started/part3.md
tayloramurphy: > I like the E(t)LT acronym because it's real easy to explain

I like it for that as well too. EtLT or E(t)LT renders nicely, connects to what people knows but is different enough to invite conversation.

sbalnojan:
Alright @aaronsteers & @tayloramurphy, I implemented all the changes, I think but merge hell might've caught me off guard here and there. Next time, smaller PRs. 

- I switched part 3 and part 4 and adapted them accordingly
- I implemented docker exec commands for the delete stuff
- I couldn't  add the --interactive config, because I couldn't get it to render nicely (which would require to kind of cut off the width. And that's some serious editing in the CLI output, not too much fun). 
- changed the language as suggested to "inline data mappings".#

docs/src/_getting-started/part4.md
```diff
@@ -0,0 +1,136 @@
...
+Added mapper 'transform-field' to your Meltano project
+Variant:        transferwise (default)
+Repository:     https://github.com/transferwise/pipelinewise-transform-field
+Documentation:  https://docs.meltano.com/concepts/plugins#mappers
+
+Installing mapper 'transform-field'...
+---> 100%
+Installed mapper 'transform-field'
+
+To learn more about mapper 'transform-field', visit https://hub.meltano.com/mappers/transform-field
```
tayloramurphy, line 38:
```suggestion
To learn more about mapper 'transform-field', visit https://docs.meltano.com/concepts/plugins#mappers
```

tayloramurphy (approved):
On the `(t)` framing, we could put a call to action to the community if they have opinions on the framing of EMLP vs EtLT etc etc. Might be too distracting for a tutorial thought. Approving - merge at will @sbalnojan

aaronsteers:
@tayloramurphy and @sbalnojan - For reference, here's the discussion where we talked through this at great length and agreed to not use "transformation" for "stream maps": https://github.com/meltano/meltano/issues/3051

(Click through to the GitLab side for historic comments.)

sbalnojan:
@aaronsteers I'm not clinging to the (t). I'm more interested in making any decision now and then be consistent. Then let's put that into the other issue/discussion you opened up (I am unable to find right now) and create a follow up to clean up the language here & there (for me).

PR #6785, opened 2022-09-22
feat: Use `NO_COLOR` env var to disable colored output

### LINKED ISSUE #6482: feature: Add support for `NO_COLOR` env var
Latest spec:

- [ ] Parse a `NO_COLOR` env var per instructions at https://no-color.org
- [ ] Optionally, also accept `--no-color` at the CLI.
- [ ] Pass this setting to logging configs.
- [ ] Pass this setting to `click`.
- [ ] Disable or remove color from `rich` exception logging, as needed.

Notes:

- This is not to be a feature flag or a Meltano setting, and therefor will not be able to be set using `meltano config`.
- This should have the equivalent affect as customizing `logging.yaml`. (Bug fixes in `color=false` implementation should solve for both methods of setting.)

edgarrmondragon:
# Summary

- ~New `meltano --no-color` option to disable color~. Not in scope. See https://github.com/meltano/meltano/pull/6785#issuecomment-1287386670
- New `NO_COLOR` environment variable used to disable color
- Disables color in the default logs
- Disables color in logs even if `meltano.core.logging.console_log_formatter` is called with `colors: true`

Closes https://github.com/meltano/meltano/issues/6482


tests/meltano/cli/test_cli.py
```diff
@@ -167,6 +169,58 @@ def test_handle_meltano_error(self):
...
+        def _dummy_command(ctx):
+            click.echo(styled_text)
+
+        args = ["dummy"]
+        expected_text = styled_text if colors_expected else text
+
+        if no_color_flag:
+            args = ["--no-color"] + args
+
+        with cli_runner.isolated_filesystem():
```
edgarrmondragon, line 217: The random order meant that some other test was leaving an empty `meltano.yml` that was causing problems. [`isolated_filesystem`](https://click.palletsprojects.com/en/8.1.x/testing/#file-system-isolation) just ensures  the CLI is invoked in an empty directory.

aaronsteers:
> Not included in this PR\nDisable color in the default logs. This requires modifying the default logging config to configure the Rich console object with disabled colors.

Can you say more about this?

I'm not sure I follow.

docs/src/_reference/command-line-interface.md
```diff
@@ -9,6 +9,15 @@ Meltano provides a command line interface (CLI) that makes it easy to manage you
 To quickly find the `meltano` subcommand you're looking for, use the Table of Contents in the sidebar.
 For a better understanding of command line documentation syntax, the [docopt](http://docopt.org/) standard is useful.
 
+## Global Options
+
+The following options are available for all `meltano` subcommands:
+
+* `--enviroment` - The [Meltano Environment](/concepts/environments) to use. Defaults to `dev`. Can also be set with the `MELTANO_ENVIRONMENT` environment variable.
```
tayloramurphy, line 16: @edgarrmondragon this isn't really true though, right? Commands like `init` and `ui` don't do anything with `--environment`. 

We have new subsections for the relevant commands on using `--environment` (see https://docs.meltano.com/reference/command-line-interface#using-add-with-environments) so I think we can remove this line.

tayloramurphy (changes requested):
Minor change request on docs

docs/src/_reference/command-line-interface.md
```diff
@@ -9,6 +9,15 @@ Meltano provides a command line interface (CLI) that makes it easy to manage you
...
+* `--log-config` - Path to a logging configuration file. See [Logging](/guide/logging) for more information.
+* `--log-level`: Set the log level for the command. Valid values are `debug`, `info`, `warning`, `error`, and `critical`.
+* `--no-color`: Disable colored output. Can also be set with the `NO_COLOR` environment variable.
```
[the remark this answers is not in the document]
edgarrmondragon: TIL `click` is smart enough to interpret `NO_COLOR=FALSE` as toggling off the flag.

src/meltano/core/logging/utils.py
```diff
@@ -60,23 +64,30 @@ def read_config(config_file: str | None = None) -> dict:
...
         log_level: set log levels to provided level.
+        no_color: Enabled no color mode, or None to auto detect. Defaults to None.
 
     Returns:
          dict: logging config suitable for use with logging.config.dictConfig
     """
+    if no_color:
+        formatter = rich_exception_formatter_factory(no_color=True)
+    else:
+        formatter = rich_exception_formatter_factory(color_system="truecolor")
```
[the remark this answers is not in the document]
edgarrmondragon: @WillDaSilva This is what structlog's built-in Rich traceback formatter uses:

```python
def rich_traceback(sio: TextIO, exc_info: ExcInfo) -> None:
    """
    Pretty-print *exc_info* to *sio* using the *Rich* package.
    To be passed into `ConsoleRenderer`'s ``exception_formatter`` argument.
    Used by default if *Rich* is installed.
    .. versionadded:: 21.2
    """
    sio.write("\n")
    Console(file=sio, color_system="truecolor").print(
        Traceback.from_exception(*exc_info, show_locals=True)
    )
```

https://github.com/hynek/structlog/blob/ec2fbefa39d29a5a9841c4d9744c090350b4fc83/src/structlog/dev.py#L175-L188

From the [Rich docs](https://rich.readthedocs.io/en/stable/console.html#color-systems):

> "truecolor" Can display 16.7 million colors, which is likely all the colors your monitor can display.

docs/src/_reference/command-line-interface.md
edgarrmondragon: @tayloramurphy done: 8daebcbe15ba57f9c911b8e9e14bd768a526eca3

edgarrmondragon:
> > Not included in this PR
> > Disable color in the default logs. This requires modifying the default logging config to configure the Rich console object with disabled colors.
> 
> Can you say more about this?
> 
> I'm not sure I follow.

@aaronsteers sure.

Colors in "normal" `click` echo calls are toggled by the flag added here, but colors in structured logs are still driven by `logging.yaml`, e.g. if `meltano.core.logging.console_log_formatter` is used with `colors: true`, like

```yaml
version: 1
disable_existing_loggers: false
formatters:
  structured_plain: # log format for structured plain text logs without colored output
    (): meltano.core.logging.console_log_formatter
    colors: False
  structured_colored: # log format for structured plain text logs WITH colored output
    (): meltano.core.logging.console_log_formatter
    colors: True
```

To make logs follow the flag, the formatter function would have to be aware of the project configuration. Users can still turn off color in logs via `logging.yaml`.

aaronsteers:
> To make logs follow the flag, the formatter function would have to be aware of the project configuration. Users can still turn off color in logs via `logging.yaml`.

I hate to add scope but I think the primary use case is for helping to read logs - especially in CI outputs and especially when a `logging.yml` is not present.

I do think the env var and CLI inputs are probably higher precedence order than the log config file - and should probably override that config file for console outputs, at least, I don't think that precedence must be resolved in the first pass. It would be fine for me if a custom `logging.yaml` was not overridden in this first iteration, but when `logging.yaml` is absent, then I would expect logs to respect the no-color preference.

Is there a path to implement this without blowing up scope too much?

After digging a bit through the code, I found one option I want to propose. Please correct me if I'm wrong, but it seems the problem is that when `setup_logging()` is called, we may or may not have the CLI input parsed. From this context, I want to propose we drop `--no-color` support from this PR and _only_ parse `NO_COLOR`. We can treat that as an low-lever environment configuration like `PATH` and use a helper function in `utils` or similar to read it from to `os.environ` context whenever we're making a choice of whether to supress/disable colorized output: within `setup_logging()` and any other codepaths which need it.

This approach would (hopefully) allow us to _fully_ respect `NO_COLOR` env var, which is the strong 'standard' suggested by https://no-color.org/.

Unless we can distinguish console output vs non-console output handlers, I think the expected behavior is to fully override any color preferences when `NO_COLOR` is true. The rationale here is threefold:

1. `NO_COLOR` as an env var has a higher precedent order than config files.
2. The cost of over-doing it is much less than the cost of under-doing it. Meaning, the impact of printing color when color is not desired is much worse than the converse impact of removing color in more places than the user expected.
3. Those who already have fine-grained `logging.yaml` config are less likely to need `NO_COLOR` anyway. And if they do apply it, it's probably for an important reason to override prior stated preferences.

And I feel very good about removing `--no-color` if it helps the scope, because for CI environments, the `NO_COLOR` env var implementation is much easier anyway - since it doesn't require having two versions of meltano CLI scripts.

aaronsteers (commented):
Requesting removal of `--no-color` support and focus solely on `NO_COLOR` for now. See my prior comment.

In addition to hopefully sidestepping some of the harder technical challenges order-of-operations and variable-passing, another reason to _not_ use `--no-color` is that third-party libraries we use will _only_ be able to take advantage of this setting if the user sets them with `NO_COLOR` env var - or if we write custom logic to request that behavior, which as we observing, is costly and can't promise perfect coverage.

https://no-color.org/#color-libraries-supporting-no_color-to-disable-default-color-output

We can try to force those dependent packages to use no-color one-by-one, but I think the more robust solution is to let any tool or library that supports `NO_COLOR` get that signal directly from the user. Less work for us, and more confidence in the long run that tools that respect that env var preference will get the signal to do so.

This can extend also to our plugins when they are invoked as child processes. If Meltano is invoked with `NO_COLOR=TRUE` and Meltano then spins off a child process like `dbt`, then `dbt` presumably will also get that `NO_COLOR` env var, and it will get a chance to honor it if it is supported by the tool.

cc @tayloramurphy

src/meltano/core/utils/__init__.py
```diff
@@ -170,7 +171,7 @@ def merge(src, dest):
     return dest
 
 
-def nest(d: dict, path: str, value=None, maxsplit=-1, force=False):
+def nest(d: dict, path: str, value=None, maxsplit=-1, force=False):  # noqa: WPS210
```
edgarrmondragon, line 174: We should address these complexity issues at some point.

src/meltano/core/utils/__init__.py
```diff
@@ -14,6 +14,7 @@
 from collections import OrderedDict
 from copy import deepcopy
 from datetime import date, datetime, time
+from distutils.util import strtobool
```
edgarrmondragon, line 17: Using this function from `distutils` feels a bit hacky. Let me know what you folks think.

WillDaSilva: I'd rather we just re-implement it in `meltano.core.utils`.

kgpayne: Its from `distutils` which is probably better than maintaining our own. I like how you have wrapped it though, so we can change the implementation in future if needed 👍

kgpayne (approved):
LGTM 🚢 🚀

src/meltano/core/utils/__init__.py
edgarrmondragon: @WillDaSilva fair enough: 8238fe9b67585f193c6eadd79de6714db900ddf7

src/meltano/core/utils/__init__.py
```diff
@@ -591,3 +591,55 @@ def safe_hasattr(obj: Any, name: str) -> bool:
...
+
+
+def strtobool(val: str) -> bool:
+    """Convert a string representation of truth to true (1) or false (0).
+
+    True values are 'y', 'yes', 't', 'true', 'on', and '1'; false values
+    are 'n', 'no', 'f', 'false', 'off', and '0'.  Raises ValueError if
+    'val' is anything else.
+
+    Re-implemented from distutils.util.strtobool to avoid importing distutils.
```
aaronsteers, line 603: I double checked that this is the case in the code but might be a helpful hint to have also in the docstring.

```suggestion

    Case is ignored in text comparisons.

    Re-implemented from distutils.util.strtobool to avoid importing distutils.
```

edgarrmondragon:
@aaronsteers @WillDaSilva @tayloramurphy I've addressed your feedback 🙏, and this is ready for the next round of reviews

docs/src/_reference/command-line-interface.md
```diff
@@ -9,6 +9,17 @@ Meltano provides a command line interface (CLI) that makes it easy to manage you
...
+## Global Options
+
+The following options are available for all `meltano` subcommands:
+
+* `--log-config` - Path to a logging configuration file. See [Logging](/guide/logging) for more information.
+* `--log-level`: Set the log level for the command. Valid values are `debug`, `info`, `warning`, `error`, and `critical`.
+
+## Environment Variables
+
+If the `NO_COLOR` environment variable is set, Meltano will disable colored output. See [`no_color`](https://no-color.org/) for more information.
```
tayloramurphy, line 21:
```suggestion
If the `NO_COLOR` environment variable is set to `true`, Meltano will disable colored output on the command line. See [`no_color.org`](https://no-color.org/) for more information.
```

docs/src/_reference/command-line-interface.md
```diff
@@ -9,6 +9,17 @@ Meltano provides a command line interface (CLI) that makes it easy to manage you
 To quickly find the `meltano` subcommand you're looking for, use the Table of Contents in the sidebar.
 For a better understanding of command line documentation syntax, the [docopt](http://docopt.org/) standard is useful.
 
+## Global Options
+
+The following options are available for all `meltano` subcommands:
+
+* `--log-config` - Path to a logging configuration file. See [Logging](/guide/logging) for more information.
+* `--log-level`: Set the log level for the command. Valid values are `debug`, `info`, `warning`, `error`, and `critical`.
```
tayloramurphy, line 17:
```suggestion
* `[--log-level](/reference/settings#clilog_level)` - Set the log level for the command. Valid values are `debug`, `info`, `warning`, `error`, and `critical`.
```

docs/src/_reference/command-line-interface.md
```diff
@@ -9,6 +9,17 @@ Meltano provides a command line interface (CLI) that makes it easy to manage you
 To quickly find the `meltano` subcommand you're looking for, use the Table of Contents in the sidebar.
 For a better understanding of command line documentation syntax, the [docopt](http://docopt.org/) standard is useful.
 
+## Global Options
+
+The following options are available for all `meltano` subcommands:
+
+* `--log-config` - Path to a logging configuration file. See [Logging](/guide/logging) for more information.
```
tayloramurphy, line 16:
```suggestion
* `[--log-config](/reference/settings#clilog_config)` - Path to a logging configuration file. See [Logging](/guide/logging) for more information.
```

docs/src/_reference/command-line-interface.md
```diff
@@ -9,6 +9,17 @@ Meltano provides a command line interface (CLI) that makes it easy to manage you
 To quickly find the `meltano` subcommand you're looking for, use the Table of Contents in the sidebar.
 For a better understanding of command line documentation syntax, the [docopt](http://docopt.org/) standard is useful.
 
+## Global Options
```
tayloramurphy, line 12:
```suggestion
## Global Configuration
```

Thoughts on this and then just getting rid of the Env var section?

docs/src/_reference/command-line-interface.md
```diff
@@ -9,6 +9,17 @@ Meltano provides a command line interface (CLI) that makes it easy to manage you
...
+
+The following options are available for all `meltano` subcommands:
+
+* `--log-config` - Path to a logging configuration file. See [Logging](/guide/logging) for more information.
+* `--log-level`: Set the log level for the command. Valid values are `debug`, `info`, `warning`, `error`, and `critical`.
+
+## Environment Variables
```
tayloramurphy, line 19: The following configuration is available for all `meltano` subcommands via an environment variable:

docs/src/_reference/command-line-interface.md
```diff
@@ -9,6 +9,17 @@ Meltano provides a command line interface (CLI) that makes it easy to manage you
...
+
```
tayloramurphy, line 20: @edgarrmondragon this feels like it deserves an entry on https://deploy-preview-6785--meltano.netlify.app/reference/settings#meltano-cli ?

tayloramurphy (changes requested):
@edgarrmondragon couple of comments on the docs. Sorry for the pickiness! Happy to sync on it tomorrow if that's quicker.

WillDaSilva (approved):
:shipit:

aaronsteers (approved):

edgarrmondragon:
@tayloramurphy see https://github.com/meltano/meltano/issues/6940

docs/src/_reference/command-line-interface.md
```diff
@@ -9,6 +9,21 @@ Meltano provides a command line interface (CLI) that makes it easy to manage you
...
+* [`--log-level`](/reference/settings#clilog_level) - Set the log level for the command. Valid values are `debug`, `info`, `warning`, `error`, and `critical`.
+
+The following configuration is available for all meltano subcommands via an environment variable:
+
+* `NO_COLOR` - Set this environment variable to a truthy value (`1`, `TRUE`, `t`) to disable colored output on the command line. See [`no_color.org`](https://no-color.org/) for more information.
+
+## Environment Variables
+
+If the `NO_COLOR` environment variable is set, Meltano will disable colored output. See [`no_color`](https://no-color.org/) for more information.
+
```
tayloramurphy, line 26:
```suggestion
```

tayloramurphy (approved):

docs/src/_reference/command-line-interface.md
edgarrmondragon: 😅 I missed it

PR #6793, opened 2022-09-26
test: add basic migrations test

pandemicsyn:
Note that this is based on and targets the branch [6790-migration-bug](https://github.com/meltano/meltano/tree/6790-migration-bug)


pandemicsyn:
@cjohnhanson looks like the integration tests works - its at least triggering the current failures

PR #6800, opened 2022-09-27
docs: shorten install quick start, move rest to guide

sbalnojan:
Moving things around to enable people to launch quickly! 

I think, we can teach people to use pipx and mount docker volumes, or handle Windows problem later after they get their first data pipeline moving and actually decide to dive deeper into 500 words of installation guide. So:

- Add tabs plugin to docs with css etc.
- Add a new install section which is much slimmer.
- Move the more detailed "install guide" (prev. installation.md) to the guides section and reference it.

FWIW, if someone with more CSS skills feels like improving the tabs, I#d love some support there (like making the "tabs" more apparent).


docs/src/_guide/installation-guide.md
```diff
@@ -0,0 +1,237 @@
...
+
+Now that you're successfully running Meltano, you can continue setting up your Meltano project by following the [Getting Started guide](/getting-started).
+
+Note that wherever you are asked to run the `meltano` command, you will want to run it through `docker run` as in the snippet above.
+
+
+# Troubleshooting Installation
+
+<div class="notification is-info">
+  <p>Are you having installation or deployment problems? We are here to help you. Check out <a href="/">Getting Help</a> on the different ways to get in touch with us.</p>
```
WillDaSilva, line 212: Why does the "Getting Help" link send users to the root of the doc site? Can we use some tag to have the page scroll down to the "Get Help" section at the bottom of the page at the root of the doc site?

WillDaSilva (changes requested):
I like the direction this is going, and have a few suggestions:
- The [home page](https://deploy-preview-6800--meltano.netlify.app/) sends me to [the getting started guide](https://deploy-preview-6800--meltano.netlify.app/getting-started/) first, but if I read that page first then the installation page won't have much benefit for me. I should be directed towards the installation page, and then have it lead me to the getting started guide.
- The order in the table of contents should be:
  - Meltano at a Glance
  - Installation
  - Getting Started
  - Tutorial Part 1 - Extract Data
- All of the animated consoles play in unison even when they haven't yet been revealed. Switching to a tab to find that the console animation is finished isn't ideal. That said, my hunch is that fixing this would be more trouble than it's worth, but maybe you know some relatively easy solution.
- The [installation guide](https://deploy-preview-6800--meltano.netlify.app/guide/installation-guide) refers to the getting started guide as the place to find basic installation instructions.
- It's somewhat confusing to have differing installation instructions present in `installation.md`, the getting started guide, and the installation guide. I would prefer having the getting started guide link to `installing.md` rather than giving its own instructions, and have `installation.md` and the installing guide link to each other (i.e. "for more details see the installation guide" and "for a quick overview see the installation page"). That, or a more significant restructuring, e.g. instead of splitting `installation.md` into two pages, put the advanced/detailed content into collapsible sections.
- https://deploy-preview-6800--meltano.netlify.app/getting-started/#install-meltano says it links to the "Installation guide", but actually links to `installation.md`.

WillDaSilva:
Logged https://github.com/meltano/meltano/issues/6803 to track the failing pre-comit CI job. It has nothing to do with this PR, so we should carefully ignore it (i.e. ignore it only if it's failing because of `eslint` not finding its config file).

docs/src/_guide/installation-guide.md
sbalnojan: @WillDaSilva I agree with almost all of it, I'll take a look at it. 

I do still like to have the installation separated into two parts though, 

1. basic installation (used in the Tutorial)
2. advanced installation 

The idea is that the intended reader are two different people.
1. is for the first time meltano user, and I want him to get started as fast as possible, so he shouldn't even try to install pipx if he hasn't, or mount the docker volume.

Once someone has gone through the tutorial, he is ready for (2). But in that case, I don't want to bother him with the short version. 

Maybe we just need to distinguish the two better? By name & title? Do you think that would work for you?

sbalnojan:
@WillDaSilva  take a look at it now, I think it flows quite nicely and thanks for pointing out all the other places that need fixing! 
- I changed the home page to properly reflect the new structure, putting more focus on first time users and getting started.
- Renamed the installation guide to "in-depth installation", the same name used on the home page.
- linked up everything as suggested.

For the console preview idea: I'm not javascripty enough to get it to work - I tried. In theory, the "termynals" could receive an individual event listener listening for a click on the appropriate tab and then calling the init event (they already have on listener watching out for scroll events).

docs/src/_getting-started/installation.md
```diff
@@ -5,233 +5,89 @@ layout: doc
...
-Clicking this button will show more information and give you the option to install the update right away:
+{% endtabs %}
 
-![Update Available Popup](images/installation/update-available-popup.png)
+## More Information
+To understand details of the installation, like mounting a docker volume to work long-term with the docker image, view the [detailed installation guide](/guide/installation-guide).
+## Next Steps
 
-The Meltano UI will refresh automatically once installation is complete.
+Once you're set up, head over to the tutorial to learn [how to initialize your first project ancd start to import data](/getting-started/part1).
```
WillDaSilva, line 90:
```suggestion
Once you're set up, head over to the tutorial to learn [how to initialize your first project and start to import data](/getting-started/part1).
```

WillDaSilva (approved):
Thanks for the updates @sbalnojan! I caught a typo, but aside from this I'd say this PR is good to go.

tayloramurphy (approved):
I'm approving but I do feel like the content on the landing page is "smushed' on the left side of the page (for lack of a better phrase). I like the simplification, but it does feel like the content under "First Steps" could be put into two columns under the same header. That said, good iteration, happy to get it out there.

sbalnojan:
> I'm approving but I do feel like the content on the landing page is "smushed' on the left side of the page (for lack of a better phrase). I like the simplification, but it does feel like the content under "First Steps" could be put into two columns under the same header. That said, good iteration, happy to get it out there.

I agree, details are not my strength. Is one for the next iteration.

PR #6809, opened 2022-09-28
docs: Tutorial small cleanup

pnadolny13:
@sbalnojan I fixed the link from part 1 to part 2 and a couple wording things. Let me know what you think.


WillDaSilva (approved):

sbalnojan (approved):
Yes thanks! I'm trying to watch out for such (important) details, but to be honest, that's not my strength so thanks for catching all of these, and feel free to do that again in the future ;)

PR #6819, opened 2022-09-30
fix: Fix error on `meltano config meltano set ...`

### LINKED ISSUE #6802: bug: `meltano config meltano set ...` fails
### Meltano Version

2.6.0

### Python Version

3.8

### Bug scope

### Operating System

Ubuntu

### Description

Tried to set my buffer size via `meltano config` and received this error

### Code

```python
ubuntu@ip-172-31-19-214:~/git/tap-indeed$ meltano config meltano set elt.buffer_size 104857600
2022-09-27T13:08:59.348162Z [info     ] Environment 'dev' is active
2022-09-27T13:08:59.478128Z [info     ] The default environment (dev) will be ignored for `meltano config`. To configure a specific Environment, please use option `--environment=<environment name>`.
'ProjectSettingsService' object has no attribute 'plugin'
```

WillDaSilva:
To do:
- [x] Add one or more test cases for `meltano config meltano set ...`
- [x] Determine if the usage of `self.settings.plugin` on line 158 is problematic

Closes #6802

### DIFF
--- a/src/meltano/cli/interactive/config.py
+++ b/src/meltano/cli/interactive/config.py
@@ -63,7 +63,6 @@ class InteractiveConfig:  # noqa: WPS230, WPS214
         self.project = self.ctx.obj["project"]
         self.settings = self.ctx.obj["settings"]
         self.session = self.ctx.obj["session"]
-        self.plugin = self.ctx.obj["settings"].plugin
         self.tracker = self.ctx.obj["tracker"]
         self.environment_service = EnvironmentService(self.project)
         self.max_width = max_width or 75  # noqa: WPS432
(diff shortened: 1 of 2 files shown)


edgarrmondragon:
> - [ ] Determine if the usage of self.settings.plugin on line 158 is problematic

@WillDaSilva since it's nested under the `elif source is SettingValueStore.INHERITED:` check, it's probably fine. Only plugin settings can come from the _inherited_ store.

edgarrmondragon (approved):

WillDaSilva:
> since it's nested under the elif source is SettingValueStore.INHERITED: check, it's probably fine. Only plugin settings can come from the inherited store.

Thank you for confirming. I thought it might be fine because of the check, but wanted someone with more knowledge about this to confirm :slightly_smiling_face:

PR #6820, opened 2022-09-30
test: Ensure telemetry uses latest Iglu schemas

### LINKED ISSUE #6780: feature: Ensure telemetry uses the latest Iglu schemas
### Feature scope

Other

### Description

Currently https://github.com/meltano/meltano/blob/d1db9287ddc60269f575cb150971304ad00a8130/src/meltano/core/tracking/schemas.py needs to be manually kept up to date with the schemas in https://github.com/meltano/meltano/tree/d1db9287ddc60269f575cb150971304ad00a8130/src/meltano/core/tracking/iglu-client-embedded/schemas/com.meltano

https://github.com/meltano/meltano/issues/6778 was caused by them not being kept in sync. We should make this desync practically impossible by adding a test that verifies that `schema.py` uses each latest schema. If we ever want to intentionally use an older version, we can mark the test `xfail`.

Another option would be to dynamically get the latest schema version at runtime, but compared to using a test that's less computationally efficient, more complicated, and cannot easily be overridden.

WillDaSilva:
Closes #6780

### DIFF
--- a/tests/meltano/core/tracking/test_schemas.py
+++ b/tests/meltano/core/tracking/test_schemas.py
@@ -1,0 +1,33 @@
+from __future__ import annotations
+
+from pathlib import Path
+
+from meltano.core.tracking import schemas
+
+schemas_dir = (
+    Path(__file__).parents[4]
+    / "src"
+    / "meltano"
+    / "core"
+    / "tracking"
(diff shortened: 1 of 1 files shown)


edgarrmondragon (approved):
nice!

PR #6823, opened 2022-10-02
feat: CLI and config to clean up old runs in meltano.db

### LINKED ISSUE #6771: feature: Offer a method to clean up old states in meltano.db
### Feature scope

API

### Description

Our Meltano project has a relatively large state (like 20MB), so the size of `meltano.db` grows quickly. To my best understanding, currently the `StateService` offers no way to remove old run/job records and I haven't found a clue in the config for any housekeeping, so we probably have to work around by SQLiting into the DB and delete them by ourselves. 
It would be nice if we can do this in an official way.

Can we possibly:

1. Have a CLI command to clean up old states? and/or
2. Config meltano for housekeeping, cleaning up old states at some point?

IMO, the feature makes sense even for normal-sized states, for we eventually need to deal with evergrowing `runs` table.

simonpai:
Resolves #6771.

What does this PR do:

1. Offer a CLI command `meltano state vacuum` to clean up rows of old runs in system db.
2. Offer an option in settings: `database_max_rows_per_state`, which limits how many rows of old runs per state to keep in the table. The default system behavior (which does not remove any row) is unchanged.

What's implemented:

* StateService
  * A vacuum() method
  * Clean up old rows if necessary when after adding a new row, based on configuration
* CLI: `meltano state vacuum`
* `settings.xml`: Add `database_max_rows_per_state` definition
* Pytest on StateService (`test_state_service.py`)
* Docs
  * CLI
  * Settings

There are a few things I concern about in this PR and I would appreciate your feedback:
* Is the CLI command name appropriate?
* Is the option name appropriate?
* To access meltano configuration from `StateService`, I had to pass it in in some way and thus change the signature of its constructor, but I really didn't see an elegant way to do so.
* Am I missing anything for docs? changelog?


src/meltano/core/state_service.py
```diff
@@ -31,13 +31,16 @@ class StateService:
     Currently only manages Singer state for Extract and Load jobs.
     """
 
-    def __init__(self, session: object = None):
```
WillDaSilva, line 34: Passing in the project object, and letting the `StateService` create the `ProjectSettingsService` itself rather than putting that responsibility on all of the callers would be preferable.

docs/src/_reference/settings.md
```diff
@@ -207,6 +207,21 @@ meltano config meltano set database_retry_timeout 5
 export MELTANO_DATABASE_RETRY_TIMEOUT=5
 ```
 
+### `database_max_rows_per_state`
+
+- [Environment variable](/guide/configuration#configuring-settings): `MELTANO_MAX_ROWS_PER_STATE`
```
WillDaSilva, line 212:
```suggestion
- [Environment variable](/guide/configuration#configuring-settings): `MELTANO_DATABASE_MAX_ROWS_PER_STATE`
```

docs/src/_reference/settings.md
```diff
@@ -207,6 +207,21 @@ meltano config meltano set database_retry_timeout 5
...
+- Default: (unlimited)
+
+The value limits how many old runs for each state Meltano keeps in system DB. By default, Meltano stores the state payload for each run for auditing purpose. If your project has significantly large state sizes and requires housekeeping, the option can limit the maximum rows for each state kept in system DB.
+
+#### How to use
+
+```bash
+meltano config meltano set database_max_rows_per_state 10
+
+export MELTANO_MAX_ROWS_PER_STATE=10
```
WillDaSilva, line 222:
```suggestion
export MELTANO_DATABASE_MAX_ROWS_PER_STATE=10
```

src/meltano/core/state_service.py
simonpai: I've tried this way at the beginning, but ProjectSettingsService leads to a circular dependency. Do you have another better approach we can try?

docs/src/_reference/settings.md
simonpai: fixed.

docs/src/_reference/settings.md
simonpai: fixed. thanks.

afolson:
@simonpai Thanks for contributing! This looks awesome.

docs/src/_reference/command-line-interface.md
```diff
@@ -1279,6 +1279,41 @@ echo '{"singer_state": {"project_123456_issues": "2020-01-01"}}' > gitlab_state.
...
+Delete unused historical records of old runs to release some disk space. The states of jobs are not affected.
+
+#### How to use
+
+```bash
+# Delete some historical records of old runs.
+meltano state vacuum
+
+# Delete some old runs while specifying how many rows to keep per state ID.
+meltano state vacuum --rows-to-keep <row_count>
```
tayloramurphy, line 1293:
```suggestion
meltano state delete --runs-to-keep <run_count>
```

I'm not sure it makes sense to have folks thinking in rows here. We currently do a [merge operation](https://docs.meltano.com/reference/command-line-interface#merge) for state, so deleting a specified number of rows would potentially result in bad state. 

If we keep it specific to runs that'd be less problematic. That said though, doing this _without_ `--pattern` could be really problematic too. 

@aaronsteers what do you think about requiring both the pattern and the run count for a destructive operation?

tayloramurphy (changes requested):

docs/src/_reference/command-line-interface.md
```diff
@@ -1279,6 +1279,41 @@ echo '{"singer_state": {"project_123456_issues": "2020-01-01"}}' > gitlab_state.
 meltano state set --force dev:tap-gitlab-to-target-jsonl --input-file gitlab_state.json
 ```
 
+### vacuum
```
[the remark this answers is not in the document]
simonpai: My take on this: the `state` command verbs align with the perspective of state management, while the previous runs added to the table is the implementation detailed abstracted away from that interface. If we say `state delete`, that could easily gets confused with the semantic of `state clear`. I agree with your point on `vacuum` being specific to database. However it also has the upside of stating 1. it reduce disk space usage for housekeeping purpose, and 2. it does not change the data state. That's the reason behind my choice of term `vacuum`, but I am open to any other options as long as you think it best serves the purpose.

docs/src/_reference/command-line-interface.md
```diff
@@ -1279,6 +1279,41 @@ echo '{"singer_state": {"project_123456_issues": "2020-01-01"}}' > gitlab_state.
...
+
+Delete unused historical records of old runs to release some disk space. The states of jobs are not affected.
+
+#### How to use
+
+```bash
+# Delete some historical records of old runs.
```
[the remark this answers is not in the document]
simonpai: Do you think the issue will go away if we prompt for this command? Like my argument on previous topic, I think `state clear` serves a quite different purpose which changes the actual state and adds a new row to the table, while `delete/vacuum` keep the actual state unchanged and do some housekeeping work.

docs/src/_reference/command-line-interface.md
simonpai: I agree with using the term `runs` in place of `rows`, which is more generic.

simonpai: I totally overlooked the merge operation. I guess we need to leave all latest runs with incomplete payload and maybe plus the latest n complete states in the table, and then we want to make it clear for the user on what this command really does...

docs/src/_reference/command-line-interface.md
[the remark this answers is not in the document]
kgpayne: This is great @simonpai 👍 I agree that `vacuum` is a little too database-specific, and maybe even misleading in the database context (generally the VACUUM operation reorders data, rather than deleting it, akin to defragmentation for disk drives). The docker CLI uses `prune` for its clear-up operations (deleting unused images/containers/volumes etc.) which might work for us too; it meets your 2 points. Also, many users will be familiar with both tools, so some level of consistency is advantageous. As always, naming things is the hardest part 😅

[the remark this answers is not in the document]
simonpai: `prune` sounds like the best candidate so far! I think I'll go with `meltano state prune`. I appreciate feedback from both of you. TBH, it took me a while to think about this and I'm glad to find a better answer from discussion with you guys.

PR #6852, opened 2022-10-05
docs: add giffy to title

sbalnojan:


afolson (approved):

WillDaSilva:
@sbalnojan @afolson @tayloramurphy I think this PR should be reverted for the following reasons:
- The gif has text cut off on the left, right, and bottom.
- The gif file is in the root of the repository. We can probably find a better place for it - the old gif (Meltano on a chip) was stored publicly on Google Drive iirc. As it, it's cluttering the top-level, which is already intimidating to new contributors who would like to understand the structure of the repository.
- The gif is confusing and hard to follow, especially if you are not already familiar with Meltano. I don't think a gif of a recording of Meltano being used from the CLI is a good way to appeal to those who are new to it. The rapid output from the `pip install meltano` and the logs from `meltano run ...` are particularly offputting.

Also, not related to this, but why was the Meltano on a chip gif removed anyway? Just curious.

Issue #6677, opened 2022-08-26
bug: Projects without `environments` nor `default_environment` see error where `dev` environment is not found


edgarrmondragon:
### Meltano Version

2.5.0

### Python Version

NA

### Bug scope

Configuration (settings parsing, validation, etc.)

### Operating System

NA

### Description

Users with projects created from scratch (not using `meltano init`) or started before Environments where introduced may not be using `default_environment` nor `environments` in their `meltano.yml`.

This causes an error because in `2.5.0` we made `default_environment` a project setting that defaults to `dev`. Any project without a `dev` environment, as those exemplified above, will see an error because Meltano will be looking for `dev`.

### Code

```console
$ meltano install
Environment 'dev' was not found
```

This is reproducible with the following `meltano.yml`:

```yaml
plugins:
  loaders:
  - name: target-jsonl
    variant: andyh1203
    pip_url: target-jsonl
```

### Workaround

From AJ:

There should be two ways to work around this issue.

Running `meltano environment add dev` as a one-time fix will resolve the issue, even if no other changes are made and no environment-specific settings are added. 

Adding the `--no-environment` CLI arg will override the default `dev` environment name.

pnadolny13:
@edgarrmondragon +1 I ran into this last night also.

tayloramurphy:
@edgarrmondragon what's a reasonable resolution path here? Do we default to `default_environment: none` but put a prompt for users to consider adding environments? I don't think suddenly forcing everything through `dev` is a great experience on an existing project.

edgarrmondragon:
> @edgarrmondragon what's a reasonable resolution path here? Do we default to `default_environment: none` but put a prompt for users to consider adding environments? I don't think suddenly forcing everything through `dev` is a great experience on an existing project.

@tayloramurphy Yeah, I think defaulting to `none` in `settings.yml` is the right call. New projects created with `meltano init` will still get `default_environment: dev` and the corresponding environment.

tayloramurphy:
@edgarrmondragon sounds good!

@aaronsteers @edgarrmondragon is this worth a 2.5.1? Feels like it since it's pretty big feeling bug.

aaronsteers:
@tayloramurphy - I'm on the fence a bit due to the importance of environments for many of our default behaviors.

We had some discussion on this over here:

- #3200 

If there was a reasonable way to warn or even hard fail with a specific instruction, I would feel a bit better about that path forward.

[...]

PR #6862, opened 2022-10-07
fix: Ignore default environment for some commands

### LINKED ISSUE #6834: Ignore default Meltano Environment on non-execution CLI commands
Per:

- https://github.com/meltano/meltano/issues/6677#issuecomment-1228821208

We should ignore `default_environment` for all of the following commands:

> ...these are the commands that can't apply an environment and probably shouldn't have a default environment applied:
> 
> - [add](https://docs.meltano.com/reference/command-line-interface#add)
> - [discover](https://docs.meltano.com/reference/command-line-interface#discover)
> - [environment](https://docs.meltano.com/reference/command-line-interface#environment)
> - [init](https://docs.meltano.com/reference/command-line-interface#init)
> - [install](https://docs.meltano.com/reference/command-line-interface#install)
> - [lock](https://docs.meltano.com/reference/command-line-interface#lock)
> - [remove](https://docs.meltano.com/reference/command-line-interface#remove)
> - [ui](https://docs.meltano.com/reference/command-line-interface#ui)
> - [upgrade](https://docs.meltano.com/reference/command-line-interface#upgrade)
> - [user](https://docs.meltano.com/reference/command-line-interface#user)
> - [version](https://docs.meltano.com/reference/command-line-interface#version)

> These could have an environment applied or not, but probably should follow the meltano [config](https://docs.meltano.com/reference/command-line-interface#config) behavior of the user specifying it explicitly:
>
> - config
> - [job](https://docs.meltano.com/reference/command-line-interface#job)
> - [schedule](https://docs.meltano.com/reference/command-line-interface#schedule)
> - [select](https://docs.meltano.com/reference/command-line-interface#select)
> - [state](https://docs.meltano.com/reference/command-line-interface#state)

Which then leaves these as applying `default_environment` automatically:

> These could be run without an environment and should auto-apply the default if available:
> 
> - [elt](https://docs.meltano.com/reference/command-line-interface#elt)
> - [invoke](https://docs.meltano.com/reference/command-line-interface#invoke)
> - [test](https://docs.meltano.com/reference/command-line-interface#test)
>
> This should really not be run without an environment, because behaviors like auto-state tracking won't work as expected when no environment is set:
> 
> - [run](https://docs.meltano.com/reference/command-line-interface#run)

WillDaSilva:
Given the urgency of #6677, and the relative ease of #6834, I decided to quickly put together an implementation.

In this implementation, commands which require an environment call `activate_environment`, and those which will only use an environment if it is explicitly provided (e.g. via `--environment`) call `activate_explicitly_provided_environment`. By default no environment is required nor used, which comes with the side benefit of avoiding the "default environment activated" log message for most commands.

Closes #6834

### DIFF
--- a/src/meltano/cli/cli.py
+++ b/src/meltano/cli/cli.py
@@ -11,6 +11,7 @@ import click
 import meltano
 from meltano.cli.utils import InstrumentedGroup
 from meltano.core.behavior.versioned import IncompatibleVersionError
+from meltano.core.error import MeltanoConfigurationError
 from meltano.core.logging import LEVELS, setup_logging
 from meltano.core.project import Project, ProjectNotFound
 from meltano.core.project_settings_service import ProjectSettingsService
@@ -120,7 +121,9 @@ def cli(  # noqa: WPS231
         sys.exit(3)
 
 
-def activate_environment(ctx: click.Context, project: Project) -> None:
+def activate_environment(
+    ctx: click.Context, project: Project, required: bool = False
+) -> None:
     """Activate the selected environment.
 
     The selected environment is whatever was selected with the `--environment`
@@ -132,6 +135,12 @@ def activate_environment(ctx: click.Context, project: Project) -> None:
     """
     if ctx.obj["selected_environment"]:
         project.activate_environment(ctx.obj["selected_environment"])
+    elif required:
+        raise MeltanoConfigurationError(
+            reason="A Meltano environment must be specified",
+            instruction="Set the `default_environment` option in "
+            "`meltano.yml`, or the `--environment` CLI option",
+        )
 
 
 def activate_explicitly_provided_environment(
--- a/src/meltano/cli/cli.py
+++ b/src/meltano/cli/cli.py
@@ -123,6 +123,9 @@ def cli(  # noqa: WPS231
 def activate_environment(ctx: click.Context, project: Project) -> None:
     """Activate the selected environment.
 
+    The selected environment is whatever was selected with the `--environment`
+    option, or the default environment (set in `meltano.yml`) otherwise.
+
     Args:
         ctx: The Click context, used to determine the selected environment.
         project: The project for which the environment will be activated.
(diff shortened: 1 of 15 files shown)


tayloramurphy:
@WillDaSilva love this PR coming through. We'll need to update the docs in the CLI reference section to make it explicit about which ones require an environment or not. 

Maybe outside the scope of this PR, but I'm wondering if some short of small table at the start would be useful where we list attributes of the command such as "Uses `default_environment`", "Edits meltano.yml", etc.

At a minimum though we should update the text of every command to have a one-liner on whether or not `default_environment` is used. 

@aaronsteers what else would you like to see here?

src/meltano/cli/cli.py
```diff
@@ -123,3 +118,38 @@ def cli(  # noqa: WPS231
             "For more details, visit https://docs.meltano.com/guide/installation#upgrading-meltano-version"
         )
         sys.exit(3)
+
+
+def activate_environment(ctx: click.Context, project: Project) -> None:
+    """Activate the selected environment.
```
aaronsteers, line 124: @WillDaSilva - What do you think of slightly expanding this docstring to make it clear default environments are also applied? Maybe:

```suggestion
    """Activate the requested environment or a default environment if nothing requested.
```

WillDaSilva, line 124:
```suggestion
    """Activate the selected environment.

    The selected environment is whatever was selected with the `--environment`
    option, or the default environment (set in `meltano.yml`) otherwise.
```

WillDaSilva: Thanks for the suggestion. I updated it in b46ebc6 with the elaboration on the next line, because I couldn't fit a satisfactory summary all in the first line.

WillDaSilva:
> I don't know if run requires an enviornment as of now, but if not, can we implement that in this PR or should it be saved for a later iteration?

@aaronsteers Implemented in 7baebb88080e4a6c8b805b0dd0524a22fdadf802 - `meltano run` will now error if there is no environment.

tayloramurphy:
@WillDaSilva @aaronsteers I've updated AJ's comment with my preferred text. Thanks for well thought out prompts @aaronsteers !

aaronsteers (approved):
Noyce. 👍

PR #6863, opened 2022-10-07
Revert "docs: add giffy to title"

sbalnojan:
@WillDaSilva Ok, valid points. Sounds like I need to do a round 2. 

FWIW, we're focusing in on the ELT part of Meltano, and the chip is a representation of the DataOps infrastructure.

With a positioning that moves more towards the "Your CLI for ELT+", I think its almost mandatory for us to also show this "CLI". 

But you're right, the way I've done it might be too confusing. I'll take another shot at it next week maybe akin to https://fly.io/ (without gif just a static CLI).


WillDaSilva (approved):
Thanks @sbalnojan

Issue #6804, opened 2022-09-27
bug: `melta.no` does not support HTTPS


WillDaSilva:
### Meltano Version

N/A

### Python Version

NA

### Bug scope

Other

### Operating System

N/A

### Description

[http://melta.no works](http://melta.no).

[https://melta.no does not work](https://melta.no):

![image](https://user-images.githubusercontent.com/11428666/192583754-f7201304-acae-4fd9-94ad-5115ae09edf0.png)

From @tayloramurphy in https://github.com/meltano/meltano/pull/6748#issuecomment-1248406910:

> I've emailed the domain provider to see about getting an SSL Certificate. http://melta.no/ does work and forwards appropriately from what I see. FYI @therebbie I put an inquiry through the Marcaria contact form to go to the hello meltano address.

therebbie:
@WillDaSilva I thought we ran into this a couple of weeks ago and via a different issue (not the one listed above).

@tayloramurphy the hello address forwards (I believe) to Zendesk for such things. In this case, Marcaria responded thusly:
> To add HTTPS to a domain, you need hosting. The SSL issuer needs to verify that the domain's host is valid. Maybe check with your current hosting provider about the option to host [melta.no](http://melta.no/), but with just URL forward is not possible even with frame-forward; it will show a not secure connection for [melta.no](http://melta.no/) and [meltano.com](http://meltano.com/)'s website.

visch:
fwiw @therebbie  Google Domains auto sets up an SSL cert for you an will forward your subdomain (or domains) to whatever pages you want. Maybe others do this too, but might be easier then setting up something on aws and managing it 🤷

WillDaSilva:
> To add HTTPS to a domain, you need hosting. The SSL issuer needs to verify that the domain's host is valid. Maybe check with your current hosting provider about the option to host [melta.no](http://melta.no/), but with just URL forward is not possible even with frame-forward; it will show a not secure connection for [melta.no](http://melta.no/) and [meltano.com](http://meltano.com/)'s website.

@therebbie @tayloramurphy So if I understand the situation correctly, we are currently using our domain registrar Marcaria to provide URL forwarding instead of managing the DNS directly with our regular DNS provider?

[...]

PR #6865, opened 2022-10-08
chore: Configure Netlify redirects for `melta.no`

WillDaSilva:
Relates to #6804

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -11,3 +11,15 @@
 
 # [build.environment] # not sure if we'll need to specify a version
   # RUBY_VERSION = "some.version.of.ruby"
+
+[[redirects]]
+  from = "http://melta.no/*"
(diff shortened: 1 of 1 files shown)


netlify.toml
```diff
@@ -11,3 +11,15 @@
 
 # [build.environment] # not sure if we'll need to specify a version
   # RUBY_VERSION = "some.version.of.ruby"
+
+[[redirects]]
+  from = "http://melta.no/*"
+  to = "http://docs.meltano.com/guide/troubleshooting/:splat"
```
pandemicsyn, line 17: could just redirect directly to https and cut down the # of redirects that would happen.

pandemicsyn (approved):
:shipit:

netlify.toml
WillDaSilva: Good point. We already redirect to HTTPS for other domains for the docs site.

PR #6875, opened 2022-10-12
docs: add common problems

sbalnojan:
Added two common problems for troubleshooting to the docs with links to resources to resolve them.

### DIFF
--- a/docs/src/_guide/troubleshooting.md
+++ b/docs/src/_guide/troubleshooting.md
@@ -30,7 +30,17 @@ If you have a question about Meltano, are having trouble getting it to work, or 
     </div>
 </div>
 
-# Common Troubleshooting Issues
+# Common Issues
+
+*Problem: "Why do **incremental runs** produce duplicate data?"* 
+
+View this [Slack discussion](https://meltano.slack.com/archives/C013EKWA2Q1/p1659024288228009). Tl;dr: Singer takes an "at least once" approach to replication, so if you're encounting this, it might be intended behavior.
+
+*Problem: "My **runs take too long**."* 
+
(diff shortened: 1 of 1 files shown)


docs/src/_guide/troubleshooting.md
```diff
@@ -30,7 +30,17 @@ If you have a question about Meltano, are having trouble getting it to work, or
     </div>
 </div>
 
-# Common Troubleshooting Issues
+# Common Issues
+
+*Problem: "Why do **incremental runs** produce duplicate data?"* 
+
+View this [Slack discussion](https://meltano.slack.com/archives/C013EKWA2Q1/p1659024288228009). Tl;dr: Singer takes an "at least once" approach to replication, so if you're encounting this, it might be intended behavior.
```
WillDaSilva, line 37: Linking to Slack is unfriendly to newcomers who may not have (and not want) an account on the Meltano Slack.

WillDaSilva, line 37: Users who are on this page are probably often not having a good time, and so may take this casual tone (e.g. using "tl;dr:") poorly. They may feel as if they and their problems are not being taken seriously.

tayloramurphy:
```suggestion
Singer takes an "at least once" approach to replication, so if you're encountering this, it might be intended behavior. [This issue](https://github.com/MeltanoLabs/Singer-Working-Group/issues/13) is a good summary of the current state and a proposal to change this behavior.
```

tayloramurphy (approved):

docs/src/_guide/troubleshooting.md
tayloramurphy: @WillDaSilva I was able to rewrite it w/o the TL;DR.

WillDaSilva (approved):

PR #6911, opened 2022-10-18
feat: remote state backends

### LINKED ISSUE #5981: feat: Add support for non-DB state backends (`s3`, `dynamodb`, etc.)

cjohnhanson:
Related discussion:

- https://github.com/meltano/meltano/discussions/6952

Feature README:

- https://github.com/meltano/meltano/blob/5981-state-backends/docs/src/_concepts/state_backends.md

closes: #5981


docs/src/_concepts/state_backends.md
```diff
@@ -0,0 +1,95 @@
...
+The state backend settings for your Meltano project can be configured much the same as any other Meltano setting. For some state backends, the only required setting is `state_backend.uri`. 
+By default, `state_backend.uri` is set to the special keyword string `systemdb`, which will tell Meltano to store state in the same backend database as it stores other project data. 
+
+It can be changed by running `meltano config meltano set state_backend.uri <URI for desired state backend>` or by directly editing a project's `meltano.yml` to add the following:
+```yaml
+state_backend:
+	uri: <URI for desired state backend>
+```
+For some state backends and in some environments, `uri` is the only setting that needs to be configured in order to use a state backend other than the default `systemdb` backend. For instance, to store state directly on the local filesystem at the path `${MELTANO_ROOT_DIR}/.meltano/state`, simply run `meltano config meltano set state_backend.uri file:///${MELTANO_ROOT_DIR}/.meltano/state`. 
+Then Meltano will store the state for a given `state_id` at the path `${MELTANO_ROOT_DIR}/.meltano/state/<state_id>/state.json`.
```
WillDaSilva, line 35:
```suggestion
For some state backends and in some environments, `uri` is the only setting that needs to be configured in order to use a state backend other than the default `systemdb` backend. For instance, to store state directly on the local filesystem at the path `${MELTANO_PROJECT_ROOT}/.meltano/state`, simply run `meltano config meltano set 'state_backend.uri file:///${MELTANO_PROJECT_ROOT}/.meltano/state'`. Note the single quotes which prevent the early expansion of the environment variable.
Then Meltano will store the state for a given `state_id` at the path `${MELTANO_PROJECT_ROOT}/.meltano/state/<state_id>/state.json`.
```

docs/src/_concepts/state_backends.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Locking
+
+Because the `systemdb` state backend utilizes a transactional database, it can rely on the database's transaction logic to prevent conflicts among multiple runs of the same pipeline.
+For some other state backends, Meltano implements its own simple locking mechanism.
+This locking mechanism utilizes reasonable defaults but you may wish to configure the locking settings differently. 
+You can do this via two additional top-level `state_backend` settings: `state_backend.lock_timeout_seconds` and `state_backend.lock_retry_seconds`.
+
+When Meltano tries to read or write state for a given `state_id`, it will try to acquire a _lock_ on the state data for that `state_id`.
+For example, using the local filesystem state backend with `state_backend.uri` set to `file:///${MELTANO_ROOT_DIR}/.meltano/state`, it will check to see if a file exists at the path `${MELTANO_ROOT_DIR}/.meltano/state/<state_id>/lock`.
```
WillDaSilva, line 45:
```suggestion
For example, using the local filesystem state backend with `state_backend.uri` set to `file:///${MELTANO_PROJECT_ROOT}/.meltano/state`, it will check to see if a file exists at the path `${MELTANO_PROJECT_ROOT}/.meltano/state/<state_id>/lock`. 
```

schema/meltano.schema.json
```diff
@@ -27,6 +27,67 @@
...
+              "type": "object",
+              "description": "Configuration for Azure Blob Storage state backend.",
+              "properties": {
+                "connection_string": {
+                  "type": "string",
+                  "description": "The azure connection string to use for connecting to Azure Blob Storage."
+                  }
+                }
+            },
+          "gs": {
```
WillDaSilva, line 57: I've seen "gcs" used as an acronym for this, but not "gs". Is there precedence for this?

schema/meltano.schema.json
```diff
@@ -27,6 +27,67 @@
...
+              },
+            "lock_timeout_seconds": {
+              "type": "integer",
+              "description": "The number of seconds a lock on a state file is considered valid."
+              },
+            "lock_retry_seconds": {
+              "type": "integer",
+              "description": "The number of seconds to wait between retrying lock acquisition."
+              },
+            "az": {
```
WillDaSilva, line 47: I haven't seen "az" used as an acronym for Azure Blob Storage before. Is there precedence for this? `azure` is used elsewhere in this PR for this setting.

WillDaSilva (commented):
I have no major concerns with what's here. Really glad to see this coming along nicely!

schema/meltano.schema.json
cjohnhanson: Good callout. Yeah, the URI schema for google cloud storage is `gs` but for the settings I think it makes sense to use `gcs` for clarity and consistency with the `extras`. Updated.

schema/meltano.schema.json
cjohnhanson: I think there's some precedence in Azure's docs, but agreed that it should be `azure` for our purposes. Updated.

aaronsteers:
From 2022-11-03 discussion:

1. Environment-level support todo:
   - Cody to check if this works when set the environment level.
   - Undo `env_specific: True` and instead use `kind: password`
   - Here or elsewhere: add back full support for meltano settings that are allowed to be environment specific.
1. Integration tests

docs/src/_concepts/state_backends.md
```diff
@@ -0,0 +1,95 @@
...
+## Installation
+
+To cut down on extraneous dependencies, Meltano imports cloud storage client libraries lazily and these libraries are not included in the base `meltano` package.
+
+Only the default system database state backend _or_ the local filesystem state backend can be used as part of the base `meltano` distribution.
+
+To use a cloud storage backend, install Meltano using one of the following (extras)[https://peps.python.org/pep-0508/#extras]:
+- `meltano[s3]` to use the AWS S3 state backend.
+- `meltano[azure]` to use the Azure Blob Storage state backend.
+- `meltano[gcs]` to use the Google Cloud Storage state backend.
```
edgarrmondragon, line 22: Might be good to document these in `docs/src/_guide/advanced-topics.md` too

src/meltano/core/state_store/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    def _managers(
+        self,
+    ) -> dict[
+        str,
+        type[DBStateStoreManager]
+        | type[LocalFilesystemStateStoreManager]
+        | type[WindowsFilesystemStateStoreManager]
+        | type[AZStorageStateStoreManager]
+        | type[GCSStateStoreManager]
+        | type[S3StateStoreManager],
```
edgarrmondragon, line 54: Would it make sense to use a type alias or some type of generic that references the superclass `StateStoreManager` here?

Not a big deal, though.

cjohnhanson: Yeah I know this is pretty clunky. I initially had a type alias for this but since it required using the `type[]` syntax outside of an annotation (i.e. on the right hand side of an `=` rather than the right hand side of a `:`) it wouldn't work in <3.9.

What would using a Generic look like here?

docs/example-library/meltano-run/index.md
```diff
@@ -29,7 +29,7 @@ meltano --environment=dev config target-postgres set default_target_schema publi
 ```
 
 Note that sensitive config options such as `password` [aren't stored in your `meltano.yml` but in `.env` instead](https://docs.meltano.com/guide/configuration#configuration-layers). Print out the config to make sure it looks as expected:
-
+p
```
edgarrmondragon, line 32:
```suggestion

```

src/meltano/core/state_store/__init__.py
edgarrmondragon: @cjohnhanson maybe generics are not even required and something like

```python
    @property
    def _managers(self) -> dict[str, type[StateStoreManager]:
        ...
```

might work because all the unioned types are derived from `StateStoreManager`. Wdyt?

cjohnhanson: @edgarrmondragon yeah I initially had that but it fails mypy. It looks like we can use the generic `Mapping[str, type[StateStoreManager]]` though. I've updated this to use that instead,

docs/src/_concepts/state_backends.md
```diff
@@ -0,0 +1,95 @@
...
+
+The state backend settings for your Meltano project can be configured much the same as any other Meltano setting. For some state backends, the only required setting is `state_backend.uri`.
+By default, `state_backend.uri` is set to the special keyword string `systemdb`, which will tell Meltano to store state in the same backend database as it stores other project data.
+
+It can be changed by running `meltano config meltano set state_backend.uri <URI for desired state backend>` or by directly editing a project's `meltano.yml` to add the following:
+```yaml
+state_backend:
+    uri: <URI for desired state backend>
+```
+For some state backends and in some environments, `uri` is the only setting that needs to be configured in order to use a state backend other than the default `systemdb` backend. For instance, to store state directly on the local filesystem at the path `${MELTANO_PROJECT_ROOT}/.meltano/state`, simply run `meltano config meltano set 'state_backend.uri file:///${MELTANO_PROJECT_ROOT}/.meltano/state'`. Note the single quotes which prevent the early expansion of the environment variable.
```
edgarrmondragon, line 77:
```suggestion
For some state backends and in some environments, `uri` is the only setting that needs to be configured in order to use a state backend other than the default `systemdb` backend. For instance, to store state directly on the local filesystem at the path `${MELTANO_SYS_DIR_ROOT}/state`, simply run `meltano config meltano set 'state_backend.uri file:///${MELTANO_SYS_DIR_ROOT}/state'`. Note the single quotes which prevent the early expansion of the environment variable.
```

docs/src/_concepts/state_backends.md
```diff
@@ -0,0 +1,95 @@
...
+Then Meltano will store the state for a given `state_id` at the path `${MELTANO_PROJECT_ROOT}/.meltano/state/<state_id>/state.json`.
```
edgarrmondragon, line 78:
```suggestion
Then Meltano will store the state for a given `state_id` at the path `file:///${MELTANO_SYS_DIR_ROOT}/state/<state_id>/state.json`.
```

docs/src/_concepts/state_backends.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Locking
+
+Because the `systemdb` state backend utilizes a transactional database, it can rely on the database's transaction logic to prevent conflicts among multiple runs of the same pipeline.
+For some other state backends, Meltano implements its own simple locking mechanism.
+This locking mechanism utilizes reasonable defaults but you may wish to configure the locking settings differently.
+You can do this via two additional top-level `state_backend` settings: `state_backend.lock_timeout_seconds` and `state_backend.lock_retry_seconds`.
+
+When Meltano tries to read or write state for a given `state_id`, it will try to acquire a lock on the state data for that `state_id`.
+For example, using the local filesystem state backend with `state_backend.uri` set to `file:///${MELTANO_PROJECT_ROOT}/.meltano/state`, it will check to see if a file exists at the path `${MELTANO_PROJECT_ROOT}/.meltano/state/<state_id>/lock`.
```
edgarrmondragon, line 88:
```suggestion
For example, using the local filesystem state backend with `state_backend.uri` set to `file:///${MELTANO_SYS_DIR_ROOT}/state`, it will check to see if a file exists at the path `file:///${MELTANO_SYS_DIR_ROOT}/state/<state_id>/lock`.
```

edgarrmondragon (commented):
Some docs suggestions now that https://github.com/meltano/meltano/pull/6628 has been merged

edgarrmondragon (approved):

WillDaSilva (approved):

pdebelak:
👋 - question about this. I'm currently blocked on upgrading meltano from version 2.1.0 due to using Ubuntu 18.04 with an old version of sqlite so I thought that setting a different state backend would allow me to not care about the sqlite version. It appears that setting the state backend uri to a file location doesn't stop meltano from running migrations on its sqlite database anyway. Is that expected?

edgarrmondragon:
@pdebelak yes, that's unfortunately expected at this point. I wonder if now that Meltano supports secrets backends, Meltano should allow users to disable database migrations, or disable the backend db entirely.

cc @aaronsteers @tayloramurphy

aaronsteers:
@edgarrmondragon, @pdebelak - We have a bit more to do before we can fully replace systemdb.

- [x] State Backends
- [x] Job collision prevention (locking handled
via state backend implementation)
- [ ] Settings+Secrets Backends
- [ ] Logging Backend

However, since logging is "write only" as of now, perhaps we could disable the backend, in theory at least. If so, we'd have to fail if config get/set is performed against it, and we'd skip consideration of systemdb when using the auto-store decision tree for getting/setting config values.

It would probably take a lot of testing and debugging to get right, but in theory it may be possible... 🤔

aaronsteers:
@edgarrmondragon, @pdebelak - We have a bit more to do before we can fully replace systemdb.

- [x] State Backends
- [x] Job collision prevention (locking handled
via state backend implementation)
- [ ] Settings+Secrets Backends
- [ ] Logging Backend

However, since logging is "write only" as of now, perhaps we could disable the backend, in theory at least. If so, we'd have to fail if config get/set is performed against it, and we'd skip consideration of systemdb when using the auto-store decision tree for getting/setting config values.

It would probably take a lot of testing and debugging to get right, but in theory it may be possible... 🤔

PR #6814, opened 2022-09-28
feat: Support `meltano init` into an empty directory

ReubenFrankel:
Resolves #6806

### Checklist
- [x] Implement feature
- [x] Add/update tests
- [x] Add/update docs

Issue #6806, opened 2022-09-28
feature: Allow `meltano init` to create a project in-place


ReubenFrankel:
### Feature scope

CLI (options, error messages, logging, etc.)

### Current behaviour

Supply a project name interactively:
```
meltano init
```

Supply a project name to the command:
```
meltano init <project-name>
```

### Optimisation
Both of the above approaches require the user to `cd` into the resulting directory that is created for you:

```sh
meltano init project
cd project
```

It would be nice to be able to create a project in-place:

```
meltano init .
```

Here, the project contents should be created as normal in the current working directory. The command output should be adjusted to account for in-place operation (e.g. remove `cd <project-name>` line under the `Next steps` section).

Running within non-empty directories should check for potential file conflicts and preemptively fail the operation, or even detect whether there is an existing Meltano project present (borrow logic from attempting to run a `meltano` command outside of a project - assuming presence of `meltano.yml`?).

### Example use-case
If a user wanted to version control a new project via a remote repository (GitHub or similar), creating it in-place would avoid having to copy/move the contents of the project directory into the cloned empty repository directory.

Current:
```sh
git clone git@github.com:<user>/my-meltano-project.git

meltano init temp-project
cp -a temp-project/. ../my-meltano project

cd my-meltano-project

git add .
git push

# optionally, clean-up
rm -rf temp-project
```

Proposed;
```sh
git clone git@github.com:<user>/my-meltano-project.git
cd my-meltano-project

meltano init .

git add .
git push

# no cleanup necessary
```

### Comparison to `git init`
Initialise a Git repository in a new directory:
![image](https://user-images.githubusercontent.com/60552974/192675002-632e7bc7-d8e1-4fb2-a917-501282240a16.png)

This is the same behaviour as `meltano init`.

Initialise a Git repository in an existing directory:
![image](https://user-images.githubusercontent.com/60552974/192675129-df127aff-152b-4f9b-bbab-34241d9541ac.png)
![image](https://user-images.githubusercontent.com/60552974/192675157-6189ec08-fd11-4a79-9709-8fe76bf6764f.png)

No argument and `.` result in the same behaviour, whereas currently `meltano init` with no arguments initiates interactive setup.

---

Let me know what you think - happy to open a PR as always! :smile:

sbalnojan:
@ReubenFrankel agreed, stumbled over this multiple times. This behaviour makes the onboarding harder, I tend to first create a new repository and then read the init docs. Also this would allow to:

[...]

PR #6927, opened 2022-10-24
feat: Add `--force` option to ignore existing files/directories in an non-empty project directory with `meltano init`

### LINKED ISSUE #7077: Feature request: Support `meltano init` in non-empty directory
Follows from

- https://github.com/meltano/meltano/pull/6814, 

which was authored by @ReubenFrankel and which allows `meltano init` to target existing (empty) directories.

This new feature would _also_ add support for non-empty directories.

Safety considerations:

If the operation would overwrite any critical project files such as `meltano.yml`, then the `init` operation should fail _before_ beginning any file or directory write operations. (Aka, it should not fail mid-operation.)

Optionally, a `--force` option could be utilized to bypass checks and simply write all files and overwrite existing directory contents.

## Handling `.gitkeep` files

When we check for meltano init safety, we probably want to ignore the presence of .gitkeep files and directories themselves. The argument goes that, while adding a new meltano project to an existing repo would not have those conflicts, the tutorial use cases will have those conflicts.

In order to streamline tutorial use cases, we would not want the user (or the tutorial maintainer) to have to first delete all the .gitkeep files before they are able to run or re-run meltano init.

ReubenFrankel:
One of the primary use-cases addressed by #6806 was to be able to create an empty repository in GitHub, clone into a directory and immediately `meltano init` into it. I completely forgot about this when implementing #6814, so this is the follow-up PR! :sweat_smile: 

See [this Slack thread](https://meltano.slack.com/archives/C013Z450LCD/p1666396209017099) for more information.

### DIFF
--- a/src/meltano/cli/initialize.py
+++ b/src/meltano/cli/initialize.py
@@ -26,10 +26,16 @@ path_type = click.Path(file_okay=False, path_type=Path)
 @click.pass_context
 @click.argument("project_directory", required=False, type=path_type)
 @click.option(
+    "-f",
+    "--force",
+    help="Ignore existing files/directories within the project directory, if it exists.",
+    is_flag=True,
+)
+@click.option(
     "--no_usage_stats", help="Do not send anonymous usage stats.", is_flag=True
 )
 @database_uri_option
-def init(ctx, project_directory: Path, no_usage_stats):
+def init(ctx, project_directory: Path, force: bool, no_usage_stats):
     """
     Create a new Meltano project.
 
@@ -52,7 +58,7 @@ def init(ctx, project_directory: Path, no_usage_stats):
 
     init_service = ProjectInitService(project_directory)
 
-    project = init_service.init()
+    project = init_service.init(force=force)
     init_service.echo_instructions(project)
 
     # since the project didn't exist, tracking was not initialized in cli.py
--- a/src/meltano/core/project_init_service.py
+++ b/src/meltano/core/project_init_service.py
@@ -34,12 +34,15 @@ class ProjectInitService:
         except ValueError:
             pass
 
-    def init(self, activate: bool = True, add_discovery: bool = False) -> Project:
+    def init(
+        self, activate: bool = True, add_discovery: bool = False, force: bool = False
+    ) -> Project:
         """Initialise Meltano Project.
 
         Args:
             activate: Activate newly created project
             add_discovery: Add discovery.yml file to created project
+            force: Ignore existing files/directories within the project directory, if it exists
 
         Returns:
             A new Project instance
@@ -50,7 +53,7 @@ class ProjectInitService:
         try:
             self.project_directory.mkdir()
         except FileExistsError as ex:
-            if any(self.project_directory.iterdir()):
+            if not force and any(self.project_directory.iterdir()):
                 raise ProjectInitServiceError(
                     f"Directory '{self.project_directory}' not empty."
                 ) from ex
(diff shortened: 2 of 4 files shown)


tayloramurphy (changes requested):
Documentation update on this please :) https://docs.meltano.com/reference/command-line-interface#init

Also - if there is a matching file / folder, I'm assuming it would get overwritten, correct? Would be good to call that out in the docs.

ReubenFrankel:
@tayloramurphy Given the contents of a new project, what are your thoughts on overwriting vs throwing an error on a conflict? I was leaning towards the latter since there is already protection against running `meltano init` within an existing Meltano project, and I didn't really want to start overwriting things without explicit user confirmation. Since `meltano init` doesn't prompt on file conflicts currently anyway, maybe `--force` is the wrong option name?

tayloramurphy:
> Given the contents of a new project, what are your thoughts on overwriting vs throwing an error on a conflict? I was leaning towards the latter since there is already protection against running `meltano init` within an existing Meltano project, and I didn't really want to start overwriting things without explicit user confirmation. Since `meltano init` doesn't prompt on file conflicts currently anyway, maybe `--force` is the wrong option name?

@ReubenFrankel `--force` is probably the wrong one argument in this case. What about something like:

`meltano init --ignore-files --force`

where `--ignore-files` does what's implemented here and writes the Meltano files ignoring existing ones, but errors if there is a conflict. `--force` would then just blast away everything if there is a conflict. 

I'd also be open to `--ignore-existing` or something else if there's a better option.

ReubenFrankel:
@tayloramurphy

> where --ignore-files does what's implemented here and writes the Meltano files ignoring existing ones, but errors if there is a conflict.

There were some comments raised in the original issue about improving onboarding. I think it would be a more seamless experience for new users if `meltano init` could create a project in *any* directory (barring conflicts), like `git init` can. This would then remove the need for a new `--ignore-existing` or similar option. Do you think that could work, or does it just complicate things too much?

> --force would then just blast away everything if there is a conflict.

Not sure where this leaves `--force` as described here. Maybe out-of-scope for this PR?

aaronsteers:
@ReubenFrankel, @tayloramurphy - I'm okay dropping `--force` and just making this the default behavior, so long as this default is still 'safe' and that we check that there aren't any other collisions, for instance that `meltano.yml` does not already exist.

The challenge here though is that our default implementation installs a bunch of files and paths. After getting 50% through the `init`, if we fail on the 5th file (for instance), we would not want to leave the target folder in a bad or incomplete state.

@ReubenFrankel - Do you think we can ensure safety of the command before running it, so that '--force' or similar "opt in" behavioral flags are not needed?

If we do need an opt in though, then I'm totally fine with using something like `--ignore-existing`, as I can understand that `--force` implies files would be overriden in case of conflict - which is not the primary use case we're trying to handle here.

tayloramurphy:
> The challenge here though is that our default implementation installs a bunch of files and paths. After getting 50% through the init, if we fail on the 5th file (for instance), we would not want to leave the target folder in a bad or incomplete state.

Yeah, the difference of `git init` is that it's creating a single hidden directory whereas Meltano is a bit more aggressive about what it's creating. I'm with AJ on assuring safety and cleanup to keep it as a default, but if not, then making it an explicit opt in.

ReubenFrankel:
@aaronsteers @tayloramurphy Yeah, I'm confident we can implement a check for potential conflicts before anything is actually created. I'll close this PR and open a separate one - appreciate your input! :smile:

PR #6938, opened 2022-10-26
fix: Handle case where setting metadata is `None`

WillDaSilva:
Before:
```
$ meltano config meltano set state_backend.uri file:///${MELTANO_ROOT_DIR}/.meltano/state

'NoneType' object has no attribute 'is_redacted'
```

After:
```
$ meltano config meltano set state_backend.uri file:///${MELTANO_ROOT_DIR}/.meltano/state
Meltano setting 'state_backend.uri' was set in `meltano.yml`: 'file:////.meltano/state'
```

Instead of setting the non-existent setting, would we rather it error cleanly?

Closes #6937

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -41,6 +41,8 @@ ignore =
     # pycodestyle
     # Allow for line breaks where Black decides are valid
     W503
+    # Let Black put whitespace before :
+    E203
     # Allow for long lines since we delegate to Black for enforcement
     E501
     # Allow custom exceptions whose name ends with "Exception"
--- a/src/meltano/cli/interactive/config.py
+++ b/src/meltano/cli/interactive/config.py
@@ -211,9 +211,7 @@ class InteractiveConfig:  # noqa: WPS230, WPS214
 
     @staticmethod
     def _value_prompt(config_metadata):
-        if (not config_metadata["setting"]) or config_metadata[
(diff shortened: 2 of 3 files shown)


tayloramurphy:
@WillDaSilva is the actual problem here that when you do a config for a setting that doesn't exist it gives this error? Looks like this is coming up because EDK things can use the config block however they want technically, right?

WillDaSilva:
> is the actual problem here that when you do a config for a setting that doesn't exist it gives this error?

@tayloramurphy Yes. Whether or not we want to permit setting these settings is unknown to me, but we definitely shouldn't be erroring in this way.

I asked in the PR description:

> Instead of setting the non-existent setting, would we rather it error cleanly?

If we do want it to error cleanly, I can update this PR to do so.

tayloramurphy:
> If we do want it to error cleanly, I can update this PR to do so.

@WillDaSilva I'd say have it execute but give a warning that "this setting is not known to Meltano".

WillDaSilva:
> I'd say have it execute but give a warning that "this setting is not known to Meltano".

There's a nice parallel between that approach and what we do when a setting that does not exist is retrieved, which is emit a warning and return `None`.

edgarrmondragon (approved):

PR #6964, opened 2022-11-07
feat: Add `--cwd` CLI option

### LINKED ISSUE #6943: feature: `--cwd` CLI option
### Feature scope

### Description

As seen in some other CLIs, it'd be nice to have a `--cwd` option which takes a path to a directory as an argument, and runs the program as-if it were invoked from that directory. This could be implemented by calling [`os.chdir`](https://docs.python.org/3/library/os.html#os.chdir) right after the CLI options are parsed if the `--cwd` option is provided.

WillDaSilva:
Closes #6943

### DIFF
--- a/src/meltano/cli/cli.py
+++ b/src/meltano/cli/cli.py
@@ -3,7 +3,9 @@
 from __future__ import annotations
 
 import logging
+import os
 import sys
+from pathlib import Path
 from typing import NoReturn
 
 import click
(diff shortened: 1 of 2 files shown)


edgarrmondragon (approved):
I re-ran the [flaky tests](https://github.com/meltano/meltano/issues/6965) and they passed.

:shipit:

PR #6977, opened 2022-11-10
docs: Link to Meltano Cloud from Production doc

DouweM:
Related to https://github.com/meltano/internal-marketing/issues/750

### DIFF
--- a/docs/src/_guide/production.md
+++ b/docs/src/_guide/production.md
@@ -26,10 +26,13 @@ you can skip steps 1 through 3 and refer primarily to the "Containerized Meltano
 We also provide a [Helm Chart](https://github.com/meltano/helm-meltano) for deploying a containerized instance of the Meltano UI to [Kubernetes](https://kubernetes.io).
 More on that [in the Kubernetes section](#kubernetes).
 
-## SaaS Hosting Options
+## Managed Hosting Options
 
-Meltano currently does not offer project hosting as a paid offering.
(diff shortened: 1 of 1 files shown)


DouweM (commented):
Edit: accidentally put in this comment, but apparently I can't delete it

aaronsteers (approved):
👍

WillDaSilva (approved):

edgarrmondragon (approved):

sbalnojan (commented):
@DouweM Are you intentionally hiding the cloud offering in a subpage on the docs? It's fine to have it there, but IMHO it should go straight into the header as important announcement (preferrably in bolded letters ;)). 

It's a good thing to have more people know about it. Even new open source users will like to know when they can go with an optional cloud offering. And building up the waiting list is also a good thing.

DouweM:
@sbalnojan Not intentionally hiding it, but this PR was just to make sure that the one place where we already talk about managed hosting options was accurate.

The big announcement will come later this month: https://github.com/meltano/internal-marketing/issues/466 (we have enough people for the Alpha right now, but definitely want more people on the waitlist for Beta and GA), at which point I like the idea of a header banner. 

Until then, I do already want everyone who starts evaluating Meltano to know this is going to exist, hence the new button on the homepage, and I think we should mention it as an alternative to self-managing wherever makes sense. Are there other spots in the docs where you'd add it? This section comes to mind: https://docs.meltano.com/getting-started/#deploy-your-pipelines-in-production.

PR #6984, opened 2022-11-14
docs: fix tap-github

sbalnojan:
Hotfix the tutorial by pinning to the "old" non-default singer-io version of the tap.

### DIFF
--- a/docs/src/_getting-started/part1.md
+++ b/docs/src/_getting-started/part1.md
@@ -96,19 +96,19 @@ cd my-meltano-project
 
 ## Add an Extractor to Pull Data from a Source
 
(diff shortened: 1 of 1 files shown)


WillDaSilva (approved):
This looks good as a quick fix. What would it take to make the tutorial work with the default variant? Do we want to do that, and if so, is there an issue open for that?

sbalnojan:
> This looks good as a quick fix. What would it take to make the tutorial work with the default variant? Do we want to do that, and if so, is there an issue open for that?

There already is an issue for making it more self-contained, and I do think, if we make a change it makes sense to get rid of GitHub as well (as it is a little stepping stone since it requires the user to have an account and get an auth token). 

https://github.com/meltano/meltano/discussions/6950#discussioncomment-4051518

PR #6988, opened 2022-11-15
chore: Fix type hints for vars defaulting to `None`

WillDaSilva:
Found and replaced this with the regex `:(\s*(\w+\b(?<!None))\s*\|?)+=\s*None`, then ran pre-commit to apply Black, PyUpgrade, isort, etc.

CI will fail for this PR because of Flake8. We can ignore these linter alerts for now, since they are not introduced by this PR, and this PR would become bloated if it attempted to address them.


edgarrmondragon (approved):
Nice!

PR #6997, opened 2022-11-16
fix: correct meltano config behavior when explicit environment is provided

### LINKED ISSUE #6512: bug:  Conflicting behavior flags in `meltano config set`
### Meltano Version

latest

### Python Version

NA

### Bug scope

### Operating System

any

### Description

From Slack: https://meltano.slack.com/archives/C01TCRBBJD7/p1659032672217559?thread_ts=1658965446.758529&cid=C01TCRBBJD7

With the example invocation:

```console
meltano --environment=prod config target-csv set destination_path ./output
```

The expectation is that meltano would store the value within the environment config. Instead, the value is stored at the top-level meltano config.

## Workaround

The correct behavior can be achieved by also overriding `--store` as in:

```
meltano --environment=prod config target-csv set destination_path ./output --store=meltano_environment
```

## Proposed solution

The default behavior (without specifying `--store` or by using `--store=auto`) should be updated to prefer `meltano_environment` over `meltano_yml` if `--environment` is set to a valid environment name.

## Alternative Approaches

Another option is to remove or deprecate `--store=meltano_environment` in deference to the `--environment` flag being specified or not.

In that case `--store=meltano_yml` would indicate _either_ the top-level yaml declarations (if no environment is set) _or_ the environment-specific section of the yaml (if an environment is set).

Note:

- There are probably good reasons why this alternative wouldn't work, for instance if internally we need to treat these as two different stores and due to the fact that internally we would step through each layer with the top-level and environment-level yaml each being a different important part of that resolution.

### Code

_No response_

cjohnhanson:
Closes #6512


edgarrmondragon (approved):

src/meltano/cli/config.py
```diff
@@ -31,6 +32,62 @@
...
+
+    @wraps(func)
+    def _wrapper(*args, **kwargs):
+        store = kwargs.pop("store")
+        if store not in {SettingValueStore.MELTANO_YML, SettingValueStore.MELTANO_ENV}:
+            return func(*args, **kwargs, store=store)
+        ctx = _get_ctx_arg(*args)
+        if ctx.obj["is_default_environment"]:
+            return func(*args, **kwargs, store=SettingValueStore.MELTANO_YML)
+        return func(*args, **kwargs, store=SettingValueStore.MELTANO_ENV)
```
WillDaSilva, line 86: We can reduce the number of return statements here:

```suggestion
        store = (
            SettingValueStore.MELTANO_YML
            if ctx.obj["is_default_environment"]
            else SettingValueStore.MELTANO_ENV
        )
        return func(*args, **kwargs, store=store)
```

WillDaSilva (approved):

PR #7007, opened 2022-11-21
chore: Accept `client_id` from optional `MELTANO_CLIENT_ID` env var

WillDaSilva:
To do:
- [X] Testing

Relates to https://github.com/meltano/cloud-planning/issues/85

### DIFF
--- a/src/meltano/core/tracking/contexts/environment.py
+++ b/src/meltano/core/tracking/contexts/environment.py
@@ -6,9 +6,11 @@ import os
 import platform
 import uuid
 from collections import defaultdict
+from contextlib import suppress
 from datetime import datetime
 from pathlib import Path
 from typing import Any
+from warnings import warn
 
 import psutil
 from cached_property import cached_property
@@ -25,6 +27,19 @@ logger = get_logger(__name__)
 release_marker_path = Path(__file__).parent / ".release_marker"
 
 
+def _get_environment_context_uuid() -> uuid.UUID:
+    with suppress(KeyError):
+        uuid_str = os.environ["MELTANO_CONTEXT_UUID"]
+        try:
+            return uuid.UUID(uuid_str)
+        except ValueError:
+            warn(
+                f"Invalid telemetry environment context UUID {uuid_str!r} "
+                "from $MELTANO_CONTEXT_UUID - a random UUID will be used"
+            )
+    return uuid.uuid4()
+
+
 class EnvironmentContext(SelfDescribingJson):
     """Environment context for the Snowplow tracker."""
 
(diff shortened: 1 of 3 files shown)


pnadolny13 (approved):
@WillDaSilva conceptually this looks good to me! I'll let @aaronsteers do a proper review of the code though.

src/meltano/core/tracking/contexts/environment.py
```diff
@@ -25,6 +27,20 @@
...
+        uuid_str = os.environ["MELTANO_CONTEXT_UUID"]
+        try:
+            return uuid.UUID(uuid_str)
+        except ValueError:
+            warn(
+                f"Invalid telemetry environment context UUID {uuid_str!r} "
+                "from $MELTANO_CONTEXT_UUID - a random UUID will be used",
+                RuntimeWarning,
+            )
+    return uuid.uuid4()
```
aaronsteers, line 41: Should we also set the value in `os.environ` so that child processes will inherit it?

Something like the below?

```suggestion
def _init_environment_context_uuid() -> uuid.UUID:
    if "MELTANO_CONTEXT_UUID" in os.environ:
        uuid_str = os.environ["MELTANO_CONTEXT_UUID"]
        try:
            return uuid.UUID(uuid_str)
        except ValueError:
            warn(
                f"Invalid telemetry environment context UUID {uuid_str!r}"
                "from $MELTANO_CONTEXT_UUID - a random UUID will be used",
                RuntimeWarning,
            )

    uuid_str = uuid.uuid4()
    os.environ["MELTANO_CONTEXT_UUID"] = uuid_str
    return uuid_str
```

WillDaSilva: @aaronsteers We don't want the child process to inherit `MELTANO_CONTEXT_UUID`, for reasons explained here: https://github.com/meltano/meltano/pull/7015#issuecomment-1332306898:

> If we re-used an environment context UUID, then we would not be able to easily distinguish between different environment contexts from the same session. That is to say, two executions of meltano, one nested below the other in the process tree, would appear to be the same execution.

I considered deleting it from `os.environ` after reading the value, but that could get messy if, for whatever reason, another `EnvironmentContext` is instantiated. Perhaps I should just make the class a singleton; I can't think of a legitimate use for instantiating a second instance.

WillDaSilva: Resolved in a 1:1 Zoom call on 2022-12-12.

WillDaSilva:
@aaronsteers ae1a9248f46f1c530c677fcfdb3b26635cc46ff9 implements the changes we discussion. The env var `MELTANO_CLIENT_ID` is used to set the client ID instead of setting the environment context UUID. I'll update https://github.com/meltano/infra/pull/169 to reflect these changes soon.

aaronsteers (approved):

PR #7010, opened 2022-11-21
fix: Re-create plugin venv if Python executable is missing

### LINKED ISSUE #7008: bug: Missing Python executable from plugin venv causes error during `meltano install`
### Meltano Version

2.10.0

### Python Version

3.9

### Bug scope

### Operating System

Linux Mint 20.3

### Description

1. Install Python (e.g. using `pyenv`).
2. Use that version of Python to run `meltano install`.
3. Uninstall that version of Python.
4. Install a new version of Python (e.g. going from version 3.9.X to 3.9.Y)
5. Use that version of Python to run `meltano install` again:

```
Installing X plugins...
[...]

[Errno 2] No such file or directory: '<project directory>/.meltano/<installed plugin>/venv/bin/python'
```

The venvs managed by Meltano assume that the underlying Python executable that `venv/bin/python` links to will not be deleted. In reality it may be, and if it is we should detect this and automatically replace the venv (as if the `--clean` flag has been provided).

### Code

_No response_

WillDaSilva:
These changes make it so that if the Python executable used by a Meltano-managed venv is deleted, the venv will be recreated on the next `meltano install` that affects it as if the `--clean` flag had been supplied, rather than erroring.

Closes #7008

### DIFF
--- a/tests/meltano/core/test_venv_service.py
+++ b/tests/meltano/core/test_venv_service.py
@@ -117,7 +117,7 @@ class TestVenvService:
         # Make sure the venv exists already
         await subject.install(["example"], clean=True)
 
-        original_link_target = subject.python_path.readlink()
+        original_link_target = os.readlink(subject.python_path)
         try:
             # Simulate the deletion of the underlying Python executable by
             # making its symlink point to a file that does not exist
--- a/tests/meltano/core/test_venv_service.py
+++ b/tests/meltano/core/test_venv_service.py
@@ -117,17 +117,18 @@ class TestVenvService:
         # Make sure the venv exists already
         await subject.install(["example"], clean=True)
 
-        original_link_target = os.readlink(subject.python_path)
-        try:
-            # Simulate the deletion of the underlying Python executable by
-            # making its symlink point to a file that does not exist
-            subject.python_path.unlink()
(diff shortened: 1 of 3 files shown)


tests/meltano/core/test_venv_service.py
```diff
@@ -117,6 +117,18 @@ async def test_requires_clean_install(self, project, subject: VenvService):
         # Make sure the venv exists already
         await subject.install(["example"], clean=True)
 
+        original_link_target = subject.python_path.readlink()
```
edgarrmondragon, line 120: This was added in 3.9, so maybe we can port [the current 3.11 implementation](https://github.com/python/cpython/blob/3.11/Lib/pathlib.py#L1081-L1087):

```python
import os
import sys
from pathlib import Path

if sys.version_info < (3, 11):
    readlink = Path.readlink()
else:
    def readlink(self):
        """Return the path to which the symbolic link points."""
        if not hasattr(os, "readlink"):
            raise NotImplementedError("os.readlink() not available on this system")
        return self._from_parts((os.readlink(self),))

...

        original_link_target = readlink(python_path)
```

edgarrmondragon (changes requested):

edgarrmondragon (approved):

PR #7015, opened 2022-11-22
chore: Set environment context UUID for subprocesses

WillDaSilva:
This PR extends https://github.com/meltano/meltano/pull/7007, and should not be merged before it. With https://github.com/meltano/meltano/pull/7007, we can set the environment context UUID using an env var. With this PR, we that env var is automatically set by Meltano when invoking Meltano or a plugin, and is reported using the new `parent_context_uuid` field within version 1-1-0 of the environment context.

Once this PR has been approved, but before it is merged, I will publish the new environment context schema version to Snowcat Cloud.

Meltano can invoke a plugin, which then calls into Meltano (e.g. to list schedules). Now telemetry from the child Meltano process can more easily be related to telemetry from the parent process. This was previously possible using the data from the process tree, but it was difficult.

Most of the changes in this PR were done to make it so that importing the `Tracker` class doesn't import much else. The need to import the tracker in a couple new spots was causing circular imports.

Relates to https://github.com/meltano/cloud-planning/issues/85

### DIFF
--- a/src/meltano/core/tracking/iglu-client-embedded/schemas/com.meltano/environment_context/jsonschema/1-1-0
+++ b/src/meltano/core/tracking/iglu-client-embedded/schemas/com.meltano/environment_context/jsonschema/1-1-0
@@ -1,9 +1,9 @@
 {
     "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
-    "description": "Schema for a Snowplow context describing the environment in which a Meltano project is running",
+    "description": "Schema for a Snowplow context describing the environment in which the Meltano process is executing",
     "self": {
         "vendor": "com.meltano",
-        "name": "project_context",
+        "name": "environment_context",
         "format": "jsonschema",
         "version": "1-1-0"
     },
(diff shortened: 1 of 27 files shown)


src/meltano/core/tracking/iglu-client-embedded/schemas/com.meltano/environment_context/jsonschema/1-1-0
```diff
@@ -0,0 +1,180 @@
...
+    "type": "object",
+    "properties": {
+        "context_uuid": {
+            "description": "A UUID that uniquely identifies an instance of this context",
+            "type": "string",
+            "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$",
+            "minLength": 36,
+            "maxLength": 36
+        },
+        "parent_context_uuid": {
```
pnadolny13, line 19: Sharing the main diff of this file vs the 1-0-0 version.

Screen Shot 2022-11-30 at 9 17 22 AM

pnadolny13:
@WillDaSilva this PR looks good to me although I dont 100% follow the use case for this attribute yet. Not saying we shouldnt implement this but just saying that I dont have full context. I saw this comment https://github.com/meltano/cloud-planning/issues/85#issuecomment-1317559771 but I thought that was a theoretical situation that doesnt exist, is this to cover a future case that were planning to support soon? What the benefit of having a parent context uuid attribute vs continuing to pass the same context uuid to all child processes?

I think the main question is what we consider to be a single "session". Is it 1-1 with a meltano process or 1-1 with a CLI execution (i.e. if I run a CLI command that calls other stuff its still 1 single cli "session" context vs 1 CLI command turning into many "sessions" contexts based on subprocesses).

For the `dbt-ext` case are we saying that plugins would send telemetry events itself? I'm trying to think through the benefit of using a parent context ID vs just setting the normal context ID to that value. So a case would be setting the uuid in an env var and propagating to each level - Env var context uuid-> cloud context uuid -> meltano context uuid -> dbt-ext context uuid. So grouping by context uuid would give me all events for that "session" across cloud/meltano core/plugin/extension. Otherwise we're going to end up grouping by context uuid and parent context uuid, which isnt a problem I'm only wondering what benefit we get. Again - I'm not pitching a different implementation, just I'm asking to understand.

For the meltano calling meltano case we could technically have an infinite chain of meltano calling itself, in which case the parent context IDs would link them all together. Is that the idea?

cc @tayloramurphy

WillDaSilva:
> I think the main question is what we consider to be a single "session". Is it 1-1 with a meltano process or 1-1 with a CLI execution (i.e. if I run a CLI command that calls other stuff its still 1 single cli "session" context vs 1 CLI command turning into many "sessions" contexts based on subprocesses).

@pnadolny13 I define a "session" as a logical execution of Meltano, regardless of how many processes are used to run it.

If we re-used an environment context UUID, then we would not be able to easily distinguish between different environment contexts from the same session. That is to say, two executions of `meltano`, one nested below the other in the process tree, would appear to be the same execution.

We want to be able to relate these executions. To know that one caused the other, and to be able to interpret the data collected from the child processes in the context of the execution of the parent. We do not want to conflate the nested executions, and treat them as the same process/environment.

Distinguishing between environment contexts while still being able to relate them together is why `parent_context_uuid` exists. The ability to set the `context_uuid` using an env var (implemented in #7007) has a different use-case entirely: it is supported so that Meltano Cloud can relate the telemetry emitted from it directly with the telemetry emitted by `meltano` running under it. We currently do not collect any telemetry from Meltano Cloud directly - we only collect that which runs from `meltano` running under it at this time.

> For the dbt-ext case are we saying that plugins would send telemetry events itself?

No, but if `meltano` is run by the plugin then `meltano` will send telemetry with the `parent_context_uuid` set.

> I'm trying to think through the benefit of using a parent context ID vs just setting the normal context ID to that value.

If we set the context ID to that of the parent, then we lose the ability to easily distinguish the data from the parent and child processes.

> So a case would be setting the uuid in an env var and propagating to each level - Env var context uuid-> cloud context uuid -> meltano context uuid -> dbt-ext context uuid.

I'm not sure what you mean by this. This PR does not have much to do with setting the context UUID. That would be #7007.

> For the meltano calling meltano case we could technically have an infinite chain of meltano calling itself, in which case the parent context IDs would link them all together. Is that the idea?

Yes.

pnadolny13:
@WillDaSilva thanks for the detail, I think I'm on the same page as you now! This looks good to me from a functional perspective 👍  but again I'll leave it to AJ or whoever else to approve the actual code changes.

aaronsteers (approved):

PR #7017, opened 2022-11-22
fix: Remove premature deserialization of settings in the CLI

edgarrmondragon:
See in Slack https://meltano.slack.com/archives/C03GKHWS0HM/p1669128539584859

### DIFF
--- a/src/meltano/cli/interactive/config.py
+++ b/src/meltano/cli/interactive/config.py
@@ -2,9 +2,6 @@
 
 from __future__ import annotations
 
-import contextlib
-import json
-
 import click
 from jinja2 import BaseLoader, Environment
 from rich.console import Console, Group
@@ -368,9 +365,6 @@ class InteractiveConfig:  # noqa: WPS230, WPS214
 
     def set_value(self, setting_name, value, store, interactive=False):
         """Set value helper function."""
-        with contextlib.suppress(json.JSONDecodeError):
-            value = json.loads(value)
-
         settings = self.settings
         path = list(setting_name)
         try:
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -22,6 +22,7 @@ jobs:
         - { integration_test: "meltano-migrations", needs_postgres: false}
         - { integration_test: "meltano-state-s3", needs_postgres: false}
         - { integration_test: "meltano-state-local", needs_postgres: false}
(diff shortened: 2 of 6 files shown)


src/meltano/cli/interactive/config.py
```diff
@@ -368,9 +365,6 @@ def configure_all(self):
 
     def set_value(self, setting_name, value, store, interactive=False):
         """Set value helper function."""
-        with contextlib.suppress(json.JSONDecodeError):
-            value = json.loads(value)
```
WillDaSilva, line 372: @edgarrmondragon @kgpayne Why was this here in the first place?

tayloramurphy (approved):
Approving - I absolutely _love_ these integration examples. Super clear what's happening.

src/meltano/cli/interactive/config.py
kgpayne: It was added by this [MR](https://gitlab.com/meltano/meltano/-/merge_requests/1764/diffs) back in Gitlab world. We do accept json-formatted strings using `meltano set` for (object and array types), so this was likely added to solve for that use case. As @edgarrmondragon mentions, we do cast by kind at a later stage now, so this may just be an older artefact 🤷‍♂️

edgarrmondragon: @WillDaSilva I found https://github.com/meltano/meltano/commit/6c912d65e58017c781df212bba1b464b29a12ebb and https://gitlab.com/meltano/meltano/-/merge_requests/1764/.

Though that work was done after `object` and `array` were introduced https://github.com/meltano/meltano/commit/fcda3c948c572a28c16dd09a2480a4bfc0f1aae4 and https://github.com/meltano/meltano/commit/12362f93f01dcacb54251e5d18c2928f55c12dd9.

So, I'm gonna do some testing with `meltano config ... set` for schemas to see if anything's off.

edgarrmondragon: Setting the `_schema` extra works fine too, as any normal `object` 🤔

@kgpayne @WillDaSilva it seems safe to merge. Wdyt?

kgpayne (approved):

src/meltano/cli/interactive/config.py
kgpayne: LGTM 🚢

WillDaSilva (approved):

PR #7019, opened 2022-11-23
feat: Support env vars in plugin `pip_url`

### LINKED ISSUE #7018: feature: Support env vars in plugin `pip_url`
### Feature scope

### Description

We currently support env vars in plugins' `pip_url` field, but only for `$MELTANO__PYTHON_VERSION`:
- https://github.com/meltano/meltano/issues/2529 (implemented by https://gitlab.com/meltano/meltano/-/merge_requests/2256)

I don't see any reason why we shouldn't expand any env var in the `pip_args` field. One reason in particular why we should is because there are no good ways to supply the credentials to access a private Python package index except within its URL. Since env vars generally cannot be used within `pip_url` currently, that requires credentials be added to `meltano.yml` as plain text.

See also:
- https://github.com/meltano/cloud-docs/issues/2#issuecomment-1324304952
- #3422 (should be done after this issue has been addressed)

### LINKED ISSUE #7031: Make `meltano-env-precedence` example more easily callable/demoable
_Originally posted by @aaronsteers in https://github.com/meltano/meltano/pull/7019#discussion_r1033893644_
      
This simplifies reproducibility by declaring a couple `echo` commands which print out the respective values.

```py
#...
plugins:
  utilities:
  - name: example
    # ...
    settings:
    - name: echo_text
      env: ECHO_TEXT
    config:
      echo_text: ${STACKED}
    commands:
      echo-stacked:        
        # Echoed text is 1234 (no environment) or 12345 (in 'prod')
        executable: echo
        args: "Stacked env var value is: ${STACKED}"
      echo-config:
        # FIXME: Should be same as echo_stacked 1234.
        #        Currently prints partially resolved: '${STACKED}4'
        executable: echo
        args: "Config value is: ${ECHO_TEXT}"
    env:
```

Example usage shows env vars are working correctly when referenced as env vars:

```console
cd docs/example-library/meltano-env-precedence
poetry run meltano invoke example:echo-stacked
> Stacked env var value is: 1234
poetry run meltano --environment=prod invoke example:echo-stacked
> Stacked env var value is: 12345
```

However the stacked referenced does not appear to be correctly resolved when the config value references it:

```console
poetry run meltano invoke example:echo-config
> Echoed text is: ${STACKED}4
poetry run meltano --environment=prod invoke example:echo-config
> Echoed text is: ${STACKED}5
```

WillDaSilva:
Closes #7018

Closes #7031

### DIFF
--- a/docs/example-library/meltano-env-precedence/ending-meltano.yml
+++ b/docs/example-library/meltano-env-precedence/ending-meltano.yml
@@ -14,7 +14,21 @@ plugins:
     # Note that `meltano install` runs with no active environment,
     # so 3 and 5 are absent from the expanded string.
     pip_url: -e env-var-in-pip-url-example-${STACKED}  # 'env-var-in-pip-url-example-124'
-    executable: example
+    settings:
+    - name: echo_text
+      env: ECHO_TEXT
+    config:
+      echo_text: ${STACKED}
+    commands:
+      echo-stacked:
+        # Echoed text is 1234 (no environment) or 12345 (in 'prod')
+        executable: echo
+        args: "Stacked env var value is: ${STACKED}"
+      echo-config:
+        # FIXME: Should be same as echo_stacked 1234.
+        #        Currently prints partially resolved: '${STACKED}4'
+        executable: echo
+        args: "Config value is: ${ECHO_TEXT}"
     env:
       # Level 4: plugin-level `env:`
       # Inherits from a environment-level `env:` if an environment is active
--- a/docs/example-library/meltano-env-precedence/ending-meltano.yml
+++ b/docs/example-library/meltano-env-precedence/ending-meltano.yml
@@ -1,0 +1,39 @@
+version: 1
+project_id: faa14098-8fdc-4336-850e-96db154b7990
+send_anonymous_usage_stats: false
+env:
+  # Level 2: top-level `env:`
+  # Inherits from terminal context
+  LEVEL_NUM: '2'                  #  '2'
+  STACKED: ${STACKED}2            # '12'
+plugins:
+  utilities:
+  - name: example
+    namespace: example
+    # Env vars in `pip_url` are expanded during `meltano install`.
+    # Note that `meltano install` runs with no active environment,
+    # so 3 and 5 are absent from the expanded string.
+    pip_url: -e env-var-in-pip-url-example-${STACKED}  # 'env-var-in-pip-url-example-124'
+    executable: example
+    env:
+      # Level 4: plugin-level `env:`
+      # Inherits from a environment-level `env:` if an environment is active
+      # Inherits directly from top-level `env:` if no environment is active
+      LEVEL_NUM: '4'              #    '4'
+      STACKED: ${STACKED}4        # '1234'
+environments:
+- name: prod
+  config:
+    plugins:
+      utilities:
+      - name: example
+        env:
+          # Level 5: environment-level plugin `env:`
+          # Inherits from (global) plugin-level `env:`
+          LEVEL_NUM: '5'              #     '5'
+          STACKED: ${STACKED}5        # '12345'
+  env:
+    # Level 3: environment-level `env:`
+    # Inherits from top-level `env:`
+    LEVEL_NUM: '3'                #   '3'
+    STACKED: ${STACKED}3          # '123'
--- a/.flake8
+++ b/.flake8
@@ -154,8 +154,8 @@ per-file-ignores =
         D102
         D103
         D104
-	D105
-	D107
+        D105
+        D107
         # String does contain unindexed parameters
         P103
         # Allow underscored number name pattern (e.g. tap_covid_19)
@@ -164,8 +164,8 @@ per-file-ignores =
         WPS204
         # Don't warn on too many arguments in test methods
         WPS211
-	# Don't warn on too many local variables in test methods
-	WPS210
+        # Don't warn on too many local variables in test methods
+        WPS210
         # Don't warn on too many `assert` statements in test methods
         WPS218
         # Found too many expressions
@@ -194,8 +194,8 @@ per-file-ignores =
         WPS425
         # Found too long name
         WPS118
-	# Avoid false positive for implicit raw strings
-	WPS342
+    # Avoid false positive for implicit raw strings
+    WPS342
     !tests/*:
         T
     scripts/*:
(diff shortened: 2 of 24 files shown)


WillDaSilva:
Is there anywhere we document which fields support env vars? If so, it'll need to be updated by this PR.

edgarrmondragon:
> Is there anywhere we document which fields support env vars? If so, it'll need to be updated by this PR.

@WillDaSilva maybe https://docs.meltano.com/guide/configuration#environment-variable-expansion

WillDaSilva:
> maybe [docs.meltano.com/guide/configuration#environment-variable-expansion](https://docs.meltano.com/guide/configuration#environment-variable-expansion)

Thanks @edgarrmondragon. That reminds me that this'll need to support interpolation from the `env` field at the `meltano.yml` level, the `environment` level, the `plugin` level, and the `environment.plugin` level. Should be straightforward.

tayloramurphy:
@WillDaSilva can you add an update to the docs around this too? In pt 2 on https://docs.meltano.com/guide/configuration#configuration-layers we say:

> Inside values, [environment variables can be referenced](https://docs.meltano.com/guide/configuration#expansion-in-setting-values) as $VAR (as a single word) or ${VAR} (inside a word).

Which links to https://docs.meltano.com/guide/configuration#expansion-in-setting-values We also have https://docs.meltano.com/guide/configuration#environment-variable-expansion which comes before expansion in settings. I think a short new section in that part of the docs detailing where in the meltano.yml we can specify an environment variable and expect it to be expanded appropriately would work.

sbalnojan:
Fwiw @tayloramurphy I think this is a very demoworthy feature.

aaronsteers:
@WillDaSilva - Seems like there's a few things happening here and I want to confirm.

Is this right?:

- New: Support env vars in plugin `pip_url`
- Fix (or New): `env:` mappings now resolve at all levels of the `meltano.yml` level: top level, environment level, plugin level, and environment.plugin level

The reason I ask is because we've be incrementally trying to make sure that all `env: {}` mappings correctly resolve and inherit from each other. I'm not clear from this PR if we fully resolve the `env` inheritance issues or what the current state is on that.

cc @cjohnhanson

WillDaSilva:
> Is this right?:
> - New: Support env vars in plugin pip_url
> - Fix (or New): env: mappings now resolve at all levels of the meltano.yml level: top level, environment level, plugin level, and environment.plugin level

@aaronsteers Yes to the first point, no to the second. I have not made changes to the `env` mapping in this PR.

WillDaSilva:
@tayloramurphy I've updated the docs for this in https://github.com/meltano/meltano/pull/7030

docs/example-library/meltano-env-precedence/ending-meltano.yml
```diff
@@ -0,0 +1,39 @@
...
+plugins:
+  utilities:
+  - name: example
+    namespace: example
+    # Env vars in `pip_url` are expanded during `meltano install`.
+    # Note that `meltano install` runs with no active environment,
+    # so 3 and 5 are absent from the expanded string.
+    pip_url: -e env-var-in-pip-url-example-${STACKED}  # 'env-var-in-pip-url-example-124'
+    executable: example
+    env:
```
aaronsteers, line 18: This simplifies reproducibility by declaring a couple `echo` commands which print out the respective values.

```suggestion
    pip_url: -e env-var-in-pip-url-example-${STACKED}  # 'env-var-in-pip-url-example-124'
    settings:
    - name: echo_text
      env: ECHO_TEXT
    config:
      echo_text: ${STACKED}
    commands:
      echo-stacked:        
        # Echoed text is 1234 (no environment) or 12345 (in 'prod')
        executable: echo
        args: "Stacked env var value is: ${STACKED}"
      echo-config:
        # FIXME: Should be same as echo_stacked 1234.
        #        Currently prints partially resolved: '${STACKED}4'
        executable: echo
        args: "Config value is: ${ECHO_TEXT}"
    env:
```

Example usage shows env vars are working correctly when referenced as env vars:

```console
cd docs/example-library/meltano-env-precedence
poetry run meltano invoke example:echo-stacked
> Stacked env var value is: 1234
poetry run meltano --environment=prod invoke example:echo-stacked
> Stacked env var value is: 12345
```

However the stacked referenced does not appear to be correctly resolved when the config value references it:

```console
poetry run meltano invoke example:echo-config
> Echoed text is: ${STACKED}4
poetry run meltano --environment=prod invoke example:echo-config
> Echoed text is: ${STACKED}5
```

I hope this is helpful. I authored and tested the above in codespaces. I could put this in a separate PR but it seems that would probably just slow you down. Lmk.

aaronsteers: Note: In order to invoke this, I had to comment out the `pip_url`. Is there a way to make this 'real' so it is runnable but also demonstrate the use of env vars? Maybe pass the name of a real repo, pypi image, or local path?

Even after commenting out `pip_url`, Meltano will still complain about 'plugin not installed' if it can't find the executable. Easy fix for that is to delete `executable: example` (it's not required) or replace with `executable: echo`.

aaronsteers: @WillDaSilva - Per your comment here, I understand the scope of this PR is not to resolve precedence/resolution issues broadly. Totally fine to hold off on the above suggestions and keep these for a later PR.

aaronsteers: A little late, I just saw the related bug report here:

- https://github.com/meltano/meltano/issues/7023

afolson (approved):

docs/example-library/meltano-env-precedence/ending-meltano.yml
WillDaSilva: > I hope this is helpful.

It was, thanks!

154f1c6cf4a5051d07eea95f52f1f817c4112c30

I was a little surprised that `meltano invoke` cannot invoke executables from outside the plugin venv. I made a minimal Python module that provides an `echo` command to work around that.

WillDaSilva: @aaronsteers I also opened an issue to track the problem you found with `meltano invoke example:echo-config`: 
- https://github.com/meltano/meltano/issues/7034

cjohnhanson (approved):
Lg2me

edgarrmondragon (approved):

PR #7026, opened 2022-11-25
fix: Permit `env` under plugins in `meltano.yml` schema

### LINKED ISSUE #7025: bug: `meltano.yml` JSON schema does not permit plugin-level `env`
### Meltano Version

2.10.0

### Python Version

NA

### Bug scope

### Operating System

N/A

### Description

```yaml
plugins:
  utilities:
  - name: example
    namespace: example
    env:
      KEY: VALUE
```

The above minimal `meltano.yml` converts to the following JSON:

```json
{
  "plugins": {
    "utilities": [
      {
        "name": "example",
        "namespace": "example",
        "env": {
          "KEY": "VALUE"
        }
      }
    ]
  }
}
```

According to [`schemas/meltano.schema.json`](https://github.com/meltano/meltano/blob/55c378460ee39f6ad1ccadc7e38a77992066a550/schema/meltano.schema.json) this is invalid. You can verify this at https://jsonschemalint.com/#!/version/draft-07/markup/json. The error is:

```
.plugins.utilities[0] should NOT have additional properties
```

This issue appears to be caused by the combination of `allOf` with `additionalProperties`: https://github.com/meltano/meltano/blob/55c378460ee39f6ad1ccadc7e38a77992066a550/schema/meltano.schema.json#L587

This issue applies to all plugin types, both at the top-level and under the `environments` property.

### Code

_No response_

WillDaSilva:
I have added `env` as a permitted property under each plugin type except for file bundles. Additionally I ran the file through the default VS Code JSON autoformatter, mostly to fix the inconsistent indentation.

I'll make a follow-up PR to add a JSON formatter to `pre-commit`.

Closes #7025

### DIFF
--- a/schema/meltano.schema.json
+++ b/schema/meltano.schema.json
@@ -27,42 +27,42 @@
       "description": "A database URI for the project system database. Defaults to a SQLite file stored at .meltano/meltano.db",
       "default": "sqlite:///$MELTANO_PROJECT_ROOT/.meltano/meltano.db"
     },
-      "state_backend": {
-        "type": "object",
-          "description": "Configuration for state backend.",
+    "state_backend": {
+      "type": "object",
+      "description": "Configuration for state backend.",
+      "properties": {
+        "uri": {
+          "type": "string",
+          "description": "The URI pointing to the backend to use.",
+          "default": "systemdb"
+        },
+        "lock_timeout_seconds": {
+          "type": "integer",
+          "description": "The number of seconds a lock on a state file is considered valid."
+        },
+        "lock_retry_seconds": {
+          "type": "integer",
+          "description": "The number of seconds to wait between retrying lock acquisition."
+        },
+        "azure": {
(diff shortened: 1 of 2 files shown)


edgarrmondragon (approved):

schema/meltano.schema.json
```diff
@@ -688,6 +722,7 @@
               "settings_group_validation": {},
               "commands": {},
               "requires": {},
+              "env": {},
```
tayloramurphy, line 725: @WillDaSilva do we need an update to https://docs.meltano.com/reference/plugin-definition-syntax ?

WillDaSilva: Yes. Thanks for calling that out.

WillDaSilva: e6fef197dc1305198675682ac19cb6964a89fa88

aaronsteers (approved):

Issue #7004, opened 2022-11-20
Support xcopy-esque venv installation


aaronsteers:
## Assumptions

1. We're running on Docker with explicit control of the base image.
1. Meltano is already installed.
1. The container may or may not have a `.meltano` folder created as of yet.
1. We have cached one or more venv folder(s) running on the same exact Docker image (meaning: also same exact version of Pip+Python).
1. We are controlling for changes in `pip_url` via a hash or other constraint.

## Requirements

1. Assuming all of the above have been controlled for, allow 'near-instant' plugin installation via xcopy into the correct location. **OR:**
2. Allow Meltano to read from a different already-existing directory when searching for a specific plugin's `venv`.

## Related

- The GitHub `python-setup` action proves this is possible, with near-instant bootup of projects and no installation needed when the precheck of the installation definition hashes matches an existing cache.
- This tackles a similar challenge as #2089 and related issues, but instead of skipping installs we try to make them near-instantaneous.

aaronsteers:
@z3z1ma - I think you mentioned you have a working implementation via PEX? Is that right?

https://pex.readthedocs.io/en/v2.1.123/

z3z1ma:
@aaronsteers yes I do. I actually plan to drop the project on github soon so you can dig in yourself!

aaronsteers:
Very timely: Dagster is also moving to PEX for some of their Python installs.

https://dagster.io/blog/fast-deploys-with-pex-and-docker

>  What is PEX?
> Short for Python Executable, [pex](https://pex.readthedocs.io/) is a tool that bundles Python packages into files called pex files. These are executable files that carry Python packages and some bootstrap code within them...
> 
> Using pex in combination with S3 for storing the pex files, we built a system where the fast path avoids the overhead of building and launching Docker images.

aaronsteers:
I got this working today with `shiv`, another alternative to `pex.

Prereqs: a valid Meltano project, with `meltano` and `pipx` installed

### Steps to reproduce

## 0 - Setup some vars

Just so the later code can be generic and easier to read:

```
export PLUGIN_NAME=target-duckdb
export PIP_URL=target-duckdb
```

### Step 1: Build the `pyz` file.

```sh
pipx run shiv -c $PLUGIN_NAME -o ./$PLUGIN_NAME.pyz $PIP_URL
```

Which resolves to:

```console
pipx run shiv -c target-duckdb -o ./target-duckdb.pyz target-duckdb
```

[...]

PR #7030, opened 2022-11-28
docs: Env var expansion within `pip_url`

### LINKED ISSUE #3422: Document how to use alternative PyPI indexes
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3506

Originally created by @WillDaSilva on 2022-05-18 16:36:46

---

https://docs.meltano.com/concepts/plugins#project-plugins

The docs linked above explains that you need to tell Meltano where to obtain a Python package for a plugin, and that it can be identified by its name on PyPI:
> where to find the package, typically a pip package identified by its name on PyPI, public or private Git repository URL, or local directory path,

In this case "on PyPI" refers to the default PyPI server at https://pypi.org/. I was not able to find an explanation for how one would use an alternative PyPI index server, e.g. from a mirror, or an internal company server (possibly provided by Nexus, or even something like [simplepypi](https://github.com/steiza/simplepypi)).

WillDaSilva:
The `PIP_INDEX_URL` env var could also be used to access alternative Python package indexes without having credentials in plain text, but that may mess with our currently-implemented venv caching strategy (which is flawed), so I didn't mention it here.

Relates to https://github.com/meltano/meltano/pull/7019

Closes https://github.com/meltano/meltano/issues/3422

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -154,8 +154,8 @@ per-file-ignores =
         D102
         D103
         D104
-	D105
-	D107
+        D105
+        D107
         # String does contain unindexed parameters
         P103
         # Allow underscored number name pattern (e.g. tap_covid_19)
@@ -164,8 +164,8 @@ per-file-ignores =
         WPS204
         # Don't warn on too many arguments in test methods
         WPS211
-	# Don't warn on too many local variables in test methods
-	WPS210
+        # Don't warn on too many local variables in test methods
+        WPS210
         # Don't warn on too many `assert` statements in test methods
         WPS218
         # Found too many expressions
@@ -194,8 +194,8 @@ per-file-ignores =
         WPS425
         # Found too long name
         WPS118
-	# Avoid false positive for implicit raw strings
-	WPS342
+    # Avoid false positive for implicit raw strings
+    WPS342
     !tests/*:
         T
     scripts/*:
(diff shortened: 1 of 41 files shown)


edgarrmondragon (approved):

WillDaSilva:
> is there an issue for [the flaws in our venv caching strategy]?

@edgarrmondragon No. I've been mulling it over as I think about https://github.com/meltano/meltano/issues/7004, and while I was working on `venv_service`, but there's no issue specifically for it. I'm not sure we ought to open one until we believe that a solution (or significant improvement) is possible. Also in practice this does not appear to be a common source of friction.

The fact that we cannot know for certain what the dependency tree of a package is before installing it on that specific machine makes it difficult, if not impossible, to provide perfect venv caching. The looseness of `pip_url` also makes this challenging.

One of the issues with our venv caching strategy currently is that information that can drastically affect what's in the resulting venv can come from outside of the `pip_url`, or can be presented within the `pip_url` in different forms. The former leads to cache hits that should be misses, and the latter leads to cache misses that should be hits.

If we could refactor it such that it is no long a simple string field, but is instead a `dependencies` list for which each entry is a dictionary with specific information (e.g. the exact version, whether it's a pip dependency or something else, the PyPI server, etc.), then we may be able to make significant improvements here. In that case we'd need to ensure that no env vars can be access during installation - all relevant info must come from the `dependencies` section.

tayloramurphy (approved):

tayloramurphy:
Approved - note that the docs preview didn't run for this. The content looks good though!

afolson (approved):

PR #7035, opened 2022-11-29
docs: Update custom extractor Poetry instructions

EMCP:
the build and publish command have been changed in poetry

### DIFF
--- a/docs/src/_tutorials/custom-extractor.md
+++ b/docs/src/_tutorials/custom-extractor.md
@@ -470,7 +470,7 @@ your tap directly to PyPI.
 
 1. Create an account with [PyPI](https://pypi.org).
 2. Create a PyPI API token for use in automated publishing. (Optional but recommended.)
-3. Run `poetry build` from within your repo to build 
+3. Run `poetry build` from within your repo to build.
 4. Run `poetry publish` to push your latest version to the PyPI servers.
 
 ### Test a `pip` install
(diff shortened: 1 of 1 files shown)


EMCP:
@afolson , thanks for the quick response.. hope this format works otherwise I can use the multi line ```bash approach

```bash
like
so
```

afolson (approved):

afolson:
@edgarrmondragon Is this the best way to do this now? Looks like `--build` was deprecated a bit ago.

edgarrmondragon (approved):

docs/src/_tutorials/custom-extractor.md
```diff
@@ -470,8 +470,8 @@ your tap directly to PyPI.
 
 1. Create an account with [PyPI](https://pypi.org).
 2. Create a PyPI API token for use in automated publishing. (Optional but recommended.)
-3. Run `poetry --build publish` from within your repo to build and push your latest version
-   to the PyPI servers.
+3. Run `poetry build` from within your repo to build
```
edgarrmondragon, line 473:
```suggestion
3. Run `poetry build` from within your repo to build.
```

edgarrmondragon (changes requested):

edgarrmondragon:
> @edgarrmondragon Is this the best way to do this now? Looks like `--build` was deprecated a bit ago.

@afolson yup, the official docs now use commands `build` and `publish`

tayloramurphy (approved):

edgarrmondragon (approved):

edgarrmondragon:
Thanks @EMCP!

PR #7046, opened 2022-12-02
docs: Fix internal links and `meltano config` example code block in state backends page

edgarrmondragon:


cjohnhanson (approved):
Looks good, thanks for the quick fix.

PR #7057, opened 2022-12-06
fix: Ensure `environment_name_hash` is collected

WillDaSilva:
In https://github.com/meltano/meltano/pull/6862 we defered the activate of environments. The `Tracker` was implicitly relying on the environment being activated before it was instantiated, or no environment being activated at all.

Unfortunately fixing that implicit dependency would be more effort than I'm willing to commit to this at this time. Instead, I have made it so that activating an environment will update `environment_name_hash` in the `ProjectContext` in the `Tracker`.

New `ProjectContext` objects created past that point will, as before, get the active environment information from the `Project` object.

I have verified that this works using Snowplow Micro locally. Before this change `environment_name_hash` was always `null`. After this change, `environment_name_hash` is set to the expected hashed string when an environment is active, and `null` when no environment is active. I would encourage at least one reviewer to double check this.

Closes https://github.com/meltano/internal-data/issues/51

### DIFF
--- a/src/meltano/cli/utils.py
+++ b/src/meltano/cli/utils.py
@@ -584,26 +584,31 @@ def activate_explicitly_provided_environment(
         activate_environment(ctx, project)
 
 
-class CliEnvironmentAction(Enum):
-    """Enum of the different ways in which a Meltano environment can be activated."""
+class CliEnvironmentBehavior(Enum):
+    """Enum of the different Meltano environment activation behaviours."""
 
-    activate = auto()
-    activate_required = auto()
-    activate_explicitly_provided = auto()
+    # Use explicit environment, or `default_environment`, or fail.
+    environment_required = auto()
+
+    # Use explicit environment, or `default_environment`, or no environment.
+    environment_optional_use_default = auto()
+
+    # Use explicit environment, or no environment; ignore `default_environment`.
+    environment_optional_ignore_default = auto()
 
 
-def enact_environment_action(
-    action: CliEnvironmentAction | None,
+def enact_environment_behavior(
+    behavior: CliEnvironmentBehavior | None,
     ctx: click.Context,
 ) -> None:
     """Activate the environment in the specified way."""
-    if action is None:
+    if behavior is None:
         return
-    if action is CliEnvironmentAction.activate:
+    if behavior is CliEnvironmentBehavior.environment_optional_use_default:
         activate_environment(ctx, ctx.obj["project"], required=False)
-    elif action is CliEnvironmentAction.activate_required:
+    elif behavior is CliEnvironmentBehavior.environment_required:
         activate_environment(ctx, ctx.obj["project"], required=True)
-    elif action is CliEnvironmentAction.activate_explicitly_provided:
+    elif behavior is CliEnvironmentBehavior.environment_optional_ignore_default:
         activate_explicitly_provided_environment(ctx, ctx.obj["project"])
 
 
@@ -613,18 +618,18 @@ class InstrumentedCmdMixin:
     def __init__(
         self,
         *args,
-        environment_action: CliEnvironmentAction | None = None,
+        environment_behavior: CliEnvironmentBehavior | None = None,
         **kwargs,
     ):
         """Initialize the `InstrumentedCmdMixin`.
 
         Args:
             args: Arguments to pass to the parent class.
-            environment_action: The action to take regarding the activation of
-                the Meltano environment for this command.
+            environment_behavior: The behavior to use regarding the activation
+                of the Meltano environment for this command.
             kwargs: Keyword arguments to pass to the parent class.
         """
-        self.environment_action = environment_action
+        self.environment_behavior = environment_behavior
         super().__init__(*args, **kwargs)
 
 
@@ -634,7 +639,7 @@ class InstrumentedGroupMixin(InstrumentedCmdMixin):
     def invoke(self, ctx: click.Context):
         """Update the telemetry context and invoke the group."""
         ctx.ensure_object(dict)
-        enact_environment_action(self.environment_action, ctx)
+        enact_environment_behavior(self.environment_behavior, ctx)
         if ctx.obj.get("tracker"):
             ctx.obj["tracker"].add_contexts(CliContext.from_click_context(ctx))
         super().invoke(ctx)
@@ -658,7 +663,7 @@ class InstrumentedCmd(InstrumentedCmdMixin, click.Command):
     def invoke(self, ctx: click.Context):
         """Invoke the requested command firing start and events accordingly."""
         ctx.ensure_object(dict)
-        enact_environment_action(self.environment_action, ctx)
+        enact_environment_behavior(self.environment_behavior, ctx)
         if ctx.obj.get("tracker"):
             tracker = ctx.obj["tracker"]
             tracker.add_contexts(CliContext.from_click_context(ctx))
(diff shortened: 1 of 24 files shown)


tayloramurphy:
@WillDaSilva if it fixes it then I'm happy! I'd like to have us eventually get some kind of testing in place so we can detect when fields are accidentally dropped. I'm not hitting the approve button since there are no docs changes and I can't review this code - cc @aaronsteers

WillDaSilva:
> I'd like to have us eventually get some kind of testing in place so we can detect when fields are accidentally dropped.

@tayloramurphy @aaronsteers I recall the entire telemetry refactor was done in a rush for the 2.0 release, and we still haven't addressed some of the tech debt we accumulated in the process.

Right now we've got tests that ensure that we don't generate "bad" events by running Snowplow Micro in the tests, but for cases like this it's tricky since `null` is a valid value for `environment_name_hash`. Additional unit testing would not have caught this either, since each testable piece was working fine in isolation, but when these pieces were instantiated in the wrong order this bug was the result.

@meltano/engineering Meltano core/CLI refactor proposal/rant below:

The main flaw that I think led to this is that the `Tracker` takes the `Project` instance as an argument (and instantiates the `ProjectContext` using that `Project` instance), but the `Project` instance is mutable, and will sometimes have its environment changed after instantiation. I have two non-mutually-exclusive solutions in mind:

1. Make `Project` immutable - its environment must be specified when it is initialized.

    When `Project` instances are immutable, anything that needs one can instantiate one with `Project.find()`, except that is redundant because `Project.find(project_root)` and `Project(project_root)` are practically the same. We can remove `Project.find` in favor of having `Project.__init__` automatically find the project if `project_root` is `None`. Note that `Project.__init__` is currently only used within `Project.find` and `ProjectInitService.init`, and both can be replaced easily.

    One issue that would arise with immutable `Project` instances is that `Project` acts like a singleton: `Project.find`, which is used almost exclusively instead of `Project.__init__`, will cache its results by default (referred to as "activating" the project). Once a project has been activated `Project.find` will always return that project object, regardless of what project root directory is provided to `Project.find`.

    This seems like a bug, but that fact that it hasn't been an issue yet lends support to the idea that we aren't ever using multiple projects per execution, and thus `Project` could be made into a true singleton rather than this hybrid approach. That said, I don't like making parameterized classes into singletons, and if `Project` were a singleton then it would be difficult to ever refactor Meltano core into a daemon, so I'd rather just have it not be a singleton at all, and instead have it re-instantiate each time as necessary. It shouldn't be particularly costly outside of the UI; every call to `Project.find` save for one is done for the UI.

    Since we're deprecating the UI anyway, I don't feel too bad about making its performance marginally worse on newer versions of Meltano. If we want to avoid that slight performance regression, we can do so by having the `Project` instance be cached within `meltano.api`, and have it be passed around or re-instantiated everywhere else. That would essentially quarantine the design flaw to the deprecated module, while letting us freely improve the rest. (Or we can use the context-based approach - see below.)

2. Make `Tracker` not parameterized by `Project`.

    The difficultly with this is that the `Tracker` needs access to settings like `send_anonymous_usage_stats`, for which it needs a `ProjectSettingsService`, for which it needs a project, but we can simply have it instantiate a new `Project` instance instead of being passed one.

    If it is not parameterized by `Project`, then we'd still need some way to get the `ProjectContext`. I propose we make `ProjectContext` work like `ExceptionContext` by being instantiated as-needed by `Tracker.contexts`. This could only be done efficiently if `Project` instances are immutable, and are cached by the caller. They would be cached by the caller if we adopt the context model I briefly proposed at the end of the engineering all-hands meeting on 2022-12-06.

    <details><summary>Aside</summary>There I was specifically talking about accessing elements from a Meltano manifest without ambiguity (decided by the caller), but the same pattern can be applied to the <code>Project</code> class. Presumably with these changes in place we could also make a <code>ProjectSettingsService</code> be an element of <code>Project</code>, and then eventually replace the <code>ProjectSettingsService</code> with simple reads from the manifest. That would require a no-environment manifest, which I haven't yet planned to do, but I suspect we'll want it eventually since some commands do not permit the use of Meltano environments at all. A similar approach could be used for <code>PluginSettingsService</code>.</details>

    In such a scenario, `meltano.cli` would make calls into `meltano.core` within a Python context block (i.e. using a `with` statement), and within that block a single `Project` instance would be used by everything. This lets us avoid making `Project` into a singleton, while also saving us from having to re-instantiate it ever within the `with` block. Obtaining the `Project` instance within one of these contexts would instead be done using a class method of `Project` which would have no parameters.

Of course we can and should add additional integration tests, such as a test which runs various commands with and without environments active, then checks Snowplow Micro to ensure that the `environment_name_hash` values collected are as expected. But getting coverage via integration tests for these sorts of things is tedious and time-consuming. Oftentimes important, but good testing is no substitute for good design.

I'll update this PR with a heavily parameterized Pytest test that uses the `snowplow` fixture to test `environment_name_hash` as described above, as well as any other similar elements I think should be tested like that if I find any.

aaronsteers:
@WillDaSilva - Thanks very much for the detailed writeup.

I have enough experience with telemetry systems in Amazon Digital Video to know there are a near-infinite number of edge cases for how telemetry can fail. I very much appreciate your honest evaluation of how far we can get with unit testing.

My own hunch is that the Manifest-based serialization and deserialization methods related to #6944 could open up new avenues for us in terms of how we construct the project.

Perhaps in the future, Meltano commands would follow this flow:

1. Meltano compiles the project, with an environment input or a `null` environment.
1. Meltano raises a compile error or a compile warning if the project cannot be compiled.
1. Meltano initializes a `Project` object from the compiled result, with one or more project environments attached.
1. Actions are then performed against the fully-resolved, fully-compiled `Project` object - reducing the number of edge cases.

In the first two steps, we have a `Tracker` and maybe a `Compiler` object, but no `Project`. By step 3, we should have `Tracker` and `Project`.

Whether or not the above is the exactly correct or best flow, do you agree we should delay major refactoring until the manifest `compile` operation is available as a starting point?

If yes, then I think it becomes easier to make `Project` _not_ a singleton.

aaronsteers:
(My response part 2)

Could we / should we inject a set of Snowplow data tests into the already existing integration tests for `example-projects`? Challenge there is that there are still so many cases which could slip through, since there are a wide swath of 'valid but incorrect' telemetry outputs. Perhaps we could steal/borrow @edgarrmondragon's pytest-snapshot approach? In that case, we'd record and validate the telemetry for one or more example project, and then commit the result to the repo. Then if the the telemetry events change at all, someone would need to review and re-commit the updated snapshot.

cjohnhanson (approved):
Looks good (pending potential naming changes per @aaronsteers), thanks for taking the time to talk through it in Slack. 

:ship:

src/meltano/cli/utils.py
```diff
@@ -528,44 +530,141 @@ def check_dependencies_met(
...
+    else:
+        activate_environment(ctx, project)
+
+
+class CliEnvironmentAction(Enum):
+    """Enum of the different ways in which a Meltano environment can be activated."""
+
+    activate = auto()
+    activate_required = auto()
+    activate_explicitly_provided = auto()
```
aaronsteers, line 592: Per our conversation, some naming ideas. Longer, but doesn't require special training and speaks for itself.

```
activate_required            > environment_required
# use default_environment or explicit environment. otherwise, fail. (meltano run)

activate                     > environment_optional_use_default
# use default_environment or explicit environment. otherwise, no environment

activate_explicitly_provided > environment_optional_ignore_default
# ignore default_enviornment. only active environment if explicitly requested. otherwise, no environment.
```

cc @cjohnhanson

aaronsteers (commented):
👍 

I finished my line-by-line review. Just one suggestion, inline, about naming.

aaronsteers (approved):

src/meltano/cli/utils.py
WillDaSilva: Done in 7b779c4.

I also renamed `CliEnvironmentAction` to `CliEnvironmentBehavior`, along with other related `action` -> `behavior` changes, to reflect the updated names.

PR #7069, opened 2022-12-10
refactor: Use a custom build backend

### LINKED ISSUE #7068: Run custom build code when creating wheel
Context: We currently have a `build.py` file which is not used. If we instruct Poetry to use it, it will be included in the sdist, and build the webapp (among other things) into the wheel when the wheel is either built from source directly or from the sdist.

The problem with `build.py` is that when Poetry uses `build.py` it builds a platform specific wheel, rather than a pure wheel. We currently build a pure wheel, and want to continue to do so.

The crux of this issue is to find some way to run custom code when a wheel is being constructed (either directly from source or from an sdist). This code should at least build the UI webapp.

Resolving this issue will unblock:
- #6876 
- #6566
- #6488

---

~~Created this draft issue to track work with `poetry` team on how to get the build.py process in a good place. Today this is not compatible with a generic 'pure' wheel that is not platform-specific.~~

Converted to a normal issue in light of this becoming a likely blocker for:
- #6876 

We need to be able to go beyond what Poetry's build-system allows us to do. We currently do so with a mix of a mostly-dead-code `Makefile`, and a build workflow in `.github/`.

After spending a few hours looking into this again today, I've reaffirmed my belief that this won't be possible using `build.py`. We don't have any practical way to hook into Poetry's decision to build a plat-wheel, since it runs `build.py` (and Setuptools) in a subprocess. We could (and in the past I have) use the extreme flexibility of Setuptools to build the wheel, and then kill the parent process (Poetry), but this is a terribly hacky approach that I didn't open a PR for since I was not confident it in.

That said, I do believe we can achieve the desired effect (running arbitrary code when the wheel is built) by using a custom build backend that is a wrapper around `poetry-core`. Poetry won't use the custom build backend when using `poetry build`, but `pip wheel . --no-deps` will.

WillDaSilva:
See #7068 for context.

- This works properly with editable installs through Poetry or pip, though it'll only include the webapp and schemas when build using pip or some other Python packaging frontend that supports PEP 517 build backends.
- This works properly with non-editable installs through pip. Note that `poetry install` always produces an editable install.
- Source distributions (sdists) produced by Poetry are as they should be: they include everything necessary to build the wheel. Note that pip does not create sdists.
- Wheels produced by pip (either from source or using the sdist created by Poetry) have the webapp and schemas built into them.
- Wheels produced by Poetry are as they were before these changes were implemented. They lack the webapp and schemas unless they were added to the source tree prior to building the wheel.

Closes #7068


Makefile
```diff
@@ -116,16 +116,19 @@ freeze_db: ## Freeze the Meltano database to support DB migration during upgrade
 
 # sdist:
 # Build the source distribution
-# Note: plese use `sdist-public` for the actual release build
+# Note: please use `sdist-public` for the actual release build.
+# Note: despite being called "sdist", this builds both the sdist and wheel.
 sdist: freeze_db bundle ## Build the Meltano sdist for development
```
WillDaSilva, line 121: Can `freeze_db` and `bundle` be removed?

edgarrmondragon: `freeze_db` updates `db.lock`, which is included in the wheel so maybe that has to run as part of the build too?

I don't know if that's only relevant for `meltano upgrade`, though.

WillDaSilva: I've added a step to the GitHub workflow that previously called `sdist_public` to freeze the DB. With that done, it no longer has any reason to use the Makefile, so I replaced the call to `make sdist_public`. Now nothing in the repository relies on the Makefile.

edgarrmondragon (approved):

cjohnhanson (approved):
:+1:

PR #7090, opened 2022-12-14
docs: fix config name for poetry

adherr:
Running the command as specified yields
```text
Setting virtualenv.in-project does not exist
```

Also, some text that looks like tags were being swallowed on the PR doc page
image


WillDaSilva (approved):
Thanks @adherr!

PR #7094, opened 2022-12-15
feat: Allow annotations in Meltano project files

### LINKED ISSUE #7092: Allow vendor-specific annotations in `meltano.yml`
Discussed in https://github.com/meltano/meltano/discussions/7053

We should update `meltano.schema.json` to permit an object with arbitrary fields (i.e. `"additionalProperties": true`) within any object in the schema that does not already impose restrictions on the permitted additional properties. That is to say:
- If the object has `"additionalProperties": true`, then it can have `annotations`, and no changes are required to the schema.
- If the object has `"additionalProperties": false`, then it can have `annotations` and we must explicitly add it to the properties.
- If the object has `additionalProperties` set to anything else (e.g. limiting it to strings as `env` does), then annotations are not supported, and no change to the schema is necessary.

So the only places in the schema that need to be updated are where it says `"additionalProperties": false`.

Documentation around what this field is for should be added. It should be made clear that the core Meltano library/CLI does not use this field, but it may be used for Meltano Cloud, or third-party tools.

The documentation should also state that we impose no limitation on how it can be used except for requiring that the top-level of each annotation object may only contain objects as properties, and that we recommend that third-party tools keep their annotations under a sensible key, such as the name of the tool.

Testing should be done by adding annotations liberally to integration tests, and ensuring that they are ignored.

WillDaSilva:
Closes #7092

### DIFF
--- a/docs/example-library/meltano-annotations/ending-meltano.yml
+++ b/docs/example-library/meltano-annotations/ending-meltano.yml
@@ -1,0 +1,62 @@
+version: 1
+default_environment: dev
+project_id: abd310b5-698b-4244-a109-9490047fa2ec
+send_anonymous_usage_stats: false
+annotations:
+  example-annotation:
+    a_key: a value
+  another-annotation-category:
+    best_season: Autumn
+    the_purpose_of_this_annotation: Nothing much
+plugins:
+  extractors:
+  - name: tap-gitlab
+    variant: meltanolabs
+    pip_url: git+https://github.com/MeltanoLabs/tap-gitlab.git
+    annotations:
+      example-annotation-3: {}
+    config:
+      annotations:
+        example-annotation-4: {}
+      projects: meltano/meltano
+      start_date: '2022-11-01T00:00:01Z'
+    select:
+    - commits.*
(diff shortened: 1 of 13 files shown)


WillDaSilva:
@meltano/engineering This PR currently implements support for annotations essentially anywhere within Meltano project files. It works, but as discussed in the codesharing call on 2022-12-15, it's a mess, and we aren't aware of any straightforward ways of improving it that would be worth the time commitment. Additionally, it's unclear how the Manifest should handle them when it hoists stuff from an environment up to the top level.

Thoughts on reducing the scope down to permitting an annotations field at the top-level of `meltano.yml` only, rather than allowing annotations anywhere? Later on it'll be easy enough to add support for annotations elsewhere (i.e. by doing what this PR currently does), but removing support for it somewhere is a breaking change, so I'm inclined to only permit them everywhere now if we have a good way to implement that.

cjohnhanson (approved):
As discussed in codeshare: lg2me

WillDaSilva:
Superseded by:
- #7115

PR #7097, opened 2022-12-16
feat: Feature flag to disable calls to Hub and `discovery.yml` outside of `add` and `discover`

### LINKED ISSUE #7095: Feature flag to disallow Meltano calling Hub outside of `add` and `discover`
### Meltano Version

2.11.1

### Python Version

NA

### Bug scope

### Operating System

NA

### Description

With the recent switch to API Gateway and Netlify redirects, users that had whitelisted the now legacy Hub API (`https://hub.meltano.com/meltano/api/v1/`) are seeing failures because of the redirect to the Gateway endpoint.

See https://meltano.slack.com/archives/C01TCRBBJD7/p1671112207278089.

---

Back when I implemented `meltano add` from Hub, I decided to allow calls outside of `add` and `discover`, so users not yet onboarded to lockfiles would still be able to use their plugins without having to add those to a custom `discovery.yml`.

This can be disabled (as a breaking change) in our next major release (3.0).

### Code

_No response_

edgarrmondragon:
Closes #7095 

- [x] Tests
- [x] Docs


edgarrmondragon:
> @edgarrmondragon - I think we have to hold this until 3.0, if it would break existing users (and there are a lot) who have not yet locked their plugin files. Can you confirm?

@aaronsteers yes, invoking a plugin that isn't locked would result in a failure with this approach

edgarrmondragon:
@aaronsteers Perhaps this can be introduced earlier with a feature flag?

aaronsteers:
> @aaronsteers Perhaps this can be introduced earlier with a feature flag?

Great idea!

Maybe something like `require_plugin_lock_files` or `plugin_locks_required`?

I also considered a generic for:

- https://github.com/meltano/meltano/issues/6754

As someone noted in Slack, a possible mechanism for this could be forcing `hub_api_root` and `discovery_url` both to be `false`.

- https://docs.meltano.com/reference/settings#hub_api_root
- https://docs.meltano.com/reference/settings#discovery_url

That generic solve to restrict network connectivity would break new plugin additions though, so I ruled it out as a good solve here. The ideal state in 3.0, I think, is that you wouldn't be able to run plugins unless they had a local definition file. Making an opt-in flag now for this requirement is nice, because we can start recommending this flag as a best practice, and when we get close to 3.0 we can recommend it as a pre-upgrade step.

WillDaSilva:
> Perhaps this can be introduced earlier behind a feature flag?

Sounds good to me.

> because we can start recommending this flag as a best practice, and when we get close to 3.0 we can recommend it as a pre-upgrade step.

@edgarrmondragon @aaronsteers We could also make not having it enabled a lint issue:
- #7285

aaronsteers:
Not a high priority per se, but I've updated the PR description to reflect the feature flag-based scope discussed above, which enables this to ship before 3.0 without breaking existing users.

docs/src/_reference/settings.md
```diff
@@ -597,3 +597,10 @@ Snowplow collector endpoints to be used if the [`send_anonymous_usage_stats` set
 - Default: `False`
 
 Causes an exception to be raised if an environment variable is used within the project's Meltano configuration but that environment variable is not set.
+
+### <a name="ff-plugin-locks-required"></a>`ff.plugin_locks_required`
+
+- [Environment variable](/guide/configuration#configuring-settings): `MELTANO_FF_PLUGIN_LOCKS_REQUIRED`
+- Default: `False`
+
+When this flag is enabled, plugins will not be lookup up outside of `meltano add` and `meltano discover` operations. So, for example, calling `meltano run` will fail if a [lock file](/concepts/plugins#lock-artifacts) is not present for one of the plugins.
```
tayloramurphy, line 606:
```suggestion
When this flag is enabled, plugins will only use [lock files](/concepts/plugins#lock-artifacts) to determine the settings, installation source, etc with the exception of the `meltano add` and `meltano discover` operations. This means that calling `meltano run` will fail if a lock file is not present for one of the plugins.
```

tayloramurphy (approved):
Minor docs change, otherwise approving.

src/meltano/core/project_plugins_service.py
```diff
@@ -32,14 +34,18 @@
...
-    INHERITED = "inherited"
+    NONE = 0
+    DISCOVERY = enum.auto()
+    HUB = enum.auto()
+    CUSTOM = enum.auto()
+    LOCKFILE = enum.auto()
+    INHERITED = enum.auto()
+
+    ANY = DISCOVERY | HUB | CUSTOM | LOCKFILE | INHERITED
+    LOCAL = ~HUB
```
WillDaSilva, line 48: Nice!

src/meltano/core/bundle/settings.yml
```diff
@@ -231,5 +231,9 @@ settings:
 - name: ff.strict_env_var_mode
   kind: boolean
   value: False
+# Whether invoke, and run operations if a plugin is not custom or it is not locked.
+- name: ff.plugin_locks_required
+  kind: boolean
+  value: false
```
WillDaSilva, line 237: In addition to adding the feature flag here, it should be added to the `meltano.yml` schema at `src/meltano/schemas/meltano.schema.json`.

WillDaSilva (commented):
This is great! I have one request, but aside from that this looks good to go

src/meltano/core/bundle/settings.yml
edgarrmondragon: Done! 772de9d

WillDaSilva (approved):

PR #7099, opened 2022-12-16
chore: Use Flakeheaven

### LINKED ISSUE #7039: Use `flakeheaven` as a wrapper around `flake8`
https://wemake-python-styleguide.readthedocs.io/en/latest/pages/usage/integrations/flakeheaven.html

Flakeheaven allows us to establish a baseline when running Flake8 on the entirety of the Meltano repository. On subsequent runs it will only fail if we break a Flake8 rule that wasn't previously broken.

Currently we do something akin to this by only running `pre-commit` on changes files. This prevents us from using https://pre-commit.ci, since that is only able to run on the entire repository.

By using Flakeheaven (and its pre-commit hook) around Flake8 we can:
- Use https://pre-commit.ci instead of our custom `pre-commit` GitHub workflow.
- Progressively address Python lint issues in a granular way. We should no longer run into the problem of having to fix every lint issue in a (large) file just because we made a small change in it.

The Flakeheaven baseline file will need to be committed into the repository. I'm not yet sure how we/CI should update it.

WillDaSilva:
Closes #7039


edgarrmondragon (approved):

edgarrmondragon:
@WillDaSilva :shipit: It'll be nice to have this

PR #7104, opened 2022-12-16
fix: Support default boto3 auth for S3 state backend when AWS creds are unconfigured

### LINKED ISSUE #7103: bug: S3StateStoreManager cannot delete state when state_backend.s3.aws_* settings are not present
### Meltano Version

2.11.1

### Python Version

3.10

### Bug scope

API

### Operating System

meltano/meltano:v2.11.1 image in ECS fargate

### Description

When no `aws_secret_access_key` or `aws_access_key_id` settings are present in a Meltano project, the `S3StateStoreManager.client` attribute does not get initialized. Currently this is a rare circumstance, since the relevant `AWS_*` environment variables are aliased to the Meltano settings. However, there are some circumstances (e.g. running in ECS) where we should still fallback to boto3's internal logic for initiating a client (we already do this for reading and writing objects, but listing and deleting objects require a standalone client outside of the read/write logic). Once https://github.com/meltano/meltano/pull/7078 is released though, this will potentially cause more issues for end users that are directly using the `AWS_*` environment variables for S3 state backend configuration.

### Code

_No response_

cjohnhanson:
Closes #7103


src/meltano/core/state_store/s3.py
```diff
@@ -84,6 +84,9 @@ def client(self):
                 raise InvalidStateBackendConfigurationException(
                     "AWS access key ID configured, but no AWS secret access key."
                 )
+            else:
+                session = Session()
```
WillDaSilva, line 88: `boto3.Session` is not imported at this point. Why is the import happening within the function?

WillDaSilva (changes requested):
In addition to the in-line issue I commented on, I think it would also be nice if you could update the `client` property to be a `cached_property` instead of a `property`, and then return the client object instead of setting it to `self._client`.

src/meltano/core/state_store/s3.py
cjohnhanson: Ah, good catch. 

We want to lazily import boto3, since it's an optional dependency that should only be needed when `state_backend.uri` is set to a URI with `s3://` scheme. Importing within the method is obviously fragile (especially since this `else` branch is difficult to fully test for every potential method of providing credentials); is there a better pattern for this?

cjohnhanson:
@WillDaSilva agreed that this approach for "caching" the property is a bit inelegant compared to `@functools.cached_property` and may also have performance implications that I'm unaware of-- is there a clean 3.7-compatible approach you had in mind for using `cached_property` here?  Unless I'm mistaken,`@functools.cached_property` was new as of 3.8 and unfortunately we're still supporting 3.7 (for at least another ~6 months).

WillDaSilva:
> is there a clean 3.7-compatible approach you had in mind for using `cached_property` here?

@cjohnhanson Yes, apologies for not being explicit about it in my previous comment. We currently have a dependency on the `cached_property` package, so you can use `from cached_property import cached_property`. It's used in a few spots throughout the codebase, and works like `functools.cached_property`.

edgarrmondragon (approved):

WillDaSilva (approved):

PR #6954, opened 2022-11-02
feat: Support `meltano init` in a non-empty directory

ReubenFrankel:
Addresses discussion in #6927.

### Requirements
- Check that all files that would be created by `meltano init` do not conflict with any existing files
- Check any existing directories that would be created by `meltano init` have write permissions
- If any checks fail, error with a helpful message before creating any project files

### Checklist

- [x] Check for conflicting files
- [x] Check for existing directories without write permissions
- [x] Add/update tests
- [ ] Add/update docs

PR #7105, opened 2022-12-16
feat: Support `meltano init` in a non-empty dir

### LINKED ISSUE #7077: Feature request: Support `meltano init` in non-empty directory

edgarrmondragon:
- Fails with a clear message if `meltano.yml` exists in the target directory
- With `--force`, overwrites `meltano.yml` and skips any other files that would be created by `meltano init`.

Closes #7077


tayloramurphy (approved):
This looks great! 

FYI @sbalnojan once this is merged we get the codespace demo updated. It changes the post-install instructions to exclude `cd <directory>` which was a point of confusion in the current iteration.

WillDaSilva (approved):

ReubenFrankel:
Does this invalidate https://github.com/meltano/meltano/pull/6954?

edgarrmondragon:
> Does this invalidate #6954?

@ReubenFrankel I think so. This implementation was a bit different and with a more limited scope, but I feel like the main problem is fixed now.

PR #7121, opened 2022-12-20
chore: Use Poetry v1.3 lockfile

WillDaSilva:


edgarrmondragon:
@WillDaSilva we need to downgrade virtualenv or change its constraint to `<=`

WillDaSilva:
Thanks for the pointer @edgarrmondragon. Why do we even have `virtualenv` pinned?

edgarrmondragon:
> Thanks for the pointer @edgarrmondragon. Why do we even have `virtualenv` pinned?

@WillDaSilva Nox uses it under the hood to create session venvs. It's probably safe to remove it from the constraints, though.

edgarrmondragon (approved):

Issue #6876, opened 2022-10-12
Feature proposal: meltano-manifest.json


aaronsteers:
Related to:

- https://github.com/meltano/meltano/discussions/6944
- https://github.com/meltano/meltano/discussions/6766

As discussed in office hours on 2022-10-12.

Feature goals/requirements:

1. Stability. Since this would serve as the base for many interop/integration scenarios, we want to make the format as stable as possible.
2. Self-contained. There should be no need to reference external files or services when using this format. So, for example, lock file content would be embedded so no external references are needed.
3. Pre-calculated inheritance, precedence, and override effects. The reader shouldn't need to know how inheritance works, or understand any other internal Meltano business logic.
4. No secrets. This artifact essentially needs to be treated as 'code' and may be passed around in less-trusted contexts. As such, it should _not_ contain: env vars from local OS or terminal context, env vars from `.dotenv`, settings values from `systemdb` or other external settings/secrets providers.
5. Simple as possible. Bonus points if the new json manifest is able to be validated against existing JSON Schema rulesets, without creating or maintaining net new data structures.

So, given these guidelines, here's a possible path forward:

1. Start with contents and structure of the `meltano.yml` file itself.
2. Merge in all data from `include_paths` declarations.
3. Super-populate each plugin definition in the global context:
   1. Inject all properties from the lock file, except of course, those properties overridden.
   2. Calculate the value of each entry of the plugin's `config`.
   3. Optionally, we can store 'extra' info about how the evaluation was performed under a `meta` or `vendor` key that is explicitly non-stable or at least explicitly free-form.
   4. Environment variable declarations in plugin config should be left unresolved until runtime. If we are confident that the value can be resolved without leaking any sensitive info, the predicted evaluation could optionally be rendered under a `meta` or `vendor` key, without losing fidelity of the config's env var reference.
5. Repeat the above process for each declared Meltano Environment.
6. Inject other top-level entities (`jobs`, `env`, `schedules`, etc.) into the environment declarations if applicable, so that each environment definition is standalone. 

Variations:

1. We optionally could give an environment name to the 'global' or `--no-environment` behavior, so that the top-level file is just the environments declaration: `version: 1 \n environments: [...] \n <EOF>`. This reduces the clutter in the file, and readers would only deserialize the environment definition they need in the given context.
2. We could optionally create separate manifest files per environment so that the manifest is specific to what is needed exactly in a given context.

pnadolny13:
A use case we discussed in office hours:

[...]

PR #7123, opened 2022-12-20
feat(core): Optionally ignore missing env vars during expansion

### LINKED ISSUE #7122: Optionally ignore missing env vars during expansion
Currently, if an env var is not set in the provided environment dictionary, the `meltano.core.utils.expand_env_vars` function will replace it with the empty string by default. If `raise_if_missing` is `True` then it will raise an exception instead.

I would like it to be possible to optionally have it leave the missing environment  variable alone, rather than replacing it with the empty string. This could be controlled with a boolean parameter `ignore_if_missing`, which would default to `False` to preserve the current default behaviour, and would be mutually exclusive with `raise_if_missing`.

Relates to:
- #6876

WillDaSilva:
In addition to addressing #7122, this also includes some performance improvements that have a significant effect when using `expand_env_vars` on many large env dictionaries, as will be done by the implementation for [Meltano manifest](https://github.com/meltano/meltano/issues/6876).

Some work to wrangle the complexity of `expand_env_vars` was made, but it's widely used by legacy code, which makes it difficult to improve significantly.

For instance, `expand_env_var` could previous be called on any type, and it would return it. This is now disallowed by the type hint, but still supported in the code itself. Hopefully this will discourage these bad practices.

Another questionable aspect of `expand_env_vars` that had to be preserved is that it will perform a recursive tree-traversal on any mapping structure it is provided as `raw_value`. This dangerous/slow/complex behaviour can now be opted out of by setting the `flat` paramter to `True`, which asserts that the `raw_value` being provided has a flat structure (i.e. no nested dicts).

Closes #7122

### DIFF
--- a/src/meltano/core/utils/__init__.py
+++ b/src/meltano/core/utils/__init__.py
@@ -15,7 +15,7 @@ from collections import OrderedDict
 from copy import deepcopy
 from datetime import date, datetime, time
 from pathlib import Path
-from typing import Any, Callable, Iterable, TypeVar, overload
+from typing import Any, Callable, Iterable, Mapping, Sequence, TypeVar, overload
 
 import flatten_dict
 from requests.auth import HTTPBasicAuth
@@ -23,7 +23,6 @@ from requests.auth import HTTPBasicAuth
 from meltano.core.error import MeltanoError
 
 logger = logging.getLogger(__name__)
-T = TypeVar("T")
 
 TRUTHY = ("true", "1", "yes", "on")
 REGEX_EMAIL = r"(^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$)"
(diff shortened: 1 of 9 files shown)


src/meltano/core/utils/__init__.py
```diff
@@ -478,42 +477,111 @@ def __init__(self, env_var: str):
...
+    *,
     raise_if_missing: bool = False,
-):
-    if isinstance(raw_value, dict):
-        return {
-            key: expand_env_vars(val, env, raise_if_missing)
-            for key, val in raw_value.items()
-        }
-    elif not isinstance(raw_value, str):
+    ignore_if_missing: bool = False,
```
edgarrmondragon, line 488: wdyt about refactoring these args to use a single flag instead:

1. EmptyStr
2. Raise
3. Ignore

WillDaSilva: I had thought about doing that, but didn't want to break every spot that uses the pattern:

```python
with project_settings_service.feature_flag(
    FeatureFlags.STRICT_ENV_VAR_MODE, raise_error=False
) as strict_env_var_mode:
    expand_env_vars(..., raise_if_missing=strict_env_var_mode)
```

`ProjectSettingsService.feature_flag` returns a bool which is passed in directly. We could map that bool to the appropriate enum value at every call site, but that's a bit messy.

One hacky approach we could take is to make it an int enum, have 0 correspond to the default behaviour, and have 1 correspond to raising if missing. That way a bool passed in would work without any changes. What do you think about that @edgarrmondragon?

edgarrmondragon: @WillDaSilva Yeah good point, quick search reveals 16 uses of `raise_if_missing=...` 😬 . I'm ok with the hack, with a note to refactor later perhaps?

WillDaSilva: @edgarrmondragon Implemented in cf2333d95a0c51a17173ca9b9fd880903ffd59fa. I'm not sure if it's an improvement or not. If you have strong feelings one way or the other we can keep or revert the changes.

edgarrmondragon: @WillDaSilva I think it's a good compromise to avoid having mutually exclusive params 👍

edgarrmondragon (approved):

PR #7124, opened 2022-12-21
chore: Enable type checking in `meltano.core.schedule`

edgarrmondragon:


src/meltano/core/schedule.py
```diff
@@ -86,8 +86,10 @@ def elt_schedule(self) -> bool:
         return not self.job
 
     @property
-    def elt_args(self) -> list[str]:
-        """Return the list of arguments to pass to the elt command, if the schedule is an elt schedule.
+    def elt_args(self) -> list[str | None]:
+        """Return the list of arguments to pass to the elt command.
+
+        Only valid f the schedule is an elt schedule.
```
WillDaSilva, line 92:
```suggestion
        Only valid if the schedule is an elt schedule.
```

src/meltano/core/schedule.py
```diff
@@ -8,7 +8,7 @@
 from meltano.core.job import Job as StateJob
 from meltano.core.job import JobFinder as StateJobFinder
 
-CRON_INTERVALS = {
+CRON_INTERVALS: dict[str | None, str | None] = {
```
WillDaSilva, line 11:
```suggestion
CRON_INTERVALS: dict[str, str | None] = {
```

WillDaSilva (approved):

PR #7130, opened 2022-12-21
chore: Enable type checking in `meltano.core.state_service`

edgarrmondragon:


src/meltano/core/state_service.py
```diff
@@ -110,8 +110,8 @@ def validate_state(state: dict[str, Any]):
     def add_state(
         self,
         job: Job | str,
-        new_state: str | None,
-        payload_flags: Payload = Payload.STATE,
```
WillDaSilva, line 114: Does this not work? It's more clear like this, since it specifies the enum which defines the meaning of the int value meaning.

edgarrmondragon: @WillDaSilva Yeah, `Payload` here is a subclass of this

https://github.com/meltano/meltano/blob/1fec4b6304e10949849769ff849bbfdf2d62be0d/src/meltano/core/sqlalchemy.py#L32-L38

and not `enum.IntFlag`, so its members are integers instead of instances of `Payload`.

WillDaSilva: That's mildly confusing, but at least the type info is accurate now.

WillDaSilva (approved):

PR #7145, opened 2023-01-04
fix: Setting nested setting key no longer clears sibling keys

### LINKED ISSUE #7132: bug: nested configs are cleared when using space as delimiter and when using `--interactive`
### Update (@aaronsteers)

Per the issue thread below, we've identified that the issue is present in the behavior of the CLI syntax with space as delimiter but not when dot is used. The proposal is to make these two function identically:

```bash
meltano config testy-mctest set file_configs.format json       # the expected behavior (does not reset sibling config)
meltano config testy-mctest set file_configs format json       # bug: clears sibling nested config
```

---------

### Original Issue description (@pnadolny13)

When using `--interactive` to update nested settings it overwrites the entry config set. For example:

```yaml
  - name: tap-file
    config:
      file_configs:
        dataset_name: test_file
        format: csv
        path: /my/path/
```

Then I run `meltano config tap-file set --interactive`, navigating to the setting `file_configs.format` (in dot notation) as defined in the plugin definition lock file, updating to `json`, meltano overwrites the whole nested config with only the new value:

```yaml
  - name: tap-file
    config:
      file_configs:
        format: json
```

Slack thread with video showing bug https://meltano.slack.com/archives/C03GKHWS0HM/p1671746236351709

meltano, version 2.12.0

kgpayne:
Closes #7132


src/meltano/cli/config.py
```diff
@@ -350,6 +350,9 @@ def reset(ctx, store):
 @_use_meltano_env
 def set_(ctx, setting_name, value, store, interactive):
```
WillDaSilva, line 351: Adding type hints for this function would help clarify what's going on here.

tayloramurphy (approved):
Confirmed locally this seems to fix the bug!

src/meltano/cli/config.py
kgpayne: Agreed! Done ✅

kgpayne:
@edgarrmondragon I think this might have been a side-effect of the ruamel onboarding 🤔 The root cause was a type comparison between `dict` and `ordereddict` that was [failing](https://github.com/meltano/meltano/blob/main/src/meltano/core/utils/__init__.py#L218), triggering an overwrite of the existing values. To resolve I loosened up the check by evaluating if either comparator is a subclass of the other, in addition to exact matches. WDYT?

src/meltano/cli/config.py
```diff
@@ -348,8 +348,17 @@ def reset(ctx, store):
 )
 @click.pass_context
 @_use_meltano_env
-def set_(ctx, setting_name, value, store, interactive):
+def set_(
+    ctx: click.core.Context,
+    setting_name: tuple,
```
WillDaSilva, line 353:
```suggestion
    setting_name: tuple[str, ...],
```

edgarrmondragon:
> @edgarrmondragon I think this might have been a side-effect of the ruamel onboarding 🤔 The root cause was a type comparison between `dict` and `ordereddict` that was [failing](https://github.com/meltano/meltano/blob/main/src/meltano/core/utils/__init__.py?rgh-link-date=2023-01-04T16%3A06%3A34Z#L218), triggering an overwrite of the existing values. To resolve I loosened up the check by evaluating if either comparator is a subclass of the other, in addition to exact matches. WDYT?

@kgpayne yeah that sounds ok. Comments seem to be preserved well within the `config` dict anyway. This is great work @kgpayne!

edgarrmondragon (approved):

src/meltano/core/utils/__init__.py
```diff
@@ -170,6 +170,14 @@ def merge(src, dest):
     return dest
 
 
+def are_similar_types(left, right):
+    return (
+        issubclass(left.__class__, right.__class__)
+        or issubclass(right.__class__, left.__class__)
+        or type(left) == type(right)  # noqa: WPS516
+    )
```
WillDaSilva, line 178: Types should generally be compared using `is`, but in this case the type comparison isn't necessary since `issubclass(class_x, class_x)` is `True`.

That said, we can use `isinstance` plus `type` here to avoid accessing the dunder attribute `__class__`.

```suggestion
def are_similar_types(left, right):
    return isinstance(left, type(right)) or isinstance(right, type(left))
```

kgpayne: TIL`isinstance` caters for inheritance, while checking for equality of `type` does not 😅 Thanks @WillDaSilva !

WillDaSilva (approved):

PR #7174, opened 2023-01-16
fix: Support default auth for GCS and Azure state backends

cjohnhanson:
Basically the same as https://github.com/meltano/meltano/pull/7104 but for GCS and Azure.


src/meltano/core/state_store/azure.py
```diff
@@ -2,9 +2,42 @@
 from __future__ import annotations
 
 import re
+from contextlib import contextmanager
+
+from cached_property import cached_property
```
WillDaSilva, line 7: Using `functools.cached_property` when it is available instead of the backport helps mypy determine the type of the property.

```suggestion
import re
import sys
from contextlib import contextmanager

if sys.version_info >= (3, 8):
    from functools import cached_property
else:
    from cached_property import cached_property
```

edgarrmondragon (approved):

src/meltano/core/state_store/azure.py
```diff
@@ -2,9 +2,42 @@
 from __future__ import annotations
 
 import re
+from contextlib import contextmanager
+
+from cached_property import cached_property  # type: ignore
```
WillDaSilva, line 7: @cjohnhanson I'm not sure why my [previous suggestion](https://github.com/meltano/meltano/pull/7174#discussion_r1071635748) was ignored. I'll re-raise it here:

```suggestion
import re
import sys
from contextlib import contextmanager

if sys.version_info >= (3, 8):
    from functools import cached_property
else:
    from cached_property import cached_property
```

This seems like a better approach than using `# type: ignore`.

src/meltano/core/state_store/google.py
```diff
@@ -2,9 +2,42 @@
 from __future__ import annotations
 
 import re
+from contextlib import contextmanager
+
+from cached_property import cached_property  # type: ignore
```
WillDaSilva, line 7: Same as my other suggestion above.

```suggestion
import re
import sys
from contextlib import contextmanager

if sys.version_info >= (3, 8):
    from functools import cached_property
else:
    from cached_property import cached_property
```

src/meltano/core/state_store/azure.py
cjohnhanson: Yeah good callout, I hit `Commit suggestion` but didn't follow up on it to make sure after resolving, must not have gone through. 

I also changed the import in `s3.py` to this pattern for good measure.

PR #7179, opened 2023-01-17
chore: Collect select set env vars in telemetry

### LINKED ISSUE #7175: Add is_codespaces to telemetry environment context
Just like [we set `is_ci_environment` when the `CI` env var is set](https://github.com/meltano/meltano/blob/main/src/meltano/core/tracking/contexts/environment.py#L57), we should set `is_codespaces` (or something to that effect) when `CODESPACES` is set (see [docs](https://docs.github.com/en/codespaces/developing-in-codespaces/default-environment-variables-for-your-codespace)).

@tayloramurphy It'd be interesting to compare how far people get into the funnel with codespaces vs having to install locally. On the one hand, the barrier is lower so some people that click the button may be less motivated to make it to the end, but on the other hand, it should be easier to just quickly follow the steps and get to "wow". We may run into the issue that we currently consider any usage of less than 5min a bot, and that these codespaces projects may be treated as one-offs instead of being reused to form the company's official Meltano projects, so they'll never turn active. It'll be good to have the option of treating new codespaces projects differently from new local projects in our reporting.

WillDaSilva:
Diff between `environment_context` version `1-1-0` and `1-2-0`:

```diff
*** src/meltano/core/tracking/iglu-client-embedded/schemas/com.meltano/environment_context/jsonschema/1-1-0     2023-01-16 15:30:55.005363075 -0500
--- src/meltano/core/tracking/iglu-client-embedded/schemas/com.meltano/environment_context/jsonschema/1-2-0     2023-01-17 21:57:07.315052616 -0500
***************
*** 6,10 ****
          "name": "environment_context",
          "format": "jsonschema",
!         "version": "1-1-0"
      },
      "type": "object",
--- 6,10 ----
          "name": "environment_context",
          "format": "jsonschema",
!         "version": "1-2-0"
      },
      "type": "object",
***************
*** 40,43 ****
--- 40,57 ----
              "type": "boolean"
          },
+         "notable_flag_env_vars": {
+             "description": "Select environment variable flags",
+             "type": "object",
+             "propertyNames": {
+                 "type": "string"
+             },
+             "additionalProperties": {
+                 "description": "Boolean values of select env vars, or null if the truth value of the env var could not be determined",
+                 "type": [
+                     "boolean",
+                     "null"
+                 ]
+             }
+         },
          "python_version": {
              "description": "The version of Python in use",
***************
*** 163,166 ****
--- 177,181 ----
          "is_dev_build",
          "is_ci_environment",
+         "notable_flag_env_vars",
          "python_version",
          "python_implementation",

```

This will let us see if Meltano is being run in a codespace, among other things. The `is_ci_environment` field has been left in for backwards compatibility, but if you'd prefer @pnadolny13 we can remove it in favor of this new more general approach, and bump the version to `2-0-0` instead.

TODO:
- [ ] Publish the new `environment_context` schema to SnowcatCloud.

Closes #7175


src/meltano/core/tracking/contexts/environment.py
```diff
@@ -44,9 +44,19 @@ def _get_parent_context_uuid_str() -> str | None:
...
+    ci_markers = {"GITHUB_ACTIONS", "CI"}
+    notable_flag_env_vars = {"CODESPACES", *ci_markers}
+
+    @classmethod
+    def _true_notable_flag_env_vars(cls) -> Iterable[str]:
+        for env_var_name in cls.notable_flag_env_vars:
+            env_var_value = os.environ.get(env_var_name, "0")
+            with suppress(ValueError):
+                if strtobool(env_var_value):
+                    yield env_var_name, env_var_value
```
edgarrmondragon, line 56: So we're only checking _if_ the value is truthy, and then only send the raw string value to snowplow?

WillDaSilva: f520ad8 makes it so that all *set* env vars listed are collected. Their values can be `true`, `false`, or `null`, depending on if they're truthy, falsey, or something else respectively.

edgarrmondragon (approved):

pnadolny13 (approved):
@WillDaSilva Looks good to me! I agree that I'd prefer to keep is_ci_environment around, at least for now. We can definitely have a 2.0.0 schema change down the line to remove it though.

Issue #7168, opened 2023-01-16
bug: Meltano schema forbids `env` field in `environment[].plugins[]`


WillDaSilva:
### Meltano Version

2.13.0

### Python Version

3.10

### Bug scope

Configuration (settings parsing, validation, etc.)

### Operating System

N/A

### Description

[As documented here](https://docs.meltano.com/guide/configuration#specifying-environment-variables), Meltano supports using a plugin-specific environment-specific `env` field:

```yaml
environments:
- name: dev
  env:
    # environment level env
    MY_ENV_VAR: environment_level_env_var
  config:
    plugins:
      extractors:
        - name: tap-google-analytics
          variant: meltano
          env:
            # environment level plugin env
            MY_ENV_VAR: environment_level_plugin_env_var
```

This is not currently permitted by `schemas/meltano.schema.json`.

Relates to:
- #7148, because schema validate is performed during manifest generation

tayloramurphy:
@WillDaSilva thanks for opening and working on this!

WillDaSilva:
This doesn't seem to be a problem with jsonschema draft 7, which is what the Meltano schema uses. Unfortunately the jsonschema validator used by VS Code seems to ignore that, and checks the schema using the latest validator.

We may want to bump the Meltano schema to jsonschema draft 2020-12 (the latest version as of now), and use [`unevaluatedProperties`](https://json-schema.org/understanding-json-schema/reference/object.html#unevaluated-properties) to greatly simplify the schema. This would fix this issue as a side-effect, since we'd be able to remove `"additionalProperties": false` in each of the `environments[].plugins[]` objects.

PR #7180, opened 2023-01-17
chore: delete `meltano.core.utils.validate_json_schema`

WillDaSilva:
This appears to be misplaced. It only works when run from the Meltano repository, rather than from an installed copy of Meltano. It is a utility script for developers, rather than something that belongs under `src/meltano/`.

That would give reason to move it into `scripts/`, but instead I am deleting it because it does not appear to do anything more than the equivalent of running `check-jsonschema --schemafile schema/discovery.schema.json src/meltano/core/bundle/discovery.yml`.

Note that the `jsonschema` package/tool can be used instead of `check-jsonschema`, but recent versions of `jsonschema` recommend the use of `check-jsonschema` instead as `jsonschema`'s included CLI is deprecated.

Relates to #7168


edgarrmondragon (approved):

WillDaSilva:
I'll open a PR soon that adds a check for the `discovery.yml` file this checks to pre-commit, which currently checks the misc `meltano.yml` files within the repository.

PR #7188, opened 2023-01-18
feat: Add header with Meltano command to Hub request

### LINKED ISSUE #7185: feature: Add Meltano command information to Hub API request
### Feature scope

### Description

When Meltano sends requests to the Hub API, currently sets the `X-Project-ID` and `User-Agent` headers: https://github.com/meltano/meltano/blob/193a3ae4c7adfa29e102a65d31b7e865d9eea349/src/meltano/core/hub/client.py#L113-L125

However, it'd be good to include the CLI command where the request is coming from.

edgarrmondragon:
Closes #7185

### DIFF
--- a/src/meltano/core/hub/client.py
+++ b/src/meltano/core/hub/client.py
@@ -210,12 +210,7 @@ class MeltanoHubService(PluginRepository):  # noqa: WPS214
             The prepared request.
         """
         request = requests.Request(method, url)
-
-        try:
-            click_context = click.get_current_context(silent=True)
-        except RuntimeError:
-            logger.debug("Could not get current click context")
-            click_context = None
+        click_context = click.get_current_context(silent=True)
 
         if click_context:
             request.headers.update(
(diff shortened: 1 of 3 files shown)


src/meltano/core/hub/client.py
```diff
@@ -198,6 +199,48 @@ def plugin_endpoint(
...
+        try:
+            click_context = click.get_current_context(silent=True)
+        except RuntimeError:
+            logger.debug("Could not get current click context")
+            click_context = None
+
+        if click_context:
+            request.headers.update(
+                {"X-Meltano-Command": click_context.command_path},
+            )
```
WillDaSilva, line 223: This is probably the best way to do this currently, but it'll cause some pain if we ever decide to separate the CLI from the core.

edgarrmondragon: Yeah, I'm not a fan of calling `click` here but decoupling might require a not-so-small refactor to the `ProjectPluginsService` to make it aware of the request context.

edgarrmondragon: (or having a singleton Meltano context in which the CLI sets some attributes)

WillDaSilva (approved):

tayloramurphy (approved):

PR #7190, opened 2023-01-18
fix: Make fully missing env var settings `None`

### LINKED ISSUE #7189: bug: env vars in `meltano.yml` cause exception while running meltano install since v2.13.0
### Meltano Version

2.13.0

### Python Version

3.9

### Bug scope

### Operating System

Windows 10 (using WSL2), same issue occurs on Ubuntu 22.04 VM

### Description

[Slack thread](https://app.slack.com/client/TFG99TU9K/C01TCRBBJD7/thread/C01TCRBBJD7-1674066405.617239)

Using Meltano's Docker image I set up env vars in the `meltano.yml` file to utilize dynamically created database connections (using docker compose) cross-projects. This worked like a charm in `v2.12.0`, however since `v2.13.0` I'm receiving an exception during the meltano install step in the Dockerfile.

This exception occurs:
```
invalid literal for int() with base 10: ''
```

For example, the meltano.yml contains the following loader:

```
loaders:
  - name: target-postgres
    variant: transferwise
    pip_url: pipelinewise-target-postgres
    config:
      host: ${DB_HOST}
      port: ${DB_PORT}
      user: ${DB_USER}
      password: ${DB_PASS}
```

And in the docker-compose.yml I set up the env vars:

```
services:
  meltano:
    <<: *meltano-image
    command: ui
    environment:
      DB_HOST: ${DB_HOST}
      DB_PORT: ${DB_PORT}
      DB_USER: ${DB_USER}
      DB_PASS: ${DB_PASS}
```

If I remove the env vars from the meltano.yml, the exception doesn't occur.
Also, if I comment out the `meltano install` command from the `Dockerfile`, run the container and `bash` into it, the exception doesn't occur if I manually execute `meltano install`. In that case the env vars are also passed correctly, which I checked using the `meltano config target-postgres list` command.

The `Dockerfile` for reference:

```
# registry.gitlab.com/meltano/meltano:latest is also available in GitLab Registry
ARG MELTANO_IMAGE=meltano/meltano:latest-python3.9
FROM $MELTANO_IMAGE

RUN apt-get update && apt-get install libpq-dev -y

WORKDIR /project

# Install 
# Install any additional requirements
COPY ./requirements.txt .
RUN pip install -r requirements.txt

# Copy over Meltano project directory
COPY . .

RUN meltano install

# Don't allow changes to containerized project files
# ENV MELTANO_PROJECT_READONLY 1

ENTRYPOINT ["meltano"]
```

### Code

_No response_

WillDaSilva:
Previously string settings which were env vars (e.g. `"${DB_PORT}"`) would be replaced by `None` if the env var was missing during expansion. This was unintentionally changed in #7123 when the function which expanded env vars was changed.

This PR restores the old behaviour in `src/meltano/core/settings_service.py::SettingsService.get_with_metadata` where it was being relied on.

This PR does not re-implement the old behaviour in the function which expands env vars, since the old behaviour there seems incorrect/surprising.

Closes #7189

### DIFF
--- a/src/meltano/core/settings_service.py
+++ b/src/meltano/core/settings_service.py
@@ -360,6 +360,9 @@ class SettingsService(ABC):  # noqa: WPS214
                 env=expandable_env,
                 if_missing=EnvVarMissingBehavior(strict_env_var_mode),
             )
+            # https://github.com/meltano/meltano/issues/7189#issuecomment-1396112167
+            if value and not expanded_value:  # The whole string was missing env vars
+                expanded_value = None
 
             if expanded_value != value:
                 metadata["expanded"] = True
--- a/tests/meltano/core/plugin/test_plugin_settings.py
+++ b/tests/meltano/core/plugin/test_plugin_settings.py
@@ -538,7 +538,7 @@ class TestPluginSettingsService:
 
         assert config["var"] == "hello world!"
         assert config["foo"] == "42"
-        assert config["missing"] == ""
+        assert config["missing"] is None
         assert config["multiple"] == "rock paper scissors"
         assert config["info"] == "tap-mock"
 
@@ -939,4 +939,4 @@ class TestPluginSettingsService:
             [FEATURE_FLAG_PREFIX, str(FeatureFlags.STRICT_ENV_VAR_MODE)], False
         )
         subject.set("stacked_env_var", "${NONEXISTENT_ENV_VAR}")
-        assert subject.get("stacked_env_var") == ""
+        assert subject.get("stacked_env_var") is None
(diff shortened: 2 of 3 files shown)


edgarrmondragon (approved):

cjohnhanson (approved):
Seems like a reasonable fix for the relevant issue. 

Possibly relevant: long-term (but possibly not until 3.x) we want to be "strict" about environment variable references and raise errors whenever an environment variable is referenced but not set. As of now, this behavior is experimentally supported via the `STRICT_ENV_VAR_MODE` feature flag.

PR #7198, opened 2023-01-19
feat: Replace `venv` with `virtualenv`

### LINKED ISSUE #7152: Proposed: Should we add explicit `virtualenv` dependency
Follows from:

- https://github.com/meltano/meltano/issues/6882

We used to recommend _first_ installing `virtualenv`, then using `virtualenv` to install Meltano. This meant that Meltano was able to use `virtualenv` for plugin installation even without any explicit dependency declaration.

Today, if the user installs with `pipx`, then they also get `virtualenv` installed, since `pipx` ships with `virtualenv`.

Question: Should we just create an explicit `virtualenv` dependency so we can be sure it is available during plugin installation?

A valid scenario where `pip` would be fine for Meltano installation would be on a new+dedicated docker image. Presumably, that would be a safe time to install with `pip` and skip `pipx` installation - without requiring virtualenv isolation for Meltano itself. However, even in those cases, Meltano still needs `virtualenv` for plugin installation, and plugin installations would likely fail unless the image already contains `virtualenv` installed.

edgarrmondragon:
Closes #7152


src/meltano/core/venv_service.py
```diff
@@ -227,7 +227,10 @@ async def create(self) -> Process:
         """
         logger.debug(f"Creating virtual environment for '{self.namespace}/{self.name}'")
         try:
-            return await exec_async(sys.executable, "-m", "venv", str(self.venv))
+            return await exec_async(
+                os.path.join(os.path.dirname(sys.executable), "virtualenv"),
+                str(self.venv),
+            )
```
WillDaSilva, line 233: It may be worth exploring the use of `virtualenv` as a Python library rather than a CLI program: https://virtualenv.pypa.io/en/latest/user_guide.html#virtualenv.cli_run

edgarrmondragon: @WillDaSilva Yeah, the problem with that is we lose the concurrency of `exec_async`. I might dive a bit in the source of `virtualenv` to see to how hard it'd be to implement an async API for the `Session` object.

WillDaSilva: Good point. We likely won't be able to do it truly concurrently without using multiple processes. With that in mind my guess is that using the CLI will be the best approach in terms of effort versus reward.

src/meltano/core/venv_service.py
```diff
@@ -227,7 +227,10 @@ async def create(self) -> Process:
         """
         logger.debug(f"Creating virtual environment for '{self.namespace}/{self.name}'")
         try:
-            return await exec_async(sys.executable, "-m", "venv", str(self.venv))
+            return await exec_async(
+                os.path.join(os.path.dirname(sys.executable), "virtualenv"),
```
WillDaSilva, line 231: I see you pushed a commit that switched this over to using `pathlib`, and then reverted it. Why?

WillDaSilva (approved):

src/meltano/core/venv_service.py
edgarrmondragon: Pushed it without testing locally. Had to cast the `Path` instance to a string.

PR #7203, opened 2023-01-24
fix: "must be run inside a Meltano project" error

### LINKED ISSUE #7202: bug: Error message only says `selected_environment` when running outside of a project
### Meltano Version

2.12.0, 2.13.0, 2.14.0

### Python Version

NA

### Bug scope

### Operating System

N/A

### Description

Reproduction steps:
- Install Meltano `v2.14.0`, `v2.13.0`, or `v2.12.0`
- Run `meltano invoke example`  outside of a Meltano project directory

Expected output:

```

`meltano invoke` must be run inside a Meltano project.
Use `meltano init <project_directory>` to create one.
```

Actual output:

```

'selected_environment'
```

When using other commands (e.g. `config`) the error message may differ. Another one I saw was just `'is_default_environment'` instead of `'selected_environment'`.

---

It appears the cause is `meltano.cli.utils.activate_environment` and `meltano.cli.utils.activate_explicitly_provided_environment` accessing the `ctx.obj` dictionary without handling the case where the expected keys are missing.

### Code

_No response_

WillDaSilva:
While editing `tests/meltano/cli/test_cli.py` I also took the liberty of adding 5 seconds to the performance test, since it's a somewhat frequent cause of flakyness in our tests due to the Windows runners running slower. This isn't a good solution (if it can even be considered one), but it's probably worth doing until we have better performance testing in-place.

Closes #7202

### DIFF
--- a/tests/meltano/cli/test_cli.py
+++ b/tests/meltano/cli/test_cli.py
@@ -3,13 +3,13 @@ from __future__ import annotations
 import platform
 import re
 import shutil
+import subprocess
 from pathlib import Path
 from time import perf_counter_ns
 
 import click
 import pytest
 import yaml
-from click.testing import CliRunner
 from structlog.stdlib import get_logger
 
 import meltano
@@ -210,7 +210,7 @@ class TestCli:
             assert Path().resolve() == dirpath
 
     @pytest.mark.parametrize(
-        "command",
+        "command_args",
         (
             ("invoke", "example"),
             ("config", "example"),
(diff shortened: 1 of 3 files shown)


tests/meltano/cli/test_cli.py
```diff
@@ -208,6 +209,43 @@ def test_cwd_option(self, cli_runner, project, tmp_path: Path, pushd):
...
+            ("environment", "list"),
+            ("add", "utility", "example"),
+        ),
+    )
+    def test_error_msg_outside_project(
+        self,
+        tmp_path: Path,
+        command_args: tuple[str, ...],
+    ):
+        # Unless this test runs before every test that uses a project, we
```
edgarrmondragon, line 227: Maybe [`isolated_filesystem`](https://click.palletsprojects.com/en/8.1.x/api/#click.testing.CliRunner.isolated_filesystem) could help here?

edgarrmondragon (approved):

PR #7212, opened 2023-01-25
fix: Do not log env vars passed to subprocesses

### LINKED ISSUE #7184: bug: secrets are leaked when Meltano invokes subprocesses using log-level debug
We're not currently obfuscating any secrets in the `Env` dictionary that is dumped to the logs when running with `log-level=debug`. I'm guessing https://github.com/meltano/meltano/issues/7167 is related, but flagging as a new issue since it's tied to a specific subcommand.

edgarrmondragon:
It was originally added in https://github.com/meltano/meltano/commit/3603cd558613c5ae2a511f9860e0b599d12faf63

### DIFF
--- a/src/meltano/core/plugin_invoker.py
+++ b/src/meltano/core/plugin_invoker.py
@@ -426,7 +426,6 @@ class PluginInvoker:  # noqa: WPS214, WPS230
             popen_env = {**self.env(), **env}
             popen_args = self.exec_args(*args, command=command, env=popen_env)
             logging.debug(f"Invoking: {popen_args}")
-            logging.debug(f"Env: {popen_env}")
 
             try:
                 yield (popen_args, popen_options, popen_env)


WillDaSilva (commented):
If a user wishes to know exactly what env vars are being passed to a plugin, is there a way for them to do that after this change has been made?

edgarrmondragon:
> If a user wishes to know exactly what env vars are being passed to a plugin, is there a way for them to do that after this change has been made?

@WillDaSilva Other than plugin-specific via `meltano config <plugin-name> list`, no there isn't. I'm thinking `meltano run` should have a dry-run option that dumps env contents cc @tayloramurphy @aaronsteers

tayloramurphy:
@edgarrmondragon dropping the line is probably the quickest - it is a bit of a bummer that we lose a lot of valuable context with it. Ideally it'd be nice to still have the dictionary (or something equivalent) with sensitive values just asterisk'd out `*********` but I realize that's a bigger lift.

edgarrmondragon:
> @edgarrmondragon dropping the line is probably the quickest - it is a bit of a bummer that we lose a lot of valuable context with it. Ideally it'd be nice to still have the dictionary (or something equivalent) with sensitive values just asterisk'd out `*********` but I realize that's a bigger lift.

yeah, @aaronsteers summarizes the difficulties will in https://github.com/meltano/meltano/issues/7184#issuecomment-1402918632.

GitHub actions automatically mask repo and org secret values, but also allow custom masks at the action toolkit level with `core.setSecret` and at the workflow level with `echo "::add-mask::`.

aaronsteers:
> If a user wishes to know exactly what env vars are being passed to a plugin, is there a way for them to do that after this change has been made?

I think the way we'd suggest users approach this would be with the `--print-var` arg sent to the CLI.

https://docs.meltano.com/reference/command-line-interface#debugging-plugin-environment


> ### Debugging plugin environment [#](https://docs.meltano.com/reference/command-line-interface#debugging-plugin-environment)
>
> When debugging plugin configuration, it is often useful to view environment variables being provided to a plugin at runtime. This can be achieved using --log-level=debug but for readability and convenience, the meltano invoke command also supports printing individual environment variables to stdout at runtime:
> 
> ```
> # Print the runtime value PLUGIN_ENVIRONMENT_VARIABLE_1:
> meltano invoke --print-var <PLUGIN_ENVIRONMENT_VARIABLE_1> <PLUGIN_NAME>
> 
> # The option supports printing multiple variables as well.
> 
> # # Print the runtime values of both PLUGIN_ENVIRONMENT_VARIABLE_1 and PLUGIN_ENVIRONMENT_2 on separate lines:
> meltano invoke --print-var <PLUGIN_ENVIRONMENT_VARIABLE_1> --print-var <PLUGIN_ENVIRONMENT_VARIABLE_2>
> ```

aaronsteers:
> > If a user wishes to know exactly what env vars are being passed to a plugin, is there a way for them to do that after this change has been made?
> 
> I think the way we'd suggest users approach this would be with the `--print-var` arg sent to the CLI. We'd need to add support for this arg in more commands, though, since I think today it might only be supported in `invoke`.


Logged as new issue:

- https://github.com/meltano/meltano/issues/7219

WillDaSilva (approved):
Approved because #7219 has been opened

aaronsteers (approved):

PR #7213, opened 2023-01-25
fix: Omit local variables in `rich` tracebacks by default and add `show_locals` option to log formatter

### LINKED ISSUE #7167: bug: Secrets leak in logs if `asyncio.create_subprocess_exec()` fails (or if log-level=debug)
### Meltano Version

2.13.0

### Python Version

3.8

### Bug scope

### Operating System

Ubuntu

### Description

Meltano failed for me in production and during the printing of the huge line of stack traces (Which most of the traces are pretty useless to me while I"m in "user mode" unless I"m running in debug mode).

Ideally envs wouldn't print out. I was running `meltano run dashboard` which is a job name

```
 /venv/lib/python3.9/site-packages/meltano/core/plugin_invoker.py:453 in      │
│ invoke_async                                                                 │
│                                                                              │
│   450 │   │   │   popen_options,                                             │
│   451 │   │   │   popen_env,                                                 │
│   452 │   │   ):                                                             │
│ ❱ 453 │   │   │   return await asyncio.create_subprocess_exec(               │
│   454 │   │   │   │   *popen_args,                                           │
│   455 │   │   │   │   **popen_options,                                       │
│   456 │   │   │   │   env=popen_env,                                         │
│                                                                              │
│ ╭───────────────────────────────── locals ─────────────────────────────────╮ │
│ │          args = ('--discover',)                                          │ │
│ │        kwargs = {                                                        │ │
│ │                 │   'stdout': -1,                                        │ │
│ │                 │   'stderr': -1,                                        │ │
│ │                 │   'universal_newlines': False                          │ │
│ │                 }                                                        │ │
│ │    popen_args = [                                                        │ │
│ │                 │                                                        │ │
│ │                 '/builds/autoidm/prodder/.meltano/extractors/tap-gmail/… │ │
│ │                 │   '--config',                                          │ │
│ │                 │                                                        │ │
│ │                 '/builds/autoidm/prodder/.meltano/run/tap-gmail/tap.c25… │ │
│ │                 │   '--discover'                                         │ │
│ │                 ]                                                        │ │
│ │     popen_env = {                                                        │ │
│ │                 │                                                        │ │
│ │                 'TAP_GMAIL_PERSONAL_OAUTH_CREDENTIALS_REFRESH_TOKEN':    │ │
│ │                 'refreshtoken_listd_here… │ │
│ │                 │   'TAP_GMAIL_USER_ID': 'me',                           │ │
│ │                 │   'CI_PROJECT_NAMESPACE': 'autoidm',                   │ │
│ │                 │   'GITLAB_USER_ID': 'userid',                         │ │
│ │                 │   'CI_RUNNER_VERSION': '15.6.0~beta.186.ga889181a',    │ │
│ │                 │   'FF_SKIP_NOOP_BUILD_STAGES': 'true',                 │ │
│ │                 │   'CI_SERVER_NAME': 'GitLab',                          │ │
│ │                 │   'CI_RUNNER_DESCRIPTION':                             │ │
│ │                 '2-blue.shared.runners-manager.gitlab.com/default',      │ │
│ │                 │   'GITLAB_USER_EMAIL': 'dvisch@autoidm.com',           │ │
│ │                 │   'CI_SERVER_REVISION': 'fd1954e655c',   
```

### Code

_No response_

edgarrmondragon:
- [x] Add a `show_locals` option to `meltano.core.logging.console_log_formatter` so local vars can be omitted to avoid leaking secrets.
- [x] ~Capture more known errors to avoid printing full tracebacks.~ Not generally valuable for `meltano run` since there was a `logger.error(str(err), exc_info=True)` causing the traceback to be printed.

Closes #7167


WillDaSilva:
> Add a `show_locals` option to `meltano.core.logging.console_log_formatter` so local vars can be omitted to avoid leaking secrets.

@edgarrmondragon Does this only apply to logs being sent to the console? It'd be nice if every log handler could redact secrets from their messages.

edgarrmondragon:
> @edgarrmondragon Does this only apply to logs being sent to the console? It'd be nice if every log handler could redact secrets from their messages.

@WillDaSilva the `console_log_formatter` can be used with any handler, the console/stream handler is only the suggested use. So, yes, any handler that uses this formatter can omit local variables with this change. Other formatters don't have this problem AFAICT, i.e. they don't print local variables.

src/meltano/core/logging/formatters.py
```diff
@@ -47,27 +57,15 @@ def rich_exception_formatter_factory(
     def _traceback(sio, exc_info) -> None:
         sio.write("\n")
         Console(file=sio, color_system=color_system, no_color=no_color).print(
-            Traceback.from_exception(*exc_info, show_locals=True)
+            Traceback.from_exception(
+                *exc_info,
+                show_locals=show_locals,
```
edgarrmondragon, line 62: Defaulting this to `False` means logs will be a bit less verbose and won't contain local variables of arbitrary functions. Users can still enable verbose tracebacks (see changes to the logging guide).

edgarrmondragon: @kgpayne I think you first made the call to default `show_locals` to `True`, so pinging you to make sure this looks good to you.

src/meltano/core/logging/output_logger.py
```diff
@@ -202,7 +202,7 @@ def redirect_logging(self, ignore_errors=()):
         except ignored_errors:  # noqa: WPS329
             raise
         except Exception as err:
-            logger.error(str(err), exc_info=True)
```
edgarrmondragon, line 205: `exc_info=True` is not too useful here, since it'll point to a `raise RunnerError` if one of the blocks failed.

src/meltano/core/logging/formatters.py
kgpayne: @edgarrmondragon I think I was following examples from the Rich/structlog docs, so sure it wasn't a decision made with a strong use-case in mind. It made debugging more useful, which was the aim of the Rich formatting feature. However I appreciate, as a default, it shouldn't be used in prod. I know our docs suggest multiple logging configs, but this makes me think it would be good to associate logging features with environments somehow 🤔 I.e. By default in Meltano I want to see locals in dev, but not prod (without necessarily needing to configure my own logging configs). Just a thought 😅

src/meltano/core/logging/formatters.py
```diff
@@ -27,8 +35,9 @@
 
 
 def rich_exception_formatter_factory(
-    color_system: str = "auto",
+    color_system: Literal["auto", "standard", "256", "truecolor", "windows"] = "auto",
```
kgpayne, line 38: 🙌

kgpayne (approved):

src/meltano/core/logging/formatters.py
edgarrmondragon: @kgpayne Thanks for the context!

> but this makes me think it would be good to associate logging features with environments somehow

Agreed. I recall folks requesting to be able to put the contents of `logging.yaml` in `meltano.yml`.

WillDaSilva: Without having put much thought into it, being able to put the contents of `logging.yaml` into `meltano.yml` seems valuable.

docs/src/_guide/logging.md
```diff
@@ -38,12 +38,13 @@ disable_existing_loggers: false
...
     format: "[%(asctime)s] [%(process)d|%(threadName)10s|%(name)s] [%(levelname)s] %(message)s"
-  structured_plain: # log format for structured plain text logs without colored output
+  structured_plain_no_locals: # log format for structured plain text logs without colored output and without local variables
     (): meltano.core.logging.console_log_formatter
     colors: False # also disables `rich` traceback formatting
-  structured_colored: # log format for structured plain text logs WITH colored output
+    show_locals: False # disables local variable logging in tracebacks (which be very verbose and leak sensitive data)
+  structured_locals: # log format for structured plain text logs WITH local variables
     (): meltano.core.logging.console_log_formatter
-    colors: True # also enables traceback formatting with `rich`
```
WillDaSilva, line 46: Why remove `colors` from here?

WillDaSilva (approved):

docs/src/_guide/logging.md
edgarrmondragon: 'twas a mistake 😅

docs/src/_guide/logging.md
```diff
@@ -38,12 +38,13 @@ disable_existing_loggers: false
...
+    show_locals: True # enables local variable logging in tracebacks (can be very verbose and leak sensitive data)
```
edgarrmondragon, line 47:
```suggestion
    colors: True # also enables traceback formatting with `rich`
    show_locals: True # enables local variable logging in tracebacks (can be very verbose and leak sensitive data) 
```

Issue #7052, opened 2022-12-06
bug: Docs still recommends ELT command


pnadolny13:
I updated the title for this issue to make it more general, we still recommend the elt command a lot in the docs when run would be better.

- Replicate data - https://docs.meltano.com/guide/integration
- Orchestrate - https://docs.meltano.com/guide/orchestration
- Getting started guide - https://docs.meltano.com/getting-started/#schedule-pipelines-to-run-regularly
- etc.

Orignal description

> The getting started guide still recommends using the legacy extractor/loader/transform=skip syntax for scheduling pipelines.
> 
> https://docs.meltano.com/getting-started/#schedule-pipelines-to-run-regularly
> 
> The orchestrate data guide is updated to use job/task syntax https://docs.meltano.com/guide/orchestration.
> 
> Related to slack thread https://meltano.slack.com/archives/CMN8HELB0/p1670337451689919. We should update the GSG to use the new syntax because some users are getting confused with the mix of elt/run syntax.
> 

cc @sbalnojan

tayloramurphy:
FYI @sbalnojan @afolson

pnadolny13:
@tayloramurphy @sbalnojan I updated this issue because its more widespread than just the getting started guide. I think this is urgent from my view.

I would expect new users at this point to basically be unaware of the `elt` command because `run` is preferred. Maybe I'm wrong in that assumption. If I'm not then we should make the docs reflect it. It feels painful to see new users start their project with `elt`.

sbalnojan:
@pnadolny13 @tayloramurphy Sadly I agree :-/ I keep on stumbling over a lot of people who think the elt command is the command to use meltano. And I'm having a hard time saying "please don't use that" because it's quite widespread.

tayloramurphy:
@sbalnojan @pnadolny13 to be clear, we're not deprecating the `elt` command so it's not necessarily a bad thing that folks are using it. 

That said, we can likely reframe `elt` as useful for more advanced users as there are things you can do with `elt` that you can't do with `run`. (We've also talked about renaming this to just `el` in a future major release). So I'm supportive of reworking the docs to guide users to `meltano run` as it's broadly more powerful for ELT+. 

> It feels painful to see new users start their project with elt

Agreed - it should be on `run`. I just want us to be considered about how and where we talk about the `elt` command.

pnadolny13:
[...]

PR #7214, opened 2023-01-25
docs: remove/rename major elt references

sbalnojan:
As discussed in https://github.com/meltano/meltano/issues/7052, I removed/moved a few of the major mentions of the "meltano elt" command, making "meltano run" the prime suspect for doing your first bigger things with Meltano.

I also moved a few of the references to running Singer taps in isolation etc. to the advanced topics. Those IMHO shouldn't be inside the "Replicate Data" area. That should focus just on Meltano, and should be inclusive to e.g. airbyte taps.


docs/src/_guide/integration.md
```diff
@@ -272,28 +245,12 @@ meltano invoke <plugin> [PLUGIN_ARGS...]
...
-
-If using a custom tap, ensure that the tap declares the `state` capability as described in the [plugin capabilities reference](/reference/plugin-definition-syntax#capabilities).
-
 ### Testing Specific Failing Streams
 
 When extracting several streams with a single tap, it may be challenging to debug a single failing stream.
 In this case, it can be useful to run the tap with just the single stream selected.
 
-Instead of duplicating the extractor in `meltano.yml`, try running `meltano elt` with the [`--select` flag](/reference/command-line-interface#parameters-2).
+Instead of duplicating the extractor in `meltano.yml`, try running `meltano run` with the [`--select` flag](/reference/command-line-interface#parameters-2).
```
WillDaSilva, line 253: `meltano run` has no `--select` flag.

docs/src/_guide/integration.md
```diff
@@ -237,12 +210,12 @@ This command returns the merged result of the latest completed state plus any ne
...
-This can happen when a [`meltano elt`](/reference/command-line-interface#elt) execution is aborted before a particular stream completes.
+This can happen when a [`meltano run`](/reference/command-line-interface#run) execution is aborted before a particular stream completes.
 Partial state records can also be inserted manually via [`meltano state merge'](/reference/command-line-interface#state).
 
 Unlike [`meltano state merge`](/reference/command-line-interface#state),[`meltano state set`](/reference/command-line-interface#state) will insert a complete record, which causes meltano to ignore any previous state records, whether completed or partial.
 
-Note that if you already have a state file you'd like to use, it can be provided explicitly using [`meltano elt`](/reference/command-line-interface#elt)'s `--state` option or the [`state` extractor extra](/concepts/plugins#state-extra).
+Note that if you already have a state file you'd like to use, it can be provided explicitly using [`meltano run`](/reference/command-line-interface#run)'s `--state` option or the [`state` extractor extra](/concepts/plugins#state-extra).
```
WillDaSilva, line 218: `meltano run` has no `--state` option.

WillDaSilva (changes requested):

docs/src/_guide/integration.md
```diff
@@ -37,12 +37,12 @@ On a shell, you can explicitly `export` environment variables, that will be pass
...
+meltano run ...
 
-TAP_FOO_BAR=bar TAP_FOO_BAZ=baz meltano elt ...
+TAP_FOO_BAR=bar TAP_FOO_BAZ=baz meltano run ...
 ```
 
-To verify that these environment variables will be picked up by Meltano as you intended, you can test them with [`meltano config <plugin>`](/reference/command-line-interface#config) before running `meltano elt`.
+To verify that these environment variables will be picked up by Meltano as you intended, you can test them with [`meltano config <plugin>`](/reference/command-line-interface#config) before running `meltano run`.
 
 If you're using [`meltano schedule`](/reference/command-line-interface#schedule) to [schedule your pipelines](/guide/orchestration), you can specify environment variables for each pipeline in your [`meltano.yml` project file](/concepts/project#meltano-yml-project-file), where each entry in the `schedules` array can have an `env` dictionary:
```
tayloramurphy, line 48: @sbalnojan just below this is the non-job format for schedules, would you want to update that as well? The current format uses `elt` behind the scenes while the jobs format would use `run`.

tayloramurphy (changes requested):

docs/src/_guide/integration.md
```diff
@@ -37,12 +37,12 @@ On a shell, you can explicitly `export` environment variables, that will be pass
 ```bash
 export TAP_FOO_BAR=bar
```
tayloramurphy, line 38: Perhaps this is best for another issue / PR, but we probably shouldn't recommend folks manage env vars this way. They can use the `env:` key to export environment variables!

sbalnojan:
@tayloramurphy @WillDaSilva thanks! I made the changes. I just cut out a bit of these parts completely, I think for a "replicate data" guide the "pipeline specific configuration" is waaaay too much overhead, so it's now inside the advanced topics (those get way less traffic so I didn't bother to rewrite the schedule syntax for now).

tayloramurphy:
@sbalnojan it looks like Will's comments still need addressing around the available options for run vs elt.

sbalnojan:
@tayloramurphy hm so now everything should be addressed, I think I removed all parts referenced by Will, and I incorporated the comments from @pnadolny13 on the ticket.  If something is still missing, please highlight the line, otherwise I won't be able to find it :-)

docs/src/_guide/integration.md
```diff
@@ -15,69 +15,17 @@ and keeps track of [the incremental replication state](#incremental-replication-
...
 
 ## Plugin configuration
 
-As described in the [Configuration guide](/guide/configuration#configuration-layers), [`meltano elt`](/reference/command-line-interface#elt) will determine the configuration of the extractor, loader, and (optionally) transformer by looking in [**the environment**](/guide/configuration#configuring-settings), your project's [**`.env` file**](/concepts/project#env), the [system database](/concepts/project#system-database), and finally your [**`meltano.yml` project file**](/concepts/project#meltano-yml-project-file), falling back to a default value if nothing was found.
+As described in the [Configuration guide](/guide/configuration#configuration-layers), [`meltano run`](/reference/command-line-interface#run) will determine the configuration of the extractor, loader, and (optionally) transformer by looking in [**the environment**](/guide/configuration#configuring-settings), your project's [**`.env` file**](/concepts/project#env), the [system database](/concepts/project#system-database), and finally your [**`meltano.yml` project file**](/concepts/project#meltano-yml-project-file), falling back to a default value if nothing was found.
```
WillDaSilva, line 23: The configuration layers section of the configuration guide describes these things for `meltano elt`, not `meltano run`. Even if it were updated to say `meltano run` instead of `meltano elt`, I'm not sure that that would make it true.

WillDaSilva (changes requested):

docs/src/_guide/integration.md
WillDaSilva: @sbalnojan Two possible quick fixes:
- Update `guide/configuration#configuration-layers` to use `meltano run` instead of `meltano elt` (preferable).
- Remove the link to `guide/configuration#configuration-layers`. The rest of the paragraph seems mostly fine (if a bit light on details) without it.

In either case you'll need to rephrase this part:
> will determine the configuration of the extractor, loader, and (optionally) transformer

sbalnojan:
@WillDaSilva I will close it.

PR #7221, opened 2023-01-25
docs: Ensure search button is always accessible from nav bar/menu

DouweM:


tayloramurphy:
@sbalnojan anything else we can cut from the header here? Even with @DouweM's change i can't see the search button by default.

WillDaSilva:
@DouweM @tayloramurphy What are you seeing on your screens? It looks fine to me in Firefox and Chrome in desktop and phone views.

There is a brief period when the screen is ~1100px wide where the search button has been pushed off the right side of the screen, but it hasn't switched to phone view yet. I think a fix for that would be to change our phone view `min-width` from its current value `1023px` to `1200px`.

tayloramurphy:
@WillDaSilva my default width is right at 1120 px. At that width I see the following currently:

Screenshot 2023-01-26 at 11 34 57 AM

It seems like from 1024 px to about 1224 px the search button is cut off. to see all the content I have to be at 1334 px

WillDaSilva:
Increasing the `$navbar-breakpoint` Bulma variable seems to have done the trick. Reference: https://bulma.io/documentation/components/navbar/#variables

@tayloramurphy Could you please check https://deploy-preview-7221--meltano.netlify.app/getting-started/installation to confirm that the issue has been fixed?

tayloramurphy:
@WillDaSilva yep - seems to fix the bug generally. It now goes to the hamburger menu by default for me, but better than the alternative. @sbalnojan maybe in another PR we can change some more text or update the headers somewhat to make fewer options.

tayloramurphy (approved):

PR #7236, opened 2023-01-30
fix: Invoke `virtualenv` as a module

edgarrmondragon:
To fix

```
No such file or directory: '/usr/bin/virtualenv'
```

seen in https://github.com/MeltanoLabs/target-postgres/actions/runs/4046145129/jobs/6958562915

Note that this doesn't affect `pipx install meltano` installations nor the docker images.


edgarrmondragon:
@kgpayne this should probably go out in a patch release 🙏

WillDaSilva (approved):

PR #7238, opened 2023-01-30
chore: run `poetry lock --check` with `pre-commit`

WillDaSilva:
Also some minor style updates to the `pre-commit` config file.


edgarrmondragon:
My only concern is with `pre-commit autoupdate` not working properly with poetry hooks https://github.com/pre-commit/pre-commit/issues/2507.

TL;DR: `pre-commit autoupdate` reverts `1.3.2` to `1.3.0`:

```
$ gh repo clone https://github.com/python-poetry/poetry
$ cd poetry
$ git describe --tags --abbrev=0
1.3.0
```

WillDaSilva:
@edgarrmondragon Wow. I went through the issue tracker, and the dismissiveness of this real (and imo serious bug) is somewhat astounding.

Thanks for calling attention to this. Now I know to be more wary of `pre-commit autoupdate`.

I'll replace this hook with a local that does the same as the one provided by Poetry, which will prevent `pre-commit autoupdate` from operating on it.

edgarrmondragon (approved):

PR #7240, opened 2023-01-31
chore: Update PyPI classifiers

WillDaSilva:


pyproject.toml
```diff
@@ -7,7 +7,13 @@ license = "MIT"
...
 repository = "https://github.com/meltano/meltano"
 classifiers = [
-  "Programming Language :: Python :: 3",
+  "Development Status :: 5 - Production/Stable",
+  "Programming Language :: Python :: 3 :: Only",
+  "Programming Language :: Python :: 3.7",
+  "Programming Language :: Python :: 3.8",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
```
edgarrmondragon, line 16: Some of of these seem to be added automatically: 
Screenshot 2023-01-30 at 20 04 16

It's probably ok to be explicit, though

edgarrmondragon (approved):

pyproject.toml
WillDaSilva: Oh that's strange - I added them explicitly because I recall seeing them missing from Meltano on PyPI. Perhaps I was looking at a different view, or maybe I just misremembered.

edgarrmondragon (approved):

PR #7245, opened 2023-01-31
chore: Remove superfluous `package.json` and `yarn.lock`

WillDaSilva:
Both `docs/` and `src/webapp` have their own `package.json` and lock file. The one at the root of the repository is unncessary, and seemingly unused. The repeated dependencies are difficult to manage.

The testing section of the prerequisites docs needed to be updated to reflect that you now have to run `cd src/webapp` before running `yarn`, but as I was updating them I realized that they didn't make sense. Since we've deprecated the web UI, I don't want to spend time figuring out how to run those Cypress tests, then documenting them. The quick fix to reduce harm is to remove the inaccurate documentation.


cjohnhanson (approved):
🚢

PR #7250, opened 2023-01-31
ci: Only run CI tests when relevant files are changed

WillDaSilva:


.github/workflows/test.yml
```diff
@@ -1,9 +1,21 @@
 name: Tests
 
 on:
-  pull_request: {}
+  pull_request:
+    paths:
+      - src/**
```
edgarrmondragon, line 6:
```suggestion
      - src/**
      - tests/**
```

.github/workflows/test.yml
```diff
@@ -1,9 +1,21 @@
...
+      - noxfile.py
+      - pyproject.toml
+      - .github/workflows/test.yml
+      - .github/workflows/resources/constraints.txt
   push:
     branches: [main]
+    paths:
+      - src/**
```
edgarrmondragon, line 14:
```suggestion
      - src/**
      - tests/**
```

.github/workflows/test.yml
WillDaSilva: :facepalm:

.github/workflows/test.yml
WillDaSilva: Thanks for catching this

edgarrmondragon (approved):

PR #7252, opened 2023-01-31
ci: Misc test workflow improvements

WillDaSilva:
- Run `mypy` and `coverage` with Python 3.10
- Use `nox` to run `coverage`


.github/workflows/test.yml
```diff
@@ -73,32 +73,19 @@ jobs:
...
-    - name: Configure Poetry
-      run: |
-        poetry env use "${{ matrix.python-version }}"
-        poetry debug info
+    - name: Setup Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v4.5.0
+      with:
+        python-version: ${{ matrix.python-version }}
+        architecture: x64
+        cache: 'poetry'
```
edgarrmondragon, line 88: IIRC the reason I used the `pip` cache is because this workflow doesn't really run `poetry install`, but instead `pip install` under the hood via Nox.

edgarrmondragon (approved):

PR #7256, opened 2023-01-31
chore: Update `environment_context` 1-2-0 to work with SnowcatCloud

WillDaSilva:
SnowcatCloud uses an older version of jsonschema that doesn't allow the `propertyNames` field.


pnadolny13 (approved):
👍

PR #7258, opened 2023-01-31
docs: Fix `virtualenvs.in-project` setting in custom extractor debugging doc

davidwitk:
The provided code snippet did not work for me, saying `Setting virtualenv.in-project does not exist`. Googling for the error took me here and helped me to find the correct setting: https://stackoverflow.com/a/62071321


WillDaSilva (approved):
Thanks @chodera!

PR #7261, opened 2023-02-01
feat: Meltano Cloud CLI

### LINKED ISSUE #7226: Preview Meltano Cloud CLI
Prior discussions:
- https://github.com/meltano/cloud-planning/discussions/33
- https://meltano.slack.com/archives/C03GKHWS0HM/p1671478362577159

---

I propose we create a feature branch based off of `main` called `cloud`, which will be kept up to date with `main`.

On this branch we would develop Meltano Cloud features that need a CLI, or otherwise interact with `meltano` in some way.

Meltano Cloud alpha/beta users would install this branch as opposed to the regular Meltano from PyPI by using `pipx install git+https://github.com/meltano/meltano.git@cloud`.

Running `meltano upgrade` would not work because it tries to upgrade using `meltano` from PyPI. As such, the `upgrade` command should be temporarily disabled.

Likewise, building the UI requires Yarn & Node v16 as system dependencies (among other things), and so because installing directly from source normally forces the user to build the UI, we should skip building the UI in `scripts/build.py` on this feature branch, and ensure that `meltano` works fine without it (i.e. the `ui` command should error with a clear message). The full removal of the UI should wait until Meltano v3 at least, and so will not be part of this branch.

A first iteration of this branch should provide the `meltano cloud` command group as a minimal Click CLI. Initially it may have a simple no-op command under it, but in the future we expect to add several commands including `log`, `login`, `run`, etc.

In particular, there would be immediate value for Meltano Cloud Alpha users in being able to re-run their scheduled jobs. This is currently possible through a curl command, but we could integrate it into the Cloud CLI. This would be done as a follow-up.

We will likely want to implement the Meltano Cloud CLI in `src/meltano-cloud/` as a separate Python package. In addition to being imported by `meltano` to provide the `meltano cloud` CLI, this package would also provide its own CLI as `meltano-cloud`.

Ideally `meltano` can import from `meltano-cloud`, but `meltano-cloud` cannot import from `meltano`. Maintaining this separation essentially reduces `meltano-cloud` to a simple CLI that handles HTTPS requests and responses to/from Meltano Cloud API endpoints. `meltano-cloud` would then be usable on its own as a lightweight utility, and would not interact with local Meltano projects.

WillDaSilva:
Documented at `src/cloud-cli/README.md`

To do:
- [x] Run Cloud CLI tests with Nox
- [x] Make `cloud` a protected branch
- [x] Remove unused `importlib-metadata` dependency
- [x] Switch `cloud-cli` from proprietary license to MIT (per https://github.com/meltano/internal-general/issues/466)

Follow ups:
- https://github.com/meltano/meltano/issues/7266

Closes #7226


src/meltano/cli/cli.py
```diff
@@ -64,7 +63,7 @@ def main(self, *args, **kwargs) -> NoReturn:
     type=click.Path(exists=True, file_okay=False, resolve_path=True, path_type=Path),
     help="Run Meltano as if it had been started in the specified directory.",
 )
-@click.version_option(version=meltano.__version__, prog_name="meltano")
+@click.version_option(prog_name="meltano")
```
edgarrmondragon, line 66: Would this help `click` figure out the version slightly faster?

```suggestion
@click.version_option(prog_name="meltano", package_name="meltano")
```

https://github.com/pallets/click/blob/cba52fa76135af2edf46c154203b47106f898eb3/src/click/decorators.py#L460-L475

src/meltano/cli/upgrade.py
```diff
@@ -34,9 +34,16 @@ def upgrade(ctx, project):
 
     \b\nRead more at https://docs.meltano.com/reference/command-line-interface#upgrade
     """
-    engine, _ = project_engine(project)
-    upgrade_service = UpgradeService(engine, project)
-    ctx.obj["upgrade_service"] = upgrade_service
+    raise Exception(
```
edgarrmondragon, line 37: Would a more specific exception make sense here?

```suggestion
    raise CliError(
```

edgarrmondragon (approved):

src/cloud-cli/README.md
```diff
@@ -0,0 +1,96 @@
+# Meltano Cloud CLI
```
cjohnhanson, line 1: Great docs -- thanks for taking the time to write this out.

cjohnhanson (approved):
Great work! We should probably create the follow-up issue with info on how this will be merged into `main` when the time comes. That way we have somewhere to track any hypothetical drift that we may have to introduce between now and then. :shipit:

PR #7264, opened 2023-02-02
chore: Enable type checking in other core modules

edgarrmondragon:
See https://github.com/meltano/meltano/issues/6491


WillDaSilva (approved):
Love to see it

PR #7269, opened 2023-02-03
chore: Remove duplicate schema files in `schema/`

### LINKED ISSUE #7267: Move schemas to `src/meltano/schemas/`
Currently we rely on the wheel build process to ensure the schemas are included in the Meltano package, similar to what we do for the UI.

This results in development pains as described here:
- https://github.com/meltano/meltano/pull/7148#issuecomment-1414658980

Ideally we would simply store the schemas under `src/meltano/schemas/` instead of `schema/`, but the problem with that is that https://github.com/SchemaStore/schemastore expects them to be located at `schema/` on `main`.

To fix this, we can:
- [x] Merge a PR here to copy the schemas from `schema/` to the `src/meltano/schemas/`: https://github.com/meltano/meltano/pull/7268
- [x] Merge a PR at https://github.com/SchemaStore/schemastore to have them use the new location: https://github.com/SchemaStore/schemastore/pull/2775
- [ ] Merge a PR here to delete the `schema/` directory now that it is no longer being used: https://github.com/meltano/meltano/pull/7269

With this done, and the UI being removed in Meltano 3.0, we may even be able to get rid of the hacky custom build process entirely!

WillDaSilva:
The schemas are now located in `src/meltano/schemas`.

This PR will remain a draft until the following has been merged:
- https://github.com/SchemaStore/schemastore/pull/2775

Closes: #7267


edgarrmondragon:
Just need make sure @pnadolny13 is not using these in a way that moving them would break something in some workflow

WillDaSilva:
[This GitHub org search](https://github.com/search?q=org%3Ameltano+meltano.schema.json&type=code) suggests that it's only being used in this repository, or somewhere outside of GitHub. Running this by @pnadolny13 first is a good call.

pnadolny13 (approved):
@WillDaSilva looks good to me! I dont have any workflows tied to this.

PR #7279, opened 2023-02-07
docs: change ref

sbalnojan:
I messed up parts of the links in  #7274, this fixes it, now completely IMHO.

(The ToC on the RHS displays weird links in the current docs thanks to  #7274).

### DIFF
--- a/docs/src/_reference/glossary.md
+++ b/docs/src/_reference/glossary.md
@@ -7,7 +7,7 @@ weight: 5
 
 {% for item in site.data.glossary %}
 
-<h2 id="#{{ item.shorthand }}">{{ item.term }}</h2>
+<h2 id="{{ item.shorthand }}">{{ item.term }}</h2>
 
 <p>{{ item.definition }}
 {% if item.link %}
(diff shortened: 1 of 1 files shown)


docs/src/_reference/glossary.md
```diff
@@ -7,7 +7,7 @@ weight: 5
 
 {% for item in site.data.glossary %}
 
-<h2>{{ item.term }} <a href="https://docs.meltano.com/reference/glossary#{{ item.shorthand }}">#</a></h2>
+<h2 id="#{{ item.shorthand }}">{{ item.term }}</h2>
```
WillDaSilva, line 10: As it is currently, it creates links like https://deploy-preview-7279--meltano.netlify.app/reference/glossary##python-venv (note the double `#`). This occurs in Firefox and Chrome.

Maybe removing the `#` would fix that? I haven't tested this change - it's just a guess.

```suggestion
<h2 id="{{ item.shorthand }}">{{ item.term }}</h2>
```

WillDaSilva (changes requested):

WillDaSilva (approved):

PR #7283, opened 2023-02-07
feat: Lazy install plugins on invocation

JulesHuisman:
# Lazy install
I was playing around with the idea of lazy installing plugins if there are present in your Meltano project, but the executable is missing. Instead of displaying the missing "Missing plugin run meltano install xyz to install it", Meltano would install the plugin for you and re-invoke it instantly. 

This would remove the necassity of always running Meltano install on a new project. It would also make the use of Meltano in ephemeral environments easier, as you no longer need to have all plugin pre-installed. 

When running the following command `rm -rf .meltano && meltano invoke dagster` it looks like this:
image

# Catalog file issue
I am running into an issue however, when running the following command: `rm -rf .meltano && meltano --log-level=debug run tap-carbon-intensity target-jsonl` the code reaches the discovery command and tries to install the tap. However, here some weird bugs occur which I can't quite figure out. 

```
Cannot start plugin tap-carbon-intensity: Catalog discovery failed: invalid catalog: [Errno 2] No such file or directory: '/home/jules/projects/temp/project/.meltano/run/tap-carbon-intensity/tap.properties.json'
```
It probably has to do something with the fact it tries to write the catalog here: https://github.com/meltano/meltano/blob/af049cc941884c1f19356b92697056e20bd2b2ef/src/meltano/core/plugin/singer/tap.py#L376, while the file is open here: https://github.com/meltano/meltano/blob/af049cc941884c1f19356b92697056e20bd2b2ef/src/meltano/core/plugin/singer/tap.py#L405

Do you think the lazy install is feasable, or is it too much of a hassle?

### DIFF
--- a/src/meltano/cli/invoke.py
+++ b/src/meltano/cli/invoke.py
@@ -173,29 +173,9 @@ async def _invoke(
                     *plugin_args,
                 )
             else:
-                try:
-                    handle = await invoker.invoke_async(*plugin_args, command=command_name)
-                    with propagate_stop_signals(handle):
-                        exit_code = await handle.wait()
-                except ExecutableNotFoundError:
-                    # If the plugin is not installed, try to install it
-                    logging.info(f"Installing plugin '{invoker.plugin.name}'...")
-
-                    # Try to install the plugin
-                    install_service = PluginInstallService(
-                        project=project,
-                        plugins_service=invoker.plugins_service,
-                        clean=True
-                    )
-                    await install_service.install_plugin_async(invoker.plugin)
-
-                    # Log that the plugin is installed
-                    logging.info(f"Plugin '{invoker.plugin.name}' installed. Retrying invocation...")
-
-                    # Try to invoke the plugin again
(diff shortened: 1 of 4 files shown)


aaronsteers:
@JulesHuisman - First of all - this is very cool. Thanks for contributing! 

> However, here some weird bugs occur which I can't quite figure out.
> 
> ```
> Cannot start plugin tap-carbon-intensity: Catalog discovery failed: invalid catalog: [Errno 2] No such file or directory: '/home/jules/projects/temp/project/.meltano/run/tap-carbon-intensity/tap.properties.json'
> ```

It sounds like the 'discovery' operation might be failing silently, failing on the next step (sync), then installing, and then failing on retry of sync because the catalog doesn't exist.

> Do you think the lazy install is feasable, or is it too much of a hassle?

I think it is definitely feasible, but I wonder if it would be easier to introduce some type of "is_installed()" check as condition for inline install, rather than triggering install upon failure. Wdyt?

Issue #6386, opened 2022-07-08
Allow env vars to be set for jobs


rabidaudio:
See discussion in slack: https://meltano.slack.com/archives/C01TCRBBJD7/p1657233220239139

Consider example:

```yml
jobs:
- name: gitlab
  tasks:
  - dbt-postgres:seed tap-gitlab target-postgres dbt-postgres:snapshot dbt-postgres:run dbt-postgres:test
 ```
 
 
`meltano run gitlab` will run all the dbt commands for the whole project rather than allowing me to specify only the specific models that have changed (e.g. `DBT_MODELS=+gitlab+`).

I realize this can be solved by schedules:

```yml
schedules:
- name: gitlab
  interval: '@daily'
  job: gitlab
  env:
    DBT_MODELS: +gitlab+
```

However it seems to me that this information belongs to the job itself, not the schedule; that is even if I were to manually run `meltano run gitlab` outside of the `@daily` schedule, I’d still expect it to apply my configuration.

Proposal: Allow `env`s to be set on a job basis:

```yml
jobs:
- name: gitlab
  tasks:
  - dbt-postgres:seed tap-gitlab target-postgres dbt-postgres:snapshot dbt-postgres:run dbt-postgres:test
  env:
    DBT_MODELS: +gitlab+
    TARGET_POSTGRES_BATCH_SIZE_ROWS: 100 # could also be useful to override plugin configs
```

Precedence can get tricky here, but I would expect:

meltano environment < job < schedule < .env < system envrionment

rabidaudio:
By `a < b` I meant b has higher precedence than a but I can see how that was ambiguous. Yes, I agree that schedules should have higher precedence than jobs.

I'm a little surprised that system env vars are overwritten by environments (for example, I might do `FOO=override meltano run`), but I don't think it's a problem and I'm sure y'all have considered it more than I have

pnadolny13:
@tayloramurphy @rabidaudio I landed on this while working on https://github.com/meltano/squared/issues/289 where I was trying to get away from exporting env vars before running a command like in this case https://github.com/meltano/squared/blob/64cffc745eba6721fb97490d5aa14f6f7a970afc/data/orchestrate/dag_definition.yml#L146 when using schedules/jobs. I was a able to solve it without schedule/job env vars but was about to request the same feature before I found this issue so I was giving it some thought also:

[...]

PR #7284, opened 2023-02-07
fix: Use `$MELTANO_ENVIRONMENT` set in `.env`

### LINKED ISSUE #7280: bug: `meltano config <plugin>` ignores `MELTANO_ENVIRONMENT` environment variable
### Meltano Version

2.15.0

### Python Version

3.11

### Bug scope

### Operating System

macOS Monterey 12.6

### Description

We run meltano using environment variables for as much configuration as possible. I have found what I think is a bug, in that I have the meltano environment set via an environment variable, but the `meltano config <plugin>` command ignores it.

Relevant sections of `.env`:
```shell
MELTANO_ENVIRONMENT=dev
```

and `meltano.yml`
```yaml
include_paths:
- ./environments/dev.meltano.yml
plugins:
  extractors:
  - name: tap-spreadsheets-anywhere
    variant: ets
    pip_url: git+https://github.com/ets/tap-spreadsheets-anywhere.git@34f135b45a87b468f3375c72f9d451ae6525fae7
  - name: tap-s3-mongo
    inherit_from: tap-spreadsheets-anywhere
```

and from `environments/dev.meltano.yml`
```yaml
environments:
  - name: dev
    config:
      plugins:
        loaders:
          - name: target-redshift-mongo
            config:
              schema_mapping:
                payments:
                  target_schema: payments
                profiles:
                  target_schema: customers
```

Expected behavior: I expect that running `meltano config target-redshift-mongo` here (with meltano environment set to `dev` via environment variables) includes the schema mapping defined for this plugin in the dev environment.

Observed behavior: It does not include the schema mapping. But, if I pass `--environment=dev` explicitly as a command-line argument, it does include it.

```shell
# note: MELTANO_ENVIRONMENT=dev is set in .env
$ meltano config target-redshift-mongo
2023-02-07T12:43:14.151935Z [info     ] The default environment 'dev' will be ignored for `meltano config`. To configure a specific environment, please use the option `--environment=<environment name>`.
{
  "host": "<SECRET>",
  "port": <SECRET>,
  "dbname": "<SECRET>",
  "user": "<SECRET>",
  "password": "<SECRET>",
  "s3_bucket": "<SECRET>",
  "default_target_schema": "<SECRET>",
  "copy_options": "EMPTYASNULL BLANKSASNULL TRIMBLANKS TRUNCATECOLUMNS TIMEFORMAT 'auto' COMPUPDATE OFF STATUPDATE OFF",
  "batch_size_rows": 100,
  "flush_all_streams": false,
  "parallelism": 0,
  "max_parallelism": 16,
  "disable_table_cache": false,
  "add_metadata_columns": true,
  "hard_delete": false,
  "data_flattening_max_level": 0,
  "primary_key_required": false,
  "validate_records": false,
  "skip_updates": false,
  "slices": 1
}
```

```shell
# note: MELTANO_ENVIRONMENT=dev is set in .env
$ meltano --environment=dev config target-redshift-mongo
2023-02-07T12:45:17.010451Z [info     ] Environment 'dev' is active
{
  "host": "<SECRET>",
  "port": <SECRET>,
  "dbname": "<SECRET>",
  "user": "<SECRET>",
  "password": "<SECRET>",
  "s3_bucket": "<SECRET>",
  "default_target_schema": "<SECRET>",
  "copy_options": "EMPTYASNULL BLANKSASNULL TRIMBLANKS TRUNCATECOLUMNS TIMEFORMAT 'auto' COMPUPDATE OFF STATUPDATE OFF",
  "batch_size_rows": 100,
  "flush_all_streams": false,
  "parallelism": 0,
  "max_parallelism": 16,
  "schema_mapping": {
    "payments": {
      "target_schema": "payments"
    },
    "profiles": {
      "target_schema": "customers"
    }
  },
  "disable_table_cache": false,
  "add_metadata_columns": true,
  "hard_delete": false,
  "data_flattening_max_level": 0,
  "primary_key_required": false,
  "validate_records": false,
  "skip_updates": false,
  "slices": 1
}
```

If I change the `default_environment` in `meltano.yml` (so that it's not the same as the environment I have defined in `.env`) I get the same behavior.

### Code

_No response_

WillDaSilva:
Context on Slack: https://meltano.slack.com/archives/C03GKHWS0HM/p1675776097120869

I decided to add some special handling for `MELTANO_ENVIRONMENT` to fix this, rather than try for some kind of generic solution for respecting env vars early on in the CLI. This was partially because we already rely on special handling for this env var and its related options/settings, so any sort of generic solution would have to content with the special handling that already exists.

I pulled out the existing logic for detecting which Meltano environment was selected into a function `detect_selected_environment`, and updated it to check `project.dotenv_env` for `MELTANO_ENVIRONMENT` as a fallback if it wasn't set terminal environment. The order of precedence is:
1. `$MELTANO_ENVIRONMENT` set in the terminal
2. `$MELTANO_ENVIRONMENT` set in `.env`
3. The `--environment` CLI option.

@tayloramurphy Please let me know if that precedence is correct.

Closes #7280

### DIFF
--- a/docs/src/_getting-started/index.md
+++ b/docs/src/_getting-started/index.md
@@ -127,7 +127,7 @@ This will allow you to use [`git diff`](https://git-scm.com/docs/git-diff) to ea
 
 ## View and Activate Your Environments
 
-As part of creating your Meltano project, we automatically added your first [environments](/concepts/environments) called `dev`, `staging` and `prod`. This allows you to define configurations specific to the environment in which you're running your project. Theres also a [`default_environment`](https://docs.meltano.com/concepts/environments#default-environments) setting in the `meltano.yml` that get automatically set to `dev`, you can list and change the active environment using:
+As part of creating your Meltano project, we automatically added your first [environments](/concepts/environments) called `dev`, `staging` and `prod`. This allows you to define configurations specific to the environment in which you're running your project. There's also a [`default_environment`](https://docs.meltano.com/concepts/environments#default-environments) setting in the `meltano.yml` that get automatically set to `dev`, you can list and change the active environment using:
 
 1. List your available environments:
 
@@ -147,7 +147,7 @@ As part of creating your Meltano project, we automatically added your first [env
    $env:MELTANO_ENVIRONMENT="dev"
    ```
 
-   Alternatively you can include the `--environment=dev` argument to each meltano command. You should now see a log message that says `Environment 'dev' is active` each time you run a meltano command.
+   Alternatively you can include the `--environment=dev` argument to each `meltano` command. You should now see a log message that says `Environment 'dev' is active` each time you run a `meltano` command.
 
 1. [optional] Add a new environment:
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -250,6 +250,7 @@ pyflakes = [
   "-WPS407",
   "-WPS410", # Allow metadata variable (__all__)
   "-WPS412",
+  "-WPS414", # Allow unpacking assignments
   "-WPS420", # Allow `pass` keywords in `except` bodies to prevent falling through to another `except`
   "-WPS429",
   "-WPS430",
--- a/src/meltano/cli/cli.py
+++ b/src/meltano/cli/cli.py
@@ -154,9 +154,9 @@ def detect_selected_environment(
     Returns:
         The selected environment, and whether it is the default environment.
     """
-    environment = cli_environment or os.environ.get(
+    environment = os.environ.get(
         PROJECT_ENVIRONMENT_ENV,
-        project.dotenv_env.get(PROJECT_ENVIRONMENT_ENV, None),
+        project.dotenv_env.get(PROJECT_ENVIRONMENT_ENV, cli_environment),
     )
     if cli_no_environment or (environment and environment.lower() == "null"):
         logger.info("No Meltano environment was selected")
--- a/src/meltano/cli/cli.py
+++ b/src/meltano/cli/cli.py
@@ -15,7 +15,7 @@ from meltano.cli.utils import InstrumentedGroup
 from meltano.core.behavior.versioned import IncompatibleVersionError
 from meltano.core.error import EmptyMeltanoFileException
 from meltano.core.logging import LEVELS, setup_logging
-from meltano.core.project import Project, ProjectNotFound
+from meltano.core.project import PROJECT_ENVIRONMENT_ENV, Project, ProjectNotFound
 from meltano.core.project_settings_service import ProjectSettingsService
 from meltano.core.tracking import Tracker
 from meltano.core.tracking.contexts import CliContext
@@ -51,11 +51,7 @@ class NoWindowsGlobbingGroup(InstrumentedGroup):
     "--log-config", type=str, help="Path to a python logging yaml config file."
 )
 @click.option("-v", "--verbose", count=True, help="Not used.")
-@click.option(
-    "--environment",
-    envvar="MELTANO_ENVIRONMENT",
-    help="Meltano environment name.",
-)
+@click.option("--environment", help="Meltano environment name.")
 @click.option(
     "--no-environment", is_flag=True, default=False, help="Don't use any environment."
 )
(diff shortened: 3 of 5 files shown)


tayloramurphy:
@WillDaSilva I'm referencing my comment in https://github.com/meltano/meltano/issues/6386#issuecomment-1195887590 when I talk about this. From there we laid out:

```
- environment level plugin env # highest
- schedule level env
- job level env
- root level plugin env 
- environment level env 
- root level env
- .env file
- terminal env # lowest
```

```
command line flag # highest
environment level plugin env
... same as above ...
.env file
terminal  # lowest
```

My logic being that if you're setting something at run time in the command, you want that to take precedence over everything. @aaronsteers will you validate my logic and structure here please?

WillDaSilva:
@tayloramurphy We previously used [Click](https://click.palletsprojects.com/en/8.1.x/options/#values-from-environment-variables) for this, and Click makes the env var take precedence over the CLI option. We can change the precedence per your request, but we should be aware that that would make this a change to the existing behaviour, rather than a mere addition.

tayloramurphy:
@WillDaSilva ok, good to know. Let's get @aaronsteers's buy-in on this. It seems like if we have a current behavior we need to validate it's documented properly in https://docs.meltano.com/guide/configuration#specifying-environment-variables and then flag this sort of change for 3.0 since it's breaking.

aaronsteers:
@tayloramurphy and @WillDaSilva - I always defer to [this site](https://clig.dev/#configuration) for reference:

> image

I agree with this order, and I think if a user specifally sets `--environment=prod`, that should override the terminal context and `.env` file.

We should probably get this into our own handbook too.

tayloramurphy:
@aaronsteers to be clear then, what you're proposing based on that site is:

```
- command line flag # highest
- environment level plugin env
- schedule level env
- job level env
- root level plugin env 
- environment level env 
- root level env
- terminal 
- .env file  # lowest
```

I'm fine with that as well - I just want us to commit to something clearly and then make progress on making our stated target reality.

WillDaSilva:
@tayloramurphy @aaronsteers @edgarrmondragon I've updated the precedence and added a test for it here: ebdb940ddecf34254c02b24573cfab165a88f971

aaronsteers:
> @tayloramurphy and @WillDaSilva - I always defer to [this site](https://clig.dev/#configuration) for reference:
> 
> > image

@WillDaSilva - Side-by-side, your code in https://github.com/meltano/meltano/commit/ebdb940ddecf34254c02b24573cfab165a88f971 comment matches the above perfectly, I think:

> Precedence is:
> 1. The `--environment` CLI option
> 2. Env var from the shell
> 3. Env var from `.env`
> 4. Default environment from `meltano.yml`
> 5. `None`

----------

> @aaronsteers to be clear then, what you're proposing based on that site is:
> 
> ```
> - command line flag # highest
> - environment level plugin env
> - schedule level env
> - job level env
> - root level plugin env 
> - environment level env 
> - root level env
> - terminal 
> - .env file  # lowest
> ```

I need to double-check some of these before I can say yes to the above. For this PR though, we are just handling the "--environment" CLI arg and `MELTANO_ENVIRONMENT`, which I think are not applicable at the other levels.

tayloramurphy:
@aaronsteers happy to spin it out. We've talked a lot about this and I feel like we keep losing our previous decisions on what the target state is supposed to be. This issue:

* https://github.com/meltano/meltano/issues/6386

Has a Miro board along with several threads where we hashed out the order of precedence. I will need help to get everyone on the same page with this cc @seth-meltano 

----

For this PR though I think the proposed precedence looks good and I'll approve.

tayloramurphy (approved):

docs/src/_getting-started/index.md
```diff
@@ -127,7 +127,7 @@ This will allow you to use [`git diff`](https://git-scm.com/docs/git-diff) to ea
 
 ## View and Activate Your Environments
 
-As part of creating your Meltano project, we automatically added your first [environments](/concepts/environments) called `dev`, `staging` and `prod`. This allows you to define configurations specific to the environment in which you're running your project. Theres also a [`default_environment`](https://docs.meltano.com/concepts/environments#default-environments) setting in the `meltano.yml` that get automatically set to `dev`, you can list and change the active environment using:
+As part of creating your Meltano project, we automatically added your first [environments](/concepts/environments) called `dev`, `staging` and `prod`. This allows you to define configurations specific to the environment in which you're running your project. There's also a [`default_environment`](https://docs.meltano.com/concepts/environments#default-environments) setting in the `meltano.yml` that get automatically set to `dev`, you can list and change the active environment using:
```
aaronsteers, line 130: Do I understand correctly that this PR doesn't impact how `default_environment` does or does not get applied? For instance, `default_environment` would today be ignored for `meltano config` but respected for `meltano run`.

My observation is that this part looks unchanged in the PR (which is a good thing for preserving those behaviors) and I just want to confirm that expectation is correct.

@WillDaSilva - Do I understand that correctly?

WillDaSilva: @aaronsteers That is correct. This PR has no impact on the application of `default_environment`.

aaronsteers: Great. Thanks for confirming.

aaronsteers (approved):
🙌

PR #7148, opened 2023-01-05
feat: Meltano manifest

WillDaSilva:
Todo:
- [x] Unit tests (aside: maybe we should run coverage on Meltano during the integration tests, and include that data our coverage stats - I'll open another issue for this if there's any interest)
- [x] Misc schema updates
- [x] Maybe some easy way to mute the schema warnings?

Closes #6876
Closes #7147

PR #7287, opened 2023-02-08
feat: Add support for `annotations` to provide documentation and 3rd-party configurations in `meltano.yml`

### LINKED ISSUE #7215: feature: annotations in `meltano.yml` at depth
Follow-up to:
- https://github.com/meltano/meltano/issues/7092
- https://github.com/meltano/meltano/pull/7094
- https://github.com/meltano/meltano/discussions/7053

---

While implementing #7092 we encountered difficulty with supporting `annotations` fields within `meltano.yml` at depth. This was because many places within the code had to handle/ignore the new field. To avoid a messy sprawling implementation, we opted to only support annotations at the top-level of `meltano.yml` in that iteration.

This issue is for the original goal of #7092, which is to support the `annotations` field in as many places as it makes sense (detailed in the original issue).

This can likely be achieved by adding a step to loading project files which strips out all annotations. Annotations aren't meant to be handled by `meltano` itself, so having `meltano` receive the project files without them seems ideal. If we later decide that we want to e.g. pass along annotations to a plugin, then we should probably try to implement that by leveraging the [manifest](#7148), which will preserve the annotations (n.b. the current implementation of manifest generation does its own parsing of the Meltano project files, so this step of removing annotations should not affect it).

The challenge will be writing them back to the project files. As we strip them out we'll need to record where they came from (the file, and a location within) so that we can re-insert them afterwards.

WillDaSilva:
Annotations can now go within any dictionary in `meltano.yml` except for the `env` dictionaries, and the `plugins` dictionary. The former because it wouldn't make sense, and the latter because of technical difficulties with supporting annotations in that location in particular (because `ProjectFiles` has special handling for it).

Closes #7215


docs/example-library/meltano-annotations/ending-meltano.yml
```diff
@@ -13,7 +13,11 @@ plugins:
   - name: tap-gitlab
     variant: meltanolabs
     pip_url: git+https://github.com/MeltanoLabs/tap-gitlab.git
+    annotations:
+      example-annotation-plugins-extractors-tap-gitlab: {}
     config:
+      annotations:
+        example-annotation-plugins-extractors-tap-gitlab-config: {}
```
aaronsteers, line 20: (1 of 2): I would like to keep `annotations` out of `config`, here and below. Because a `config` value can be anything, it's not clear that `annotations` isn't the name of a setting. Also - I can't imagine any use case where the `config` level is interesting to add annotations, rather than annotating one layer above (on the plugin in this case).

What do you think?

docs/example-library/meltano-annotations/ending-meltano.yml
```diff
@@ -22,15 +26,37 @@ plugins:
...
     config:
       do_timestamp_file: false
 environments:
 - name: dev
+  annotations:
+    example-annotation-environments-dev: {}
+  config:
+    annotations:
+      example-annotation-environments-dev-config: {}
+    plugins:
```
aaronsteers, line 40: Similar to above, I would rather not have `annotations` supported under `config:` because I'd like to reserve that for settings' names. All use cases that could be applied here could presumably be applied at the parent level, aka on the environment declaration in this case.

aaronsteers (changes requested):

aaronsteers:
@WillDaSilva - This looks really good. Just one request above re: excluding from `config:` blocks. 

Thanks for putting this together!

docs/example-library/meltano-annotations/ending-meltano.yml
WillDaSilva: Agreed. Marking this conversation as resolved to keep this matter limited to a the single thread above.

docs/example-library/meltano-annotations/ending-meltano.yml
WillDaSilva: That makes sense. I can remove the annotations from the example, and from the schema for the configs, but unfortunately the current implementation would still remove them from the merged project files.

To avoid this I think we'll need something like what has been implemented in #7148 under `meltano.core.manifest.jsonschema`, which can tell us every location a particular reference within a jsonschema appears. In this case we'd limit `ProjectFiles.extract_annotations` to only extract those which are located where a ref to `#/$defs/annotations` appears in the schema.

tayloramurphy: I'll echo what AJ said - annotations at the config level could be very confusing about whether it's a configurable setting or not.

src/meltano/core/project_files.py
```diff
@@ -360,3 +386,68 @@ def _copy_yaml_attributes(  # noqa: WPS210
...
+    def extract_annotations(cls, data: CommentedMap | CommentedSeq) -> None:
+        """Extract annotations, and store them for later re-insertion.
+
+        Args:
+            data: The data from which to extract all "annotations" keys. It
+                will be modified in-place.
+        """
+        extracted = list(cls._extract_annotations(data=data, path=()))
+        if not cls._annotations:
+            cls._annotations = extracted
```
aaronsteers, line 400: @WillDaSilva - Looking more at internals here now that #6876 has shipped.... 🎉 

I'm sure you looked into this already and I think it would be helpful for me to better understand the context - and hopefully this is not a dumb question. 😅 Can you help me understand what was blocking the treatment of annotations as a property of the constructed classes which is simply ignored by them, but still serialized and deserialized in the same way as other properties?

I see we are extracting and restoring annotations but I guess I am missing context on why they cannot travel with the rest of the instantiated objects.

WillDaSilva: @aaronsteers That was the first thing I tried, but I ruled it out because the way we build `Canonical` instances would require we deal with `annotations` at every level in the class hierarchy. Having every `Canoncial` subclass implement special handling for an attribute that `meltano` shouldn't ever touch was too messy:
- https://github.com/meltano/meltano/pull/7094

That said, I can probably give `Canonical` a metaclass that handles this before `__init__` gets called. It's invisible magic, which could trip someone up (possibly myself) in the future, but at least it contains the logic that handles annotations to a single location, and saves us from having to think about them in most cases.

I'll try that approach again now, and see what problem I run into. I wouldn't want to settle for a sub-par approach when a significantly better approach is possible.

WillDaSilva: Thanks for asking about this. A second attempt at my first approach (this time using a metaclass) worked, and it's much simpler! I've implemented it in 86ef0e0.

docs/example-library/meltano-annotations/ending-meltano.yml
WillDaSilva: Resolved in 86ef0e0

tayloramurphy:
@WillDaSilva can you add a docs update to https://docs.meltano.com/concepts/project#annotations indicating where annotations is/isn't allowed?

WillDaSilva:
> can you add a docs update to [docs.meltano.com/concepts/project#annotations](https://docs.meltano.com/concepts/project#annotations) indicating where annotations is/isn't allowed?

@tayloramurphy Thanks for reminding me to do that. Updated in 0ab36ec

tayloramurphy (approved):
Thank you for the addition @WillDaSilva !

edgarrmondragon (approved):

docs/src/_concepts/project.md
```diff
@@ -312,6 +312,17 @@ annotations:
 
 The core Meltano library and CLI never access the `annotations` field. To access it, one must read `meltano.yml`. Nothing within an `annotations` field should be thought of as part of Meltano's own configuration - it is merely extra data that Meltano permits within its configuration files.
 
+Annotations are supported in the following locations within `meltano.yml`:
+
+- At the top level
+- In a job definition
+- In a schedule definition
+- In an environments definition
```
aaronsteers, line 320: Sorry, total nit 🤓 

```suggestion
- In an environment definition
```

WillDaSilva: Thanks!

aaronsteers (commented):
I didn't do a deep dive into the code, but I'm approving after a review of the docs, and based on the generated examples from pre-refactor.

@WillDaSilva - One think is just to confirm that you were able to regenerate the sample files after metaclass update, and that they passed a manual spot check similar to pre-metaclass refactor.

Assuming all looks good on the manual inspection, I'd say ship it! 🚢

aaronsteers (approved):

WillDaSilva:
> One think is just to confirm that you were able to regenerate the sample files after metaclass update, and that they passed a manual spot check similar to pre-metaclass refactor.

@aaronsteers What sample file are you referring to? The integration test?

aaronsteers:
> > One think is just to confirm that you were able to regenerate the sample files after metaclass update, and that they passed a manual spot check similar to pre-metaclass refactor.
> 
> @aaronsteers What sample file are you referring to? The integration test?

Sorry for the confusion. Was referring to these sample files generated from squared and shared in slack: https://meltano.slack.com/archives/C03GKHWS0HM/p1676061575297269?thread_ts=1675995126.993639&cid=C03GKHWS0HM

WillDaSilva:
To enable a clearer review, I'm going to open a follow-up PR to address this issue:
- https://github.com/meltano/meltano/issues/7330

PR #7294, opened 2023-02-09
fix: Avoid casting settings before env var expansion

### LINKED ISSUE #7293: bug: Cannot set integer settings with environment variables in meltano.yml
### Meltano Version

2.15.2

### Python Version

3.8

### Bug scope

### Operating System

Linux/Docker (using the image `meltano:v2.15.2-python3.8` pushed a few hours ago)

### Description

I have been using an environment variable to configure the port of my mysql tap with something like the below in my meltano.yml.

```yaml
plugins:
  extractors:
    - name: tap-mysql
      variant: transferwise
      pip_url: pipelinewise-tap-mysql
      config:
        host: '${MY_TAP_HOST}'
        port: '${MY_TAP_PORT}'
```

This use to work but appears to be broken now, presumably due to https://github.com/meltano/meltano/pull/7232. Now you'll get an error ending with

```
│ /venv/lib/python3.8/site-packages/meltano/core/setting_definition.py:374 in cast_value           │
│                                                                                                  │
│   371 │   │   │   if self.kind == SettingKind.BOOLEAN:                                           │
│   372 │   │   │   │   return utils.truthy(value)                                                 │
│   373 │   │   │   elif self.kind == SettingKind.INTEGER:                                         │
│ ❱ 374 │   │   │   │   return int(value)                                                          │
│   375 │   │   │   elif self.kind == SettingKind.OBJECT:                                          │
│   376 │   │   │   │   value = json.loads(value)                                                  │
│   377 │   │   │   │   if not isinstance(value, dict):                                            │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
ValueError: invalid literal for int() with base 10: '${MY_TAP_PORT}'
```

I realize there are other ways I could accomplish this, e.g. setting TAP_MYSQL_PORT and leave this out of meltano.yml. I mostly want to confirm if this is intended behavior and see if there are any workarounds that allow me to express something similar within meltano.yml.

### Code

_No response_

WillDaSilva:
This fixes a regression in #7232. In that PR logic was added to cast settings within the setting store. This was needed because some settings were not retrieved through a settings service in order to avoid infinite regression (i.e. the `get` method in a setting service cannot generally call itself).

A consequence of those changes was that now we are casting setting values before env var expansion has occurred. To address the regression, this PR makes the casting logic from #7232 disabled by default, and it is only enabled when accessing settings from within the `get` methods of a setting service (where we cannot rely on the regular setting casting).

A cleaner approach would be preferable, but this should be sufficient to address the regression while preserving the bug fix that #7232 implemented.

Closes #7293

### DIFF
--- a/src/meltano/core/settings_service.py
+++ b/src/meltano/core/settings_service.py
@@ -351,7 +351,8 @@ class SettingsService(ABC):  # noqa: WPS214
         # Can't do conventional SettingsService.feature_flag call to check;
         # it would result in circular dependency
         strict_env_var_mode, _ = source.manager(self.project_settings_service).get(
-            f"{FEATURE_FLAG_PREFIX}.{FeatureFlags.STRICT_ENV_VAR_MODE}"
+            f"{FEATURE_FLAG_PREFIX}.{FeatureFlags.STRICT_ENV_VAR_MODE}",
+            cast_value=True,
         )
         if expand_env_vars and metadata.get("expandable", False):
             metadata["expandable"] = False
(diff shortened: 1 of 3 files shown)


WillDaSilva:
The new test may be flaky: https://github.com/meltano/meltano/actions/runs/4131188538/jobs/7138602617

I have yet to reproduce a failure like this.

edgarrmondragon (approved):

kgpayne (approved):

cjohnhanson (approved):

PR #7299, opened 2023-02-09
fix: Respect `requests`/`curl` custom Certificate Authorities env vars

### LINKED ISSUE #7298: bug: SSL: CERTIFICATE_VERIFY_FAILED - Meltano >= 2.14.0 hub commands stop working with custom CA Root Certificates
### Meltano Version

2.14.0

### Python Version

3.10

### Bug scope

Other

### Operating System

Linux

### Description

Starting with v2.14.0, meltano hub interactions have switched to manual request preparation with `Session.prepare_request`, which results in the problem where the standard `requests` and `curl` environment variables are now no longer being honoured unless the environment variables are explicitly merged into the requests session again during the explicit request preparation.

https://github.com/meltano/meltano/pull/7188
https://requests.readthedocs.io/en/latest/user/advanced/#prepared-requests

The impacted environment variables are:

- `HTTP_PROXY`
- `HTTPS_PROXY`
- `REQUESTS_CA_BUNDLE`
- `CURL_CA_BUNDLE`

This breaks `meltano discover` and `meltano add` for users who must use a custom Certificate Authority Root Store for whatever reasons:
1.  using Zero Trust Network Proxy with SSL Inspection. 
2. using http debugging (e.g. `mitmproxy`)
3.  people who may run their own private "hubs" with private PKIs

In the case of explicitly configured (non-transparent) proxy set-ups, the HTTP communication with the hub also no longer flows through the manually configured proxy after this change, due to `HTTP_PROXY` and `HTTPS_PROXY` variables being ignored now as well.

To reproduce:

1. Install meltano v2.14.0 or higher
2. Install an SSL intercepting proxy e.g. `pip install mitmproxy`
3. Launch `mitmproxy` and install the mitmproxy Root CA cert https://docs.mitmproxy.org/stable/concepts-certificates/
4. Configure `requests` to use the mitmproxy Root CA cert: `export REQUESTS_CA_BUNDLE=${HOME}/.mitmproxy/mitmproxy-ca-cert.pem`
5. Set HTTP_PROXY and HTTPS_PROXY vars: `export http_proxy=http://localhost:8080` `export https_proxy=http://localhost:8080/`
4. Run `meltano discover` 

## Expected result (behaviour with meltano 2.13.0 and earlier)

- `meltano discover` succeeds and prints all available plugins from the hub
- all HTTP traffic to the hub is flowing throw `mitmproxy` correctly, honouring the mitmproxy custom Root CA cert

## Actual result (behaviour with meltano 2.14.0 and higher)

`meltano discover` fails in two ways:

1. HTTP request to the hub are no longer flowing through the `mitmproxy` due to `HTTP_PROXY` and `HTTPS_PROXY` variables being missing from the request session
2. Even when somehow managing to steer the HTTP requests to the hub through the proxy (e.g. with transparently proxying with iptables https://docs.mitmproxy.org/stable/howto-transparent-vms/), the request still fails with `[SSL: CERTIFICATE_VERIFY_FAILED]` due to `REQUESTS_CA_BUNDLE` env var not being merged into the requests session.

Related: https://gitlab.com/meltano/sdk/-/merge_requests/256

### Code

_No response_

edgarrmondragon:
Related:

- Closes https://github.com/meltano/meltano/issues/7298
- Fixes a regression introduced in https://github.com/meltano/meltano/pull/7188

### DIFF
--- a/src/meltano/core/hub/client.py
+++ b/src/meltano/core/hub/client.py
@@ -229,8 +229,17 @@ class MeltanoHubService(PluginRepository):  # noqa: WPS214
         Raises:
             HubConnectionError: If the Hub API could not be reached.
         """
+        prep = self._build_request("GET", url)
+        settings = self.session.merge_environment_settings(
+            prep.url,
+            None,
+            None,
+            None,
+            None,
+        )
+
         try:
-            return self.session.send(self._build_request("GET", url))
+            return self.session.send(prep, **settings)
         except requests.exceptions.ConnectionError as connection_err:
             raise HubConnectionError("Could not reach Meltano Hub.") from connection_err
 
--- a/tests/meltano/core/hub/test_meltano_hub_service.py
+++ b/tests/meltano/core/hub/test_meltano_hub_service.py
@@ -6,6 +6,7 @@ import click
 import mock
 import pytest
 from requests import HTTPError, Response
+from requests.adapters import BaseAdapter
 
 from meltano.cli import cli
 from meltano.cli.discovery import discover
@@ -136,3 +137,24 @@ class TestMeltanoHubService:
             get_context.return_value = None
             request = subject._build_request("GET", "https://example.com")
             assert "X-Meltano-Command" not in request.headers
+
+    def test_custom_ca(self, project, monkeypatch):
+        send_kwargs = {}
+
+        class _Adapter(BaseAdapter):
+            def send(self, request, **kwargs):
+                nonlocal send_kwargs
+                send_kwargs = kwargs
+
+                response = Response()
+                response._content = b'{"name": "tap-mock", "namespace": "tap_mock"}'
+                response.status_code = 200
+                return response
+
+        mock_url = "hub://meltano"
+        hub = MeltanoHubService(project)
+        hub.session.mount(mock_url, _Adapter())
+
+        monkeypatch.setenv("REQUESTS_CA_BUNDLE", "/path/to/ca.pem")
+        hub._get(mock_url)
+        assert send_kwargs["verify"] == "/path/to/ca.pem"


WillDaSilva (commented):
Do you think it would be worth adding a test to ensure such a regression does not occur again?

edgarrmondragon:
> Do you think it would be worth adding a test to ensure such a regression does not occur again?

@WillDaSilva yeah, the requests API doesn't make it easy but I'll try to figure out something

WillDaSilva (approved):

cwegener:
> > Do you think it would be worth adding a test to ensure such a regression does not occur again?
> 
> @WillDaSilva yeah, the requests API doesn't make it easy but I'll try to figure out something

If it can be an integration/e2e test, somebody did start writing a pytest fixture to spin up mitmproxy instances ... :grin: 
Not sure how the built-in Root CA cert function would work in this context though.

https://github.com/sangoma/pytest-mitmproxy/blob/5d657ac0b58b67fc063091b26fe180a5fe2e1a7a/pytest_mitmproxy.py

edgarrmondragon:
> > > Do you think it would be worth adding a test to ensure such a regression does not occur again?
> > 
> > 
> > @WillDaSilva yeah, the requests API doesn't make it easy but I'll try to figure out something
> 
> If it can be an integration/e2e test, somebody did start writing a pytest fixture to spin up mitmproxy instances ... 😁 Not sure how the built-in Root CA cert function would work in this context though.
> 
> [sangoma/pytest-mitmproxy@`5d657ac`/pytest_mitmproxy.py](https://github.com/sangoma/pytest-mitmproxy/blob/5d657ac0b58b67fc063091b26fe180a5fe2e1a7a/pytest_mitmproxy.py)

I ended up just using a custom adapter: [meltano/meltano@`9c99938` (#7299)](https://github.com/meltano/meltano/pull/7299/commits/9c99938e3ae7aacf3c926fa12053f0a469e6e1d9)

But, yeah we'll probably need some fixture like that to test no network calls are made.

PR #7301, opened 2023-02-10
docs: refresh to prefer 'run' and 'invoke' over 'elt', and 'utilities' over 'transformers' and 'orchestrators'

pnadolny13:
I was primarily trying to update references to transformers/orchestrators and elt so we recommend utilities and `run` but ended up refreshing a bunch of other stuff too:
- Update transformers + orchestrators to utilities. I think I made it clear that everything is still supported but that we're transitioning to calling everything utilities
- Prefer meltano run over meltano elt. Our main examples used elt so we were funneling new users towards that command, prefer run moving forward and elt is for existing users or specific use cases where they need it right now.
- Replicate data page (integrations)
  - I added examples for what settings metadata should look like for replication method and id. Personally I always have trouble finding examples and end up back in the getting started guide.
  - Reorganized to move elt specific stuff down further on the page

-  I also cleaned up anything I found along the way including:
    - rewording some stuff related to state merging
    - a few typos
    - updated outdated container name and version references


edgarrmondragon (approved):

docs/src/_guide/configuration.md
```diff
@@ -25,13 +25,13 @@ Meltano itself can be configured as well. To learn more, refer to the [Settings
...
    - Note that configuration for Meltano itself is stored at the root level of `meltano.yml`.
    - You can use [Meltano Environments](/concepts/environments) to manage different configurations depending on your testing and deployment strategy. If values for plugin settings are provided in both the top-level plugin configuration _and_ the environment-level plugin configuration, the value at the environment level will take precedence.
-3. **Your project's [**system database**](/concepts/project#system-database)**, which (among other things) stores configuration set using [`meltano config <plugin> set`](/reference/command-line-interface#config) or [the UI](/reference/ui) when the project is [deployed as read-only](/reference/settings#project-readonly).
+3. **Your project's [state backend](/concepts/state_backends)**, which (among other things) stores configuration set using [`meltano config <plugin> set`](/reference/command-line-interface#config) or [the UI](/reference/ui) when the project is [deployed as read-only](/reference/settings#project-readonly).
```
aaronsteers, line 34: The header here is "To determine the values of settings, Meltano will look in 4 main places" so I think this is not actually correct and should be reverted. While state backends can replace systemdb for state storage, they don't affect settings storage at this time.

aaronsteers (dismissed):
Fantastic updates, @pnadolny13. I reviewed the full text and noted just one suggested change.

docs/src/_guide/configuration.md
pnadolny13: @aaronsteers ahh thanks for the review! I hadn't consider that.

docs/src/_concepts/plugins.md
```diff
@@ -688,16 +696,26 @@ export TAP_GITLAB__VARS='{"schema": "{{ env_var(''DBT_SOURCE_SCHEMA'') }}"}'
...
-Meltano supports [Apache Airflow](https://airflow.apache.org/) out of the box, but can be used with any tool capable of reading the output of [`meltano schedule list --format=json`](/reference/command-line-interface#schedule) and executing each pipeline's [`meltano elt`](/reference/command-line-interface#elt) command on a schedule.
+Meltano supports [Apache Airflow](https://airflow.apache.org/) out of the box, but can be used with any tool capable of reading the output of [`meltano schedule list --format=json`](/reference/command-line-interface#schedule) and executing each pipeline's [`meltano run`](/reference/command-line-interface#run) command on a schedule.
 
 When the `airflow` orchestrator is added to your project using [`meltano add`](/reference/command-line-interface#add),
 its related [file bundle](#file-bundles) will automatically be added as well.
 
 ### Transformers
 
-Transformers are [pip packages](https://pip.pypa.io/en/stable/) used by [`meltano elt`](/reference/command-line-interface#elt) as part of [data transformation](/guide/transformation).
+<div class="notification is-warning">
```
tayloramurphy, line 713: Thoughts on included the same warning for files bundles? Long-term I think we'd want to have those all EDK-based even if all it does is inject some files.

docs/src/_concepts/project.md
```diff
@@ -190,28 +190,25 @@ extractors:
 Plugin [commands](/reference/command-line-interface#commands) are defined by the `commands` property. The keys are the name of the command and the values are the arguments to be passed to the plugin executable. These can contain dynamic references to [configuration](/guide/configuration) using the [Environment variable form](/guide/configuration#environment-variables) of the configuration option.
 
 ```yaml{3-7}
```
tayloramurphy, line 192:
```suggestion
```yaml
```
row highlighting doesn't work with this style unfortunately. And having it makes it not highlight appropriately!

tayloramurphy (changes requested):

docs/src/_concepts/plugins.md
pnadolny13: @tayloramurphy I didnt know what our long term vision was for those so I just left them. I'd have to think about that a little more 🤔 .

tayloramurphy: That's fine for now. Can you open an internal discussion on that when you get a chance and we can get alignment on it?

tayloramurphy (approved):

docs/src/_concepts/plugins.md
pnadolny13: @tayloramurphy I created https://github.com/meltano/internal-general/discussions/469 to discuss it.

aaronsteers (approved):
👍

PR #7315, opened 2023-02-15
docs: Escaping `$` in `meltano config <plugin> set`

### LINKED ISSUE #3065: docs: Setting config value with $ in the value fails due to Bash Expansion
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3137

Originally created by @vischous on 2022-01-04 22:31:32

---

Steps to reproduce
1.  `meltano config target-snowflake set password "@$a"` 

env file contains

`TARGET_SNOWFLAKE_PASSWORD="@"` 

Should be set to `@$a`

--

Two reasons to do this in my opinion
1. Security - bash history shouldn't have your passwords inside of them
2. To prevent bash from expanding the values

I think we could do something like wait for STDIN data to come through to fix this (just like what happens if you run `cat` )

edgarrmondragon:
Closes https://github.com/meltano/meltano/issues/3065

### DIFF
--- a/docs/src/_reference/command-line-interface.md
+++ b/docs/src/_reference/command-line-interface.md
@@ -239,6 +239,14 @@ If multiple plugins share the same name, you can provide an additional `--plugin
 meltano config --plugin-type=<type> <plugin> ...
 ```
 
+When setting a config value that contains the character `$`, you can avoid expansion by
+escaping it with `\` or using single quotes:
+
+```bash
+meltano config <plugin> set <name> "@\$a"
+meltano config <plugin> set <name> '@$a'
+```
+
 #### Nested properties
 
 Nested properties can be set (and unset) by specifying a list of property names:


docs/src/_reference/command-line-interface.md
```diff
@@ -239,6 +239,14 @@ If multiple plugins share the same name, you can provide an additional `--plugin
 meltano config --plugin-type=<type> <plugin> ...
 ```
 
+When setting a config value that contains the character `$`, you can avoid expansion by
+escaping it with `\` or using single quotes:
+
+```bash
+meltano config <plugin> set <name> "@\$a"
+meltano config <plugin> set <name> '@$a'
```
WillDaSilva, line 247: What is the `@` for?

edgarrmondragon: Just a placeholder for any special character that's not expanded by the shell

kgpayne (approved):

WillDaSilva (approved):

tayloramurphy (approved):

Issue #7306, opened 2023-02-13
Insert settings and config into manifest directly, rather than derived env vars


WillDaSilva:
Context:
- https://github.com/meltano/meltano/pull/7148#issuecomment-1428445422
- https://github.com/meltano/meltano/pull/7148#issuecomment-1428660571

Environment variables like `MELTANO_<setting name>` or `TAP_ABC_<setting name>` should be provided at run-time, rather than contained in the manifest. Instead, the manifest should contain the settings/config directly, with all default values included when the project files do not specify something else.

WillDaSilva:
Some findings:
- We aren't yet properly handling the merging of plugin lockfiles into plugins which use `inherit_from`.
- Sometimes collisions occur when dealing with plugin lockfiles because some setting keys will be dot-expanded, while others will have `.` in their name. I'm not yet sure what the root cause is. Relates to https://github.com/meltano/meltano/issues/7270

WillDaSilva:
@aaronsteers I've created an issue to separately address the aforementioned issue around setting key dot-expansion:
- https://github.com/meltano/meltano/issues/7319

aaronsteers:
@WillDaSilva - Is it okay to treat this as a relatively low priority for now? While the large list of env vars may be redundant, I don't see immediate issues or blockers being raised from this, and I'd be okay to keep low pri for now.

Wdyt?

aaronsteers:
I think part of the framing for me would be:

1. Does the manifest accurately carry the state of the project? (I think yes.)
2. In reconstructing from the manifest, does Meltano have sufficient context to properly override declared env vars with runtime env vars? (I think also yes.)

I might be misreading or misinterpreting this point:

> ...the manifest should contain the settings/config directly, with all default values included when the project files do not specify something else

but I'd be okay if _unset_ project-level default settings and feature flags were omitted from the manifest altogether. A good example of settings we'd prefer to omit if unset would be deprecated settings at https://docs.meltano.com/reference/ui#meltano-ui-server - such as `ui.bind_host`, `ui.bind_port`, etc.  

As I'm writing this, I also recognize that it could be a lot of effort to distinguish between 'set' and 'unset' values - hence again why I think it's okay to not prioritize highly right now if the existing resolutions will accomplish our core objectives without any major defects.

PR #7321, opened 2023-02-15
chore: Disable `meltano compile` command

WillDaSilva:
Context:
- https://github.com/meltano/meltano/issues/7306#issuecomment-1432086410

To avoid anyone depending on the manifest feature before we are confident that we can avoid making breaking changes to it, this PR disables the `meltano compile` command. Manifests can continue to be used internally, but once we allow users to generate them via the CLI, and depend on the data they contain, we will have to be careful about what we change.


WillDaSilva:
Closing following a discussion with @aaronsteers. Another PR will be opened shortly to make it clear that the `compile` command is experimental.

PR #7322, opened 2023-02-15
docs: State that the `meltano compile` command is in beta

WillDaSilva:
Supersedes:
- https://github.com/meltano/meltano/pull/7321

### DIFF
--- a/docs/src/_reference/command-line-interface.md
+++ b/docs/src/_reference/command-line-interface.md
@@ -122,6 +122,10 @@ The `add` command does not run relative to a [Meltano Environment](https://docs.
 
 ## `compile`
 
+<div class="notification is-warning">
+    <p>The compile command is currently experimental, and subject to change without corresponding semantic version updates.</p>
+</div>
+
 ### How to use
 
 Generally, the `compile` command need not be executed manually, as it will be run automatically by Meltano as-needed.
--- a/docs/src/_reference/command-line-interface.md
+++ b/docs/src/_reference/command-line-interface.md
@@ -123,7 +123,7 @@ The `add` command does not run relative to a [Meltano Environment](https://docs.
 ## `compile`
 
 <div class="notification is-warning">
-    <p>The compile command is currently experimental, and subject to change without corresponding semantic version updates.</p>
+    <p><a href="https://github.com/meltano/meltano/discussions/7323">The compile command is currently in beta</a>, and subject to change without corresponding semantic version updates.</p>
 </div>
 
 ### How to use
--- a/src/meltano/cli/compile.py
+++ b/src/meltano/cli/compile.py
@@ -18,9 +18,7 @@ from meltano.core.tracking import Tracker
 from meltano.core.tracking.contexts import CliEvent
 
 
-@cli.command(
-    cls=InstrumentedCmd, short_help="Compile a Meltano manifest. (experimental)"
-)
+@cli.command(cls=InstrumentedCmd, short_help="Compile a Meltano manifest. (beta)")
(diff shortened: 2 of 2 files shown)


aaronsteers (approved):
Nice 👍

aaronsteers:
Totally optional, but @WillDaSilva , what do you think of linking to a github discussion to collect feedback and share back updates?

WillDaSilva:
> Totally optional, but @WillDaSilva , what do you think of linking to a github discussion to collect feedback and share back updates?

@aaronsteers A GitHub discussion about what? This PR before we merge it (and whether we should require the `experimental` setting to be `true`), or something else related to the broader manifest topic?

aaronsteers:
> > Totally optional, but @WillDaSilva , what do you think of linking to a github discussion to collect feedback and share back updates?
> 
> @aaronsteers A GitHub discussion about what? This PR before we merge it (and whether we should require the `experimental` setting to be `true`), or something else related to the broader manifest topic?

Here you go:

- https://github.com/meltano/meltano/discussions/7323

docs/src/_reference/command-line-interface.md
```diff
@@ -122,6 +122,10 @@ The `add` command does not run relative to a [Meltano Environment](https://docs.
 
 ## `compile`
 
+<div class="notification is-warning">
+    <p>The compile command is currently experimental, and subject to change without corresponding semantic version updates.</p>
```
aaronsteers, line 126: We could add more copy or just link under the work "experimental".

```suggestion
    <p>The compile command is currently [experimental](https://github.com/meltano/meltano/discussions/7323), and subject to change without corresponding semantic version updates.</p>
```

tayloramurphy: @aaronsteers thoughts on calling it something other than _experimental_? An experiment to me indicates something we have a 50/50 shot on rolling back. Compile is something we'll very likely keep and continue to iterate on. 

Perhaps we could say the structure is not yet stable? Or is there a better phrasing here?

WillDaSilva: @tayloramurphy Thoughts on calling it a "feature preview"? Or we could say that the feature is in beta.

tayloramurphy: @WillDaSilva I guess Beta makes sense - we could also just have a 0.X.X semver on it generally with no promise on when the manifest will be 1.0, right?

WillDaSilva: I don't think we should have versions for features. Let's hope that Meltano itself never becomes so complicated as to warrant that.

tayloramurphy: @WillDaSilva sorry, not for the feature itself, but the schema of the generated file could be versioned right? https://docs.getdbt.com/reference/artifacts/manifest-json dbt-core v1.3 produces schema v7 while 1.4 produces v8. Something like that could happen, right?

WillDaSilva: @tayloramurphy The manifest and `meltano.yml` share a schema and version number. The `meltano.yml` version number is hard-coded as `1`. Ideally we won't add a manifest-specific version number, as that would mean the schema for the manifest and `meltano.yml` differ.

WillDaSilva: Resolved in b38571b

tayloramurphy: Ah ok, thanks for that context.

tayloramurphy (approved):

PR #7324, opened 2023-02-16
feat: Meltano Cloud Run

edgarrmondragon:
Related:
- Closes https://github.com/meltano/infra/issues/270

TODO:
- [x] Add tests
- [x] Confirm API spec


WillDaSilva (commented):
Looks good to me, but I'll hold off on actually approving it till it's ready to go.

src/cloud-cli/meltano/cloud/client/__init__.py
```diff
@@ -0,0 +1,164 @@
...
+        Returns:
+            The response JSON.
+
+        Raises:
+            MeltanoCloudError: If the response status is not OK.
+        """
+        url = f"{self.api_url}{path}"
+        headers = {
+            "Content-Type": "application/json",
+            "User-Agent": f"Meltano Cloud CLI/v{version('meltano')}",
```
WillDaSilva, line 121: Caching the result of `version('meltano')` is probably worth doing so that we aren't reading it from disk for each request.

edgarrmondragon: Maybe, although there's currently only 1 request per invocation, so I'm not sure we'd be winning a lot from caching. Or I'm probably missing something 😅

WillDaSilva: Nope. That's my bad.

edgarrmondragon:
@cjohnhanson In retrospect, the use of `aiohttp` here feels like overkill so I'm happy to fallback to `requests` if you prefer. Wdyt?

src/cloud-cli/meltano/cloud/cli/run.py
```diff
@@ -0,0 +1,94 @@
...
+    Args:
+        job_or_schedule: The job or schedule identifier.
+        environment: The environment to run in.
+        project_id: The project identifier.
+    """
+    click.echo("Running a Meltano project in Meltano Cloud.")
+
+    api_key = get_env_var("MELTANO_RUNNER_API_KEY")
+    runner_secret = get_env_var("MELTANO_RUNNER_SECRET")
+    tenant_resource_key = get_env_var("MELTANO_TENANT_RESOURCE_KEY")
```
cjohnhanson, line 82: Could we use these env vars instead? My plan for login is to just incorporate these changes after they merge; the config I wrote uses `MELTANO_CLOUD_` as the env var prefix and we're using `organization_id` as the user-facing name for tenant resource key (https://meltano.slack.com/archives/C03GKHWS0HM/p1676062063363619)
```suggestion
    api_key = get_env_var("MELTANO_CLOUD_RUNNER_API_KEY")
    runner_secret = get_env_var("MELTANO_CLOUD_RUNNER_SECRET")
    tenant_resource_key = get_env_var("MELTANO_CLOUD_ORGANIZATION_ID")
```

src/cloud-cli/meltano/cloud/cli/run.py
```diff
@@ -0,0 +1,94 @@
...
+        msg = f"Environment variable {name} is not set."
+        raise click.UsageError(msg)
+
+    return value
+
+
+async def run_project(
+    api_key: str,
+    runner_secret: str,
+    tenant_resource_key: str,
```
cjohnhanson, line 37:
```suggestion
    organization_id: str,
```

src/cloud-cli/meltano/cloud/cli/run.py
```diff
@@ -0,0 +1,94 @@
...
+    project_id: str,
+    environment: str,
+    job_or_schedule: str,
+) -> dict:
+    """Run a project in Meltano Cloud.
+
+    Args:
+        api_key: The API key to use for authentication.
+        runner_secret: The runner secret to use for authentication.
+        tenant_resource_key: The tenant resource key.
```
cjohnhanson, line 47:
```suggestion
        organization_id: The organization ID.
```

src/cloud-cli/meltano/cloud/cli/run.py
```diff
@@ -0,0 +1,94 @@
...
+        project_id: The project identifier.
+        environment: The environment to run in.
+        job_or_schedule: The job or schedule identifier.
+
+    Returns:
+        The new run details.
+    """
+    async with MeltanoCloudClient() as client:
+        return await client.run_project(
+            tenant_resource_key,
```
cjohnhanson, line 57:
```suggestion
            organization_id,
```

src/cloud-cli/meltano/cloud/cli/run.py
cjohnhanson, line 82: I'm planning on using `MELTANO_CLOUD_` as the env var prefix for cloud-cli-specific config. Could we update these for consistency? Then users will also be able to set these once in their config file rather than having to set these each time.
```suggestion
    api_key = get_env_var("MELTANO_CLOUD_RUNNER_API_KEY")
    runner_secret = get_env_var("MELTANO_CLOUD_RUNNER_SECRET")
    organization_id = get_env_var("MELTANO_CLOUD_ORGANIZATION_ID")
```

src/cloud-cli/meltano/cloud/cli/run.py
```diff
@@ -0,0 +1,94 @@
...
+
+    api_key = get_env_var("MELTANO_RUNNER_API_KEY")
+    runner_secret = get_env_var("MELTANO_RUNNER_SECRET")
+    tenant_resource_key = get_env_var("MELTANO_TENANT_RESOURCE_KEY")
+
+    result = asyncio.run(
+        run_project(
+            api_key,
+            runner_secret,
+            tenant_resource_key,
```
cjohnhanson, line 88:
```suggestion
            organization_id,
```

cjohnhanson (changes requested):
Looks good! A few suggested changes for consistency with #7338--once this merges I'll just pull these changes into that branch, I've already incorporated this approach to `MeltanoCloudClient` locally and it's only a few lines of difference to use the auth and config in #7338, so no other big changes needed here to merge this into the `cloud login` changes.

cjohnhanson:
> @cjohnhanson In retrospect, the use of `aiohttp` here feels like overkill so I'm happy to fallback to `requests` if you prefer. Wdyt?

 @edgarrmondragon I'm fine either way. I don't foresee many use cases for async requests, in theory `meltano cloud ...` commands shouldn't ever do anything more complex than parsing input, using it to make a blocking API request, and then formatting the response; so there probably aren't going to be many use cases where async would give us much benefit. That said, I've already made the `MeltanoCloudClient` changes for #7338 locally and they plug in to this `MeltanoCloudClient` really easily. Only other change on my end for moving from requests to `aiohttp` would be to update the `logout` command to make the request with `aiohttp` instead of `requests`. On the whole, probably more work to change this PR to use `requests` than to change #7338 to use `aiohttp` so I'm good with sticking to `aiohttp`.

cjohnhanson (approved):

WillDaSilva (commented):
@edgarrmondragon Do you intend to add tests at this time?

edgarrmondragon:
> @edgarrmondragon Do you intend to add tests at this time?

@WillDaSilva yeah, thanks for asking. Just pushed them: 4a0052ec0bbd19c902afa5a00cf46413a1c6f2a1

cjohnhanson (approved):
Re-approving after addition of tests (obviously pending resolution of currently failing tests--but it looks like those are just a missing dev dependency in the testing env). Thanks @WillDaSilva for calling that out.

WillDaSilva (approved):
💯

PR #7338, opened 2023-02-20
feat: CLI commands `meltano cloud login` and `meltano cloud logout`

### LINKED ISSUE #7200: feat(CLI): `meltano cloud login` and `meltano cloud logout`
# Success Criteria
- Users can run `meltano cloud login` and `meltano cloud logout`
- If the user is currently "logged in" `meltano cloud login` indicates this and returns successfully
- If the user is not currently "logged in", `meltano cloud login` directs the user to complete a login flow. Upon successful completion, the `meltano` CLI is provided an authorization token to be stored locally.
- If the user is currently "logged in" `meltano cloud logout` requests for cloud.meltano.com to invalidate the locally stored token and then deletes the token from the local machine.
- If the user is not currently "logged in", `meltano cloud logout` indicates this and returns successfully.

# Context
For all upcoming `meltano cloud ...` commands, the meltano CLI will need to be able to authenticate against cloud.meltano.com as part of any relevant API requests. 

# Resources (some links are internal)

- Depends on https://github.com/meltano/infra/issues/314 and https://github.com/meltano/infra/issues/315
- Older issue this one replaces: https://github.com/meltano/infra/issues/95
- Beta signup flow: https://github.com/meltano/infra/issues/336
- Backend mapping of projects+resources: https://github.com/meltano/infra/issues/314

cjohnhanson:
Implements `login` and `logout` commands for Meltano Cloud. 

Resolves https://github.com/meltano/meltano/issues/7200


WillDaSilva:
@cjohnhanson I've changed the base branch from `main` to `cloud`. Feel free to close and re-create the PR if you'd rather make it clean.

aaronsteers:
@cjohnhanson - does this pr also include the logout capability?

I linked to the issue but want to make sure I'm not messing anything up if this isn't 1:1 with that.

src/cloud-cli/pyproject.toml
```diff
@@ -41,6 +41,7 @@ python = ">=3.7,<3.12"
 requests = "^2.28.0"
 structlog = "^22.3.0"
 platformdirs = "^3.0.0"
+flask = "^2.1"
```
WillDaSilva, line 44: @cjohnhanson Why Flask?

cjohnhanson:
> @cjohnhanson - does this pr also include the logout capability?
> 
> I linked to the issue but want to make sure I'm not messing anything up if this isn't 1:1 with that.

@aaronsteers yes, this has logout capability as well -- though there will be a one-line Pulumi update that needs to go out to before this, I realized in testing that you have to provide a redirect URI for the logout and we'll need to add it as an allowed URI in Cognito.

src/cloud-cli/tests/cli/test_cli.py
```diff
@@ -1,11 +1,13 @@
...
 
+from pathlib import Path
+
 from click.testing import CliRunner
 
 from meltano.cloud.cli import cloud as cli
 
 
-def test_cli():
+def test_cli(tmp_path: Path):
```
edgarrmondragon, line 10: I think this fixture isn't used

pyproject.toml
```diff
@@ -95,6 +95,8 @@ tzlocal = "^4.2.0"
 uvicorn = {extras = ["standard"], version = "^0.17.6"}
 virtualenv = "^20.17.1"
 werkzeug = ">=2.1,<=2.1.3"
+platformdirs = "^3.0.0"
+pytest-structlog = "^0.6"
```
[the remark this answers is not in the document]
cjohnhanson: Yeah, good catch.

src/cloud-cli/meltano/cloud/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Settings
+
+The following settings are configurable:
+
+#### `auth_callback_port`
+
+As part of authentication, `meltano-cloud` runs a lightweight, short-lived HTTP
+server on `localhost` to handle authentication callbacks. By default, this
+server runs on port 8080 but can be configured to another port using this setting.
```
WillDaSilva, line 26: Is this out of date? I thought port 9999 was being used? This port number also appears in a couple other spots.

cjohnhanson: yep, that is out of date. Updated. 

The only other place I could find it was in unit tests, which is fine. It can use any port locally but cognito will throw an error if you try to redirect to anything besides 9999 on localhost.

edgarrmondragon (approved):

src/cloud-cli/meltano/cloud/README.md
```diff
@@ -0,0 +1,85 @@
...
+- Command line flag: `--auth-callback-port`
+- Environment variable: `MELTANO_CLOUD_AUTH_CALLBACK_PORT`
+
+#### `base_url`
+
+This is the base URL used for API calls to Meltano Cloud, `https://api.meltano.cloud`
+by default.
+
+- Command line flag: `--base-url`
+- Environment variable: `MELTANO_CLOUD_BASE_URL`
```
aaronsteers, line 37: Should this have the `internal.` prefix?

```suggestion
This is the base URL used for API calls to Meltano Cloud, `https://internal.api.meltano.cloud`
```

edgarrmondragon (changes requested):

src/cloud-cli/meltano/cloud/README.md
```diff
@@ -0,0 +1,85 @@
+# Meltano Cloud CLI
+
+`meltano-cloud` is a standalone CLI tool for interacting with the Meltano Cloud API.
+
+## Configuration
+
+`meltano-cloud` can be configured via
+`meltano-cloud config set <setting name> <setting value>`.
+
```
aaronsteers, line 9: I was thinking that we would preserve `meltano-cloud config set` for project settings and secrets that would be stored in the cloud.

Is that still the case, or would this also be used for the below options like `auth_callback_port`?

We need to think a bit about collisions and behavior of get/set if so.

src/cloud-cli/meltano/cloud/README.md
```diff
@@ -0,0 +1,85 @@
...
+- Command line flag: `--base-auth-url`
+- Environment variable: `MELTANO_CLOUD_BASE_AUTH_URL`
+
+#### `organization_id`
+
+This is the default ID for the Meltano Cloud organization to use in interacting with
+the Meltano Cloud API.
+
+- Command line flag: `--organization-id`
+- Environment variable: `MELTANO_CLOUD_ORGANIZATION_ID`
```
aaronsteers, line 53: Do we need this to be configurable? 

If not, I prefer to drop it and avoid needing to support permanent user-facing organization IDs, which aren't yet a thing we're supporting/declaring for Cloud users here: https://github.com/meltano/cloud-docs/blob/main/docs/concepts.md

src/cloud-cli/meltano/cloud/api/client/__init__.py
```diff
@@ -94,32 +108,56 @@ def session(self) -> ClientSession:
...
+                del self.session.headers[key]
+
+    @asynccontextmanager
+    async def authenticated(self) -> t.AsyncIterator[None]:
+        """Provide context for API calls which require authentication.
+
+        Yields:
+            None
+        """
+        if not self.auth.logged_in():
```
edgarrmondragon, line 137:
```suggestion
        if not await self.auth.logged_in():
```

edgarrmondragon (changes requested):

src/cloud-cli/meltano/cloud/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+#### `project_id`
+
+For project-specific commands, this is the project ID to use by default in interacting
+with the Meltano Cloud API.
+
+- Command line flag: `--project-id`
+- Environment variable: `MELTANO_CLOUD_PROJECT_ID`
```
aaronsteers, line 61: This is potentially confusing due to the name collision of Meltano Core `project_id` and the (internal) "Cloud Project ID". I prefer not to have any user-facing references to Cloud Project ID unless their clearly delineated as something different.

Can we remove from user-facing docs or maybe give a new name to this concept?

The obvious suggestion of `cloud_project_id` does result in the unfortunate "`MELTANO_CLOUD_CLOUD_PROJECT_ID`", but if we can work around this, I'd be fine with just calling it `cloud_project_id` - since that is different enough in my mind to still be clear from a variety of contexts.


FWIW, we don't have the ID listed here yet, but we should add it if we are going to make it user-facing: https://github.com/meltano/cloud-docs/blob/main/docs/concepts.md

src/cloud-cli/meltano/cloud/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Setting Value Resolution
+
+`meltano-cloud` resolves configuration setting values as follows:
+
+- Use a value provided in a command-line flag, for example
+  `meltano-cloud --project-id <project_id> <subcommand>`.
+- Use a value provided in an environment variable, for example `MELTANO_CLOUD_PROJECT_ID`.
```
aaronsteers, line 69: Ditto comment above re "project id".

From this context, a user would likely assume their project_id from `meltano.yml` would work here. We probably should do one of the following:

1. Use `--cloud-project-id` to be explicit, OR
2. Take out the reference to project ID, OR
3. Rename to something different.

aaronsteers (dismissed):

WillDaSilva:
The first time I try to login, I get the following error in my browser (Firefox) console:

```
Scripts may not close windows that were not opened by script.
```

It closes as expected on all other attempts.

After authentication the CLI hangs.

WillDaSilva:
@cjohnhanson I've left my review suggestions in another PR:
- https://github.com/meltano/meltano/pull/7380

They're mostly small stuff, but a few things I consider important have change there too (e.g. restricting the file permissions on the config file which contains the secret tokens). I've also addressed all of the lint issues.

src/cloud-cli/meltano/cloud/README.md
cjohnhanson: I'd like to leave it configurable for ease of testing, but a middle path might be to remove from docs and not explicitly support it.

src/cloud-cli/meltano/cloud/README.md
cjohnhanson: That makes sense -- this line in the docs is actually outdated, as we had settled in another discussion on configuring exclusively via env vars and flags, since there are very few cloud-specific config options. I'll remove this here.

src/cloud-cli/meltano/cloud/README.md
cjohnhanson: Yep, will update.

cjohnhanson:
> The first time I try to login, I get the following error in my browser (Firefox) console:
> 
> ```
> Scripts may not close windows that were not opened by script.
> ```
> 
> It closes as expected on all other attempts.

@WillDaSilva good callout. I've added a `You may close this window.` message to the login/logout pages for any browsers that won't run that `window.close()` call for whatever reason.

cjohnhanson:
@aaronsteers I went ahead and just removed the settings section for the docs. Users shouldn't be configuring any of those configurable options (but we should still keep them around for testing purposes), with the one exception of `MELTANO_CLOUD_PROJECT_ID` which isn't actually used anywhere in this PR. The `meltano cloud run` command has `--project-id` as its own flag, and I think it's reasonable to stick to that pattern in the short term until we settle on a top-level "default" project to use when one isn't provided to subcommands that need it, maybe with an associated env var like `MELTANO_CLOUD_DEFAULT_CLOUD_PROJECT`

src/cloud-cli/meltano/cloud/api/config.py
```diff
@@ -0,0 +1,171 @@
...
+    def from_config_file(cls, config_path: os.PathLike | str) -> MeltanoCloudConfig:
+        """Initialize the configuration from a config file.
+
+        Args:
+            config_path: Path to the config file.
+
+        Returns:
+            A MeltanoCloudConfig
+        """
+        with open(config_path, encoding="utf-8") as config_file:
```
edgarrmondragon, line 126:
```suggestion
        with Path(config_path).open(encoding="utf-8") as config_file:
```

src/cloud-cli/meltano/cloud/api/config.py
```diff
@@ -0,0 +1,171 @@
...
+        try:
+            return cls.from_config_file(config_path or cls.find_config_path())
+        except (FileNotFoundError, json.JSONDecodeError):
+            config = cls()
+            config.write_to_file()
+            return config
+
+    def refresh(self) -> None:
+        """Reread config from config file."""
+        with open(self.config_path, encoding="utf-8") as config_f:
```
edgarrmondragon, line 151:
```suggestion
        with Path(self.config_path).open(encoding="utf-8") as config_f:
```

src/cloud-cli/tests/api/auth/test_callback_server.py
```diff
@@ -0,0 +1,71 @@
...
+        self, subject: Flask, client: FlaskClient, config: MeltanoCloudConfig
+    ):
+        response: TestResponse = client.get(
+            "/tokens?id_token=meltano-cloud-testing&access_token=meltano-cloud-testing"
+        )
+        config.refresh()
+        assert response.status_code == 204
+        assert config.access_token == "meltano-cloud-testing"  # noqa: S105
+        assert config.id_token == "meltano-cloud-testing"  # noqa: S105
+        with open(config.config_path) as config_file:
```
edgarrmondragon, line 52:
```suggestion
        with Path(config.config_path).open() as config_file:
```

src/cloud-cli/tests/api/auth/test_callback_server.py
```diff
@@ -0,0 +1,71 @@
...
+    ):
+        config.access_token = "meltano-cloud-testing"  # noqa: S105
+        config.id_token = "meltano-cloud-testing"  # noqa: S105
+        config.write_to_file()
+        response: TestResponse = client.get("/logout")
+        assert response.status_code == 200
+        config.refresh()
+        assert config.id_token is None
+        assert config.access_token is None
+        with open(config.config_path) as config_file:
```
edgarrmondragon, line 68:
```suggestion
        with Path(config.config_path).open() as config_file:
```

src/cloud-cli/tests/api/test_cloud_config.py
```diff
@@ -0,0 +1,102 @@
...
+            "organization_id": f"{self._val_prefix}organization-id",
+            "project_id": f"{self._val_prefix}project-id",
+            "id_token": f"{self._val_prefix}id-token",
+            "access_token": f"{self._val_prefix}access-token",
+        }
+
+    @pytest.fixture(scope="function")
+    def config_path(self, tmp_path: Path, config_dict: dict):
+        config_file_path = tmp_path / "meltano-cloud.json"
+        with open(config_file_path, "w") as config_file:
```
edgarrmondragon, line 40:
```suggestion
        with Path(config_file_path).open("w") as config_file:
```

src/cloud-cli/tests/api/test_cloud_config.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def test_refresh(
+        self,
+        subject: MeltanoCloudConfig,
+        config_path: Path,
+        config_dict: dict,
+    ):
+        self.config_assertions(subject)
+        os.remove(config_path)
+        with open(config_path, "w") as config_file:
```
edgarrmondragon, line 84:
```suggestion
        with Path(config_path).open("w") as config_file:
```

src/cloud-cli/tests/api/test_cloud_config.py
```diff
@@ -0,0 +1,102 @@
...
+    def test_write_to_file(
+        self,
+        monkeypatch: pytest.MonkeyPatch,
+        subject: MeltanoCloudConfig,
+        config_path: Path,
+    ):
+        self.config_assertions(subject)
+        subject.organization_id = "organization-id-changed"
+        subject.write_to_file()
+        with open(subject.config_path) as config_file:
```
edgarrmondragon, line 100:
```suggestion
        with Path(subject.config_path).open() as config_file:
```

edgarrmondragon (changes requested):
@cjohnhanson I think the change from `access_token` to `id_token` is causing the auth check to get into an infinite loop. The `oauth2/userInfo` seems to require the `access_token`.

src/cloud-cli/meltano/cloud/README.md
aaronsteers: That makes sense. Wdyt of renaming with an "internal" prefix like "MELTANO_CLOUD_INTERNAL_ORG_ID"?

In that case, we can worry less about whether it's in the docs or not, since it's more self-documenting as non-user-facing.

src/cloud-cli/tests/api/auth/test_callback_server.py
WillDaSilva: @edgarrmondragon Why prefer `Path.open` over `open`?

edgarrmondragon: Really just a nit but I like the API better, and I guess I've been looking at Ruff-linted code 🙈

WillDaSilva: I see: https://beta.ruff.rs/docs/rules/#flake8-use-pathlib-pth

> `open()` should be replaced by `Path.open()`

I'm not a huge fan of that rule in particular (`PTH123`), but Ruff can automatically make it consistent, and I value consistency, so I'll support it in most cases.

Unfortunately it appears that `Path.open` doesn't support all of the parameters that `open` does. Notably it's missing the `opener` parameter, which is useful for opening a file with non-default permission bits set. We open the Cloud CLI config/secrets file for writing with 0600 as its permissions using that.

edgarrmondragon (approved):
It's working! 😄

WillDaSilva (approved):
:shipit:

PR #7343, opened 2023-02-22
feat: Support arrow key text navigation during interactive config

### LINKED ISSUE #7342: feature: Support arrow key text navigation during interactive config
### Feature scope

### Description

Currently when using interactive config, the arrow keys are interpreted as raw values, rather than as navigation controls:

Examples:

Pressing the up key to try to set the prompt to the last value entered:

```
New value: ^[[A
```

Pressing the left key repeatedly to try to add a missing quote:

```
New value: example"^[[D^[[D^[[D^[[D^[[D^[[D^[[D^[[D
```

Ideally arrow keys pressed during interactive config would result in typical text navigation behaviour.

WillDaSilva:
Closes #7342

### DIFF
--- a/src/meltano/cli/interactive/config.py
+++ b/src/meltano/cli/interactive/config.py
@@ -2,6 +2,11 @@
 
 from __future__ import annotations
 
+# NOTE: Importing the readline module enables the use of arrow
+#       keys for text navigation during interactive config.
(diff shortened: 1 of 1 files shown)


tayloramurphy (commented):
Love this change 😄

edgarrmondragon (approved):
TIL `readline`!

PR #7347, opened 2023-02-24
fix: Error in `meltano add` when using env var proxies (e.g. `HTTP_PROXY`)

### LINKED ISSUE #7346: bug: `meltano add` throws an error on `2.16.0` when using proxies
https://github.com/meltano/meltano/blob/dc04f5eb69837220fcf62542913bc92f64eb2722/src/meltano/core/hub/client.py#L235

second arg(proxies) should be `{}`, otherwise it throws Nonetype error from requests/sessions.py when you run `meltano add loader ?`

image

edgarrmondragon:
Closes https://github.com/meltano/meltano/issues/7346

### DIFF
--- a/src/meltano/core/hub/client.py
+++ b/src/meltano/core/hub/client.py
@@ -230,7 +230,7 @@ class MeltanoHubService(PluginRepository):  # noqa: WPS214
         prep = self._build_request("GET", url)
         settings = self.session.merge_environment_settings(
             prep.url,
-            None,
+            {},
             None,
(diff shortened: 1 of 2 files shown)


WillDaSilva (approved):
@edgarrmondragon Do you think this warrants a patch release?

edgarrmondragon:
> @edgarrmondragon Do you think this warrants a patch release?

@WillDaSilva yeah, that'd be nice

PR #7358, opened 2023-03-02
fix: Add error and log for locking unknown plugins

### LINKED ISSUE #7356: bug: `meltano lock` doesn't complain when plugin is spelled wrong
### Meltano Version

2.16.1

### Python Version

3.9

### Bug scope

### Operating System

macOS

### Description

I originally noticed in this slack thread https://meltano.slack.com/archives/C01TCRBBJD7/p1677603089734759?thread_ts=1677524811.910379&cid=C01TCRBBJD7 that the lock command exits with code 0 and no logs if I have a misspelled plugin name i.e. `meltano lock foo-bar`. I would expect it to fail if the tap isnt installed in my project. This was a little misleading because it seemed like it successfully locked when really it was just successfully exiting with nothing to do.

Suggestions:
- Either log something helpful like "nothing to do, no matching plugins by name `foo-bar`"
- Or fail with a similar message like "Plugin `foo-bar` not found, failed to lock."

### Code

_No response_

pnadolny13:
Closes https://github.com/meltano/meltano/issues/7356

I have it raising an error right now but I could change it to log instead if we want it to pass. I thought an error made more sense but I do see that install currently just logs `Installing 0 plugins...` in a similar situation of not finding the specified plugin.

### DIFF
--- a/src/meltano/cli/lock.py
+++ b/src/meltano/cli/lock.py
@@ -77,6 +77,10 @@ def lock(
 
     tracked_plugins = []
 
+    if not plugins:
+        raise CliError("No matching plugin(s) found")
+
+    click.echo(f"Locking {len(plugins)} plugin(s)...")
     for plugin in plugins:
         descriptor = f"{plugin.type.descriptor} {plugin.name}"
         if plugin.is_custom():
--- a/src/meltano/cli/lock.py
+++ b/src/meltano/cli/lock.py
@@ -78,7 +78,9 @@ def lock(
     tracked_plugins = []
 
     if not plugins:
-        raise CliError("No matching plugin(s) found")
+        tracker.track_command_event(CliEvent.aborted)
+        errmsg = "No matching plugin(s) found"
+        raise CliError(errmsg)
 
     click.echo(f"Locking {len(plugins)} plugin(s)...")
     for plugin in plugins:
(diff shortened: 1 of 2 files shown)


src/meltano/cli/lock.py
```diff
@@ -77,6 +77,10 @@ def lock(
 
     tracked_plugins = []
 
+    if not plugins:
+        raise CliError("No matching plugin(s) found")
```
edgarrmondragon, line 81: 1. We may want to track the event (@WillDaSilva can confirm)
2. Errors are usually easier to read when the string is not repeated in the traceback and the final output message

```suggestion
        tracker.track_command_event(CliEvent.aborted)
        errmsg = "No matching plugin(s) found"
        raise CliError(errmsg)
```

edgarrmondragon (commented):
Beat me to it @pnadolny13, thanks this is great!

src/meltano/cli/lock.py
pnadolny13: @edgarrmondragon thanks! Based on Will's thumbs up I assumed he liked your suggestion so I went ahead and pulled it in.

edgarrmondragon (approved):

src/meltano/cli/lock.py
```diff
@@ -77,6 +78,12 @@ def lock(
 
     tracked_plugins = []
 
+    if not plugins:
+        tracker.track_command_event(CliEvent.aborted)
+        errmsg = "No matching plugin(s) found"
+        raise CliError(errmsg)
```
WillDaSilva, line 84:
```suggestion
        raise CliError("No matching plugin(s) found")
```

WillDaSilva (approved):

tayloramurphy (approved):

PR #7366, opened 2023-03-06
docs: gsg to use utility dbt-postgres vs transformer

pnadolny13:
@sbalnojan let me know what you think about these updates:

- switch to using utility dbt-postgres over the transformer + file bundle
- it also looks like on part 2 we're pulling commits from `sbalnojan/meltano-lightdash` then in part 3 we're using `sbalnojan/meltano-example-el`. Does it make a difference? I noticed the lightdash repo only gave me 1 record vs the example-el gave me many. I just noticed that it switched between parts. It might be less confusing for users if we chose one for the full tutorial.

### DIFF
--- a/docs/src/_getting-started/part1.md
+++ b/docs/src/_getting-started/part1.md
@@ -216,7 +216,7 @@ This will add the configuration to your [`meltano.yml` project file](/concepts/p
           config:
             start_date: '2022-01-01'
             repositories:
-              - sbalnojan/meltano-lightdash
+              - sbalnojan/meltano-example-el
   ```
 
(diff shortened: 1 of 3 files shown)


tayloramurphy:
@pnadolny13 I'm merging https://github.com/meltano/meltano/pull/7619 which has some of these changes. Will you review this PR and either close or update it?

pnadolny13:
@tayloramurphy I'm closing since its outdated. Although I noticed one thing that https://github.com/meltano/meltano/pull/7619 accidentally reverted so I created a [new PR](https://github.com/meltano/meltano/pull/7783) to add that back.

PR #7367, opened 2023-03-07
fix: Resolve variant label from Hub response

### LINKED ISSUE #7365: bug: `meltano add` logs variant as default when its not
When I add a [non-default meltanolabs target-postgres](https://hub.meltano.com/loaders/target-postgres--meltanolabs/) variant to my project from the hub it says that its the default variant in the logs even though its not the default.

Screen Shot 2023-03-06 at 4 55 36 PM

I confirmed with a request to the hub api that the results coming from the hub are correct. It seems like Meltano might not be retrieving the default name from the hub api properly.

Screen Shot 2023-03-06 at 4 58 15 PM

edgarrmondragon:
Related:
- Closes https://github.com/meltano/meltano/issues/7365

### DIFF
--- a/src/meltano/core/plugin/base.py
+++ b/src/meltano/core/plugin/base.py
@@ -327,8 +327,10 @@ class PluginDefinition(PluginRef):
         plugin_type: PluginType,
         name: str,
         namespace: str,
+        *,
         variant: str | None = None,
         variants: list | None = None,
+        is_default_variant: bool | None = None,
         **extras,
     ):
         """Create a new PluginDefinition.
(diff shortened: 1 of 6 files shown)


src/meltano/core/plugin/base.py
```diff
@@ -327,8 +327,10 @@ def __init__(
         plugin_type: PluginType,
         name: str,
         namespace: str,
+        *,
         variant: str | None = None,
         variants: list | None = None,
```
edgarrmondragon, line 332: This is legacy stuff. I'll experiment a bit to see if it can be safely removed (after this PR is merged).

WillDaSilva (approved):

PR #7374, opened 2023-03-13
chore: Limit direct use of Werkzeug to `meltano.api`

### LINKED ISSUE #7373: Limit direct Werkzeug use to `meltano.api`
When `meltano.api` is removed as part of version 3, we also want to remove any direct usage of Werkzeug so that we are no longer directly depending on it, as this has been a significant maintenance burden in the past, and is currently (as of 2023-03-13) the source of [a security issue](https://github.com/meltano/meltano/security/dependabot/2).

In preparation for this, we'll need to remove all usage of Werkzeug outside of `meltano.api`. It is currently used in 2 places within `src/meltano/core/project.py`: calling `werkzeug.utils.secure_filename(state_id)` to create the path to the jobs directory, and to create the path to the jobs logs directory.

We cannot change the behaviour at those 2 locations without breaking backwards compatibility, so switching to an alternative path sanitization library is not an option unless it can do exactly what `werkzeug.utils.secure_filename` does. Our best bet is probably to write a small utility function that exhibits the same behaviour as `werkzeug.utils.secure_filename`, and use that instead.

WillDaSilva:
Closes #7373

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -163,3 +163,6 @@ setup.py
 *.jsonl
 **/.meltano/**
 !docs/example-library/meltano-migrations/.meltano/meltano.db
+
+# Ruff
+.ruff_cache/
--- a/src/meltano/core/project.py
+++ b/src/meltano/core/project.py
@@ -14,7 +14,6 @@ from pathlib import Path
 
 import fasteners
 from dotenv import dotenv_values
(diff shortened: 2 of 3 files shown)


edgarrmondragon (approved):
nice!

PR #7379, opened 2023-03-14
feat(cli): Print cloud execution logs

edgarrmondragon:
Related:

- Depends on
  - #7200
  - #7338
- Closes https://github.com/meltano/infra/issues/323


edgarrmondragon:
I'm seeing some weird behavior:

1. Run `meltano-cloud logs print --execution-id 0678cfe097a740798682e52fb3038709`
2. Browser tab opens and promptly closes itself
3. The CLI hangs with the message `If a web browser does not open, open the following link:
https://auth-dev.meltano.cloud...`

WillDaSilva:
> Browser (Chrome) tab opens and promptly closes itself

This is by design. Personally I'm not sure if I like it.

> The CLI hangs with the message

I was seeing this until I reverted your suggestions from the base PR:
- https://github.com/meltano/meltano/pull/7338#discussion_r1136067703
- https://github.com/meltano/meltano/pull/7338#discussion_r1136071740

So I suspect it has to do with one of those changes.

edgarrmondragon:
> I was seeing this until I reverted your suggestions from the base PR:
> 
> * [feat: CLI commands `meltano cloud login` and `meltano cloud logout` #7338 (comment)](https://github.com/meltano/meltano/pull/7338#discussion_r1136067703)
> * [feat: CLI commands `meltano cloud login` and `meltano cloud logout` #7338 (comment)](https://github.com/meltano/meltano/pull/7338#discussion_r1136071740)

Yeah, although I think those are strictly necessary. The first one was how I used the API in manual tests. The second one is an asyncio bug.

cjohnhanson (approved):
Looks good! Nice to see the auth pattern in action.

src/cloud-cli/meltano/cloud/api/client/__init__.py
```diff
@@ -228,8 +250,30 @@ async def run_project(
...
+
+        Args:
+            execution_id: The execution identifier.
+
+        Yields:
+            The response object.
+        """
+        url = (
+            "/logs/v1/"
+            f"{self.config.organization_id}/{self.config.project_id}/{execution_id}"
```
WillDaSilva, line 274: @edgarrmondragon @cjohnhanson We should probably rename `organization_id` to `tenant_resource_key`, to avoid confusing these distinct concepts.

edgarrmondragon: Yeah, I agree. We're calling it different things in different contexts.

WillDaSilva (approved):

WillDaSilva:
@edgarrmondragon @cjohnhanson I think https://github.com/meltano/meltano/pull/7389 may have introduced a regression. When I run `meltano cloud login` followed by `meltano cloud logout` I get the message `Not logged in`, despite the login completing successfully. It seems like the config file is being ignored when instantiated as `MeltanoCloudConfig()` since it never calls `refresh` or otherwise does anything to set the its instance values other than `config_path`.

edgarrmondragon:
> @edgarrmondragon @cjohnhanson I think #7389 may have introduced a regression. When I run `meltano cloud login` followed by `meltano cloud logout` I get the message `Not logged in`, despite the login completing successfully. It seems like the config file is being ignored when instantiated as `MeltanoCloudConfig()` since it never calls `refresh` or otherwise does anything to set the its instance values other than `config_path`.

@WillDaSilva Yeah I can confirm, I'm investigating.

src/cloud-cli/tests/cli/test_logs.py
```diff
@@ -0,0 +1,92 @@
...
+            app_client_id="meltano-cloud-test",
+            access_token="meltano-cloud-test",  # noqa: S106
+            id_token=id_token,
+        )
+        config.write_to_file()
+        return config
+
+    @pytest.fixture
+    def client(self, config: MeltanoCloudConfig):
+        return MeltanoCloudClient(config=config)
```
WillDaSilva, line 55: These fixtures should probably go in `conftest.py` so that all Cloud CLI tests can use them. They can be overridden if needed by test modules/classes as described here: https://www.seanh.cc/2017/02/12/advanced-fixtures/

WillDaSilva (approved):

PR #7381, opened 2023-03-15
docs: move replicate troubleshooting to guide

pnadolny13:
@sbalnojan what do you think about moving the troubleshooting section from the replicate data guide to the troubleshooting guide? I was looking for a link to share with someone in slack about how to debug a tap/target pair, specifically the "isolate the connector" section but had trouble finding it because it was at the end of the replicate data guide. To me it makes sense to have these sections in the troubleshooting guide and link to them from the end of the replicate data guide. What do you think about that?


pnadolny13:
@sbalnojan any thoughts on this?

sbalnojan (approved):
@pnadolny13 I'm all for it. The troubleshooting guide should cover all cases I look for when troubleshooting!

edgarrmondragon (approved):

PR #7390, opened 2023-03-16
chore: Use ID token to configure tenant resource key and project ID

cjohnhanson:


src/cloud-cli/meltano/cloud/api/config.py
```diff
@@ -102,6 +107,84 @@ def config_path(self) -> Path:
...
+        Returns:
+            List of trks and pids in the form '<trk>::<jpid>`
+
+        """
+        with suppress(jwt.DecodeError):
+            decoded = self.decode_jwt(self.id_token)
+            trks_and_pids = decoded.get("custom:trk_and_pid")
+            if trks_and_pids:
+                return [perm.strip() for perm in trks_and_pids.split(",")]
+        return []
```
WillDaSilva, line 135: I think we should avoid exposing this beyond this class by prefixing its name with an underscore.

src/cloud-cli/meltano/cloud/api/config.py
```diff
@@ -102,6 +107,84 @@ def config_path(self) -> Path:
...
+
+    @property
+    def tenant_resource_keys(self) -> t.List[str]:
+        """Get list of tenant resource keys from id token.
+
+        Returns:
+            List of trks in id token.
+
+        """
+        return [perm.split("::")[0] for perm in self.trks_and_pids]
```
WillDaSilva, line 145:
```suggestion
    def tenant_resource_keys(self) -> set[str]:
        """Get the tenant resource keys from the ID token.

        Returns:
            The tenant resource keys found in the ID token.

        """
        return {perm.split("::")[0] for perm in self.trks_and_pids}
```

src/cloud-cli/meltano/cloud/api/config.py
```diff
@@ -102,6 +107,84 @@ def config_path(self) -> Path:
...
+
+    @property
+    def internal_project_ids(self) -> t.List[str]:
+        """Get list of project IDs from id token.
+
+        Returns:
+            List of pids in id token.
+
+        """
+        return [perm.split("::")[1] for perm in self.trks_and_pids]
```
WillDaSilva, line 155:
```suggestion
    def internal_project_ids(self) -> set[str]:
        """Get the internal project IDs from the ID token

        Returns:
            The internal project IDs found in the ID token.

        """
        return {perm.split("::")[1] for perm in self.trks_and_pids}
```

src/cloud-cli/meltano/cloud/api/config.py
```diff
@@ -102,6 +107,84 @@ def config_path(self) -> Path:
...
+            The tenant resource key.
+
+        Raises:
+            NoMeltanoCloudTenantResourceKeyError: when ID token includes no
+                tenant resource keys.
+        """
+        try:
+            return self.tenant_resource_keys[0]
+        except IndexError:
+            raise NoMeltanoCloudTenantResourceKeyError()
```
WillDaSilva, line 186: These should raise an exception if there is ambiguity about which ID/key to use, rather than taking the first one. They should only return a value if there is exactly one in the collection.

src/cloud-cli/meltano/cloud/api/config.py
```diff
@@ -102,6 +107,84 @@ def config_path(self) -> Path:
...
+    def tenant_resource_keys(self) -> t.List[str]:
+        """Get list of tenant resource keys from id token.
+
+        Returns:
+            List of trks in id token.
+
+        """
+        return [perm.split("::")[0] for perm in self.trks_and_pids]
+
+    @property
```
WillDaSilva, line 147:
```suggestion
    @cached_property
```

src/cloud-cli/meltano/cloud/api/config.py
```diff
@@ -102,6 +107,84 @@ def config_path(self) -> Path:
...
+
+        """
+        with suppress(jwt.DecodeError):
+            decoded = self.decode_jwt(self.id_token)
+            trks_and_pids = decoded.get("custom:trk_and_pid")
+            if trks_and_pids:
+                return [perm.strip() for perm in trks_and_pids.split(",")]
+        return []
+
+    @property
```
WillDaSilva, line 137:
```suggestion
    @cached_property
```

src/cloud-cli/meltano/cloud/api/config.py
```diff
@@ -4,8 +4,11 @@
 
 import json
 import os
+import typing as t
+from contextlib import suppress
 from pathlib import Path
 
+import jwt
 import platformdirs
```
WillDaSilva, line 13:
```suggestion
import sys
import typing as t
from contextlib import suppress
from pathlib import Path

import jwt
import platformdirs

if sys.version_info <= (3, 8):
    from cached_property import cached_property
else:
    from functools import cached_property
```

WillDaSilva (approved):

src/cloud-cli/meltano/cloud/api/config.py
```diff
@@ -102,6 +120,92 @@ def config_path(self) -> Path:
...
+        Returns:
+            Internal project ID.
+
+        Raises:
+            NoMeltanoCloudProjectIDError: when ID token includes no project IDs.
+            MeltanoCloudProjectAmbiguityError: when ID token includes more
+                than one project ID.
+        """
+        if len(self.internal_project_ids) > 1:
+            raise MeltanoCloudProjectAmbiguityError()
```
aaronsteers, line 183: Here and for the related error `MeltanoCloudTenantAmbiguityError`, can add printed instructions and/or other mitigation on this error? Could instead be added to the docstring, but I think a clear error message to the user could go a long way.

I think the mitigation is for user to set `MELTAN_CLOUD_INTERNAL_PROJECT_ID` value. Not documented, that I know of, but I think you mentioned that this follows automatically from our env var auto-parsing logic.

Also, can we safely print the observed keys to the user, since I believe those are non-secret and would (worst case) give them an option to learn their key and self-solve the issue if unexpected?

src/cloud-cli/meltano/cloud/api/config.py
```diff
@@ -102,6 +120,92 @@ def config_path(self) -> Path:
...
+            The tenant resource key.
+
+        Raises:
+            NoMeltanoCloudTenantResourceKeyError: when ID token includes no
+                tenant resource keys.
+            MeltanoCloudTenantAmbiguityError: when ID token includes more
+                than one tenant resource key.
+        """
+        if len(self.tenant_resource_keys) > 1:
+            raise MeltanoCloudTenantAmbiguityError()
```
aaronsteers, line 203: Ditto comment regarding the project ambiguity error:

Would be helpful if we can provide a message about how to resolve, optionally including the values we observe if it is safe to print those.

src/cloud-cli/meltano/cloud/api/config.py
cjohnhanson: Good callout. I've added a more helpful error message.

edgarrmondragon (approved):

Issue #6412, opened 2022-07-13
docs: Build example Meltano project using `meltano-map-transform` with examples for common uses


DouweM:
<details>
<summary>Original description</summary>

The README for https://github.com/MeltanoLabs/meltano-map-transform lacks usage instructions, https://docs.meltano.com/concepts/plugins#mappers has an example of `pipelinewise-transform-field` but not `meltano-map-transform`, and I happen to know that some of the docs on https://sdk.meltano.com/en/latest/stream_maps.html apply to `meltano-map-transform` as well, but a user wouldn't know that. 

Since mapping in between extracting and loading is a form of transformation, I think we should document this functionality on https://docs.meltano.com/guide/transformation, and reference it from https://docs.meltano.com/guide/integration as well, as a new user wouldn't know to find the relevant docs in https://docs.meltano.com/concepts/plugins#mappers. I also think it's worth listing the mappers supported out of the box on https://hub.meltano.com/. 

</details>

Proposal (from @aaronsteers):

1. Create a new entry in the `example-projects` repo with a set of example map transforms (hopefully covering all/most of what @sbalnojan outlines [here](https://github.com/meltano/meltano/issues/6412#issuecomment-1277145178)).
2. Link to the new usage examples in:
   - the `meltano-map-transform` readme
   - the related SDK docs page
   - on docs.meltano.com

DouweM:
@tayloramurphy I think this is more important to prioritize than https://github.com/meltano/sdk/issues/801, as users can already get a long way with our existing mapper, without needing to write a custom one.

tayloramurphy:
>  I also think it's worth listing the mappers supported out of the box on https://hub.meltano.com/.

We have https://github.com/meltano/hub/issues/589 which tracks that.

> Since mapping in between extracting and loading is a form of transformation, I think we should document this functionality on https://docs.meltano.com/guide/transformation, and reference it from https://docs.meltano.com/guide/integration as well, as a new user wouldn't know to find the relevant docs in https://docs.meltano.com/concepts/plugins#mappers.

I think that makes sense. Having a short interstitial on the integration guide is good just to get people thinking about it, but then the transformation page can be structured to have "in-flight" and "in-warehouse" transformations.  

I think it makes sense to prioritize as well @DouweM

sbalnojan:
[...]

PR #7394, opened 2023-03-17
docs: add mappers docs guide

pnadolny13:
Related to https://github.com/meltano/meltano/issues/6412#issuecomment-1471684404

Adds a guide to the docs that includes both standalone mappers and SDK based inline stream maps

- I sourced a lot of this information from https://sdk.meltano.com/en/latest/stream_maps.html and some examples from https://github.com/MeltanoLabs/meltano-map-transform/pull/117
- I put a couple examples to give a preview of what they do but I mostly wanted to link out to other docs like https://github.com/MeltanoLabs/meltano-map-transform/pull/117 and https://sdk.meltano.com/en/latest/stream_maps.html
- I felt like it was worth highlighting the difference between the two mapping methods and when you would use each, with tradeoffs. So let me know if I got anything wrong.
- Its a bit tricky to figure out the best way to refer to the two different approach. I went with "standalone mapper plugins" and "SDK 

Let me know if you have any obvious use cases I missed and what you think in general.

### DIFF
--- a/docs/src/_guide/mappers.md
+++ b/docs/src/_guide/mappers.md
@@ -175,7 +175,7 @@ No, BATCH message type does not currently support mapping functionality.
 It's mostly preference but here are some trade offs to consider:
 - Mapper plugins can only be run using `meltano run`
 - Mappers work for SDK and non-SDK connectors. If your pipeline uses 2 non-SDK connectors this is your best option.
-- SDK stream maps are part of the SDK which inheritly makes them more difficult to change. If you need custom functionality then forking a mapper plugin and customizing it is you best option. Although if your behavior is generically useful, please open an issue in the [SDK repo](https://github.com/meltano/sdk) to get it added!
+- SDK stream maps are part of the SDK which inherently makes them more difficult to change. If you need custom functionality then forking a mapper plugin and customizing it is you best option. Although if your behavior is generically useful, please open an issue in the [SDK repo](https://github.com/meltano/sdk) to get it added!
 - SDK stream maps avoid having to install an additional plugin into your project.
 - Mapper plugins alter data after its been extracted whereas SDK based stream maps on the tap side could alter the extraction behavior.
 Depending on your use case, for example deleting streams, it might be slightly more efficient to do it in the tap and avoid extracting unneeded data.
--- a/docs/src/_guide/mappers.md
+++ b/docs/src/_guide/mappers.md
@@ -1,0 +1,185 @@
+---
+title: Inline Data Mapping
+description: Alter records and streams on the fly.
+layout: doc
+weight: 5
+---
+
+Meltano gives you the flexibility to alter data passing through your EL pipelines to do precisely what you need for your use case.
+Although commonly users want to replicate their data in the most raw format, using [ELT vs ETL](https://meltano.com/blog/the-benefits-of-elt-vs-etl-what-you-need-to-know/
+), there are many use cases that require minor alterations to data on the fly.
+This is where mappers, or also referred to as inline stream maps, shine!
+
+# Features
+
+The mapper features can be viewed in 2 categories: stream level or property (record) level.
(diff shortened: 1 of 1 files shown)


docs/src/_guide/mappers.md
```diff
@@ -0,0 +1,185 @@
...
+
+# Features
+
+The mapper features can be viewed in 2 categories: stream level or property (record) level.
+Stream level features allow you to modify the stream itself, whereas the property level features allow you to modify contents of the stream.
+
+## Stream Level
+
+- Aliasing: streams can be aliased to provide custom naming downstream.
+- Filtering: streams records can be filtered based on any user-defined logic.
```
tayloramurphy, line 21:
```suggestion
- Filtering: stream records can be filtered based on any user-defined logic.
```

With the way we've got property level changes labeled as `property (record)` this sentence is suddenly very confusing to me. Would `property (column)` be more appropriate?

docs/src/_guide/mappers.md
```diff
@@ -0,0 +1,185 @@
...
+        issues:
+          repo: record['repo'].lower()
+```
+
+For more details and examples refer to the [SDK documentation](https://sdk.meltano.com/en/latest/stream_maps.html).
+
+## Standalone Mapper Plugins
+
+The other way of getting mapper functionality is to use a standalone mapper plugin that is inserted between your tap and target to do the translation `meltano run tap-csv <MAPPER> target-jsonl`.
+These can be used even whether or not your tap and/or target are SDK based, it's your choice how you implement mappers.
```
tayloramurphy, line 106:
```suggestion
These can be used even when your your tap and/or target are not SDK based.
```

docs/src/_guide/mappers.md
```diff
@@ -0,0 +1,185 @@
...
+
+The limitation with this approach is that currently you can only run these using the `meltano run` command.
```
tayloramurphy, line 108:
```suggestion
The limitation with this approach is that currently you can only run these using the `meltano run` command and not using `meltano elt` or `meltano invoke`.
```

docs/src/_guide/mappers.md
```diff
@@ -0,0 +1,185 @@
...
+
+## FAQs
+
+### Can I used standalone mapper plugins with the `meltano elt` command?
+
+No, currently only `meltano run` support standalone mapper plugins.
+
+### Can I use mapping features with the BATCH message type?
+
+No, BATCH message type does not currently support mapping functionality.
```
tayloramurphy, line 171:
```suggestion
No, BATCH message type does not currently support mapping functionality.
Follow along with [this discussion](https://github.com/meltano/meltano/discussions/6639#discussioncomment-3417860) where we talk about how stream maps could support BATCH messages.
```

docs/src/_guide/mappers.md
```diff
@@ -0,0 +1,185 @@
...
+An example taken from the [meltano-map-transformer](https://github.com/MeltanoLabs/meltano-map-transform/examples) repository shows the mapper altering casing of the data.
+Given this input CSV and mapper configurations:
+
+```csv
+id,first_name,last_name,email,ip_address
+1,Ethe,Book,ebook0@twitter.com,67.61.243.220
+```
+
+```yaml
+  - name: meltano-map-transformer
```
tayloramurphy, line 137:
```suggestion
plugins
  mapper:
    - name: meltano-map-transformer
```

might be useful to give the context of where this is in the file.

tayloramurphy (approved):
@pnadolny13 minor comments, but this looks _great_. Coupled with the PR on the mapper repo, I'm quite happy with these updates.

docs/src/_guide/mappers.md
pnadolny13: @tayloramurphy I agree its confusing. I updated to column.

PR #7415, opened 2023-03-22
feat: Implement `meltano cloud history`

edgarrmondragon:
Related:

* Closes https://github.com/meltano/infra/issues/321

TODOs:

- [x] Tests
- [x] Docs


src/cloud-cli/meltano/cloud/cli/history.py
```diff
@@ -0,0 +1,139 @@
...
+import tabulate
+
+from meltano.cloud.api.client import MeltanoCloudClient
+from meltano.cloud.cli.base import cloud
+
+if t.TYPE_CHECKING:
+    from meltano.cloud.api.config import MeltanoCloudConfig
+    from meltano.cloud.api.types import CloudExecution
+
+PAGE_SIZE = 100
```
WillDaSilva, line 20: What's the justification behind this page size? Does it correspond to anything configured in the backend?

src/cloud-cli/meltano/cloud/api/client.py
```diff
@@ -287,6 +287,40 @@ async def schedule_put_enabled(
                     raise MeltanoCloudError(ex.response) from ex
                 raise
 
+    async def get_execution_history(
+        self,
+        *,
+        schedule: str | None = None,
+        page_size: int = 10,
```
WillDaSilva, line 294: This default value should probably be sourced from a constant `PAGE_SIZE` used both here and within `history.py`.

src/cloud-cli/meltano/cloud/cli/history.py
```diff
@@ -0,0 +1,139 @@
...
+    table = []
+    headers = ["Execution ID", "Schedule Name", "Executed At", "Result", "Duration"]
+
+    for execution in history:
+        start_time = datetime.datetime.fromisoformat(execution["start_time"])
+
+        if execution["end_time"]:
+            end_time = datetime.datetime.fromisoformat(execution["end_time"])
+            td = end_time - start_time
+            duration = f"{td.total_seconds() / 60:.2f} minutes"
```
WillDaSilva, line 78: I think I'd prefer seeing the runtime as a whole number of seconds with no unit suffix, and then name the column `Duration (s)` or `Duration (seconds)`. Opinions? @aaronsteers @tayloramurphy

WillDaSilva (commented):
I've tested this locally, and it seems to be working well. All the CLI flags work as expected. 

A slight issue with the `--filter` flag is if I had schedules `daily` and `daily-xyz`, I wouldn't be able to use `--filter` to only get `daily`. That said, the filtering is done on the API side, so that shouldn't be dealt with in this PR.

src/cloud-cli/meltano/cloud/cli/history.py
tayloramurphy: @WillDaSilva I have a preference for something like 6m42s - but we can't really assume runs will all be that short. 

Can we just do `00:06:42` as the format? Seems like that covers most cases.

edgarrmondragon: @taylormurphy yeah, we can do that

src/cloud-cli/meltano/cloud/cli/history.py
```diff
@@ -0,0 +1,139 @@
...
+            [
+                execution["execution_id"],
+                execution["schedule_name"],
+                execution["start_time"],
+                "Success" if execution["exit_code"] == 0 else "Failed",
+                duration,
+            ],
+        )
+
+    return tabulate.tabulate(table, headers, tablefmt="github")
```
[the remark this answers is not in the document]
edgarrmondragon: I'd still like the option to output a markdown table since it can be rendered nicely in a PR comment, for example. We could have both. Wdyt @tayloramurphy @aaronsteers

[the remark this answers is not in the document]
WillDaSilva: > We could have both

I suppose that would be done with a `--table-format` option? That seems good to me, but I'd recommend we not simply pass through the format option to the underlying library, as that would mean that changing the library we use for this would be a breaking change. Instead we can have our own options like `markdown` and `terminal` (or something like that) which map to the options from the library, e.g. `github` and `rounded_outline`.

WillDaSilva (approved):

PR #7432, opened 2023-03-23
feat: Support schedule Cloud CLI options at the group and command level

WillDaSilva:
This PR:
- Updates the existing `meltano cloud schedule enable` command to use the `PUT` API endpoint from https://github.com/meltano/infra/pull/534
- Addresses the [usability concern around group-level CLI options](https://meltano.slack.com/archives/C04V2560TPE/p1679434305660249) identified by @aaronsteers during UAT

Now the following are all valid:

```sh
meltano cloud schedule enable --deployment dev --schedule daily
meltano cloud schedule --schedule daily enable --deployment dev
meltano cloud schedule --deployment dev --schedule daily enable
meltano cloud schedule --deployment dev enable --schedule daily
...
```

This pattern seems like it could be valuable across the Cloud CLI and Meltano CLI in general. For instance, it would be nice to be able to configure the `--log-level` anywhere within a command, rather than only at the top level following `meltano`. 

Two issues I see with this current approach:
- All values that can go into the CLI context must be declared in the shared dataclass, rather than having each group inherit from the dataclass and extend it with their own options.
- If an option is required, one must manually check that it has been set, and raise a `click.UsageError` otherwise. We can't use `click.option(..., required=True)` because then it would have to be specified at every level, rather than on exactly one level of the CLI.


src/cloud-cli/meltano/cloud/api/client.py
```diff
@@ -15,6 +15,11 @@
 from meltano.cloud.api.auth import MeltanoCloudAuth
 from meltano.cloud.api.config import MeltanoCloudConfig
 
+if sys.version_info >= (3, 11):
+    from typing import Self
+else:
+    from typing_extensions import Self
```
edgarrmondragon, line 21: This can go in a `t.TYPE_CHECKING` block afaict

tayloramurphy (approved):

edgarrmondragon (approved):

src/cloud-cli/meltano/cloud/api/client.py
WillDaSilva: Resolved in 4befccd. Thanks!

Issue #7383, opened 2023-03-16
Replace `flakeheaven`, `pyupgrade`, and `isort` with `ruff`


WillDaSilva:
@edgarrmondragon has proven Ruff's value and suitability across several repositories within the Meltano GitHub org.

We should be able to use it to replace `flakeheaven`, `pyupgrade`, and `isort` within the Meltano repository. This should significantly reduce the amount of time it takes for our linters to run. Additionally this will increase the number of lint issues which can be auto-fixed, and will significantly reduce the number of dev dependencies we have.

Some of the Flake8/Flakeheaven lint plugins we use are not yet supported by Ruff:
- `flake8-broken-line`, but Black handles this.
- `flake8-rst-docstrings`, but we use Markdown docstrings.
- `flake8-string-format`, but we primarily use f-strings instead of `str.format`. We use `str.format` in 4 places, and all of those should be switched to use f-strings instead (because they aren't using an `str.format`-exclusive features like unpacking an argument list into the format string).
- `flake8-typing-as-t`. We may want to open an issue https://github.com/charliermarsh/ruff requesting support for this flake8 plugin. [`flake8-import-conventions`](https://beta.ruff.rs/docs/rules/#flake8-import-conventions-icn) is supported, but it doesn't affect `from ... import ...` style import statements, only `import module as bad_alias`.
- `wemake-python-styleguide`. Personally I'm okay with losing the WPS lint rules. I find that they're just as likely to be detrimental to the quality of the code as they are to improve it. We have 452 `noqa` comments within this repository, and 344 of them are for WPS rules. Of the 136 disabled lint rules within `pyproject.toml`, 98 of them are WPS rules. There are definitely plenty of good rules we'd be losing, but the most important/impactful rules tend to find their way into other flake8 plugins and related tools.

We'll lose the Flakeheaven baseline (see https://github.com/charliermarsh/ruff/issues/1149), but Ruff can add `noqa` comments using the `--add-noqa`, so once we've switched over and addressed every new lint issue uncovered that we want to address, we can easily add a noqa comment to the remaining ones. This adds some visual noise compared to a baseline file, but has the benefit of being explicit about what rules are being ignored and where.

Ruff can also automatically remove unused noqa comments, which will make this transition easier since it saves us from having to go through 452 existing noqa comments to see which ones are no longer necessary :sweat_smile:

Ruff rule classes to enable:

```[tasklist]
### Tasks
[...]

PR #7447, opened 2023-03-26
style: Enable Ruff `SIM` (simplify) rules

WillDaSilva:
Relates to:
- #7383


src/meltano/cli/elt.py
```diff
@@ -95,7 +95,7 @@
 @click.pass_context
 @pass_project(migrate=True)
 @click_run_async
-async def elt(
+async def elt(  # noqa: WPS 408
```
edgarrmondragon, line 98:
```suggestion
async def elt(  # WPS408
```

src/meltano/core/block/future_utils.py
```diff
@@ -12,10 +12,7 @@
 
 def all_done(tasks: list[Task], done: set[Task]) -> bool:
     """Iterate through a task list checking if ALL tasks are in the done set."""
-    for idx in tasks:
-        if idx not in done:
-            return False
-    return True
+    return all(idx in done for idx in tasks)
```
edgarrmondragon, line 15: This is really nice!

edgarrmondragon (changes requested):

edgarrmondragon (approved):

PR #7463, opened 2023-03-30
feat: Add `cloud project list` and `cloud project use` commands

### LINKED ISSUE #7411: Cloud CLI to list projects, describe projects, and switch project contexts
cc @tayloramurphy 

Related to:

- https://github.com/meltano/infra/issues/513 (internal)
- https://github.com/meltano/meltano/issues/7412

WillDaSilva:
Closes #7411


docs/src/_reference/cloud-cli.md
```diff
@@ -25,6 +25,41 @@ Logging out of Meltano Cloud invalidates your login token, and deletes the local
...
+meltano cloud project activate 'Meltano Squared'
+Activated Meltano Cloud project 'Meltano Squared'
+```
+
+```sh
+meltano cloud project list
+╭──────────┬───────────────────────────────┬──────────────────────────────────────────────────────────╮
+│  Active  │ Name                          │ Git Repository                                           │
+├──────────┼───────────────────────────────┼──────────────────────────────────────────────────────────┤
+│    X     │ Meltano Squared               │ https://github.com/meltano/squared.git                   │
```
tayloramurphy, line 56: @WillDaSilva @aaronsteers we'd talked somewhere about having the referenceable name for a project to always be a lower case slug with just a-z and dashes (maybe digits too?) We can still have a pretty name, but keeping the slugified name would make the CLI syntax easier so quotes aren't necessary. 

That way the above command could be `meltano cloud project activate meltano-squared` (or whatever the slug is).

docs/src/_reference/cloud-cli.md
```diff
@@ -25,6 +25,41 @@ Logging out of Meltano Cloud invalidates your login token, and deletes the local
...
+├──────────┼───────────────────────────────┼──────────────────────────────────────────────────────────┤
+│          │ Meltano Squared               │ https://github.com/meltano/squared.git                   │
+│          │ MDS-in-a-box                  │ https://github.com/aaronsteers/meltano-demo-in-a-box.git │
+╰──────────┴───────────────────────────────┴──────────────────────────────────────────────────────────╯
+```
+
+When you run the `login` command, if you only have a single project, it will be activated by default. Otherwise, you will need to run `meltano cloud project activate` to specify which Meltano Cloud project the other `meltano cloud` commands should operate on.
+
+```sh
+meltano cloud project activate 'Meltano Squared'
```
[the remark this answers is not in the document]
WillDaSilva: I'm not fond of `switch` because it doesn't make sense when no project is active, and it feels a bit like something is being swapped.

@tayloramurphy @aaronsteers What do you think of `meltano cloud project use <project name>`?

docs/src/_reference/cloud-cli.md
tayloramurphy: @WillDaSilva makes sense to me. Will you make an issue around this so we don't lose track?

docs/src/_reference/cloud-cli.md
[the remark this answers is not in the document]
tayloramurphy: `use` could work! The other one that came to mind is `set-default` but I think that's a bit too heavy. @aaronsteers ?

docs/src/_reference/cloud-cli.md
WillDaSilva: - https://github.com/meltano/infra/issues/582

With this API change made, no changes should be required on the CLI side. We may want to add a doc update to say that normalization is performed on the name, so e.g. it isn't case sensitive, but this isn't strictly necessary since the existing docs won't become incorrect because of this change.

src/cloud-cli/meltano/cloud/cli/project.py
```diff
@@ -0,0 +1,221 @@
...
+@project_group.command("activate")
+@click.argument("identifier")
+@pass_context
+@run_async
+async def activate_project(
+    context: MeltanoCloudCLIContext,
+    identifier: str,
+) -> None:
+    """Set a project as active within the current Meltano Cloud session."""
+    projects = await _get_project_to_activate(context.config, identifier)
```
edgarrmondragon, line 205: Should we make the `identifier` arg optional and the give the user an interactive prompt to select a project if it's not set?

WillDaSilva: @edgarrmondragon I agree that an interactive prompt would be nice, and it can be done using `@click.option(..., prompt=True)`. I'd like to combine this with `type=click.Choice`, but that presents 2 challenges:

- The interactive prompt for a `click.Choice` isn't great. By default it looks like this (for an option `--name` with 3 possible values `option 1`, `option 2` and `option 3`):
  ```
  Name (option 1, option 2, option 3): 
  ```
  I'd prefer it print the options as a numbered list, and then let the user enter the number corresponding to the option they'd like to select.
- `click.Choice` needs the option when `meltano.cloud.cli.project` is run, rather than when the `project activate` function is run. It would need to be extended to take a callback function that get the possible choices. Then that callback function could make the API call to get the available projects.

WillDaSilva: Perhaps [`questionary`](https://github.com/tmbo/questionary) can help here, as detailed in this StackOverflow answer: https://stackoverflow.com/a/60425679/5946921

edgarrmondragon: Yeah, `questionary` looks like the Python equivalent of Golang's [Survey](https://github.com/go-survey/survey), which is used by `gh` and other CLIs.

I took a quick look at [Rich's support for prompts](https://rich.readthedocs.io/en/stable/prompt.html) and it's not much better than Click's.

edgarrmondragon: cc @kgpayne since he's worked with nice Rich prompts

docs/src/_reference/cloud-cli.md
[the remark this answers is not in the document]
aaronsteers: I didn't previously gave a string opinion on this, but you both make good points.

I can see `switch` being confusing if you haven't yet activated anything, and I can see an interpretation where `activate` would mean something like "deploy". 

I think "use" is nice: it's intuitive and hard to misinterpret.

src/cloud-cli/meltano/cloud/cli/project.py
WillDaSilva: Implemented in 033c58f. This was somewhat complex, and difficult to test, but the user experience is vastly improved!

docs/src/_reference/cloud-cli.md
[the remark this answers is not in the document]
WillDaSilva: The `activate` command has been renamed to `use` in 033c58f

src/cloud-cli/meltano/cloud/cli/project.py
```diff
@@ -0,0 +1,291 @@
...
+                if x["project_id"] == context.config.default_project_id
+            ),
+            {"project_name": None},
+        )["project_name"]
+        return questionary.select(
+            message="",
+            qmark="Use Meltano Cloud project",
+            choices=[x["project_name"] for x in context.projects],
+            default=default_project_name,
+        ).unsafe_ask()
```
edgarrmondragon, line 229: According to the docs this

<blockquote>

Does not catch keyboard interrupts.

</blockquote>

However, in my testing Cmd+C does seem to abort the prompt so I wonder if this is handled by Click.

edgarrmondragon (approved):

src/cloud-cli/meltano/cloud/cli/project.py
WillDaSilva: Yes, it is being handled by Click. I should add a comment to explain why I'm using `unsafe_ask`, since the name is pretty suspicious. If you use `ask` instead, it still aborts execution, but with a different message than what Click provides.

tayloramurphy (approved):
This looks great!

aaronsteers:
@WillDaSilva and @tayloramurphy 

Another alternative occurred to me besides `switch` and `use`.

What do you both think of `project select`. I think I like this a tiny bit more than `use` for the interactive use case. "Select" seems to pair well with "list" and (to me at least) seems to work well whether or not a context is already set.

Caveat: I'm behind on reading this PR and other threads. Feel free to push back if it's not a good fit. I like "use" and "list" a lot also.

tayloramurphy:
> What do you both think of project select. I think I like this a tiny bit more than use for the interactive use case. "Select" seems to pair well with "list" and (to me at least) seems to work well whether or not a context is already set, and whether or not we're directly providing the project handle at the cli or expecting the user to pick it interactively.

@aaronsteers I don't like `select` due to the collision with [pattern selection](https://docs.meltano.com/reference/command-line-interface#select). The action we're trying to accomplish with `meltano cloud project use` is to set the default project to which other commands are being applied. `use` works well for that case, even in the interactive mode. 

I could imagine a situation where we re-use `use` on a config level (I don't think we'll do this but just for arguments sake). If a user did `meltano use tap-gitlab` we could set `tap-gitlab` as the "default plugin" so future commands like `meltano select --list --all` or `meltano config set ...` would work without providing the plugin selection. Again, we won't do this, but just laying out some of my logic here. 

I'd rather we reserve `select` in the cloud context for something more similar to pattern selection and go with `use` for the project switching use case. @WillDaSilva

PR #7467, opened 2023-03-31
feat: Allow history filtering by schedule exact name, prefix and infix

edgarrmondragon:


src/cloud-cli/meltano/cloud/cli/history/command.py
```diff
@@ -47,12 +47,60 @@ def convert(
...
+        schedule_contains: The schedule name substring to match.
+
+    Returns:
+        The glog expression to use for filtering.
+
+    Raises:
+        click.UsageError: More than one filter was specified.
+    """
+    # If more than one filter is specified, fail.
+    if sum([1 for f in (schedule, schedule_prefix, schedule_contains) if f]) > 1:
```
WillDaSilva, line 69:
```suggestion
    if sum(1 for f in (schedule, schedule_prefix, schedule_contains) if f) > 1:
```

WillDaSilva (approved):

docs/src/_cloud/cloud-cli.md
```diff
@@ -89,6 +89,15 @@ $ meltano cloud history --lookback 1w
 
 # Display the last hour and a half of executions
 $ meltano cloud history --lookback 1h30m
+
+# Display executions exactly matching a schedule name
+$ meltano cloud history --schedule daily_github_to_snowflake
+
+# Display executions matching a schedule name prefix
+$ meltano cloud history --schedule-prefix daily_
```
tayloramurphy, line 97: @edgarrmondragon what do you think about following the precedent we have with `state`: https://docs.meltano.com/reference/command-line-interface#how-to-use-17

`meltano cloud history --pattern <>` 

or we could have `meltano cloud history --schedule <string that accepts * as wildcards>`

I'm just worried that having 3 different ways for access the same set of strings is confusing.

tayloramurphy: @WillDaSilva understood - but we have the precedent in state previously. Can we require quotes? Is there another option besides the asterisk?

```
# List only those state IDs that start with "dev:"
meltano state list 'dev:*'

# List only those state IDs that contain "tap-gitlab"
meltano state list --pattern '*tap-gitlab*'
```

tayloramurphy: @WillDaSilva ok, I'm coming around to it. 

Will you make an issue and/or discussion around moving to more resilient patters for `state list`? We're prioritizing state management on Cloud very soon so it would be good to get alignment the patterns we should standardizing on. This would also warrant an update to our CLI guidelines! cc @aaronsteers @seth-meltano

WillDaSilva: @tayloramurphy Sounds good. I've logged https://github.com/meltano/meltano/issues/7497, which currently includes both updating `meltano state list`, and updating the CLI guidelines. We can split out the latter into its own issue if you'd rather that doc update not be tied to the CLI update.

tayloramurphy (approved):

PR #7469, opened 2023-03-31
docs: Add Cloud Fees stub

DouweM:


docs/src/_cloud/fees.md
WillDaSilva, line 1: What do you think about calling this page "pricing"?

WillDaSilva (dismissed):

docs/src/_cloud/fees.md
```diff
@@ -0,0 +1,12 @@
+---
+title: "Cloud Fees"
+hidden: true
```
edgarrmondragon, line 3: This doesn't seem to exclude it from the ToC nor search. Is that expected?

Screenshot 2023-03-31 at 14 46 04
Screenshot 2023-03-31 at 14 45 48

DouweM: @edgarrmondragon Good catch, I just copied what we had for cloud/sla etc. It's not a huge problem if it shows up in those places, as long as it's not in the sidebar.

docs/src/_cloud/fees.md
DouweM: I'm using "Fees" to match Common Paper, which we're using to generate the contract:

Screen Shot 2023-03-31 at 15 53 51

This will have the legalese version of the human-readable https://meltano.com/pricing, and Fees sounds more legalese-y to me :)

docs/src/_cloud/fees.md
```diff
@@ -0,0 +1,12 @@
...
+---
+<div class="notification is-info">
+  <p><strong>Meltano Cloud is currently in Beta.</strong></p>
+  <p>While in Beta, functionality is not guaranteed and subject to change. <br> If you're interested in using Meltano Cloud please join our <a href="https://meltano.com/cloud/">waitlist</a>.</p>
+</div>
+
+## Meltano Cloud Fees
+
+Todo
```
tayloramurphy, line 12: @DouweM I'm not clear why we need a docs page for this when we have `meltano.com/pricing`. But I don't want to block on it.

docs/src/_cloud/fees.md
edgarrmondragon: @DouweM cool. Also, [it looks like this](https://deploy-preview-7469--meltano.netlify.app/cloud/fees)
Screenshot 2023-03-31 at 14 53 52

If that's not intended, I think `layout: doc` needs to be added to the frontmatter.

tayloramurphy (approved):

docs/src/_cloud/fees.md
WillDaSilva: Makes sense

PR #7471, opened 2023-04-01
chore: Use `pytest-httpserver` for Cloud CLI tests

### LINKED ISSUE #7468: Use `httpserver` Pytest fixture for Cloud CLI tests
The dependency on `pytest-httpserver` was introduced in:
- https://github.com/meltano/meltano/pull/7463

We can simplify our tests for the Cloud CLI by using it instead of `aioresponses`. With it, we can have a fixture which always handles auth, rather than having each test re-declare handling for that HTTP request. We can also avoid nesting code within the with block required by `aioresponses`.

To use `pytest-httpserver`, the Cloud CLI test config should be updated to use `localhost` for its auth and API base URLs.

WillDaSilva:
Closes #7468


src/cloud-cli/tests/cli/test_history.py
```diff
@@ -197,37 +164,35 @@ def test_filter_schedule(
...
-            pytest.param("1w2d", "2023-05-11T00:00:00+00:00", id="1w2d"),
-            pytest.param("1w2d3h", "2023-05-10T21:00:00+00:00", id="1w2d3h"),
-            pytest.param("1w2d3h4m", "2023-05-10T20:56:00+00:00", id="1w2d3h4m"),
+            pytest.param("1w", "2023-05-13T00:00:00 00:00", id="1w"),
+            pytest.param("1d", "2023-05-19T00:00:00 00:00", id="1d"),
+            pytest.param("1h", "2023-05-19T23:00:00 00:00", id="1h"),
+            pytest.param("1m", "2023-05-19T23:59:00 00:00", id="1m"),
+            pytest.param("1w2d", "2023-05-11T00:00:00 00:00", id="1w2d"),
+            pytest.param("1w2d3h", "2023-05-10T21:00:00 00:00", id="1w2d3h"),
+            pytest.param("1w2d3h4m", "2023-05-10T20:56:00 00:00", id="1w2d3h4m"),
```
edgarrmondragon, line 182: aiohttp and werkzeug seem to handle percent-encoding differently so it's better to rely on `pytest-httpserver` matching capabilities: https://github.com/meltano/meltano/pull/7477

edgarrmondragon (approved):

PR #7475, opened 2023-04-02
refactor: Run local auth server within same process

### LINKED ISSUE #7468: Use `httpserver` Pytest fixture for Cloud CLI tests

### LINKED ISSUE #7470: chore: run the auth callback server via, e.g., aiohttp rather than flask
We currently run the auth callback server via a dead-simple subprocess `flask` command call. This has already proven to be a bit fragile and difficult to test. Running the server async in the same process is a cleaner, more robust, and more testable solution.

WillDaSilva:
A couple of fixes for `meltano cloud` on Windows are also included in this PR.

Closes #7470


src/cloud-cli/meltano/cloud/api/auth/auth.py
```diff
@@ -3,18 +3,20 @@
...
 import webbrowser
-from contextlib import contextmanager
+from contextlib import asynccontextmanager
+from http import HTTPStatus
 from pathlib import Path
 from urllib.parse import urlencode, urljoin
 
 import aiohttp
 import click
+import importlib_resources
```
edgarrmondragon, line 17: Wdyt about falling back to `importlib.resources` for Python 3.9+:

```python

if sys.version_info < (3, 9):
    import importlib_resources
else:
    import importlib.resources as importlib_resources
```

That way, it'll eventually be auto-removed by Ruff/pyupgrade.

cjohnhanson (approved):
Looks good! I didn't do any kind of formal benchmarking/profiling, but definitely felt faster than the flask approach when I tested locally (which would make sense). Thanks for taking this on.

src/cloud-cli/meltano/cloud/api/auth/auth.py
WillDaSilva: Implemented in 9ac790e. Had to make a change from what you suggested to satisfy WPS301.

edgarrmondragon (approved):

PR #7491, opened 2023-04-04
chore: Remove unused extras `infra` and `repl`

WillDaSilva:
We have no dependency on `ansible` or `ipython`. These extras installs without error, but is are different from installing Meltano without them.


edgarrmondragon:
Should we ask around just in case anybody is installing `meltano[infra]`? I'm pretty sure there's no one but 🤷‍♂️

WillDaSilva:
> Should we ask around just in case anybody is installing meltano[infra]? I'm pretty sure there's no one but :man_shrugging:

I'll give people a heads-up in #engineering-chatter

Beyond that, I think it'll be sufficient to call this out in the release notes, not as a breaking change, but as a warning. If someone hits an error because of this, the solution is fortunately very simple: remove the useless extra.

@kgpayne Could you please make a reminder to call this out in the release notes for the next Meltano release you run?

WillDaSilva:
Warning posted in Slack: https://meltano.slack.com/archives/CKHP6G5V4/p1680634915518539

edgarrmondragon (approved):

PR #7493, opened 2023-04-05
chore: Remove cyclic imports in `meltano.cli`

WillDaSilva:
This approach of using `click.command` with `cli.add_command` is currently used by the Cloud CLI. This PR applies this approach to the rest of Meltano in order to simplify the imports in the `meltano.cli` module. Now there are no more cyclic dependencies, and we can let isort (Ruff) handle the imports.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -577,6 +577,7 @@ select = [
   "PT",  # flake8-pytest-style
   "RSE", # flake8-raise
   "SIM", # flake8-simplify
+  "TID", # flake8-tidy-imports
   "UP",  # pyupgrade
   "W",   # pycodestyle (warning)
 ]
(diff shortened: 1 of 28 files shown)


edgarrmondragon (approved):
Very nice!

Should we enforce this with [`banned-api`](https://beta.ruff.rs/docs/rules/banned-api/) like we did for the cloud CLI?

pyproject.toml
```diff
@@ -600,3 +601,7 @@ required-imports = ["from __future__ import annotations"]
 
 [tool.ruff.pydocstyle]
 convention = "google"
+
+[tool.ruff.flake8-tidy-imports.banned-api]
+"meltano.cli.cli.command".msg = "Use `click.command` and `cloud.add_command` instead"
+"meltano.cli.cli.group".msg = "Use `click.group` and `cloud.add_command` instead"
```
edgarrmondragon, line 607:
```suggestion
"meltano.cli.cli.command".msg = "Use `click.command` and `meltano.cli.cli.add_command` instead"
"meltano.cli.cli.group".msg = "Use `click.group` and `meltano.cli.cli.add_command` instead"
```

PR #7499, opened 2023-04-07
break: Remove API and web UI

### LINKED ISSUE #6959: Remove the deprecated UI and API
In https://github.com/meltano/meltano/pull/6955 we removed the documentation around the UI. At some point we can slim down the codebase and the backend database by removing that code as well.

WillDaSilva:
This change reduces the number of runtime Python dependency Meltano has from 103 to 66.

Closes #6959


src/meltano/core/setting_definition.py
```diff
@@ -170,7 +170,6 @@ def __init__(
         options: list | None = None,
         oauth: dict | None = None,
         placeholder: str | None = None,
-        protected: bool | None = None,
```
edgarrmondragon, line 173: There are plugins in the Hub that still use `protected: true`, should we keep it until we update those?

FWIW, there seem to be only 2 plugins using it: https://github.com/search?q=repo%3Ameltano%2Fhub%20protected&type=code

WillDaSilva: Good callout. The setting no longer does anything (and has a confusing name), so I'm tempted to remove it while we're making this already-breaking change.

Do you think it'd be safe to remove that setting from those plugins on the Hub ahead of the Meltano 3.0 release?

fyi @pnadolny13

edgarrmondragon: > Do you think it'd be safe to remove that setting from those plugins on the Hub ahead of the Meltano 3.0 release?

Yes, I think so.

edgarrmondragon: Started https://github.com/meltano/hub/pull/1271 to get the ball rolling on the Hub side

edgarrmondragon (approved):
I manually tested a few commands and I couldn't find any regressions :+1:

PR #7537, opened 2023-04-13
fix: Add correct type for state_backend lock settings

### LINKED ISSUE #7525: bug: State backend lock settings parsed as strings
### Meltano Version

2.16

### Python Version

3.9

### Bug scope

### Operating System

Linux - ECS, in meltano/meltano:v2.16 image

### Description

When `state_backend.lock_timeout_seconds` is explicitly set, its value gets parsed as a string, which causes an error when initializing the relevant `timedelta` during lock acquisition.

### Code

_No response_

cjohnhanson:
Also reduces default lock timeout to a more reasonable 10 seconds.

Closes #7525


WillDaSilva (approved):

WillDaSilva:
@cjohnhanson You'll have to regenerate the sample manifest files to get the integration tests passing.

aaronsteers (approved):

Issue #7001, opened 2022-11-17
feature: Optionally install only what is necessary


WillDaSilva:
### Feature scope

CLI (options, error messages, logging, etc.)

### Description

This came up in today's engineer code-sharing call. Inspired by [`ci_get_install_script.py`](https://github.com/meltano/squared/blob/c5b68af73f679c329188f3d2bbf0270a19606181/data/ci_get_install_script.py) (used [here](https://github.com/meltano/squared/blob/c5b68af73f679c329188f3d2bbf0270a19606181/.github/workflows/test.yml#L105)), it'd be great both for Meltano Cloud and the new-user experience if it were possible to have `meltano install` optionally install only what is required to run a particular job, and to have `meltano run` optionally install the plugins it needs to run a given job. Similarly we could implement this for `meltano elt` and `meltano schedule run`.

This would involve implementing:
- [ ] `meltano install --job <job name>`
- [ ] `meltano run --install`
- [ ] `meltano elt --install`
- [ ] `meltano schedule run --install`.

This can be split up into multiple issues, but I wanted to start with the proposal all together to elicit feedback.

tayloramurphy:
@pnadolny13 I believe you had an issue that was similar to this already, right? Are you able to find it?

edgarrmondragon:
There a few related issues I could find:

- https://github.com/meltano/meltano/issues/3261
- https://github.com/meltano/meltano/issues/3167

pnadolny13:
Right now schedules are 1-1 with jobs so `meltano install --job <job name>` is effectively completed by https://github.com/meltano/meltano/pull/7594.

edgarrmondragon:
With https://github.com/meltano/meltano/pull/8633, `meltano run --only-install <job name>` could be used instead of adding a new option for `meltano install --job <job name>`.

PR #7594, opened 2023-04-24
feat: Install plugins using schedule name

pnadolny13:
Related to https://github.com/meltano/meltano/issues/7001

Adds the ability to install plugins in a given schedule, helpful for workers that dont have preinstalled plugins. Personally this would be beneficial to me because it would eliminate the need for my CI install script that attempts to do the same thing [ci_get_install_script.py](https://github.com/meltano/squared/blob/main/data/ci_get_install_script.py) and also speeds up my cloud schedule startup time. I have 34 plugins to install before every cloud run so for a bunch of my schedules the install time is the same or more than my actual execution time and I think this would cut it down significantly.

Notes:
- The logic is added after the normal plugin type filtering so you can still filter for "extractors in schedule xyz"
- I confirmed that this works locally but when I went to write unit tests I wasnt quite sure how this fits in. Before I spend time on that I figured I'd get some feedback if this logic is better suited somewhere else (e.g. something along the lines of a get_job_plugins in ScheduleService or TaskSetService).


src/meltano/cli/install.py
```diff
@@ -44,6 +45,11 @@
     is_flag=True,
     help="Ignore the required Python version declared by the plugins.",
 )
+@click.option(
+    "--schedule_name",
```
tayloramurphy, line 49: Very cool PR @pnadolny13 ! Add a docs update around this as well too please 😄

pnadolny13:
@WillDaSilva @kgpayne @cjohnhanson any thoughts on whether this is the wrong place for this? I'm happy to make any changes and add tests but didnt want to spend too much time until I heard your thoughts since I'm not super familiar with the codebase.

WillDaSilva (commented):
@pnadolny13 This feature is great! A simple way for us to massively improve Meltano Cloud performance before we get around to creating images for each deployment, or cached plugin venvs/zipapps.

> any thoughts on whether this is the wrong place for this?

Ideally we wouldn't have any of this application logic in the CLI function, but it's already the case that many of the Meltano CLI functions have lots of application logic, and that's already the case in this specific CLI function, so I don't really care right now. Holding this improvement back so that we can refactor & clean up the CLI is not worth it, and it's not like this feature would make it more complicated to perform that refactoring / clean up later either.

pnadolny13:
@WillDaSilva thanks for the guidance. I updated this PR with a passing unit test, its kind of ugly because I feel like I'm forcing this code block into the CLI function but it does the job. I also added a few lines to the cli reference docs.

tests/meltano/cli/test_install.py
```diff
@@ -236,6 +236,42 @@ def test_clean_install(
...
+            install_plugin_mock.return_value = True
+            schedule_service.task_sets_service = task_sets_service
+            from meltano.core.task_sets import TaskSets
+
+            task_sets_service.add(
+                TaskSets(job_schedule.job, [tap_gitlab.name, target.name]),
+            )
+            result = cli_runner.invoke(
+                cli,
+                ["install", "--schedule_name", job_schedule.name],
```
WillDaSilva, line 263:
```suggestion
                ["install", "--schedule", job_schedule.name],
```

docs/src/_reference/command-line-interface.md
```diff
@@ -628,6 +628,9 @@ Installs dependencies of your project based on the **meltano.yml** file.
 Optionally, provide a plugin type argument to only (re)install plugins of a certain type.
 Additionally, plugin names can be provided to only (re)install those specific plugins.
 
+To only install plugins for a particular schedule specify the `--schedule_name` argument.
```
pnadolny13, line 631:
```suggestion
To only install plugins for a particular schedule specify the `--schedule` argument.
```

WillDaSilva (approved):

PR #7600, opened 2023-04-25
chore: Add initial stale bot config

### LINKED ISSUE #7328: Set up GitHub stale bot
https://github.com/marketplace/stale

As discussed in the engineering code share call on 2023-02-16, we'd want to (at least initially) configure the bot to have a light touch, and to avoid shutting out further discussion.

Example of what not to do: https://drewdevault.com/2021/10/26/stalebot.html

We can make use of `exemptLabels` to save ourselves from having to repeatedly tell the bot that an issue/PR should not be closed.

The GitHub application is configured by the file `.github/stale.yml`. Our config for it might be something like this:

```
# Number of days of inactivity before an issue becomes stale
daysUntilStale: 120

# Number of days of inactivity before a stale issue is closed
daysUntilClose: 14

# Set to true to ignore issues in a project (defaults to false)
exemptProjects: false

# Set to true to ignore issues in a milestone
exemptMilestones: true

# Set to true to ignore issues with an assignee
exemptAssignees: true

# Issues/PRs with these labels will never be considered stale
exemptLabels:
  - evergreen

# Label to use when marking an issue as stale
staleLabel: stale

# Comment to post when marking an issue as stale. Set to `false` to disable
markComment: >
  This has been marked as stale because it is unassigned, and has not had
  recent activity. It will be closed after 14 days if no further activity occurs. If
  this should never go stale, please add the `evergreen` label, or request
  that it be added.

# Comment to post when closing a stale issue. Set to `false` to disable
closeComment: false
```

WillDaSilva:
This will not lock issues, so if an issue has been closed as stale, but someone from outside the Meltano team would like to reopen it, they need only add a comment to the issue to ask that it be reopened.

Closes #7328


edgarrmondragon (approved):
Nice, not too aggressive!

WillDaSilva:
@edgarrmondragon Good idea. We can create a centralized config file in `meltano/.github` (or any other repository), and add `.github/stale.yml` to each repository we want it to operate on that extends that central config: https://github.com/probot/stale/issues/141#issuecomment-402006522

I don't believe there's any way to have a single stalebot config file apply to multiple repositories.

I'll move this config over to `.github`, and update this PR to refer to that config file.

edgarrmondragon (approved):

PR #7603, opened 2023-04-25
fix: add build workflow permissions

kgpayne:


WillDaSilva:
@kgpayne
- https://github.com/meltano/meltano/pull/7598/files

PR #7608, opened 2023-04-25
chore: add publishing environment to build.yml pypi publish

kgpayne:


kgpayne:
@WillDaSilva @edgarrmondragon FYI I removed the `environment` value from our Trusted Publisher config in PyPI to get `v2.18.0` to publish without this change, and then added it back (to `publishing`, as before) immediately afterwards. This saved me another go around the revert and version bump workflow run cycle ☺️

WillDaSilva (approved):
@kgpayne Nice workaround to save us from another unpublish republish cycle.

edgarrmondragon (approved):

PR #7639, opened 2023-05-03
docs: Promote `pipx` to install Meltano

### LINKED ISSUE #6882: docs: Encourage `pipx` more strongly in installation guide
I love the new installation guide from @sbalnojan . However, one thing we lost in the transition to the new format was the prioritization of `pipx` as the preferred installation method - and the instructions to leverage virtual environments if `pip` is used directly.

1. I would suggest we change the order of the tabs from [1: `pip`, 2: `docker`, 3: `pipx`] to instead use: [1: `pipx`, 2: `pip`, 3: `docker`]. People who want to do a 'quick test' and don't want to install pipx can still just jump over to the `pip` tab, and they can ignore the warnings/recommendations, but then if they do run into version conflicts, they remember ignoring the warnings and they know to come back and run it in one of the recommended methods.

2. The `pipx` section links to the `pipx` homepage, but it could also be helpful to link to the 'how to install pipx' content directly here: https://docs.meltano.com/guide/installation-guide#install-pipx

3. The `pip` section also currently suggests a simple installation, which on its own is very likely to lead to version conflicts. I think we should definitely recommend virtual environments on the `pip` tab, ideally with a warning about version conflicts, suggesting pip+virtualenv or pipx as the recommended path to avoid conflicts with other installed tools and libraries. To avoid blowing up the page with `virtualenv` explanation and instructions there on the `pip` tab, I think we could just add something like this with links to content off the page:

> Warning: We strongly recommend you use `pipx` for installation. If you choose to use `pip` directly, we recommend creating a dedicated virtual environment to avoid conflicts with other installed software. [Learn more here.](https://docs.meltano.com/concepts/python_virtual_environments#what-are-python-virtual-environments)

cc @tayloramurphy

WillDaSilva:
Closes https://github.com/meltano/meltano/issues/6882

### DIFF
--- a/docs/src/_getting-started/installation.md
+++ b/docs/src/_getting-started/installation.md
@@ -24,11 +24,11 @@ Meltano supports all versions of Python that have not yet reached their end-of-l
 
 {% tabs log %}
 
-{% tab log Using `pipx` %}
+{% tab log Using pipx %}
 
 Since Meltano is an application, it should always be installed into a clean virtual environment without any other packages installed alongside it.
 
-To simplify installing Meltano into a clean virtual environment, `pipx` can be used to install Meltano instead of `pip`. `pip` is a package installer that comes automatically with Python. [`pipx`](https://pypa.github.io/pipx/) is a wrapper around pip which cleanly installs executable Python applications (such as Meltano) into their own virtual environments.
+To simplify installing Meltano into a clean virtual environment, [`pipx`](https://pypa.github.io/pipx/) can be used to install Meltano instead of `pip`. `pip` is a package installer that comes automatically with Python. `pipx` is a wrapper around pip which cleanly installs executable Python applications (such as Meltano) into their own virtual environments.
 
 <div class="termy">
 
@@ -64,7 +64,7 @@ meltano, version 2.19.0
 
 {% endtab %}
 
-{% tab log Using `pip` %}
+{% tab log Using pip %}
 
 Meltano provides a Python package that can be installed using `pip`.
 
--- a/docs/src/_getting-started/installation.md
+++ b/docs/src/_getting-started/installation.md
@@ -10,20 +10,67 @@ Before you can use Meltano, you’ll need to get it installed. We have a [comple
 
 ## Install Meltano
 
-
-
-{% tabs log %}
-
-{% tab log Using Pip %}
-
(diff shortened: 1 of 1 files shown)


docs/src/_getting-started/installation.md
```diff
@@ -10,27 +10,30 @@ Before you can use Meltano, you’ll need to get it installed. We have a [comple
...
+{% tab log Using `pipx` %}
 
-Use the following command to check that you have a supported Python version installed:
+Since Meltano is an application, it should always be installed into a clean virtual environment without any other packages installed alongside it.
 
-```bash
-python --version
-```
-Currently Python 3.7, 3.8, 3.9, 3.10 and 3.11 are supported. Some plugins do not yet support Python 3.11.
+To simplify installing Meltano into a clean virtual environment, `pipx` can be used to install Meltano instead of `pip`. `pip` is a package installer that comes automatically with Python. [`pipx`](https://pypa.github.io/pipx/) is a wrapper around pip which cleanly installs executable Python applications (such as Meltano) into their own virtual environments.
```
edgarrmondragon, line 31: I tried clicking the first instance of `pipx` and noticed it's not linked to the pipx site. I guess that's what most folks reading this for the time would try to do

```suggestion
To simplify installing Meltano into a clean virtual environment, [`pipx`](https://pypa.github.io/pipx/) can be used to install Meltano instead of `pip`. `pip` is a package installer that comes automatically with Python. `pipx` is a wrapper around pip which cleanly installs executable Python applications (such as Meltano) into their own virtual environments.
```

edgarrmondragon (approved):

docs/src/_getting-started/installation.md
```diff
@@ -10,27 +10,30 @@ Before you can use Meltano, you’ll need to get it installed. We have a [comple
...
+
+<div class="notification is-warning">
+  <p>Not all plugins support the latest versions of Python.</p>
+</div>
 
-{% tab log Using Pip %}
+{% tabs log %}
 
-Meltano is pip-installable.
+{% tab log Using `pipx` %}
```
tayloramurphy, line 27:
```suggestion
{% tab log Using pipx %}
```

The backticks don't render in the header.

tayloramurphy (approved):
2 minor changes

PR #7640, opened 2023-05-03
fix: Address pagination loop in `meltano-cloud logs print` command

edgarrmondragon:
See https://github.com/meltano/infra/issues/857

### DIFF
--- a/src/cloud-cli/meltano/cloud/cli/logs.py
+++ b/src/cloud-cli/meltano/cloud/cli/logs.py
@@ -87,7 +87,7 @@ class LogsClient(MeltanoCloudClient):
                 pagination = response.get("pagination") or {}
                 new_page_token = pagination.get("next_page_token")
 
-                if not new_page_token or new_page_token == page_token:
+                if new_page_token == page_token:
                     break
 
                 page_token = new_page_token
--- a/src/cloud-cli/meltano/cloud/cli/logs.py
+++ b/src/cloud-cli/meltano/cloud/cli/logs.py
@@ -84,10 +84,13 @@ class LogsClient(MeltanoCloudClient):
                 response = await self._get_logs_page(execution_id, page_token)
                 yield response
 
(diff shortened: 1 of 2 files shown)


src/cloud-cli/meltano/cloud/cli/logs.py
```diff
@@ -84,11 +84,14 @@ async def get_logs(
                 response = await self._get_logs_page(execution_id, page_token)
                 yield response
 
-                pagination = response.get("pagination")
-                page_token = pagination.get("next_page_token") if pagination else None
-                if not page_token:
+                pagination = response.get("pagination") or {}
+                new_page_token = pagination.get("next_page_token")
+
+                if not new_page_token or new_page_token == page_token:
```
WillDaSilva, line 90:
```suggestion
                if new_page_token == page_token:
```

WillDaSilva (approved):
@edgarrmondragon Does this same bug exist with the other places in the Cloud CLI that use pagination?

src/cloud-cli/meltano/cloud/cli/logs.py
edgarrmondragon: > @edgarrmondragon Does this same bug exist with the other places in the Cloud CLI that use pagination?

@WillDaSilva No, so far only the [`get_log_events`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/logs/client/get_log_events.html) endpoint works like this. DynamoDB pagination works as expected.

edgarrmondragon:
> @edgarrmondragon Does this same bug exist with the other places in the Cloud CLI that use pagination?

@WillDaSilva No, so far only the [`get_log_events`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/logs/client/get_log_events.html) endpoint works like this. DynamoDB pagination works as expected.

PR #7654, opened 2023-05-10
feat: Command to cancel a running job from the CLI

### LINKED ISSUE #7560: Feature (Cloud CLI): Command to cancel a running job from the CLI
### Feature scope

### Description

Proposed CLI:

```
meltano cloud job stop --execution-id=<...>
```

See https://github.com/meltano/infra/issues/357

edgarrmondragon:
Docs: https://github.com/meltano/meltano/pull/7668


tayloramurphy (commented):
@edgarrmondragon what's your plan for docs changes related to this PR?

@pnadolny13 would you do some user acceptance testing on our Squared project for this PR?

edgarrmondragon:
> @edgarrmondragon what's your plan for docs changes related to this PR?
> 
> @pnadolny13 would you do some user acceptance testing on our Squared project for this PR?

@tayloramurphy I opened https://github.com/meltano/meltano/pull/7668 (doesn't make a lot of sense to add them here since they're only deployed from the `main` branch)

pnadolny13:
@tayloramurphy @edgarrmondragon I tested this and it worked as expected for me.

Some feedback as a user though:
- I got a bit tripped up on the difference between jobs vs schedules vs executions again. I dont necessarily have a recommendation but the workflow of: run schedule, list schedule execution history, print execution logs, stop job using execution id. I think jobs and schedules are the same in this context so it feels like schedule = job and job = execution.
- With my previous comment in mind, we might want to consider changing these job command to refer to "job executions" because we're interacting with execution, not jobs themselves. Maybe the help text `Interact with Meltano Cloud jobs.` to `Interact with Meltano Cloud job executions.`. 
- The `stop` help text looks like its wrong. It says ` List Meltano Cloud projects.` and `Commands:  stop  List Meltano Cloud projects.`
- The history table shows it as `FAILED` after I stop it. Do we want another result like `CANCELED` so its clear what happened? Or even clearer `CANCELED_BY_USER` if we later need other `CANCELED_BY_X` options. I think this might actually be on the API side though.

tayloramurphy:
> The history table shows it as FAILED after I stop it. Do we want another result like CANCELED so its clear what happened? Or even clearer CANCELED_BY_USER if we later need other CANCELED_BY_X options. I think this might actually be on the API side though.

I see in the tests a `STOPPED` status. We definitely want to make it clear to the user the distinction of a failure vs manual stoppage. 

> With my previous comment in mind, we might want to consider changing these job command to refer to "job executions" because we're interacting with execution, not jobs themselves. Maybe the help text Interact with Meltano Cloud jobs. to Interact with Meltano Cloud job executions..

+1 to this

pnadolny13:
>We definitely want to make it clear to the user the distinction of a failure vs manual stoppage.

@tayloramurphy ok maybe thats a bug then? 🤔  My test run history looks like this

Screen Shot 2023-05-23 at 11 01 34 AM

tayloramurphy:
> Does that need to be done before this can be merged or can we accept failed status for the time being?

@edgarrmondragon @magreenbaum we can accept a failed status so we don't block on this PR. @edgarrmondragon can you update your docs PR to reference the issue to update the reasons so we can track and users can see we're planning to update it?

edgarrmondragon:
> > Does that need to be done before this can be merged or can we accept failed status for the time being?
> 
> @edgarrmondragon @magreenbaum we can accept a failed status so we don't block on this PR. @edgarrmondragon can you update your docs PR to reference the issue to update the reasons so we can track and users can see we're planning to update it?

@tayloramurphy Done: d1ad7a002be63e844220dcb672bad082e2a2e480. Let me know if that looks ok.

pnadolny13 (approved):

edgarrmondragon:
@WillDaSilva Dependency review is failing but I'm not sure if it expects us to just downgrade those deps:

Screenshot 2023-05-26 at 9 24 00

WillDaSilva:
@edgarrmondragon Upgrade `requests` to version `2.31.0`, and the dependency review check should pass. It's failing because of:
image

Which has been fixed in version `2.31.0`: https://requests.readthedocs.io/en/latest/community/updates/

edgarrmondragon:
Fossa is timing out 🤔

edgarrmondragon:
@tayloramurphy I think Pat's done the UAT, so I'm gonna go ahead and merge tomorrow if there are no objections

tayloramurphy (approved):

PR #7678, opened 2023-05-17
docs: refresh getting started guide

pnadolny13:
I re-read through the getting started guide after someone reported seeing issues in slack https://meltano.slack.com/archives/C01TCRBBJD7/p1684327364192699 and I noticed its quite out dated especially when it comes to the dbt parts.

I've noticed that more and more people are reporting issues with dbt and airflow when using the original transformer/orchestrator plugin types but once they switch to utilities their problems are resolved. Ideally we'd get all our docs cleaned up to improve the onboarding experience.

- removed `$` from the commands because they were inconsistent and made it hard to copy and paste
- refresh some outputs and configs since the plugins have changed over time
- the tutorial starts by using `sbalnojan/meltano-lightdash` as the repo then switched to `sbalnojan/meltano-example-el`. I think this was a bug because there was no text telling the user to switch their configs.
- dbt utility instead of transformer
- the first dbt invoke command threw an error because it never told the user to run a tap/target sync after updating the selection critieria to "*". The commit column didnt exist for me.
- The docker command to start postgres needs to assign a name so later we can use it in the `exec` commands. Previously the postgres container was getting a randomly generated name.
- Also for the docker commands to drop tables we didnt provide a database name so I was getting errors like "table not found"
- Add an sample of the data for the mapper section to clarify what its doing.

TODO:
- the numbers arent rendering properly in the current version or my version but I couldnt get them to increment properly

### DIFF
--- a/docs/src/_getting-started/part1.md
+++ b/docs/src/_getting-started/part1.md
@@ -101,7 +101,7 @@ Now that you have your very own Meltano project, it's time to add [plugins](/con
 1.  Add the GitHub extractor
 
 ```bash
-$ meltano add extractor tap-github --variant=meltanolabs
+meltano add extractor tap-github --variant=meltanolabs
 ```
 <br />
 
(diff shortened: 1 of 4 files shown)


kgpayne (approved):

tayloramurphy (approved):
@pnadolny13 this looks great - thanks for getting it updated. One minor change, the very first command is `meltano --version` - can you update that the latest version?

PR #7703, opened 2023-05-25
docs: Correct name of meltano-map-transformer in add command

mjsqu:
Working on meltano v2.19.0, the following illustrates that the correct mapper name is `metlano-map-transformer`

image


edgarrmondragon:
@WillDaSilva the plugin was added to the hub as `meltano-map-transformer`, but both the package name and executable are named `meltano-map-transform`, so I had to override the executable (which by default is the same as the plugin name).

https://gitlab.com/meltano/meltano/-/merge_requests/2577

edgarrmondragon (approved):

pnadolny13 (approved):

tayloramurphy (approved):

PR #7729, opened 2023-06-01
feat: Use the default deployment by default

### LINKED ISSUE #7721: Use default deployment by default
Context:
- https://github.com/meltano/meltano/pull/7718#issuecomment-1570501300

We currently have a `meltano cloud deployment use` command which sets the default deployment to use for future commands. It does this by setting a field in the Cloud CLI config. This setting is not used by anything other than the code which displays whether a deployment is the default. It should be used whenever the `--deployment` CLI option is not provided.

WillDaSilva:
This PR makes it so that the following commands no longer require the `--deployment` option when a default deployment is configured:
- `meltano cloud run <job or schedule name>`
- `meltano schedule enable --schedule <schedule name>`

Note that this does not affect all Cloud CLI commands with a `--deployment` option. It wouldn't be appropriate to use the selected deployment for `meltano cloud history`, for instance, since that filters results, rather than controls what you're acting on.

Closes #7721

### DIFF
--- a/src/cloud-cli/meltano/cloud/cli/run.py
+++ b/src/cloud-cli/meltano/cloud/cli/run.py
@@ -51,9 +51,7 @@ async def run(
 ) -> None:
     """Run a Meltano project in Meltano Cloud."""
     deployment = (
-        deployment
-        if deployment is not None
-        else context.config.default_deployment_name
+        deployment if deployment is not None else context.config.default_deployment_name
     )
     if deployment is None:
         raise click.UsageError(
--- a/src/cloud-cli/meltano/cloud/cli/run.py
+++ b/src/cloud-cli/meltano/cloud/cli/run.py
@@ -40,7 +40,6 @@ async def run_project(
 @click.argument("job_or_schedule")
 @click.option(
     "--deployment",
-    required=True,
     help="The name of the Meltano Cloud deployment to run in.",
 )
 @pass_context
@@ -48,9 +47,18 @@ async def run_project(
 async def run(
     context: MeltanoCloudCLIContext,
     job_or_schedule: str,
-    deployment: str,
+    deployment: str | None = None,
 ) -> None:
     """Run a Meltano project in Meltano Cloud."""
+    if deployment is None:
+        if context.config.default_deployment_name is not None:
+            deployment = context.config.default_deployment_name
+        else:
+            raise click.UsageError(
+                "A deployment name is required. Please specify "
+                "one with the '--deployment' option, or specify a default"
+                "deployment by running 'meltano cloud deployment use'.",
+            )
     click.echo("Running a Meltano project in Meltano Cloud.")
 
     result = await run_project(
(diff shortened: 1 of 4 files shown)


pnadolny13:
>Note that this does not affect all Cloud CLI commands with a --deployment option. It wouldn't be appropriate to use the selected deployment for meltano cloud history, for instance, since that filters results, rather than controls what you're acting on.

@WillDaSilva personally I do think it should affect all commands with a `--deployment` option. If I have multiple deployments and I set my context to use one of them I would expect everything that I interact with would have that deployment pre-selected. I've been thinking of these deployments as completely isolated from each other. Similar to git when I chose a branch I wouldnt expect to see anything related to any other branches other than the one I'm on. 

What do you think? And to be clear I'm not saying they should all be included in this PR but thats my expectation of how it would work in an ideal state.

@tayloramurphy any thoughts on this?

pnadolny13:
@WillDaSilva everything listed in https://github.com/meltano/meltano/pull/7729#issue-1736886168 worked as expected for me!

tayloramurphy:
@pnadolny13 let's open a separate issue on adding it to non-workload commands. I'm leaning towards keeping it to what's in the implementation in this PR. 

For commands like `history`, it's not obvious to me how a user would get back to the default behavior of showing all history w/o passing an additional flag. This is akin to the `environment` flag on the main CLI that only applies to commands that are doing some kind of work on the system. 

It breaks down a bit b/c `schedule` is different on cloud vs core. 

Let's keep it as is for this PR and discuss expanding it later, especially as we add more terms to the CLI.

pnadolny13 (approved):
@WillDaSilva looks good to me! I'll open a separate issue to discuss the non-workload filtering topic. https://github.com/meltano/meltano/issues/7731

tayloramurphy (approved):

src/cloud-cli/meltano/cloud/cli/run.py
```diff
@@ -40,17 +40,25 @@ async def run_project(
...
     """Run a Meltano project in Meltano Cloud."""
+    if deployment is None:
+        if context.config.default_deployment_name is not None:
+            deployment = context.config.default_deployment_name
+        else:
+            raise click.UsageError(
+                "A deployment name is required. Please specify "
+                "one with the '--deployment' option, or specify a default"
+                "deployment by running 'meltano cloud deployment use'.",
+            )
```
edgarrmondragon, line 61: Wdyt of using an if-expression here instead of nested if statements?

```suggestion
    deployment = (
        deployment
        if deployment is not None
        else context.config.default_deployment_name
    )
    if deployment is None:
        raise click.UsageError(
            "A deployment name is required. Please specify "
            "one with the '--deployment' option, or specify a default"
            "deployment by running 'meltano cloud deployment use'.",
        )
```

edgarrmondragon (approved):

src/cloud-cli/meltano/cloud/cli/run.py
WillDaSilva: That's much nicer. Thank you for the suggestion.

PR #7754, opened 2023-06-07
docs: Update deployment details

magreenbaum:
Related to this [internal issue](https://github.com/meltano/infra/issues/1262).


pnadolny13 (approved):
@magreenbaum this is great! It sounds like we'll have auto deployments soon but until then its good to document this.

PR #7766, opened 2023-06-09
chore: History API Deployment column returns `environment_name`

magreenbaum:
Closes: https://github.com/meltano/infra/issues/1281


src/cloud-cli/meltano/cloud/api/types.py
```diff
@@ -19,6 +19,7 @@ class CloudExecution(TypedDict):
     status: str
     exit_code: int
 
+    deployment_name: str
     environment_name: str
```
WillDaSilva, line 23: Should the `environment_name` field be left behind?

magreenbaum: It's unused by the `history` cli so probably not. I went ahead and removed it.

WillDaSilva (approved):

Issue #7588, opened 2023-04-21
Indicate that the item limit was reached in Cloud CLI list tables


WillDaSilva:
Follows from:
- https://github.com/meltano/infra/issues/687#issuecomment-1517891148

In the Cloud CLI we display a table to list a few outputs, e.g. schedules. These list commands all take a `--limit` argument.

If the user is not aware of the limit argument, they may be confused as to why the output does not contain more items than the default limit. To avoid this confusion, when the limit is reached we should output a message below the table indicating that there may be additional items, and they should increase the limit argument to show more.

This message should only be printed for outputs that aren't expected to be parsed or otherwise processed by a program, i.e. only for the default table output, and not for the markdown table or JSON output formats. We may also want to print the output to `stderr` to further decrease the likelihood of it interfering with whatever the regular output was intended for.

PR #7808, opened 2023-06-15
feat: Display warning when output of `list` subcommands are truncated in cloud CLI

BraedonLeonard:
closes: https://github.com/meltano/meltano/issues/7588


BraedonLeonard:
The failing checks are due to 2 bugs:
- https://github.com/meltano/meltano/issues/7812
- https://github.com/meltano/meltano/issues/7813

src/cloud-cli/meltano/cloud/cli/base.py
```diff
@@ -38,6 +46,119 @@ def wrapper(*args, **kwargs):
...
+    """
+    List of items, along with whether or not that list is truncated.
+
+    Used to store lists that are to be printed by `print_formatted_list`.
+    """
+
+    def __init__(self, items: list[T] | None = None, truncated: bool = False):
+        """Initialize items to an empty list and truncated to False."""
+        self.items: list[T] = items if items is not None else []
+        self.truncated: bool = truncated
```
WillDaSilva, line 62:
```suggestion
from dataclasses import dataclass, field

@dataclass
class LimitedResult(t.Generic[T]):
    """
    List of items, along with whether or not that list is truncated.

    Used to store lists that are to be printed by `print_formatted_list`.
    """
    items: list[T] = field(default_factory=list)
    truncated: bool = False
```

src/cloud-cli/tests/cli/test_base.py
```diff
@@ -0,0 +1,56 @@
...
+    test_case(max_items=5, limit=6, max_page_size=10)
+
+    # Case: Limit less than number of items
+    test_case(max_items=5, limit=4, max_page_size=10)
+
+    # Case: Limit greater than number of items, multi-page
+    test_case(max_items=15, limit=16, max_page_size=10)
+
+    # Case: Limit less than number of items, multi-page
+    test_case(max_items=15, limit=14, max_page_size=10)
```
WillDaSilva, line 56: This seems like a good place to use [`pytest.mark.parameterize`](https://docs.pytest.org/en/7.2.x/how-to/parametrize.html).

The suggestion below is about what that would look like. I haven't tested the suggestion.

```suggestion
from __future__ import annotations

import pytest

from meltano.cloud.cli.base import get_paginated, run_async

@pytest.mark.parametrize(('max_items', 'limit', 'max_page_size'), (
    # Case: Page size, limit, and number of items all equal
    (10, 10, 10),
    # Case: Limit less than number of items, equal to page size
    (11, 10, 10),
    # Case: Limit greater than number of items
    (5, 6, 10),
    # Case: Limit less than number of items
    (5, 4, 10),
    # Case: Limit greater than number of items, multi-page
    (15, 16, 10),
    # Case: Limit less than number of items, multi-page
    (15, 14, 10),
))
def test_get_paginated(max_items: int, limit: int, max_page_size: int):
    async def paged_func(page_size: int, page_token: str, max_items: int):
        if not page_token:
            return {
                "results": list(range(min(page_size, max_items))),
                "pagination": {"next_page_token": str(page_size)}
                if max_items > page_size
                else None,
            }
        return {
            "results": list(
                range(int(page_token), min(int(page_token) + page_size, max_items)),
            ),
            "pagination": None,
        }

    results = run_async(get_paginated)(
        lambda page_size, page_token: paged_func(
            page_size,
            page_token,
            max_items=max_items,
        ),
        limit=limit,
        max_page_size=max_page_size,
    )

    assert results.items == list(range(min(limit, max_items)))
    assert results.truncated == (limit < max_items)
```

WillDaSilva (commented):
Thanks for DRY-ing up this code @BraedonLeonard, and for the clean implementation.

I've manually tested it with several cases, and found no issues. The warning is printed to stderr, and so is easy to separate from the regular output.

src/cloud-cli/tests/cli/test_base.py
BraedonLeonard: Good call 👍

WillDaSilva (approved):

PR #7815, opened 2023-06-16
fix: Add back support for interval: '@once'

### LINKED ISSUE #7650: feat: Readd support for schedules with `interval: once`
Context:
- https://github.com/meltano/meltano/issues/3362#issuecomment-1539064260

As part of removing the Meltano UI/API, support for schedules with `interval: once` was also removed on the basis that it was only used by the UI. This was a mistake. Such schedules can still be run using `meltano schedule run`. As such, prior to the Meltano v3 release support for `interval: once` must be added back in.

BraedonLeonard:
Closes https://github.com/meltano/meltano/issues/7650


tayloramurphy (approved):
Approving from the product side but you probably want to get a specific review from engineering

edgarrmondragon (approved):

WillDaSilva (approved):

Issue #7839, opened 2023-06-22
bug: Configured defaults are not invalidated when switching orgs/projects


BraedonLeonard:
If you set a default project or deployment with use, those chosen defaults are saved in your cloud CLI `config.json`. When you switch cloud orgs by logging out, changing your tenant resource key, and logging back in, these defaults remain saved in `config.json`. This means that when you run commands which make use of those defaults, they will use project IDs/deployment names for projects/deployments which do not exist in the new org. The same holds true if you set a default deployment and switch projects.

One approach to fix this is to simply track which org you were previously logged into and clear those defaults if it has changed. This is straighforward, but doesn't really tackle the root of the problem.

A more complete fix would be to make these concepts hierarchical, and save them as such in `config.json`. Each project would have a configured default deployment, and each org would have a configured default project. This prevents deployment names and project IDs being used in the incorrect contexts, while also preserving configured defaults while switching between projects and orgs.

PR #7847, opened 2023-06-26
fix: Create hierarchical structure for remembering org project and deployment defaults

leoroese:
Relates to https://github.com/meltano/meltano/issues/7839

We currently assume that the user's default projects and deployments are within the same organization. However, when users switch between organizations, their previous default settings persist, which can be inconsistent and confusing

@BraedonLeonard proposed a solution for creating a hierarchical structure within the config.json in order to better remember the different default settings between orgs and projects. 

Pull Request is intended to present my current approach based on Braedon's suggestion.

In this PR, the focus is solely on the creation of the new 'organizations' default object within the config. It does not yet modify any other components that depend on it. My plan is to gradually phase out the use of older dependent methods, replacing them with referencing the fields within the new object.

Screenshot 2023-06-26 at 2 08 12 PM

### DIFF
--- a/src/cloud-cli/meltano/cloud/api/config.py
+++ b/src/cloud-cli/meltano/cloud/api/config.py
@@ -117,6 +117,7 @@ class MeltanoCloudConfig:  # noqa: WPS214 WPS230
             default_project_id: The ID of the default Meltano Cloud project.
             default_deployment_name: The name of the default Meltano Cloud
                 deployment.
+            organizations_defaults: Default org settings.
         """
         self.auth_callback_port = auth_callback_port
         self.base_url = base_url
@@ -252,7 +253,11 @@ class MeltanoCloudConfig:  # noqa: WPS214 WPS230
 
     @property
     def internal_organization_default(self) -> CloudConfigOrg:
-        """Get the current tenant resource key defaults for projects and deployments"""
+        """Get the current tenant resource key defaults for projects and deployments.
+
+        Returns:
+            The current tenant resource key default project and list of projects
+        """
         if (
             self.organizations_defaults
             and self.tenant_resource_key in self.organizations_defaults
@@ -266,7 +271,11 @@ class MeltanoCloudConfig:  # noqa: WPS214 WPS230
 
     @internal_organization_default.setter
     def internal_organization_default(self, org_default: CloudConfigOrg) -> None:
-        """Sets the internal organization defaults and updates the config file"""
+        """Set the internal organization defaults and updates the config file.
+
+        Args:
+            org_default: the new default organization settings
+        """
         if not self.organizations_defaults:
             self.organizations_defaults = {}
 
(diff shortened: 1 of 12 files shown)


src/cloud-cli/meltano/cloud/api/config.py
```diff
@@ -126,6 +130,7 @@ def __init__(
         )
         self.default_project_id = default_project_id
         self.default_deployment_name = default_deployment_name
+        self.organizations_defaults = organizations_defaults
```
BraedonLeonard, line 133: I think it would be good if the naming scheme of `organizations_defaults` and `projects` were consistent, since they both represent the same kind of information (i.e. a map from IDs to defaults). Of the two, I think I'd lean towards `*_defaults` for the name, since `projects` seems to imply that it is an exhaustive list of projects, when in fact it just saves defaults for projects you've actively used.

src/cloud-cli/meltano/cloud/api/config.py
```diff
@@ -243,6 +250,63 @@ def tenant_resource_key(self) -> str:
...
+        """
+        if (
+            self.organizations_defaults
+            and self.tenant_resource_key in self.organizations_defaults
+        ):
+            return self.organizations_defaults[self.tenant_resource_key]
+
+        new_org_default = CloudConfigOrg(default_project_id=None, projects={})
+
+        self.internal_organization_default = new_org_default
```
BraedonLeonard, line 268: I would rename anything like `internal_organization_default` or `new_org_default` to be plural (i.e `*_defaults`), since the lack of a plural implies to me that this is the default internal organization, rather than the defaults for the internal organization.

src/cloud-cli/meltano/cloud/api/config.py
leoroese: I agree, good call on consistency

src/cloud-cli/meltano/cloud/cli/deployment.py
```diff
@@ -344,6 +346,23 @@ async def use_deployment(  # noqa: D103
     )
 
 
+def _set_project_default_deployment(
```
leoroese, line 349: This was reviewed by Will in a PR above this, I merged it into this since approved

src/cloud-cli/meltano/cloud/cli/deployment.py
```diff
@@ -391,12 +410,38 @@ async def create_deployment(
...
             click.secho(
-                f"Created deployment {deployment['deployment_name']!r}",
-                fg="green",
+                "Deployment already exists, and was not updated.",
+                fg="yellow",
             )
+            return
+
+        new_deployment_name = deployment["deployment_name"]
+        deployments = await client.get_deployments(page_size=2)
```
leoroese, line 420: Reviewed in another PR

src/cloud-cli/tests/cli/test_deployment.py
```diff
@@ -250,9 +260,72 @@ def test_deployment_use_by_name_interactive(
...
+
+        assert (
+            default_org_settings["projects_defaults"][default_project_id][
+                "default_deployment_name"
+            ]
+            == "legacy"
+        )
+
+    @pytest.mark.parametrize("prepared_request", ({"method": "POST"},), indirect=True)
+    def test_create_first_new_deployment(
```
leoroese, line 276: reviewed in another PR

BraedonLeonard (approved):

WillDaSilva (approved):

PR #7864, opened 2023-06-28
perf: Use Pyston to improve core performance

WillDaSilva:
https://blog.pyston.org/2022/06/08/announcing-pyston-lite-our-python-jit-as-an-extension-module/

https://blog.pyston.org/2022/09/29/announcing-3-7-3-10-support-and-a-new-direction/

Pyston-lite-autoload auto-magically (through JIT techniques) improves the performance of any compatible (CPython version 3.7-3.10 on macOS or Linux) Python environment it's installed into.

By installing it alongside Meltano, Meltano's performance seems to be significantly improved.

In the future we may want to consider also installing it into the virtual environments Meltano manages for its plugins.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -4074,4 +4074,4 @@ s3 = ["boto3"]
 [metadata]
 lock-version = "2.0"
 python-versions = ">=3.7,<3.12"
-content-hash = "e19de58d807d1be391f7e3eefa10677dd354cffa8335beb46aaca6a0912ed411"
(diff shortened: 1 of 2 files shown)


WillDaSilva:
Despite having observed a very significant performance improvement when I ran the tests locally on macOS, when running the tests in our CI workflow on Linux there doesn't seem to be an improvement.

WillDaSilva:
Closing because this is currently just cluttering the list of in-progress PRs. Using Pyston to get a performance boost for free in the future would be nice, but for now I'm not going to put more time into this.

Issue #7730, opened 2023-06-01
Reduce Cloud CLI latency by optimistically trying request before auth check


WillDaSilva:
A typical execution that hits a single API endpoint will often take around 1.5 seconds to execute. Of that, it's common for the request to authenticate at `<auth domain>/oauth2/userInfo` to take 0.47 seconds.

We should be able to avoid this for most Cloud CLI executions by avoiding the call to `<auth domain>/oauth2/userInfo`. Instead we can directly try to hit the Cloud API server, and if it returns a 403 status code, then we can go through the existing auth code wherein we check if we're logged in, and log in if not. In the case where we hit a 403 status code before checking if we had logged in, we can retry the request automatically. If the request fails with a 403 status code again, then we can produce the appropriate error message for that.

PR #7865, opened 2023-06-28
perf: Make authenticated requests without an explicit check for login

BraedonLeonard:
When you have an access and id token, requests are made with the assumption that the tokens aren't out of date. On 403, login is triggered and the request is retried automatically.

closes: https://github.com/meltano/meltano/issues/7730

### DIFF
--- a/src/cloud-cli/meltano/cloud/api/client.py
+++ b/src/cloud-cli/meltano/cloud/api/client.py
@@ -60,7 +60,6 @@ class MeltanoCloudClient:  # noqa: WPS214, WPS230
         self.auth = MeltanoCloudAuth(self.config)
         self.api_url = self.config.base_url
         self.version = version
-        # Tracks nesting depth of authenticated contexts
         self._within_authenticated: bool = False
 
     async def __aenter__(self) -> Self:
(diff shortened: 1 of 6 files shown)


src/cloud-cli/meltano/cloud/api/client.py
```diff
@@ -131,13 +132,21 @@ def headers(self, headers: dict) -> t.Iterator[None]:
     async def authenticated(self) -> t.AsyncIterator[None]:
         """Provide context for API calls which require authentication.
 
+        Authorization and ID tokens may be out-of-date. In this case, requests made
+        within the context which get a 403 response will cause re-authentication, and
+        the request will be remade.
+
+        Note: Will not work correctly if nested.
```
WillDaSilva, line 139: The technical term for this (used widely in the Python stdlib docs) is [reentrant](https://docs.python.org/3/library/contextlib.html#reentrant-context-managers)

```suggestion
        This context manager is not reentrant.
```

WillDaSilva (approved):
All of my manual tests worked as expected. This seems to save about 0.4 seconds per run after having logged in and warmed up our API servers.

PR #7869, opened 2023-06-30
chore: merge cloud-cli to main

cjohnhanson:
Opening a draft PR for the big cloud->main merge to track conflicts and changes ahead of GA launch.


WillDaSilva:
@cjohnhanson Instead of using a PR, which will cause all of the commits to be squashed thereby losing the history of all of the PRs we've merged into `cloud`, how do you feel about us just force-pushing `cloud` to `main`, thereby essentially switching the main branch over to the commits we've pushed to `cloud`. After that we can update our docs to remove references to the `cloud` branch, and eventually delete the `cloud` branch once we're pretty sure no one is still using it (e.g. in a year or so).

In either case I think we'll want to manually write the changelog for the Meltano v3 release, rather than relying on the normal auto-generated changelog.

edgarrmondragon:
@WillDaSilva I think we could also enable rebase or merge commits temporarily and merge this PR with either option

PR #7878, opened 2023-07-05
fix: Redact sensitive values in `config list` output

### LINKED ISSUE #6858: Meltano `config list` and `config set` should print settings with `kind:secret` as `(redacted)`
As reported in slack and I have confirmed locally.

Currently `meltano config set` will print the setting value in clear text after setting it, and similarly, `config list` will print the secret value in clear text as well.

Proposed changes:

1. `meltano config set` would print `(redacted)` in place of the setting value when confirming that the value has been stored.
2. `meltano config list` would print `(redacted)` in place of the current value when printing the list of all settings.

cc @tayloramurphy

ReubenFrankel:
Fix to redact values for sensitive settings (of `kind` `password` or `oauth`) in `config list`. Also minor refactor to display the redacted value in yellow across `config` commands.

`config list`

image

`config set`

image

`config set --interactive`

image

---

Closes #6858

### DIFF
--- a/docs/src/_reference/command-line-interface.md
+++ b/docs/src/_reference/command-line-interface.md
@@ -360,6 +360,15 @@ meltano config <plugin> set --interactive --extras
 meltano config <plugin> set --interactive --store=dotenv
 ```
 
+### Sensitive configuration
+Values for sensitive settings (defined as `kind: password`) are redacted in the output of the following commands:
+
+```bash
+meltano config <plugin> list
+meltano config set <name> <value>
+meltano config <plugin> set --interactive
+```
+
 ## `discover`
 
 Lists the available [discoverable plugins](/concepts/plugins#discoverable-plugins) and their [variants](/concepts/plugins#variants).
--- a/src/meltano/cli/config.py
+++ b/src/meltano/cli/config.py
@@ -284,7 +284,11 @@ def list_settings(ctx, extras: bool):
         else:
             label = f"{get_label(config_metadata)}"
 
-        current_value = click.style(f"{value!r}", fg="green")
+        current_value = click.style(
+            f"{value!r}",
+            fg="yellow" if value is not None and setting_def.is_redacted else "green",
+        )
+
         click.echo(f" current value: {current_value}", nl=False)
 
         unexpanded_value = config_metadata.get("unexpanded_value")
--- a/src/meltano/cli/config.py
+++ b/src/meltano/cli/config.py
@@ -284,9 +284,11 @@ def list_settings(ctx, extras: bool):
         else:
             label = f"{get_label(config_metadata)}"
 
+        redacted_with_value = value is not None and setting_def.is_redacted
+
         current_value = click.style(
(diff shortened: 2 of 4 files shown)


tayloramurphy:
@ReubenFrankel passwords are already appropriately redacted in interactive mode from what I can see, right? So this should mainly just touch the `config list` command, correct?

ReubenFrankel:
@tayloramurphy Yes. I made a change to display `(redacted)` in yellow across `config list`, `config set` and `config set --interactive`, but we can revert this if not necessary.

ReubenFrankel:
Should this change also redact values from `config` (no sub-command) output?

image

tayloramurphy:
I don't mind the yellow. It brings attention to it since it is doing something different. I'll approve from the Product side, but still needs engineering review.

tayloramurphy (approved):

edgarrmondragon (commented):
LGTM! The missing coverage is only on the patch (i.e. % of the changed code) and it all falls in the interactive code.

edgarrmondragon (approved):

ReubenFrankel:
> Should this change also redact values from `config` (no sub-command) output?
> 
> image

@edgarrmondragon Should we address this before merging?

edgarrmondragon:
@ReubenFrankel I think the output of `meltano config <plugin name>` is intended for debugging what config the tap or target will be fed, or even for exporting or being used by other tools. So, keeping the raw value there makes sense to me. I'd be curious what @tayloramurphy thinks.

tayloramurphy:
@edgarrmondragon I think the problem with it is that once you run it then the sensitive value is in your terminal output history. I would be ok with some psuedo-interactive action where it displays but once you exit out it removes it all. 

We do have `meltano invoke --dump=config <plugin>` for getting the actual values, which I think is the appropriate method for getting the actual values a connector sees.

edgarrmondragon:
> @edgarrmondragon I think the problem with it is that once you run it then the sensitive value is in your terminal output history. I would be ok with some psuedo-interactive action where it displays but once you exit out it removes it all.
> 
> We do have `meltano invoke --dump=config <plugin>` for getting the actual values, which I think is the appropriate method for getting the actual values a connector sees.

@tayloramurphy Ok, so thoughts on leaving the behavior of `meltano config <plugin>` as is vs redacting any sensitive values there? If the former, we could deprecate in favor of `meltano invoke --dump=config <plugin>`.

tayloramurphy:
@edgarrmondragon apologies - I misunderstood which subcommand we were talking about. 

Let's keep the scope of this PR small and open a separate issue about secret values in both `meltano config <plugin>` and `meltano invoke --dump=config <plugin>`. I wouldn't be opposed to have an update on this PR to the docs though explaining where sensitive values are redacted!

ReubenFrankel:
@tayloramurphy I can do a docs update. 🙂

edgarrmondragon (approved):
Thanks @ReubenFrankel!

PR #7902, opened 2023-07-14
docs: rm ELT messaging, move things around

sbalnojan:
This is a quick fix for the getting started guide to go along with the new website copy and the reddit campaigns.

https://github.com/meltano/internal-marketing/issues/907

It's not changing any of the content, just removing the ELT framing and cleaning up things abit.


edgarrmondragon (approved):

docs/src/_getting-started/part4.md
```diff
@@ -1,171 +0,0 @@
----
-title: Part 4 - Inline Data Mapping
```
tayloramurphy, line 2: @sbalnojan I like most of the PR except for this section. Mappers are still very relevant for this use case and fit a software engineering ICP. I would ask that we add this back but instead of hashing PII maybe we filter or duplicate some data?

tayloramurphy (changes requested):
I'd argue Mappers are still compelling. Do you think it's worth having a separate tutorial on them or keeping them in this one?

docs/src/_getting-started/part4.md
sbalnojan: @tayloramurphy, I think I disagree. For software engineers, this is a nice-to-have (and totally not needed) feature, whereas, for a data engineer, it is a must-have. 


My plan would be to:
1. Extract the complete "internal data team" tutorial, where the inline mapper would sit as well
2. And have the introductory tutorial without it.

tayloramurphy (approved):

PR #7907, opened 2023-07-15
feat: Add plugin from ref

### LINKED ISSUE #6615: feature: developing plugins locally before publishing to hub
### Feature scope

### Description

**As a user, I want to develop plugins with dependencies before publishing these plugins to the HUB**

### Background

We use a `file` bundle type plugin to distribute analytics:  `analyze-google-analytics`

We use a `file` bundle type plugin to distribute models:  `dbt-google-analytics`

We use an `extractor` type plugin to import data:  `tap-google-analytics`

To work correctly:  `analyze` requires `models` requires `tap`

## problem

The developer wants to create `analyze` and `models` at the same time, using an existing `extractor`

But `meltano add file analyze-google-analytics` cannot be built and tested until `models-google-analytics` is published on the HUB

## proposal

The developer creates the plugin definitions and adds using a local file path

NB - is `--from_ref` undocumented https://github.com/meltano/meltano/issues/3208

`meltano add file models-google-analytics --from_ref=../projects/models-google-analytics.yml`
`meltano add file analyze-google-analytics --from_ref=../projects/analyze-google-analytics.yml`

Regular plugin yml definition
`
/projects/analyze-google-analytics.yml
`

Regular plugin yml definition
`
/projects/plugins/models-google-analytics.yml
`

## Related issues

https://github.com/meltano/meltano/issues/3208

"using local hub instance" is not available for local developers https://github.com/meltano/meltano/issues/5997#issuecomment-1146358235

### LINKED ISSUE #7368: Ability to test `meltano add` on a plugin prior to a Hub PR being merged
I'd love to have a way to test `meltano add ...` on a plugin before we publish to the Hub.

Any ideas? I initially thought to use the `hub_url` overrides to target the preview deployment, but I am pretty sure we are not publishing the JSON API to Netlify. We also can't use the S3 files since many of these are run from community forks and won't have any S3 publish permissions.

Perhaps we need to implement a `--from=HUB_YAML_FILE_URL` arg in the CLI for `meltano add` or else make it process to download the yaml file from the hub directly into the location where the lockfile would be?

cc @edgarrmondragon, @pnadolny13 - I think you two are the more knowlegeble on these processes. Any ideas?

ReubenFrankel:
Closes #7368, closes #6615


ReubenFrankel:
@edgarrmondragon Do you mind taking a look at why these new tests might be failing if you have some free time? They fail intermittently locally as well, where the `cli_runner` result `stdout` and `stderr` are empty strings. Ideally would like to revert https://github.com/meltano/meltano/pull/7907/commits/60192add6a1707c26276fe412a78a8360df90636, which I pushed since that assertion also seemed indeterministic (exit code of `1` or `2` between runs).

edgarrmondragon:
> @edgarrmondragon Do you mind taking a look at why these new tests might be failing if you have some free time? They fail intermittently locally as well, where the `cli_runner` result `stdout` and `stderr` are empty strings. Ideally would like to revert [60192ad](https://github.com/meltano/meltano/commit/60192add6a1707c26276fe412a78a8360df90636), which I pushed since that assertion also seemed indeterministic (exit code of `1` or `2` between runs).

@ReubenFrankel Without diving much deeper, is it possible that something is persisted between tests and that's causing the flakiness?

Since we randomize the order of tests, we have run into that problem and the workaround is to make certain tests run in a set order:

https://github.com/meltano/meltano/blob/b2cd4f7860edffac37047ed477a40478b76e596a/tests/meltano/cli/test_add.py#L32

ReubenFrankel:
@edgarrmondragon Alright thanks, I'll take a look. :slightly_smiling_face:

ReubenFrankel:
@edgarrmondragon I think this is ready for review now (pending docs changes once behaviour is finalised), whenever you have some time. :slightly_smiling_face:

src/meltano/core/plugin/error.py
```diff
@@ -56,3 +60,27 @@ class PluginExecutionError(Exception):
...
+class InvalidPluginDefinitionError(Exception):
+    """Base exception when a plugin definition is invalid."""
+
+    def __init__(self, definition=None):
+        """Construct a new "InvalidPluginDefinitionError instance.
+
+        Args:
+            definition: Plugin definition.
+        """
+        fullargspec = inspect.getfullargspec(PluginDefinition.__init__)  # noqa: WPS609
```
edgarrmondragon, line 74: Nice!

src/meltano/cli/add.py
```diff
@@ -19,13 +22,33 @@
...
+        return
+
+    try:
+        url = urlparse(value)
+        if url.scheme and url.netloc:
+            response = requests.get(value, timeout=10)
+            response.raise_for_status()
+            content = response.text
+        else:
+            content = Path(value).read_text()
```
edgarrmondragon, line 44: This means that the option accepts two types of args:

1. URLs, e.g. `https://example.com/path/to/plugin.yaml`
2. File paths (either absolute or relative), e.g. `/Users/path/to/plugin.yaml`

right?

ReubenFrankel: Correct. Unsure if "ref" as terminology covers this, but that is the intended behaviour.

ReubenFrankel:
Since discovery functionality has been removed for upcoming v3 release, `REQUEST_TIMEOUT_SECONDS` constant is no longer available after merging. Instead I just replaced it with a hard-coded value of 10 seconds - is there something else common I can use here?

https://github.com/meltano/meltano/pull/7907/commits/661030273e9a5a1ca42d50e630e0a5fea9989917#diff-3916f7d151c1573e41e141d322bb79fe11845771f00335727d7f19715ecb8b49L41

src/meltano/cli/add.py
ReubenFrankel: - File path is intended for local development of taps to test with `meltano add`
- URL is intended for Hub PR checks (i.e. `--from-ref https://github.com/meltano/hub/blob/main/_data/meltano/extractors/tap-spotify/matatika.yml` - see https://github.com/meltano/hub/issues/1191), and maybe helps move towards a model where the plugin definition lives in the source repository - similar to how dbt Hub works IIRC (if that's something you guys are interested in)

src/meltano/cli/add.py
```diff
@@ -19,13 +22,33 @@
...
 
 
+def _load_yaml_from_ref(_ctx, _param, value: str | None) -> dict:
+    if not value:
+        return
+
+    try:
+        url = urlparse(value)
+        if url.scheme and url.netloc:
+            response = requests.get(value, timeout=10)
```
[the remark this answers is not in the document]
edgarrmondragon: The hardcoded value is ok

edgarrmondragon (commented):
Thanks @ReubenFrankel! I like the general approach of inlining the plugin definition so it becomes a custom plugin.

But one question that the tests don't seem to answer is whether the plugin can be updated by using `--from-ref` repeatedly. The workaround would be to use

```bash
# Initial installation
meltano add <plugin type> <plugin name> --from-ref https://...

# Incorporate changes from upstream definition
meltano remove <plugin type> <plugin name>
meltano add <plugin type> <plugin name> --from-ref https://...
```

I think it would make development easier if you only needed a single command to refresh the plugin definition. What do you think?

src/meltano/cli/add.py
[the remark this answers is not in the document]
edgarrmondragon: The hardcoded value is ok 👍

ReubenFrankel:
@edgarrmondragon I didn't really think about that - I agree it would be easier just to re-run the same command, but does that not go against the principles of `meltano add`? Thinking about plugins added from the Hub (non-custom): these come with a `.lock` file after `meltano add`, and then`meltano lock --update` to update. I suppose this use-case is a bit different because we *are* dealing with  custom plugins, where the plugin definition is only in the `meltano.yml` (i.e. no `.lock` file)...

I think that it might be confusing for users if `--from-ref` was to bypass the

```
Extractor 'tap-spotify' already exists in your Meltano project
```

behaviour, where all other ways of using `meltano add` do not.

That being said, I don't really have any immediate ideas, but it would be great to update a custom plugin in this way. What do you think?

ReubenFrankel:
Also, if `.lock` files are required as of Meltano v3, does this feature work then? Or is that only enforced for non-custom plugins?

edgarrmondragon:
> Or is that only enforced for non-custom plugins?

That's correct

edgarrmondragon:
> That being said, I don't really have any immediate ideas, but it would be great to update a custom plugin in this way. What do you think?

@ReubenFrankel do you mean `meltano lock --update`? Then, `--from-ref` would need to add a lockfile instead of inlining the plugin definition, and the plugin entry in `meltano.yml` would have to store the `--from-ref` value so `meltano lock --update` behaves consistently. Wdyt?

ReubenFrankel:
@edgarrmondragon I was referring to the concept of updating custom plugins only, but I guess it could work with `meltano lock`... Though, I was under the impression that `.lock` files were reserved for plugins coming from the Hub only - we would have to add an optional `ref` property for plugins in the `meltano.yml` like you say.

edgarrmondragon:
> @edgarrmondragon I was referring to the concept of updating custom plugins only, but I guess it could work with `meltano lock`... Though, I was under the impression that `.lock` files were reserved for plugins coming from the Hub only - we would have to add an optional `ref` property for plugins in the `meltano.yml` like you say. I don't know if that is too much of a change to warrant adding this feature though?

@ReubenFrankel yeah, I think there's 3 types of plugins with different expectations and flows:

1. **_100% custom_ plugins**: the user is not currently interested in publishing it to the Hub. Updates to settings, etc. are done manually in `meltano.yml`
2. **Hub plugins**: these are Hub (Meltano's or a private one) plugins that the user can and want to be updated from upstream definitions.
3. **Ref plugins (this PR)**: users want to test a plugin definition they're in the process of adding to the Hub, and they want to update the local definition after changes are made to the upstream plugin, e.g. after PR feedback in the Hub. With the current state of the feature, they'd need to do manual changes to `meltano.yml` to get those changes.

Do you think the last one above makes sense, or perhaps I'm missing something?

cc @pnadolny13 in case you have thoughts on this feature

pnadolny13:
@ReubenFrankel @edgarrmondragon I'm trying to catch up on this feature. Is my understanding right that its basically taking a meltano.yml from the tap repo and merging it into your main meltano project? Makes sense to me. I'm not sure what the normal workflow is for other users but I generally dont consider the meltano.yml in my tap repo as the source of truth (maybe thats the wrong way to think 😭 ).

>maybe helps move towards a model where the plugin definition lives in the source repository - similar to how dbt Hub works IIRC (if that's something you guys are interested in)

The only challenge with having the plugin definition live in the repo is that its formatting and content are based on what the hub needs so if that changes over time we'd need to update every plugin repo to implement it. I've been thinking more along the lines of a scraping the SDK `--about` settings and generating a hub definition from that, so the tap is decoupled from the hub syntax.

ReubenFrankel:
> With the current state of the feature, they'd need to do manual changes to `meltano.yml` to get those changes.

Or `meltano remove`, then `meltano add --from-ref` again. Otherwise, I think you covered it. Is there an argument for `meltano add --force` here though?

edgarrmondragon:
> Is my understanding right that its basically taking a meltano.yml from the tap repo and merging it into your main meltano project?

@pnadolny13 rather it takes a YAML file in the format the hub expects for a plugin.

> I've been thinking more along the lines of a scraping the SDK `--about` settings and generating a hub definition from that, so the tap is decoupled from the hub syntax.

I think this feature could be downstream from that:

1. Scrape a Hub definition from the JSON output of `--about` and save as `my_plugin.yml`
2. Add it to a project with `meltano add extractor my-plugin --from-ref ./my_plugin.yml` to test the definition with Meltano

edgarrmondragon:
> Or `meltano remove`, then `meltano add --from-ref` again. Otherwise, I think you covered it.

Thanks @ReubenFrankel!

> Is there an argument for meltano add --force here though?

Maybe 🤔. If it doesn't blow the scope of this PR, perhaps `--force` could tell Meltano to replace the plugin if it exists.

ReubenFrankel:
> 1. Scrape a Hub definition from the JSON output of `--about` and save as `my_plugin.yml`

Maybe an SDK enhancement to add `--format yaml`? Of course, that would require plugin to update its project SDK version (or built with the SDK to begin with).

> perhaps --force could tell Meltano to replace the plugin if it exists.

I *think* this is my preferred approach. I'm imagining `--force` just adds a plugin regardless of it already exists in the project or not, meaning that for existing plugins the `meltano.yml` entry is replaced - for non-custom plugins, the `.lock` file would also be regenerated. I guess you could think of this like the behaviour you would expect from running `meltano remove` into `meltano add`. That feels like a separate PR though?

edgarrmondragon:
> I _think_ this is my preferred approach. I'm imagining `--force` just adds a plugin regardless of it already exists in the project or not, meaning that for existing plugins the `meltano.yml` entry is replaced - for non-custom plugins, the `.lock` file would also be regenerated. I guess you could think of this like the behaviour you would expect from running `meltano remove` into `meltano add`. That feels like a separate PR though?

@ReubenFrankel Yeah, let's create a separate issue for `meltano add ... --force` and document for this feature how to re-add a ref plugin.

edgarrmondragon:
@ReubenFrankel Pending docs but the behavior here lgtm!

docs/docs/reference/command-line-interface.md
edgarrmondragon, line 1: Can you change

```diff
- Without a `--custom` or `--inherit-from` option, this will add the
+ Without a `--custom`, `--inherit-from` or `--from-ref` option, this will add the
```
?

ReubenFrankel: https://github.com/meltano/meltano/pull/7907/commits/43804da50dd372b463d355c85bd3f9579dda93f1

edgarrmondragon (approved):
Thanks @ReubenFrankel! I'll leave this open in case anyone from the team wants to weigh in. Otherwise, I'll try to merge this tomorrow.

ReubenFrankel:
@edgarrmondragon Awesome! Thanks for your feedback as always. :smile:

docs/docs/reference/command-line-interface.md
```diff
@@ -86,6 +86,38 @@ meltano add --custom extractor tap-covid-19
 docker run --interactive -v $(pwd):/project -w /project meltano/meltano add --custom extractor tap-covid-19
 ```
 
+To add a plugin from a [plugin definition](/concepts/project#custom-plugin-definitions) YAML file as a [custom plugin](/concepts/plugins#custom-plugins), use the `--from-ref` option referencing a URL or local path:
```
tayloramurphy, line 89: My preference would be to have this below the --inherit-from explanation as it's a more rare use case. 

I'd also like a bit of context for why users would likely want to use this. Could we add a short statement about when this functionality is useful. And perhaps we could have a separate issue and PR to add a mention in https://docs.meltano.com/tutorials/custom-extractor/ ? 

I'll approve so it's non-blocking as this is a relatively minor docs update.

tayloramurphy (approved):

docs/docs/reference/command-line-interface.md
ReubenFrankel: https://github.com/meltano/meltano/pull/7907/commits/263cc848d90de07744b950c1f07d967a119d3667
https://github.com/meltano/meltano/pull/7907/commits/57d4fb05b81601d8ea4a0b13564e9e07b4a95227

PR #7912, opened 2023-07-17
fix: Datetime parsing error in History command

edgarrmondragon:
Python < 3.11 wasn't able to parse valid ISO datetimes:

```python                                                      
>>> import datetime, sys
>>> sys.version
'3.10.12 (main, Jul 12 2023, 14:50:02) [Clang 14.0.3 (clang-1403.0.22.14.1)]'
>>> datetime.datetime.fromisoformat("2023-07-17T18:45:49.018Z")
  File "<stdin>", line 1, in <module>
ValueError: Invalid isoformat string: '2023-07-17T18:45:49.018Z'
```

But Python 3.11 changed that ([docs](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat)):

> Previously, this method only supported formats that could be emitted by [date.isoformat()](https://docs.python.org/3/library/datetime.html#datetime.date.isoformat) or [datetime.isoformat()](https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat).

```python
>>> import datetime, sys
>>> sys.version
'3.11.4 (main, Jul 11 2023, 12:59:52) [Clang 14.0.3 (clang-1403.0.22.14.1)]'
>>> datetime.datetime.fromisoformat("2023-07-17T18:45:49.018Z")
datetime.datetime(2023, 7, 17, 18, 45, 49, 18000, tzinfo=datetime.timezone.utc)
```

My suspicion is that Pydantic V2 changed the way datetimes are serialized to use the "Z" ISO format, but I couldn't find a reference to that with a quick scan.

See https://github.com/meltano/infra/issues/1719


edgarrmondragon:
click is causing mypy to fail. There's a PR out there to fix that: https://github.com/pallets/click/pull/2565

WillDaSilva (approved):
I agree that using `python-dateutil` here is a good solution. It still isn't clear to me why it's necessary, but it should make the timestamp parsing more robust.

> click is causing mypy to fail. There's a PR out there to fix that: https://github.com/pallets/click/pull/2565

In the meantime, I think we should add `# type: ignore[attr-defined]` comments as needed to make mypy pass, and open an issue or PR to remove those comments once the issue upstream has been fixed.

PR #7917, opened 2023-07-18
docs: Update instructions for providing `GIT_SSH_PRIVATE_KEY`

WillDaSilva:


docs/src/_cloud/platform.md
```diff
@@ -49,24 +49,20 @@ There are specific environment variables that are reserved for certain use-cases
 
 ### SSH Private Keys for Private Git Repository Package Access
 
-`GIT_SSH_PRIVATE_KEY` is a reserved variable that should be set if you have private repository packages.
+`GIT_SSH_PRIVATE_KEY` is a reserved variable that should be set if you have private git repositories that accessed during `meltano install` can be access via SSH.
```
edgarrmondragon, line 52:
```suggestion
`GIT_SSH_PRIVATE_KEY` is a reserved variable that should be set if you have private git repositories that are accessed during `meltano install` and can be accessed via SSH.
```

edgarrmondragon (approved):

PR #7936, opened 2023-07-21
docs: Migrate Meltano docs to Docusaurus

tayloramurphy:


edgarrmondragon:
Should we restore `docs/example-library` at least for CI to run as usual?

WillDaSilva:
> Should we restore docs/example-library at least for CI to run as usual?

@edgarrmondragon I was thinking we could just move those out of `docs/`, and into `integration/` since they were never really incorporated into the docs.

pnadolny13:
@tayloramurphy I didnt see anything to comment on in the changelog :thumbsup: . I did notice that the left side nav bar is showing some sections that are currently hidden in the existing docs e.g. Cloud->Sandbox.

tayloramurphy:
@pnadolny13 good call out on that. I'll check with Brightscout.

edgarrmondragon:
I don't know if it's important but search seems to be broken

PR #7952, opened 2023-07-26
docs: August Changelog

tayloramurphy:


tayloramurphy:
FYI @sbalnojan this is the long running PR we can add to for the next changelog.

tayloramurphy:
@sbalnojan I need to add a couple of links once the docs are updated, but this is the changelog for the month!

edgarrmondragon (approved):

sbalnojan:
One note @tayloramurphy, have you thought about adding a list of external contributors to the change log? It sounds like a nice thing since the releases on GH usually also highlight them.

tayloramurphy:
@sbalnojan not opposed - could just take a little while. Let me see what the effort is.

tayloramurphy:
Ok so where this breaks down is on the Hub and MeltanoLabs side. There's lots of updates there that we're not tracking in release and compiling the contribution activity would be pretty cumbersome. I'd rather just not do it right now as the Changelog isn't necessarily where we'd want to highlight folks anyways.


This PR Is blocked on a few others getting merged:

* #8064 
* one from Leo on the email notifications

PR #7982, opened 2023-08-01
fix: Use hyphens consistently for CLI options

### LINKED ISSUE #2253: Consistently use hyphens in CLI parameter names rather than underscores
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/2297

Originally created by @DouweM on 2020-08-21 18:02:10

---

Culprits:
- `meltano init` `--no_usage_stats`
- ~~`meltano elt` and `meltano schedule` `--job_id`~~ Changed to state-id
- `meltano upgrade` `--pip_url`

WillDaSilva:
This is not a breaking change. All old CLI options will still work thanks to the normalization the CLI now does to the CLI option names.

Closes #2253

### DIFF
--- a/src/meltano/cli/cli.py
+++ b/src/meltano/cli/cli.py
@@ -47,6 +47,11 @@ class NoWindowsGlobbingGroup(InstrumentedGroup):
     cls=NoWindowsGlobbingGroup,
     invoke_without_command=True,
     no_args_is_help=True,
+    # For backwards compatibility, accept CLI options that use underscores
+    # instead of hyphens.
+    # https://github.com/pallets/click/issues/1123#issuecomment-589989721
+    # NOTE: This CLI option normalization applies to all subcommands.
+    context_settings={"token_normalize_func": lambda x: x.replace("_", "-")},
 )
 @click.option("--log-level", type=click.Choice(LEVELS.keys()))
 @click.option(
--- a/tests/meltano/cli/test_initialize.py
+++ b/tests/meltano/cli/test_initialize.py
@@ -18,7 +18,7 @@ class TestCliInit:
             Project.find()
 
         # Create a project with the CLI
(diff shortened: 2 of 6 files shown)


tests/meltano/cli/test_initialize.py
```diff
@@ -71,7 +71,7 @@ def test_init_existing_empty_directory(self, cli_runner, tmp_path_factory, pushd
         project_dir.mkdir()
         assert project_dir.exists()
 
-        result = cli_runner.invoke(cli, ["init", "test_project", "--no_usage_stats"])
+        result = cli_runner.invoke(cli, ["init", "test_project", "--no_usage-stats"])
```
WillDaSilva, line 74: The mix of hyphens and underscores here is intentionally. Additionally, some tests are unchanged, and continue to use `--no_usage_stats` instead of `--no-usage-stats`, which ensures that the old option names still work.

src/meltano/cli/cli.py
```diff
@@ -47,6 +47,11 @@ def main(self, *args, **kwargs) -> t.NoReturn:
     cls=NoWindowsGlobbingGroup,
     invoke_without_command=True,
     no_args_is_help=True,
+    # For backwards compatibility, accept CLI options that use underscores
+    # instead of hyphens.
+    # https://github.com/pallets/click/issues/1123#issuecomment-589989721
+    # NOTE: This CLI option normalization applies to all subcommands.
+    context_settings={"token_normalize_func": lambda x: x.replace("_", "-")},
```
BraedonLeonard, line 54: This *will* lock us out of intentionally using underscores in option names in the future, but I don't expect us to ever have a need to distinguish between 2 options whose only differentiators are underscores vs hyphens

BraedonLeonard (approved):

edgarrmondragon (approved):
Nice!

PR #7984, opened 2023-08-01
feat(cli): Add `meltano config --unsafe` flag

ReubenFrankel:
Essentially bypasses behaviour changes made in #7878 to expose sensitive setting values with `--unsafe`, as per [this request on Slack](https://meltano.slack.com/archives/CKHP6G5V4/p1690300061219669) by @visch.

### DIFF
--- a/src/meltano/cli/config.py
+++ b/src/meltano/cli/config.py
@@ -131,6 +131,7 @@ def get_label(metadata) -> str:
     default="json",
 )
 @click.option("--extras", is_flag=True, help="View or list only plugin extras.")
+@click.option("--unsafe", is_flag=True, help="Expose values for sensitive settings.")
 @pass_project(migrate=True)
 @click.pass_context
 def config(  # noqa: WPS231
@@ -140,6 +141,7 @@ def config(  # noqa: WPS231
     plugin_name: str,
     config_format: str,
     extras: bool,
+    unsafe: bool,
 ):
     """
     Display Meltano or plugin configuration.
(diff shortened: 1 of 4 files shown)


ReubenFrankel:
Didn't know whether to keep the yellow styling for sensitive values when they are exposed (as a kind of reminder that the value is still sensitive) or not, so I just made it output what it otherwise would have before #7878. Pretty easy to change if anyone feels strongly.

src/meltano/cli/config.py
```diff
@@ -131,6 +131,7 @@ def get_label(metadata) -> str:
     default="json",
 )
 @click.option("--extras", is_flag=True, help="View or list only plugin extras.")
+@click.option("--unsafe", is_flag=True, help="Expose values for sensitive settings.")
```
edgarrmondragon, line 134: Wdyt about this
```suggestion
@click.option(
    "--safe/--unsafe",
    default=True,
    help="Expose values for sensitive settings.",
)
```

Then you can change instances of `not unsafe` to safe.

ReubenFrankel: 100% agree - changed in d4b7e59.

edgarrmondragon (approved):
LGTM!

src/meltano/cli/config.py
```diff
@@ -131,6 +131,12 @@ def get_label(metadata) -> str:
     default="json",
 )
 @click.option("--extras", is_flag=True, help="View or list only plugin extras.")
+@click.option(
+    "--safe/--unsafe",
```
tayloramurphy, line 135: Love this! Can we get a quick update to https://docs.meltano.com/reference/command-line-interface#config as well?

tayloramurphy (approved):

src/meltano/cli/config.py
edgarrmondragon: Docs look great. Thanks @ReubenFrankel!

PR #7992, opened 2023-08-04
feat: Add support for mappers in the invoke command

pnadolny13:
Closes https://github.com/meltano/meltano/issues/3232

I'm not sure if theres a better way to get the mapping_name from a plugin object but this properly finds the mapper/mapping from my testing. I was able to run `cat github_out.json | meltano invoke hide-github-mails` using the tutorial mapper example to hash github emails and everything worked as expected.

UPDATE: it looks like we do a similar lookup for mapping_name in https://github.com/meltano/meltano/blob/e8b96010ce86a184e21b1a2e953d84a000210ad1/src/meltano/core/project_plugins_service.py#L306


pnadolny13:
It looks like at this point the only check failing is precommit which is saying the method is too complex. I can quickly refactor that method but it will ballon the diff of this PR a bit. Let me know what you would like me to do.

src/meltano/core/project_plugins_service.py
```diff
@@ -234,29 +236,33 @@ def find_plugin(
...
+                    all_mappings = self.find_plugins_by_mapping_name(mapping_name)
+                    if len(all_mappings) > 1:
+                        raise click.ClickException(
+                            f"Ambiguous mapping name {mapping_name}, "
+                            "found multiple matches.",
+                        )
+                    return self.ensure_parent(plugin)
+        raise PluginNotFoundError(
+            PluginRef(plugin_type, plugin_name) if plugin_type else plugin_name,
+        )
```
edgarrmondragon, line 265: You can easily refactor this to something like:

```suggestion
                mapping = self._find_mapping(plugin_name, plugin)
                if mapping:
                    return mapping
        raise PluginNotFoundError(
            PluginRef(plugin_type, plugin_name) if plugin_type else plugin_name,
        )

    def _find_mapping(self, plugin_name: str, plugin: ProjectPlugin):
        mapping_name = plugin.extra_config.get("_mapping_name")
        if mapping_name == plugin_name:
            all_mappings = self.find_plugins_by_mapping_name(mapping_name)
            if len(all_mappings) > 1:
                raise click.ClickException(
                    f"Ambiguous mapping name {mapping_name}, "
                    "found multiple matches.",
                )
            return self.ensure_parent(plugin)
```

Although it's better not to use `click.ClickException` but a custom exception that inherits from `meltano.core.error.MeltanoError`

edgarrmondragon (changes requested):

src/meltano/core/project_plugins_service.py
pnadolny13: @edgarrmondragon cool thanks, I'll get it updated. I was keeping the ClickException only because of the changes I had to make in the parser. Altering it was triggering other tests to fail but now that I've got everything working I can probably work through that easier.

edgarrmondragon (approved):
Nice!

edgarrmondragon:
@pnadolny13 Should the docs in https://docs.meltano.com/guide/mappers/ be updated? I think at least the FAQ section should, wdyt?

pnadolny13:
@edgarrmondragon good point, I'll update that and see if theres other docs that can be updated too.

pnadolny13:
@edgarrmondragon I updated the docs. One final review then I think this should be good to merge!

docs/docs/guide/mappers.md
```diff
@@ -163,7 +173,7 @@ To see more examples check out the [plugin repo](https://github.com/MeltanoLabs/
 
 ### Can I used standalone mapper plugins with the `meltano elt` command?
```
edgarrmondragon, line 174:
```suggestion
### Can I use standalone mapper plugins with the `meltano elt` command?
```

edgarrmondragon (approved):

edgarrmondragon (approved):

edgarrmondragon:
:shipit:

PR #7996, opened 2023-08-04
chore: Remove `discovery.yml`

### LINKED ISSUE #7575: Remove the deprecated `discovery.yml` feature
As with all Meltano 3.0 features, this will target the `cloud` branch.

As part of this we should delete `src/meltano/schemas/discovery.schema.json`

Once this has been completed, the following issues can be closed as they will no longer be relevant:
- https://github.com/meltano/meltano/issues/6290

WillDaSilva:
To do:
- [ ] ~~Where practical, remove reliance on real plugin definitions in `tests/fixtures/core.py`~~ EDIT: Given the recent re-prioritization, I won't have time to do this. Even so, I'd rather get this PR merged in the state it's in than let it become stale.
- [x] Rename `plugin_discovery_service.py` -> `locked_definition_service.py`
- [x] Move `PluginRepository` into its own file

Closes #7575


src/meltano/core/project_plugins_service.py
```diff
@@ -82,10 +77,7 @@ def __init__(
         instruction = None
 
         if DefinitionSource.HUB in source:
-            instruction = (
-                f"Run `meltano discover {plugin.type.descriptor}` to explore available "
-                "plugins"
-            )
+            instruction = "Check Meltano Hub for available plugins"
```
edgarrmondragon, line 80: nit

```suggestion
            instruction = "Check https://hub.meltano.com/ for available plugins"
```

edgarrmondragon (approved):

src/meltano/core/project_plugins_service.py
WillDaSilva: I thought of doing that, but decided to use `Meltano Hub` instead because they might have their own Hub URL. Then again, if they are using their own Hub URL, they probably know well enough to check that instead of the Hub we host.

PR #7997, opened 2023-08-04
feat: Add `el` command as an alias of `elt` and depecrate both `elt` and the `--transform` option

### LINKED ISSUE #7617: Emit deprecation warning when `elt` command is used instead of `el`
Also emit a deprecation warning when the transform option is used.

The `el` command would have to exist before this could be completed.

Relates to:
- https://github.com/meltano/meltano/discussions/6273

edgarrmondragon:
- `meltano el` command docs: https://deploy-preview-7997--meltano.netlify.app/reference/command-line-interface/#el

Closes #7617


WillDaSilva (approved):
Clean as always @edgarrmondragon 👍

PR #8011, opened 2023-08-10
docs: Add JSON Schema information

anden-akkio:
Adds information about JSON Schema for `meltano.yml` and how to use it for things like generating Pydantic types and such. This process is useful for embedded ETL circumstances where you're modifying `meltano.yml` at runtime and want to validate the format of each write to fail fast, like we are at [Akkio](https://akkio.com). Another potential use case for even non-embedded use cases is to add an automated test that validates your `meltano.yml`, which prevents you from pushing a broken file to higher environments or even production. 

Attempted to follow PR guidelines at https://docs.meltano.com/contribute/merge, including semver PR name.


edgarrmondragon (approved):
Thanks @anden-akkio!

PR #8012, opened 2023-08-11
chore(deps): Use `pymssql` version `2.2.8`

WillDaSilva:
This fixes the Cython build issues. We should release a new version of Meltano with this. Note that for the previous version of Meltano released, the Dockerfile was updated to not install `pymssql` because of the aforementioned Cython build issue.


edgarrmondragon (approved):
The new version is missing wheels but I guess that's better than not being able to build it at all. I think Poetry is skipping the linux wheels, though: https://pypi.org/project/pymssql/2.2.8/#files

PR #8020, opened 2023-08-14
feat: Add `python` setting

### LINKED ISSUE #3333: Allow `python` interpreter overrides for specific plugins
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3411

Originally created by @aaronsteers on 2022-04-18 17:50:38

---

@edgarrmondragon suggested in https://gitlab.com/meltano/meltano/-/issues/2575#note_646844224 on `Aug 9, 2021`:

> @aaronsteers @pandemicsyn Perhaps we can just add the --python option to the venv creation call if a new python field is not null in the plugin configuration.

In theory, it should be possible to create venvs that run plugins using a different Python interpreter than Meltano itself is using.

This could be useful when Meltano itself is installed on a newer version of Python (such as 3.10) at the time of this writing, but an older version may be needed to run a specific plugin. As we expand the number of installable plugins to include everything on `https://hub.meltano.com`, the cases where specific plugins need an older (or newer) version of Python will in theory become much more common.

## Possible spec:

- Add an optional entry on plugin definitions: like `python` or (more explicit) `python_interpreter_path`.
- When not set, the python interpreter used would be the Meltano one.
- When set, the provided python interpreter would be used during plugin install (aka venv creation) instead of one Meltano is running under.

WillDaSilva:
This PR adds a `python` setting for each plugin, as well as a top-level `python` setting that provides the default for plugins which do not provide a value for this setting.

If no value is provided at the plugin-level or the top-level, the default is to use whatever version of Python Meltano is currently running on. If a value is provided, it should be a path to a Python executable, or the name of a Python executable that can be found on `$PATH`.

The specified Python executable will be used for the creation of new virtual environment, and by extension, used by the plugin subprocesses.

This allows plugins to run on a version independent from Meltano. In Meltano Cloud we may set the env var for this setting to `python3.9` so that by default plugins use the version they currently use on Meltano Cloud, chosen for its broad compatibility, while Meltano itself can run with `python3.11`. Users of Meltano Cloud (or any other Meltano host that provides multiple Python executable) can also opt to use other supported versions of Python for their plugins, whether for compatibility reasons, compliance reasons, performance reasons, or any other reason.

Relates to:
- https://github.com/meltano/infra/issues/1949

To do:
- [x] Integration test
- [x] Unit tests
- [x] Top-level `python` setting to set the default
- [x] Docs
  - [x] Top-level setting docs
  - [x] CLI examples
- [x] CLI update to support setting the `python` property when adding a plugin

Closes #3333


tayloramurphy:
@pnadolny13 you may want to take a look at this. We'll need an issue or two on the Hub side to add this to the metadata and API for a couple of plugins that we know require/recommend a specific version of python, but it's not an immediate need. 

@WillDaSilva I'll approve from the docs side, but I've done no UAT on this.

tayloramurphy (approved):
Docs approval

src/meltano/core/venv_service.py
```diff
@@ -10,93 +10,150 @@
...
-    try:
-        return PLATFORM_SPECS[system]
-    except KeyError as ex:
-        raise Exception(f"Platform {system!r} not supported.") from ex
+        if python is None:
+            python_path = sys.executable
+        elif isinstance(python, Path):
+            python_path = str(python.resolve())
+        else:
+            python_path = python if os.path.exists(python) else shutil.which(python)
```
edgarrmondragon, line 68: Would it be possible to fail early and with a better error message when `python` is a number, e.g. `3.8` (a likely user error)? Otherwise users will see:

```

stat: path should be string, bytes, os.PathLike or integer, not ScalarFloat
```

src/meltano/core/venv_service.py
```diff
@@ -228,8 +303,17 @@ async def create(self) -> Process:
         """
         logger.debug(f"Creating virtual environment for '{self.namespace}/{self.name}'")
         try:
-            return await exec_async(sys.executable, "-m", "virtualenv", str(self.venv))
```
edgarrmondragon, line 231: Why not use the `--python` option here?



[-p], [--python] | the python executable virtualenv is installed into | interpreter based on what to create environment (path/identifier) - by default use the interpreter where the tool is installed - first found wins
-- | -- | --

[-p]: https://virtualenv.pypa.io/en/latest/cli_interface.html#p
[--python]: https://virtualenv.pypa.io/en/latest/cli_interface.html#python

https://virtualenv.pypa.io/en/latest/cli_interface.html#section-discovery

edgarrmondragon (commented):
This seems to play well with `pyenv` 🎉.

If I run `pyenv local 3.11.3 3.8.17` within my project, then the Python executable is resolved to the right pyenv shim.

Only problem seems to be, if I don't set local versions and I have conflicting global pyenv shims (e.g. `pypy3.8` and `3.8.17`), then the error message is not helpful:

```

Failed to install plugin(s)
```

Under the hood, the expected pyenv error seems to be ocurring:

```python
>>> import subprocess
>>> from meltano.core.venv_service import VirtualEnv
>>> python_path = VirtualEnv._resolve_python_path("python3.8")d
>>> python_path
'/Users/edgarramirez/.pyenv/shims/python3.8'
>>> subprocess.run([python_path])
pyenv: python3.8: command not found

The `python3.8' command exists in these Python versions:
  3.8.16
  3.8.17
  pypy3.8-7.3.11

Note: See 'pyenv help global' for tips on allowing both
      python2 and python3 to be found.
CompletedProcess(args=['/Users/edgarramirez/.pyenv/shims/python3.8'], returncode=127)
```

so I wonder if this can be surfaced.

Amazing work!

src/meltano/core/venv_service.py
WillDaSilva: I wasn't aware of this option. This lets us avoid having to fall back to using the built-in `venv` instead of `virtualenv`. Thanks!

src/meltano/core/venv_service.py
WillDaSilva: Addressed in 2fe14b174a1fa3d5e2632621cb9e090757f4e3ea

src/meltano/core/venv_service.py
WillDaSilva: Addressed in 4a69f8bf5a2d7153d51b16ac4f5d05e6de8feed7

edgarrmondragon (approved):
Nice!

WillDaSilva:
@edgarrmondragon Thanks for the review. 7ef7ca5b1feb9d687ba13a776e632324e4808a6d should cause the error message from `pyenv` to be surfaced, per your suggestion.

edgarrmondragon:
> @edgarrmondragon Thanks for the review. [7ef7ca5](https://github.com/meltano/meltano/commit/7ef7ca5b1feb9d687ba13a776e632324e4808a6d) should cause the error message from `pyenv` to be surfaced, per your suggestion.

@WillDaSilva I'm probably missing something but I'm not seeing the full error output when something fails. I tried

```shell
meltano add extractor tap-readthedocs --python python3.12
```

And I expected to see the full version resolution error (`singer-sdk` doesn't support 3.12 yet), but I just got

```

Failed to install plugin(s)
```

---

That said, I think this is the historic behavior of the `meltano install` command so I don't want it to block this PR from merging.

edgarrmondragon (approved):

WillDaSilva:
@edgarrmondragon The only error message I updated to propagate through was for the case where the python version specified was found (i.e. there is a matching executable file), but isn't actually a Python executable, as is the case with:

```
pyenv: python3.8: command not found

The `python3.8' command exists in these Python versions:
  3.8.16
  3.8.17
  pypy3.8-7.3.11

Note: See 'pyenv help global' for tips on allowing both
      python2 and python3 to be found.
```

edgarrmondragon:
@WillDaSilva Ok. I can't actually reproduce the pyenv error though 😅. It might've been fixed by reverting to `virtualenv`:

```console
$ python3.10
pyenv: python3.10: command not found

The `python3.10' command exists in these Python versions:
  3.10.11
  3.10.12
  pypy3.10-7.3.12

Note: See 'pyenv help global' for tips on allowing both
      python2 and python3 to be found.
```

But `meltano add extractor tap-readthedocs --python python3.10` works fine 🤷‍♂️

Issue #7576, opened 2023-04-19
Deprecate plugin field `target_schema`


WillDaSilva:
_Originally posted by @tayloramurphy in https://github.com/meltano/meltano/issues/7548#issuecomment-1513551146_

edgarrmondragon:
Closed by https://github.com/meltano/meltano/pull/8041

PR #8041, opened 2023-08-18
fix!: Remove `target_schema` loader extra

edgarrmondragon:
See:

* https://github.com/meltano/meltano/issues/7576


WillDaSilva (approved):
Thanks @edgarrmondragon!

PR #8044, opened 2023-08-18
fix(core): Better error message when Azure connection string is missing

### LINKED ISSUE #8046: bug: Meltano crashes for Azure state backend when connection string is missing
### Meltano Version

2.20.0

### Python Version

NA

### Bug scope

### Operating System

NA

### Description

The [documentation](https://docs.meltano.com/concepts/state_backends/#azure-blob-storage) says that authentication will not be able to authenticate:

> To authenticate to Azure, Meltano will also need a [connection string](https://learn.microsoft.com/en-us/azure/storage/common/storage-configure-connection-string). You can provide this via the `state_backend.azure.connection_string` setting. If no state_backend.azure.connection_string setting is configured, Meltano will use the value of the `AZURE_STORAGE_CONNECTION_STRING` environment variable. If the connection string is not provided via either of these methods, Meltano will not be able to authenticate to Azure and any state operations will fail.

That statement is true, but Meltano will crash late with a less than useful error message.

Actually supporting a missing connection string is a separate issue: https://github.com/meltano/meltano/issues/8045

### Code

_No response_

edgarrmondragon:
Closes https://github.com/meltano/meltano/issues/8046


edgarrmondragon:
Ok, mypy is now catching the issue:

```
src/meltano/core/state_store/azure.py:100: error: Missing positional argument "account_url" in call to "BlobServiceClient"  [call-arg]
```

https://github.com/meltano/meltano/actions/runs/5907089200/job/16024403896?pr=8044#step:6:25

cjohnhanson (approved):
Sorry for the delay on this -- looks good! 🚀

Issue #8052, opened 2023-08-21
Validate input is a ULID for `meltano cloud project use --id`


WillDaSilva:
If anything other than a ULID is provided, the command will succeed, and then future commands will error with an unclear error message.

We can avoid this problem by validating that the provided value for the `--id` option is a ULID.

Context in the internal Slack: https://internalmeltano.slack.com/archives/C05N4BJEQJE/p1692561545277719

edgarrmondragon:
Closed by https://github.com/meltano/meltano/pull/8055

PR #8055, opened 2023-08-21
fix(cloud): Validate ULID inputs

edgarrmondragon:
Closes https://github.com/meltano/meltano/issues/8052


WillDaSilva (approved):
Thanks @edgarrmondragon!

PR #8078, opened 2023-08-30
docs: Fix `sidebar_class_name: hidden` in content

ReubenFrankel:
Looks like a bad find and replace from [this commit](https://github.com/meltano/meltano/pull/7936/commits/f7d65b7ac4c84117d07d08fac01def06f4e35143).

I just noticed this today, so not sure if there are any other similar occurrences.


edgarrmondragon (approved):
Thanks @ReubenFrankel!

PR #8162, opened 2023-09-18
ci: Group minor and patch Dependabot updates

WillDaSilva:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

This should reduce the number of Dependabot PRs we have to deal with, while still ensuring our CI passes for all of the dependency updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -40,7 +40,10 @@ updates:
     reviewers: [meltano/engineering]
     labels: [deps]
     groups:
-      default:
+      development-dependencies:
+        dependency-type: development
+      runtime-dependencies:
+        dependency-type: production
         update-types:
         - "minor"
         - "patch"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -29,6 +39,11 @@ updates:
       timezone: "UTC"
     reviewers: [meltano/engineering]
     labels: [deps]
+    groups:
+      default:
+        update-types:
+        - "minor"
+        - "patch"
```
edgarrmondragon, line 46:
```suggestion
    groups:
      development-dependencies:
        dependency-type: development
      runtime-dependencies:
        dependency-type: production
        update-types:
        - "minor"
        - "patch"
```

edgarrmondragon (approved):

Issue #8110, opened 2023-09-06
feature: Allow plugins to be re-added to a project


ReubenFrankel:
### Feature scope

CLI (options, error messages, logging, etc.)

### Description

As discussed in #7907, a `--force` (or similar) flag would enable a plugin to be re-added to the project, regardless of whether or not the plugin already exists. This would essentially be a shortcut for:

```
meltano remove <plugin type> <plugin name>
meltano add <plugin type> <plugin name>
```

This feature would be useful in resetting a plugin definition that perhaps isn't quite working after some user changes, updating a custom plugin that doesn't have a `.lock` file (i.e. added from `--custom` or [`--from-ref`](#7907)), or converting from a custom plugin to the Hub definition once published.

Adding a new plugin with this flag should just `meltano add` as normal.

In terms of flag name and implied behaviour:
- `--force` sort of makes sense for both new and existing plugins, but doesn't really convey *why* you might want to use it
- `--refresh`, `--reset` and `--update` make sense for existing plugins and imply the underlying behaviour, but don't really make sense for new plugins

Open to other suggestions. :slightly_smiling_face:

### Expectations

1. *Given a project with a plugin previously added from Meltano Hub*

    <details>
    <summary><code>meltano.yml</code></summary>
    
    ```yml
    plugins:
      extractors:
      - name: tap-spotify
        variant: matatika
        pip_url: git+https://github.com/Matatika/tap-spotify.git
    ```
    </details>

    <details>
    <summary><code>plugins/extractors/tap-spotify--matatika.lock</code></summary>

    ```json
    {
      "plugin_type": "extractors",
      "name": "tap-spotify",
      "namespace": "tap_spotify",
      "variant": "matatika",
      "label": "Spotify",
      "docs": "https://hub.meltano.com/extractors/tap-spotify--matatika",
      "repo": "https://github.com/Matatika/tap-spotify",
      "pip_url": "git+https://github.com/Matatika/tap-spotify.git",
      "description": "Audio Streaming and Media Services",
      "logo_url": "https://hub.meltano.com/assets/logos/extractors/spotify.png",
      "capabilities": [
        "about",
        "catalog",
        "discover",
        "schema-flattening",
        "state",
        "stream-maps"
      ],
      "settings_group_validation": [
        [
          "client_id",
          "client_secret",
          "refresh_token"
        ]
      ],
      "settings": [
        {
          "name": "client_id",
          "kind": "password",
          "label": "Client ID",
          "description": "App client ID"
[...]

PR #8169, opened 2023-09-20
feat: Added `--force-install` CLI flag for meltano `add` command.

### LINKED ISSUE #7277: feature: `--force` CLI flag for `meltano add`
### Feature scope

### Description

We currently support the `--force` CLI flag for `meltano install`. Since `meltano add` installs by default, it should also support this flag.

XshubhamX:
**Additions:**

1. Added the `--force` flag support for `add` command.
2. Added test.
3. Updated the doc for the same

cc @WillDaSilva 

Closes https://github.com/meltano/meltano/issues/7277


docs/docs/reference/command-line-interface.md
```diff
@@ -59,6 +59,8 @@ meltano add <type> <name>
 
 # For example:
 meltano add extractor tap-gitlab
+# Or you can also add a package forcefully using --force flag
```
WillDaSilva, line 62:
```suggestion
# You can ignore the required Python declared by a plugin by using --force flag
```

docs/docs/reference/command-line-interface.md
```diff
@@ -67,6 +69,8 @@ Without a `--custom`, `--inherit-from` or `--from-ref` option, this will add the
 to your [`meltano.yml` project file](/concepts/project#plugins)
 using a [shadowing plugin definition](/concepts/project#shadowing-plugin-definitions).
 
+The `--force` will ignore the required Python version declared by the plugins.
```
edgarrmondragon, line 72: Can you move this to the `Parameters` section of this command below?

tests/meltano/cli/test_add.py
```diff
@@ -793,3 +793,16 @@ def test_add_with_python_version(self, cli_runner: CliRunner):
                 ),
             )
             venv_mock.assert_called_once_with(python)
+
+    def test_add_with_force_flag(self, project: Project, cli_runner: CliRunner):
+        with mock.patch("meltano.cli.add.install_plugins") as install_plugin_mock:
+            install_plugin_mock.return_value = True
+            cli_runner.invoke(
+                cli, ["add", "transform", "tap-google-analytics", "--force"]
```
edgarrmondragon, line 801:
```suggestion
                cli,
                ["add", "transform", "tap-google-analytics", "--force"],
```

docs/docs/reference/command-line-interface.md
```diff
@@ -59,6 +59,8 @@ meltano add <type> <name>
 
 # For example:
 meltano add extractor tap-gitlab
+# Or you can also add a package forcefully using --force flag
+meltano add extractor tap-gitlab --force
```
edgarrmondragon, line 63:
```suggestion
meltano add extractor tap-gitlab --force
```

docs/docs/reference/command-line-interface.md
edgarrmondragon, line 1: Let's add a similar section to what we have for `install` above the `Note: Unlike...` note:

> If the plugin you are trying to add declares that it does not support the version of Python you are using, but you want to attempt to use it anyway, you can override the Python version restriction by providing the `--force` flag to `meltano add`.

edgarrmondragon (changes requested):
Thanks @XshubhamX! Requested a few changes.

edgarrmondragon:
You might wanna update instances of `install_plugin_mock.assert_called_*` to check the `force` parameter you're now passing it.

ReubenFrankel:
Great timing! :smile: We discussed this exact feature recently here: https://github.com/meltano/meltano/pull/7907#issuecomment-1708096810

Will this work as described in #8110?

> As discussed in #7907, a `--force` (or similar) flag would enable a plugin to be re-added to the project, regardless of whether or not the plugin already exists. This would essentially be a shortcut for:
>
> ```
> meltano remove <plugin type> <plugin name>
> meltano add <plugin type> <plugin name>
> ```
>
> This feature would be useful in resetting a plugin definition that perhaps isn't quite working after some user changes, updating a custom plugin that doesn't have a `.lock` file (i.e. added from `--custom` or [`--from-ref`](#7907)), or converting from a custom plugin to the Hub definition once published.
>
> Adding a new plugin with this flag should just `meltano add` as normal.

edgarrmondragon:
> Great timing! 😄 We discussed this exact feature recently here: [#7907 (comment)](https://github.com/meltano/meltano/pull/7907#issuecomment-1708096810)
> 
> Will this work as described in #8110
> 
> > As discussed in #7907, a `--force` (or similar) flag would enable a plugin to be re-added to the project, regardless of whether or not the plugin already exists. This would essentially be a shortcut for:
> > ```
> > meltano remove <plugin type> <plugin name>
> > meltano add <plugin type> <plugin name>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This feature would be useful in resetting a plugin definition that perhaps isn't quite working after some user changes, updating a custom plugin that doesn't have a `.lock` file (i.e. added from `--custom` or [`--from-ref`](#7907)), or converting from a custom plugin to the Hub definition once published.
> > Adding a new plugin with this flag should just `meltano add` as normal.
> 
> , or is this just a way to pass `--force` to the implicit `meltano install` step of `meltano add`?

Oh good call @ReubenFrankel. This is NOT https://github.com/meltano/meltano/issues/8110 😬.

I didn't realize we had two feature requests that suggested the same flag for different ends:

* https://github.com/meltano/meltano/issues/8110
* https://github.com/meltano/meltano/issues/7277

XshubhamX:
@WillDaSilva @edgarrmondragon Please check the PR, I have made the required changes

WillDaSilva:
@XshubhamX @edgarrmondragon @ReubenFrankel I think `--force` is fine here (if only to maintain consistency with `meltano install`), and I'd suggest we use `--reinstall` for #8110. I've left a comment on #8110 with my thoughts:
- https://github.com/meltano/meltano/issues/8110#issuecomment-1729845907

Please let me know what you all think of this approach.

ReubenFrankel:
@edgarrmondragon @WillDaSilva I'm happy to be convinced otherwise, but I actually think the opposite. :sweat_smile: https://github.com/meltano/meltano/issues/8110#issuecomment-1729855158

edgarrmondragon:
@XshubhamX Let's use `--force-install` instead of `--force`. We can get rid of the short option `-f`.

XshubhamX:
> @XshubhamX With the context given [above](https://github.com/meltano/meltano/pull/8169#issuecomment-1729739870), what do you think of using a different flag name for this feature, e.g. `--force-install`. Alternatively, we could suggest a different flag for #8110. Not requesting changes at the moment, just trying to get your thoughts on it.
> 
> cc @WillDaSilva @ReubenFrankel

I also think that using --force-install flag instead of --force flag quite fits with the already existing --no-install flag implementation.

edgarrmondragon:
@XshubhamX Given the additional context in https://github.com/meltano/meltano/issues/8110#issuecomment-1730124598, let's keep `--force` here.

Sorry for the back and forth 😅

XshubhamX:
> @XshubhamX Let's use `--force-install` instead of `--force`. We can get rid of the short option `-f`.

@edgarrmondragon  We are going ahead with this, right?

edgarrmondragon:
> > @XshubhamX Let's use `--force-install` instead of `--force`. We can get rid of the short option `-f`.
> 
> @edgarrmondragon We are going ahead with this, right?

Yeah, let's rename it :+1:

XshubhamX:
@edgarrmondragon PR is updated!

edgarrmondragon (approved):
Thanks @XshubhamX!

PR #7765, opened 2023-06-08
docs: Use tabs for meltano.yml vs meltano config or env

mjsqu:

PR #8201, opened 2023-09-30
docs: Use tabs for examples on settings page

mjsqu:
Updates the Settings page in line with my previous change (https://github.com/meltano/meltano/pull/7765) for the Plugins page, adding tabs.

Settings examples are split by the following where appropriate:

- meltano init
- meltano config
- export (environment variable)
- meltano command switch

The CSS change is the difference between a tab header being rendered like this:

```
meltano
config
```

and 

```
meltano config
```

I'm not entirely sure the correct wording for tab headers, so if a more concise word is used, the CSS update could be reverted - I'm unsure if I edited the correct spot in the CSS - and didn't assess where else it might have had an impact.


mjsqu:
A couple of extras:

- Selected tab highlighted with the cyan accent colour
- queryString - https://docusaurus.io/docs/markdown-features/tabs?current-os=ios#query-string - shifts all tabs to those of the same name, so if a user picks 'env' all tabs switch to 'env'.

tayloramurphy (approved):
Love this change! @cjohnhanson or @leoroese would either of you mind reviewing the CSS?

tayloramurphy:
@mjsqu I'm fine with the color as-is. I like the idea of the querystring param change though. Feel free to add that 👍 thanks for the PR!

leoroese (approved):

edgarrmondragon:
> @mjsqu I'm fine with the color as-is. I like the idea of the querystring param change though. Feel free to add that 👍 thanks for the PR!

@tayloramurphy That's already done, with the `meltano-tabs` query param and I confirmed it works. I think this is g2g.

PR #8213, opened 2023-10-05
fix: `meltano config <plugin> test` false-negative on Windows

### LINKED ISSUE #8212: bug: `meltano config <plugin> test` incorrectly reports a plugin configuration as invalid on Windows
```
Plugin configuration is invalid
{"type": "RECORD", "stream": "SUPPORT", "record": {"id": "2957642436270096567", "display_name": "b98ee58e372a7398b882a22abcd0b149", "email": "622474637484ce932692c2e646ba05d2", "department": "Support", "manager_id": "2636555004332212249", "manager_name": "67c835d0879ab870f0ffbc7f355b2b5b", "manager_email": "6897008ad2a760717537fbd0d8d69c9e", "2nd_level_manager_id": "2790094842804306626", "is_manager": "FALSE"}, "time_extracted": "2023-10-03T22:32:58.394266+00:00"}
```

Currently, the command uses a **negative** non-zero exit code to determine success after termination of the process by Meltano when encountering a `RECORD` message.

https://github.com/meltano/meltano/blob/4dcafc24bf0469794999950f5828943b981b80ff/src/meltano/core/plugin_test_service.py#L99-L105

It looks like there is a difference in the way Windows handles a process termination vs UNIX-based OSs:

> [A negative value `-N` indicates that the child was terminated by signal `N` (POSIX only).](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.returncode)
>
> [On POSIX OSs the method sends SIGTERM to the child. On Windows the Win32 API function TerminateProcess() is called to stop the child.](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.terminate)

It is possible that

**on UNIX OSs**, the `SIGTERM` signal is sent to the child process, and the negative of the signal number (`15`) is used as the process exit code (`-15`).
This registers as `meltano config <plugin> test` success (is a negative exit code).

**on Windows**, [`TerminateProcess()`](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-terminateprocess) is called with the resolved `SIGTERM` signal number (`15`), and then Python just returns the same value back as the process return code (`15`).
This registers as `meltano config <plugin> test` failure (is not a negative exit code).

---

*From [this](https://discuss.meltano.com/t/15889171/heyo-i-posted-https-meltano-slack-com-archives-cmn8helb0-p16) Slack thread*

ReubenFrankel:
Closes #8212

---

I'd like to refactor [`TestExtractorTestService`](https://github.com/meltano/meltano/blob/a35be0e5e131c0799f595f97bb1a63ff6a8c0679/tests/meltano/core/test_plugin_test_service.py#L44-L159) and [`TestCliConfig.test_config_test`](https://github.com/meltano/meltano/blob/a35be0e5e131c0799f595f97bb1a63ff6a8c0679/tests/meltano/cli/test_config.py#L69-L87) to spin up a real dummy process, rather than stub out return values through a mock. This means we could validate this fix on Windows through CI.


edgarrmondragon:
> I'd like to refactor [`TestExtractorTestService`](https://github.com/meltano/meltano/blob/a35be0e5e131c0799f595f97bb1a63ff6a8c0679/tests/meltano/core/test_plugin_test_service.py#L44-L159) and [`TestCliConfig.test_config_test`](https://github.com/meltano/meltano/blob/a35be0e5e131c0799f595f97bb1a63ff6a8c0679/tests/meltano/cli/test_config.py#L69-L87) to spin up a real dummy process, rather than stub out return values through a mock. This means we could validate this fix on Windows through CI.

@ReubenFrankel Let me know if I can help with that 👆

edgarrmondragon:
> @edgarrmondragon Yeah, I would appreciate some help if you have any ideas. Might unblock [meltano/edk#75 (comment)](https://github.com/meltano/edk/pull/75#issuecomment-1485849140) too if there is good pattern to follow.

@ReubenFrankel So, I don't have a windows machine with me to test but we can probably set up a dummy fixture tap that fails with a certain config value and passes otherwise. Wdyt?

ReubenFrankel:
> > @edgarrmondragon Yeah, I would appreciate some help if you have any ideas. Might unblock [meltano/edk#75 (comment)](https://github.com/meltano/edk/pull/75#issuecomment-1485849140) too if there is good pattern to follow.
> 
> @ReubenFrankel So, I don't have a windows machine with me to test but we can probably set up a dummy fixture tap that fails with a certain config value and passes otherwise. Wdyt?

The part I was struggling with was how a dummy process could be started and then properly terminated with respect to the host OS handling of `SIGTERM`, as per the [`process.terminate`](https://github.com/ReubenFrankel/meltano/blob/642d2904a998643055d70e4605c252281260a638/src/meltano/core/plugin_test_service.py#L95) call...

One idea I had was to create test resources that contain tap output indicating success (e.g. `valid.jsonl`)

```jsonl
{"type": "SCHEMA"}
{"type": "RECORD"}
{"type": "STATE"}
{"type": "SCHEMA"}
{"type": "RECORD"}
{"type": "STATE"}
{"type": "SCHEMA"}
{"type": "RECORD"}
{"type": "STATE"}
```

and failure (e.g. `invalid.jsonl`)

```jsonl
{"type": "SCHEMA"}
{"type": "SCHEMA"}
{"type": "SCHEMA"}
```

, and then mock the [`plugin_invoker.invoke_async` call](https://github.com/ReubenFrankel/meltano/blob/642d2904a998643055d70e4605c252281260a638/src/meltano/core/plugin_test_service.py#L74-L77) to stream the resource to `stdout` for the respective test. Hopefully that makes sense! :sweat_smile:

edgarrmondragon:
> I'd like to refactor [`TestExtractorTestService`](https://github.com/meltano/meltano/blob/a35be0e5e131c0799f595f97bb1a63ff6a8c0679/tests/meltano/core/test_plugin_test_service.py#L44-L159) and [`TestCliConfig.test_config_test`](https://github.com/meltano/meltano/blob/a35be0e5e131c0799f595f97bb1a63ff6a8c0679/tests/meltano/cli/test_config.py#L69-L87) to spin up a real dummy process, rather than stub out return values through a mock. This means we could validate this fix on Windows through CI.

@ReubenFrankel do you think that's a hard blocker for merging this PR?

ReubenFrankel:
I think I finally made some progress on this last night - I'll push a test POC later today and we can decide whether it's worth having or not.

ReubenFrankel:
Hmm... Working locally no problem, but looks like there is some kind of race condition in reading the emulated tap process output for `test_validate_failure_subprocess_err` specifically. Failing tests on Windows indicate there is also still an issue here (if they are to be believed)... @edgarrmondragon What do you think of this approach? It definitely feels a bit hacked in it's current state. :sweat_smile:

ReubenFrankel:
The [return code debug log](https://github.com/meltano/meltano/actions/runs/8903623347/job/24451671229#step:11:1670) seems to indicate that the process return code is always `1` on Windows for error or termination.

ReubenFrankel:
Decided to just flag if we encounter a `RECORD` message and use that to determine success, rather than rely on the process return code. I think this makes more sense anyway and should be more robust since it's now decoupled from OS-specific behaviour (although it would have been nice to see it working with the test POC implementation).

edgarrmondragon:
> return code. I think this makes more sense anyway (not sure why I with went return code initially) and should be more robust since it's now decoupled from OS-specific behaviour (although it would have been nice to see it working with the test POC implementation).

Yeah, that actually makes more sense 👍

edgarrmondragon (approved):

edgarrmondragon:
Thanks @ReubenFrankel!

PR #8228, opened 2023-10-15
feat(cli): Support setting configuration from files and STDIN

### LINKED ISSUE #8177: feature: Support passing contents of file to `meltano config <plugin> set <setting>` via stdin
### Feature scope

## Current situation

There's a few ways to set a config value at the moment:

```shell
# dot-separated setting name
meltano config meltano set cli.log_level info

# space-separated setting name
meltano config meltano set cli log_level info

# interactive
meltano config meltano set --interactive
```

## Proposal

Allow users to set a config value from a file. This is particularly  useful when the value is a multi-line string or some other value that's stored in a file somewhere.

The GitHub CLI supports this:

```console
$ gh secret set       
must pass name argument

Usage:  gh secret set <secret-name> [flags]

Flags:
  -a, --app string           Set the application for a secret: {actions|codespaces|dependabot}
  -b, --body string          The value for the secret (reads from standard input if not specified)
  -e, --env environment      Set deployment environment secret
  -f, --env-file file        Load secret names and values from a dotenv-formatted file
      --no-store             Print the encrypted, base64-encoded value instead of storing it on Github
  -o, --org organization     Set organization secret
  -r, --repos repositories   List of repositories that can access an organization or user secret
  -u, --user                 Set a secret for your user
  -v, --visibility string    Set visibility for an organization secret: {all|private|selected} (default "private")

$ gh secret set VERY_SECRET                    
? Paste your secret

$ gh secret set VERY_SECRET < private_key.txt 
✓ Set Actions secret VERY_SECRET for org/repo
```

Notice how there's 3 ways of supplying input:

1. Interactively by omitting a value.
2. Non-interactively by supplying a value to the `--body` option.
3. Non-interactively by omitting a value and reading from stdin.

The difficulty in implementing in our case is that we can't omit the value in non-interactive input because we accept multi-valued setting names:

```shell
# Meltano would think the setting is 'path.to.nested' with value
# 'setting' rather than 'path.to.nested.setting' with a value read
# from stdin
meltano config meltano set path to nested setting
```

One way to support stdin would be to use a special value to tell Meltano to read from stdin, perhaps `-`:

```console
$ meltano config my-plugin set private_key - < private_key.txt
Extractor 'my-plugin' setting 'private_key' was set in `.env`: 'dummy\n'

$ cat private_key.txt | meltano config my-plugin set private_key -
Extractor 'my-plugin' setting 'private_key' was set in `.env`: 'dummy2\n'
```

XshubhamX:
Additions:
1. New `--from-file` option.

Closes #8177


edgarrmondragon:
@XshubhamX you have some extraneous commits here. Can you rebase from `main`?

XshubhamX:
Update the PR. can you please reopen this @edgarrmondragon

tests/meltano/cli/test_config.py
```diff
@@ -65,6 +65,26 @@ def test_config_meltano_set(self, cli_runner):
             "'log_config.yml'"
         ) in result.stdout
 
+    @pytest.mark.usefixtures("project")
```
edgarrmondragon, line 68: can we add a new test (or update this existing one) to test the cases [described in the issue](https://github.com/meltano/meltano/issues/8177#issuecomment-1765061224)?

`cli.log_level` is a good case:

1. File contents: `{"log_level": "debug"}`. Command: `meltano config meltano set cli`
2. File contents: `debug`. Command: `meltano config meltano set cli.log_level`
3. File contents: `debug`. Command: `meltano config meltano set cli log_level`
```

**Review by edgarrmondragon — CHANGES_REQUESTED** [review]

File: src/meltano/cli/config.py [hunk 2]
```diff
@@ -356,7 +357,7 @@ def reset(ctx, store):
 
 
 @config.command(cls=PartialInstrumentedCmd, name="set")
-@click.option("--interactive", is_flag=True)
+@click.option("--stdin", is_flag=True)
```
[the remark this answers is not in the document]
XshubhamX: Reverted

tests/meltano/cli/test_config.py
XshubhamX: Added.

XshubhamX:
@edgarrmondragon PR is updated

src/meltano/cli/config.py
```diff
@@ -378,8 +381,38 @@ def set_(
         setting_name = tuple(setting_name[0].split("."))
 
     interaction = InteractiveConfig(ctx=ctx, store=store, extras=False)
+
     if interactive:
-        interaction.configure_all()
+        if sys.stdin.isatty():
+            # Interactive stdin flow
+            interaction.configure_all()
+        else:
```
edgarrmondragon, line 389: We got rid of the breaking change that replaced `--interactive` with `--stdin` but it feels a bit weird having to add `--interactive` to accept stdin.

Wdyt of having a `--from-file` option that accepts a filename we read the contents of:

```python
    if interactive:
        interaction.configure_all()
        ctx.exit()

    if from_file:
        setting_name += (value,)
        value = from_file.read()

    interaction.set_value(setting_name=setting_name, value=value, store=store)
```

This feels more complete, and we get `-` support out of the box:

> Command line tools are more fun if they work with files the Unix way, which is to accept `-` as a special file that refers to stdin/stdout.

https://click.palletsprojects.com/en/8.1.x/arguments/#file-arguments

XshubhamX: I checked the link, In this `-` would tell meltano that standard input will be provide but we'll still get `-` as an argument and we'll have to again remove it from setting_name. Moreover if we are not passing any file and just wan to use this `poetry run meltano --cwd ../shubh config meltano set cli log_level info`, it will still expect a file and in this case will consider info as a file.

Now either we should pass the the file as a value to this flag or the best is to just allow user to stream input using < file_location which we are doing right now, but we'll just rename the flag.

Also this --from-file can be confusing as it can convey that user has to provide file name of streaming a file, should we use `--stdin-from-file`. Now user can either pipe the input or just stream it from a file if they are using this flag.

wdyt?

src/meltano/cli/utils.py
```diff
@@ -679,3 +680,39 @@ def invoke(self, ctx):
...
+    """Update setting and value for non interactive stdin flow."""
+    current_setting += (current_value,)
+    input_data = click.get_text_stream("stdin").read()
+    try:
+        # Check if input requires parsing
+        value = json.loads(input_data)
+    except json.JSONDecodeError:
+        value = input_data
+
+    return current_setting, value
```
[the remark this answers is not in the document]
XshubhamX: Yes there is still as issue with `meltano config meltano set cli '{"log_level": "debug"}'` it sets cli.log_level = '{"log_level": "debug"}' 
Probably we should fix this such that we dont have to write custom function for parsing.I'll fix it

src/meltano/cli/config.py
edgarrmondragon: > I checked the link, In this `-` would tell meltano that standard input will be provide but we'll still get `-` as an argument and we'll have to again remove it from setting_name.

click supports `-` as a special value for `File` parameters, so there's no need for that.

> Also this --from-file can be confusing as it can convey that user has to provide file name of streaming a file, should we use `--stdin-from-file`. Now user can either pipe the input or just stream it from a file if they are using this flag.

I'm sorry I don't see how it'd cause that confusion.

```bash
# Configure a plugin with the contents of a file
meltano config tap-example set private_key --from-file private_key.txt

# Configure a plugin from stdin
uuidgen | meltano config tap-example set salt --from-file -
```

XshubhamX: These are 2 different things, This `@click.argument('input', type=click.File('rb'))` will only support the later one cause in 1st we are just providing file as an argument and it requires byte stream.

I'll say lets not make it complex and support this instead. Please share your thoughts


```
meltano config tap-example set -from-file cli < file.txt

# Configure a plugin from stdin
cat file.txt | meltano config tap-example set --from-file cli
```

and we'll just read the stdin in both cases

dresslife-shbh: @edgarrmondragon thoughts?

XshubhamX: @edgarrmondragon thoughts?

edgarrmondragon: Ok, so here's what I mean:

1. We change the scope of this issue to something like `feat: Support setting config from a file input and stdin`
2. We use `@click.argument('--from-file', type=click.File('r'))` to accept a file (in text mode). This lets us accept `-` as a filename to signify capturing stdin.
3. If `--from-file` is not null, we process the setting name to append the tail value

tests/meltano/cli/test_config.py
```diff
@@ -65,6 +66,58 @@ def test_config_meltano_set(self, cli_runner):
...
+        assert_cli_runner(result)
+
+        assert (
+            "Meltano setting 'cli.log_level' was set in `.env`: 'info'"
+        ) in result.stdout
+
+        result = cli_runner.invoke(
+            cli,
+            ["config", "meltano", "set", "--interactive", "cli.log_level"],
+            input=input_data,
```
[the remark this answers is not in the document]
XshubhamX: Done

src/meltano/cli/utils.py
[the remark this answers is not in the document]
XshubhamX: Sorry I Dint get it @edgarrmondragon

src/meltano/cli/config.py
XshubhamX: Done

src/meltano/cli/utils.py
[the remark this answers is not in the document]
edgarrmondragon: @XshubhamX Thanks for raising that!

FWIW we used to try JSON-deserializing input before but removed it cause it was a bit problematic (see https://github.com/meltano/meltano/pull/7017). So I don't want to increase the scope of this PR with that.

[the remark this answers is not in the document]
XshubhamX: @edgarrmondragon  should we retain the function that traverses nested settings in this PR or should we just accept that we will only receive text information in files rather than JSON?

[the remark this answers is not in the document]
XshubhamX: @edgarrmondragon Then we'd have to specify that files should only contain value of a specific setting and not nested setting, That we can mention in docs.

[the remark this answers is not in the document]
XshubhamX: @edgarrmondragon can you confirm here?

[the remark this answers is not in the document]
edgarrmondragon: @XshubhamX Yeah, a quick mention in the docs that the setting has to be of `object` or `array` for it to be deserialized works for me.

src/meltano/cli/config.py
```diff
@@ -372,14 +375,29 @@ def set_(
...
     if len(setting_name) == 1:
         setting_name = tuple(setting_name[0].split("."))
 
     interaction = InteractiveConfig(ctx=ctx, store=store, extras=False)
+
     if interactive:
-        interaction.configure_all()
+        if sys.stdin.isatty():
+            # Interactive stdin flow
+            interaction.configure_all()
```
edgarrmondragon: Interaction can only be explicitly requested so checking `sys.stdin.isatty()` feels redundant.

```suggestion
        interaction.configure_all()
        ctx.exit()
```

src/meltano/cli/config.py
```diff
@@ -372,14 +373,25 @@ def set_(
     value: t.Any,
     store: str,
     interactive: bool,
+    from_file,
```
edgarrmondragon, line 376:
```suggestion
    from_file: t.TextIO,
```

src/meltano/cli/config.py
```diff
@@ -372,14 +373,25 @@ def set_(
...
+        setting_name += (value,)
+        value = from_file.read()
+
+        interaction.set_value(
+            setting_name=setting_name,
+            value=value,
+            store=store,
+        )
     else:
         interaction.set_value(setting_name=setting_name, value=value, store=store)
```
edgarrmondragon, line 396: This can be simplified if we exit explicitly after configuring interactively:

```suggestion
    if interactive:
        interaction.configure_all()
        ctx.exit()

    if from_file:
        setting_name += (value,)
        value = from_file.read()

    interaction.set_value(setting_name=setting_name, value=value, store=store)
```

src/meltano/cli/config.py
```diff
@@ -372,16 +373,30 @@ def set_(
...
+        value = from_file.read()
+
+        interaction.set_value(
+            setting_name=setting_name,
+            value=value,
+            store=store,
+        )
+        ctx.exit()
+
+    interaction.set_value(setting_name=setting_name, value=value, store=store)
```
edgarrmondragon, line 399: No need to exit early here, we can reuse the setting name and value:

```suggestion
    if from_file:
        setting_name += (value,)
        value = from_file.read()

    interaction.set_value(setting_name=setting_name, value=value, store=store)
```

XshubhamX: @edgarrmondragon there are 3 different flows here, and we are also updating global variable, to prevent side effects shouldn't we explicity exit from each step. wdyt?

src/meltano/cli/config.py
```diff
@@ -372,16 +373,30 @@ def set_(
...
+
     if interactive:
         interaction.configure_all()
-    else:
-        interaction.set_value(setting_name=setting_name, value=value, store=store)
+        ctx.exit()
+
+    if from_file:
+        setting_name += (value,)
+        value = from_file.read()
```
edgarrmondragon, line 390: Wdyt about stripping any whitespace from the file contents?

```suggestion
        value = from_file.read().strip()
```

XshubhamX: Yes that could make sense

src/meltano/cli/config.py
XshubhamX: @edgarrmondragon I think it make sense to reuse these values because, after all, we are updating the global variables.

edgarrmondragon (approved):
Thanks @XshubhamX!

docs/docs/reference/command-line-interface.md
```diff
@@ -439,6 +439,25 @@ meltano config <plugin> set --interactive --extras
...
+# Configure plugin by passing file as argument
+meltano config <plugin> set nested.setting --from_file - ./file.txt
+
+# Configure plufin by passing the content of file as stdin
+cat file.txt | meltano config <plugin> set nested.setting --from_file -
+
+# File content {"setting": "info"}
+# Configure plugin by passing file as argument
+meltano config <plugin> set nested --from_file - ./file.txt
+```
```
edgarrmondragon, line 459:
````suggestion
### How to use: Read setting value from a file

The `--from-file` option can be used to read a configuration value from a file or a piped external process. This is specially useful for storing complex data like multiline strings, or passing a value generated by a different command line application.

* Settings of [kind](/reference/plugin-definition-syntax/#settingskind) `object` or `array` will be deserialized accordingly.
* The special filename `-` can be used to read from `STDIN`.

```bash
# Set setting `<name>` to the contents of `./file.txt`.
meltano config <plugin> set <name> --from-file ./file.txt

# Set setting `<name>` to the value returned by `uuidgen`.
uuidgen | meltano config <plugin> set <name> --from-file -
```
````

File: tests/meltano/cli/test_config.py [hunk 12]
```diff
@@ -65,6 +65,30 @@ def test_config_meltano_set(self, cli_runner):
             "'log_config.yml'"
         ) in result.stdout
 
+    @pytest.mark.usefixtures("project")
+    def test_config_non_interactive_stdin(self, cli_runner):
```
edgarrmondragon, line 69:
```suggestion
    def test_config_non_interactive_stdin(self, cli_runner, tmp_path):
```

tests/meltano/cli/test_config.py
```diff
@@ -65,6 +65,30 @@ def test_config_meltano_set(self, cli_runner):
...
+        result = cli_runner.invoke(
+            cli,
+            ["config", "meltano", "set", "cli.log_level", "--from-file", "-"],
+            input="info",
+        )
+        assert_cli_runner(result)
+
+        assert (
+            "Meltano setting 'cli.log_level' was set in `.env`: 'info'"
+        ) in result.stdout
```
edgarrmondragon, line 90:
```suggestion
        filepath = tmp_path.joinpath("file.txt")
        filepath.write_text("warning")

        result = cli_runner.invoke(
            cli,
            ["config", "meltano", "set", "cli.log_level", "--from-file", filepath],
        )
        assert_cli_runner(result)

        assert (
            "Meltano setting 'cli.log_level' was set in `.env`: 'warning'"
        ) in result.stdout
```

tests/meltano/cli/test_config.py
```diff
@@ -65,6 +65,32 @@ def test_config_meltano_set(self, cli_runner):
...
+
+        result = cli_runner.invoke(
+            cli,
+            ["config", "meltano", "set", "cli.log_level", "--from-file", filepath],
+        )
+        assert_cli_runner(result)
+
+        assert (
+            "Meltano setting 'cli.log_level' was set in `.env`: 'warning'"
+        ) in result.stdout
```
edgarrmondragon, line 92:
```suggestion
    def test_config_non_interactive_stdin(self, cli_runner, tap, tmp_path):
        result = cli_runner.invoke(
            cli,
            ["config", tap.name, "set", "private_key", "--from-file", "-"],
            input="content-from-stdin",
        )
        assert_cli_runner(result)

        assert (
            f"Extractor '{tap.name}' setting 'private_key' was set in `meltano.yml`: "
            "'content-from-stdin'"
        ) in result.stdout

        filepath = tmp_path.joinpath("file.txt")
        filepath.write_text("content-from-file")

        result = cli_runner.invoke(
            cli,
            ["config", tap.name, "set", "private_key", "--from-file", filepath],
        )
        assert_cli_runner(result)

        assert (
            f"Extractor '{tap.name}' setting 'private_key' was set in `meltano.yml`: "
            "'content-from-file'"
        ) in result.stdout
```

PR #8232, opened 2023-10-17
fix(cli): Interval validation for `meltano schedule set`

### LINKED ISSUE #7814: bug: `meltano schedule set` does not validate interval
### Meltano Version

2.19.1

### Python Version

3.10

### Bug scope

### Operating System

N/A

### Description

If you pass an invalid interval to `meltano schedule add`, it will error:
```bash
(meltano-dev-3.10) braedon ~/meltano/projects/test-project $ meltano schedule add csv-to-jsonl --interval fakeinterval --extractor tap-csv --loader target-jsonl
Invalid Cron expression or alias: 'fakeinterval'. Please use a valid cron expression.
```

The same is not true for `meltano schedule set`:
```bash
(meltano-dev-3.10) braedon ~/meltano/projects/test-project $ meltano schedule set csv-to-jsonl --interval fakeinterval
Updated schedule 'csv-to-jsonl'
(meltano-dev-3.10) braedon ~/meltano/projects/test-project $ meltano schedule list
[fakeinterval] elt csv-to-jsonl: tap-csv → target-jsonl x transforms
```

The same validation should be done when setting the interval with `set` and `add`.

### Code

_No response_

XshubhamX:
1. Used same validations that we use for `meltano schedule add`
https://github.com/meltano/meltano/blob/main/src/meltano/core/schedule_service.py#L216

cc @BraedonLeonard @edgarrmondragon 

Closes #7814

### DIFF
--- a/src/meltano/cli/schedule.py
+++ b/src/meltano/cli/schedule.py
@@ -7,6 +7,7 @@ import sys
 import typing as t
 
 import click
+from croniter import croniter
 
 from meltano.cli.params import pass_project
 from meltano.cli.utils import (
@@ -16,7 +17,12 @@ from meltano.cli.utils import (
 )
 from meltano.core.db import project_engine
 from meltano.core.job.stale_job_failer import fail_stale_jobs
-from meltano.core.schedule_service import ScheduleAlreadyExistsError, ScheduleService
+from meltano.core.schedule import CRON_INTERVALS
+from meltano.core.schedule_service import (
+    BadCronError,
+    ScheduleAlreadyExistsError,
+    ScheduleService,
+)
 from meltano.core.task_sets_service import TaskSetsService
 from meltano.core.utils import coerce_datetime
 
@@ -281,6 +287,13 @@ def remove(ctx, name):
     ctx.obj["schedule_service"].remove(name)
 
 
+def validate_interval(interval: str):
+    """Validate the interval for a schedule."""
+    if interval not in CRON_INTERVALS and not croniter.is_valid(interval):
+        raise BadCronError(interval)
+    return interval
+
+
 def _update_job_schedule(
     candidate: Schedule,
     job: str | None,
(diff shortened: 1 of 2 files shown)


BraedonLeonard:
Looks good to me! You'll need to add a unit test to ensure that validation is working as intended, and once you do I'll take another look and approve it 👍

src/meltano/cli/schedule.py
```diff
@@ -281,6 +287,13 @@ def remove(ctx, name):
     ctx.obj["schedule_service"].remove(name)
 
 
+def validate_interval(interval: str):
+    """Validate the interval for a schedule."""
+    if interval not in CRON_INTERVALS and not croniter.is_valid(interval):
+        raise BadCronError(interval)
+    return interval
```
edgarrmondragon, line 294: You can probably catch an invalid interval earlier by using a [`click.ParamType`](https://click.palletsprojects.com/en/8.1.x/api/#click.ParamType) for the `--interval` option. Wdyt?

XshubhamX: It works @edgarrmondragon , nice suggestion @ @

XshubhamX:
@edgarrmondragon @BraedonLeonard updated the PR

src/meltano/cli/schedule.py
```diff
@@ -350,13 +356,33 @@ def _update_elt_schedule(
...
+class CronParam(click.ParamType):
+    """Custom type definition for cron prameter."""
+
+    name = "cron"
+
+    def convert(self, value, *_):
+        """Validate and con interval."""
+        if value is None:
+            return None
+
```
edgarrmondragon, line 368: I think this code is unreachable, i.e. if `--interval` is not passed to the command, this method will not even be called and if `--interval` is indeed passed, then `value` is not `None`.

It can probably be removed:

```suggestion
```

BraedonLeonard (commented):
Look good to me 👍 I'll leave the final approval up to @edgarrmondragon

src/meltano/cli/schedule.py
XshubhamX: Make sense

edgarrmondragon (approved):

edgarrmondragon (approved):
Thanks @XshubhamX!

PR #8243, opened 2023-10-25
docs: Add "Scroll to Top" button to README

PrajwalGraj:
Added "Scroll to top" button at the bellow of the Readme.md file.Which on click will redirect us to the Top of file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,4 @@
+<div id="top"></div>
 <div align="center">
   <picture align="center" with="600">
     <source media="(prefers-color-scheme: dark)" srcset="https://user-images.githubusercontent.com/5394188/228290090-d8a1bfe4-988e-49ef-9f2f-a86f2b014536.svg">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -96,3 +97,4 @@ License
 -------
 
 This code is distributed under the MIT license, see the [LICENSE](https://github.com/meltano/meltano/blob/main/LICENSE) file.
+<p align="right">(<a href="#top">🔼 Back to top</a>)</p>
```
WillDaSilva, line 100: Thanks for this contribution @PrajwalGraj! I think that we should make it clear that this is not part of the above section by adding a horizontal bar, center the element, and remove the parentheses. I've implemented those suggestions here:

```suggestion

---

<p align="center"><a href="#top">🔼 Back to top</a></p>
```

Let me know what you think. If this seems good to you, I'll go ahead and approve & merge this.

PrajwalGraj: Thank you @WillDaSilva  for your quick response and your willingness to incorporate the changes.Iam Fine and ok with your changes, so you can go ahead and proceed🚀.
Thank you

tayloramurphy:
@PrajwalGraj thanks for suggesting this change, but with the readme being so small anyways I don't think we need this currently. If the readme gets longer we'll consider it. Thanks!

PR #8249, opened 2023-10-31
fix(cli): Display special characters in column names in the output of `meltano select --list`

### LINKED ISSUE #8244: bug: Select doesn't show % characters
### Meltano Version

3.1.0

### Python Version

3.10

### Bug scope

### Operating System

Ubuntu 20.04 (WSL)

### Description

I have a tap with properties that have a "%" in the name 

Example

```
visch@DESKTOP-9BDPA9T:~/git/meltano-projects/ait-mssql-snowflake$ meltano select tap-mssql --list
2023-10-27T15:03:27.475043Z [info     ] The default environment 'dev' will be ignored for `meltano select`. To configure a specific environment, please use the option `--environment=<environment name>`.
Legend:
        selected
        excluded
        automatic

Enabled patterns:
        *test*.*

Selected attributes:
        [selected ] dbo-test.APN
        [selected ] dbo-test.test
        [selected ] dbo-test.testcolum
 ```
 
 
 Meltano works properly with the tap and I get output like
 
 ```
 {"type": "RECORD", "stream": "dbo-test", "record": {"testcolum": "col1", "APN % Issue": "col3", "test%column": "col2"}, "version": 1698418991197, "time_extracted": "2023-10-27T15:03:11.205010Z"}
 ```
 
 So it looks to just be a UI thing
 

### Code

_No response_

### PR DESCRIPTION
**Pull request by edgarrmondragon**
Closed #8244

### CODE REVIEW

File: src/meltano/core/plugin/singer/catalog.py [hunk 1]
```diff
@@ -247,7 +247,7 @@ def path_property(path: str) -> str:
       stream[0].properties.list_items.properties.account → list_items.account
       stream[0].properties.name                          → name
     """
-    prop_regex = r"properties\.([\w\[\]\d]+)+"
+    prop_regex = r"properties\.([^.]+)+"
```
WillDaSilva, line 250: Is it okay that this also matches non-space whitespace characters such as newlines?

WillDaSilva (approved):

src/meltano/core/plugin/singer/catalog.py
edgarrmondragon: Yeah, I think it's very rare that a metadata/schema path expression would contain a newline or similar characters but this should keep

Issue #8202, opened 2023-09-30
There are a few more references to Jekyll later on the page, would be good if someone can work out if there's a new alternative or if these sections can be removed


edgarrmondragon:
There are a few more references to Jekyll later on the page, would be good if someone can work out if there's a new alternative or if these sections can be removed

_Originally posted by @mjsqu in https://github.com/meltano/meltano/pull/8200#issuecomment-1741812949_

PR #8251, opened 2023-11-05
docs: fix dbt model in part3.mdx

diegoquintanav:
Remove `raw` tag from example dbt model. Not sure if the `raw` bit is because of https://docs.meltano.com/contribute/docs#liquid-syntax-error


edgarrmondragon:
Thanks @diegoquintanav!

Liquid code is a leftover from our old Jekyll-based docs.

Related to #8202.

edgarrmondragon (approved):

PR #8195, opened 2023-09-29
feat: Added a `--merge-state` flag to `meltano run` to merge state with that of previous runs

arorarohan981:
Added a flag named --merge-state that tells Meltano whether to merge or overwrite state at the end of a run

Closes #8033

PR #8258, opened 2023-11-08
feat: Add a `--merge-state` flag to `meltano run` to merge the current pipeline state with that of the latest run

### LINKED ISSUE #8033: feature: Support a flag that tells Meltano whether to merge or overwrite state at the end of a run
### Feature scope

## Problem

Sometimes users want to run a full-refresh sync on a single stream. The user expects the bookmark for this stream to updated and that of the other streams to remain as is.

The `el`/`elt` command already supports this:

```
meltano elt --select <filter stream> --full-refresh --state-id
```

However, the `run` command can invoke more than one extractor, so it doesn't have `--select` option, so the state at the end of this pipeline

```
TAP_MYSQL__SELECT_FILTER='["table_1", "table_2"]' meltano run --full-refresh
```

only contains bookmarks for `table_1` and `table_2`.

This feature would allow it to achieve the same functionality.

## Spec

* A new CLI option to `meltano run`: `--merge-state`.

## Under the hood

* Add `merge_state` attribute to `ELBContext`

https://github.com/meltano/meltano/blob/d9a9809614c18e31029eb7acb272e0578a101842/src/meltano/core/block/extract_load.py#L69-L78

* Add the same attribute to `ELTContext` to satisfy the interface.
* Modify the `SingerTarget` plugin class to check for the `merge_state` attribute to decide whether to treat state as partial.

https://github.com/meltano/meltano/blob/d9a9809614c18e31029eb7acb272e0578a101842/src/meltano/core/plugin/singer/target.py#L165-L166

### Context

* Slack: https://meltano.slack.com/archives/C01UTUSP34M/p1692228669222549?thread_ts=1692181954.913689&cid=C01UTUSP34M
* Linen: https://www.linen.dev/s/meltano/t/15565276/hi-all-we-re-running-a-meltano-el-process-between-a-mysql-da#4bbc2aa9-e5b4-4223-a311-c8fee68ad58c

edgarrmondragon:
Closes #8033
Supersedes #8195

Docs:

* CLI reference: https://deploy-preview-8258--meltano.netlify.app/reference/command-line-interface/#parameters-3


WillDaSilva (approved):
@edgarrmondragon You're nothing if not thorough. This new CLI flag seem valuable, and I have no suggestions regarding its implementation & tests. Nice work! :shipit:

Do you think we should do a release once this goes in? The release would also include 
- https://github.com/meltano/meltano/pull/8268

edgarrmondragon:
> @edgarrmondragon You're nothing if not thorough. This new CLI flag seem valuable, and I have no suggestions regarding its implementation & tests. Nice work! :shipit:
> 
> Do you think we should do a release once this goes in? The release would also include
> 
> * [feat: Expand environment variables in array setting values #8268](https://github.com/meltano/meltano/pull/8268)

@WillDaSilva Yup, I'll be cutting a release in the coming days 😄!

PR #8266, opened 2023-11-13
chore: hide Cloud docs

tayloramurphy:
Cloud Docs are still accessible, they're now just hidden from the UI. Search still works for them.


WillDaSilva (approved):

tayloramurphy:
FYI @sbalnojan I'm making this change to the docs!

PR #8268, opened 2023-11-13
feat: Expand environment variables in array setting values

### LINKED ISSUE #3171: Nested config values in setting arrays and objects don't have environment variables expanded
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3245

Originally created by @alxthm on 2022-02-15 09:06:57

---

Related to [#2366](https://gitlab.com/meltano/meltano/-/issues/2366) 

I was having trouble using environment variables in the `meltano.yml` file, and it seems they are only supported for top-level configuration options.

If I define `SOME_ENV_VAR=SOME_VALUE` in the `.env` file and I write this configuration yaml:
```yaml
plugins:
  extractors:
    - name: tap-spreadsheets-anywhere
      variant: ets
      pip_url: git+https://github.com/ets/tap-spreadsheets-anywhere.git
      config:
        test: $SOME_ENV_VAR
        tables:
          - path: $SOME_ENV_VAR
            name: my_table
```

Then this is the actual configuration used by meltano:
```
$ meltano config tap-spreadsheets-anywhere
{
  "tables": [
    {
      "path": "$SOME_ENV_VAR",
      "name": "my_table"
    }
  ],
  "test": "SOME_VALUE"
}
```

And the `tables.path` variable is not set.

The environment variable `SOME_ENV_VAR` should expend to `SOME_VALUE` in the `tables` array as well, and we should have:
```
$ meltano config tap-spreadsheets-anywhere
{
  "tables": [
    {
      "path": "SOME_VALUE",
      "name": "my_table"
    }
  ],
  "test": "SOME_VALUE"
}
```

### Steps to reproduce

- add the configuration above in `meltano.yml`
- add the line `SOME_ENV_VAR=SOME_VALUE` to your `.env` file
- see the result with `meltano config tap-spreadsheets-anywhere`

### Relevant logs and/or screenshots
_Please use code blocks (\`\`\`) to format console output_

### Possible fixes
_If you can, link to the line of code that might be responsible for the problem or suggest a fix_

### Further regression test
_Ensure we automatically catch similar issues in the future_

- [ ] Write additional adequate test cases and submit test results

edgarrmondragon:
Closes https://github.com/meltano/meltano/issues/3171


integration/example-library/meltano-expand-envvars-in-array/index.md
```diff
@@ -0,0 +1,19 @@
...
+```shell
+meltano install
+```
+
+## Run the pipeline
+
+```shell
+ANIMAL_GROUP=birds meltano invoke tap-smoke-test
+ANIMAL_GROUP=reptiles meltano invoke tap-smoke-test
+ANIMAL_GROUP=mammals meltano invoke tap-smoke-test
```
edgarrmondragon, line 18: These expand this config

```yaml
- stream_name: $ANIMAL_GROUP
  input_filename: $MELTANO_PROJECT_ROOT/data/$ANIMAL_GROUP.jsonl
```

into different values and thus different pipelines :)

ian-lewis-d:
This exactly fits my current case where I have data coming from several environments - all with the same format. As a result I _should_ be able to use the same pipeline without duplicating it. 

Releasing this fix would assist me hugely, and many others I'm sure.

WillDaSilva (approved):
We may want to use `t.Sequence` instead of `list`, but even without that change I'm very happy to see this change. Thanks @edgarrmondragon! (and sorry for taking so long to review it!)

src/meltano/core/utils/__init__.py
```diff
@@ -477,7 +477,12 @@ def __init__(self, env_var: str):
     re.VERBOSE,
 )
 
-Expandable = t.TypeVar("Expandable", str, t.Mapping[str, "Expandable"])
+Expandable = t.TypeVar(
+    "Expandable",
+    str,
+    t.Mapping[str, "Expandable"],
+    t.List["Expandable"],
```
edgarrmondragon, line 484:
```suggestion
    t.Sequence["Expandable"],
```

edgarrmondragon:
> We may want to use `t.Sequence` instead of `list`, but even without that change I'm very happy to see this change. Thanks @edgarrmondragon! (and sorry for taking so long to review it!)

@WillDaSilva that makes sense 👍

Thanks for reviewing!

PR #8283, opened 2023-11-21
chore: Use Melty token to create version bump PR

edgarrmondragon:


edgarrmondragon:
Confirmed that this lets the version bump workflow go through: https://github.com/meltano/meltano/pull/8284

Issue #8256, opened 2023-11-07
feature: Avoid initializing the snowplow tracker when usage stats are disabled or emitters can't be initialized


jaceksan:
### Feature scope

CLI (options, error messages, logging, etc.)

### Description

I need to share the same virtual env with Meltano and dbt, because I want to orchestrate them with Dagster.

Currently, the snowplow-tracker is required by both Meltano and dbt, but they require different incompatible versions.

Please, stop initializing the tracker if the usage stats are disabled.

tayloramurphy:
@edgarrmondragon I know you did work on this. As I understand it there's not much we can do here b/c even though no events are sent the package is still there. Would we have to ship a version of Meltano w/o snowplow as a dependency to achieve this?

edgarrmondragon:
> @edgarrmondragon I know you did work on this. As I understand it there's not much we can do here b/c even though no events are sent the package is still there. Would we have to ship a version of Meltano w/o snowplow as a dependency to achieve this?

I think we should probably catch any errors raised during the initialization of the tracker, log an event at the warning level, then continue to execute the application. The biggest change would be updating all the places where the tracker is called to fire an event to make it instead a no-op.

Related:

* https://github.com/meltano/meltano/issues/7895

edgarrmondragon:
> @edgarrmondragon To avoid updating every piece of code that calls/uses the tracker, we could stub its functionality within our tracker class if the underlying Snowplow tracker fails to initialize. The code that calls the tracker doesn't need to ensure it actually runs, so no functionality should be affected, save for the telemetry being disabled of course.

@WillDaSilva That makes sense! So we could

1. Catch any exceptions when initializing emitters here: https://github.com/meltano/meltano/blob/72530e9579032aa96867f219f13948e5991c86f2/src/meltano/core/tracking/tracker.py#L116-L128
2. Log a warning that the tracker will not be initialized

With this, `Tracker.can_track` will already make calls from the cli code no-ops.

cc @jaceksan

dresslife-shbh:
@edgarrmondragon can i work on this?

edgarrmondragon:
> @edgarrmondragon can i work on this?

Yes, go for it 🙂

edgarrmondragon:
Note that this will only fix _some_ of the issues of installing dbt-core in the  virtual environment as Meltano. For example:

```bash
$ virtualenv venv
$ source venv/bin/activate
(venv) $ pip install meltano dbt-core
[...]

PR #8302, opened 2023-12-05
fix: Do not initialize snowplow tracker if anything goes wrong with it

jaceksan:
Closes issue #8256

It may be a temporary workaround -
it may be removed once https://github.com/dbt-labs/dbt-core/pull/8680 is merged.


jaceksan:
> Hi @jaceksan!
> 
> Correct me if I'm wrong but I don't think this fixes #8256 because the error does not occur at import time, but rather at _initialization_ time in
> 
> https://github.com/meltano/meltano/blob/72530e9579032aa96867f219f13948e5991c86f2/src/meltano/core/tracking/tracker.py#L116-L128
> 
> .
> What do you think? This the change in this PR fixed the problem you were having with dbt?

Hey @edgarrmondragon!
The issue I experienced was already described in one of the comments in the issue:
```
ImportError: cannot import name 'SelfDescribing' from 'snowplow_tracker'
```
This happens when you install dbt-core into a venv where you already installed Meltano.
I added dbt-core to the env created by poetry in meltano repo and executed:
```
meltano init --no-usage-stats --force .
```
It did not fail anymore.

But I am quite new in poetry, I may use it wrong.

But I think I get your point.
When you install both tools in a different way, it may NOT fail on imports and may fail later in the section you linked.
I could implement a check there with can_track(). Actually, I already tried it as an alternative solution.
I realized that it was not that easy, because can_track() calls other class methods which require some of the imports.
I could split imports and import the problematic libs as late as possible to make this alternative solution work as much as possible.
Should I extend it this way?

edgarrmondragon:
@jaceksan Do `meltano init` does work, but I fear other commands would still be broken with this change. How about `meltano install`? It imports `PluginsTrackingContext`, which depends on `SelfDescribingJson` so I'm afraid that'd still be broken.

jaceksan:
> meltano install

You are right. I re-tested it and even `meltano init` does not work. It reports the warning as expected but fails when initiating the Emitters.

I try to implement a better solution as soon as possible.

edgarrmondragon:
> @edgarrmondragon I analyzed the whole code base and I am afraid I can't solve this.
> 
> `SelfDescribingJson` is used in too many places. A couple of classes are inherited from it, which is not easy to override.
> 
> Do you have any idea how to handle such cases?

@jaceksan One option that comes to mind is

1. banning imports of any imports from `meltano.core.tracking.*` subpackages and
2. instead put all those imports in `src/meltano/core/tracking/__init__.py`,
3. wrap them in a try/except block and export a boolean constant `TRACKER_AVAILABLE` if any import fails,
4. elsewhere import things from that root module
5. use `TRACKER_AVAILABLE` to conditionally instantiate the tracker, add contexts, etc.

jaceksan:
I was notified by the related dbt-core issue that they are not going to work on it.
I try to find time to continue working on this, it blocks me from orchestrating both Meltano and dbt by Dagster.

edgarrmondragon:
> I was notified by the related dbt-core issue that they are not going to work on it.
> I try to find time to continue working on this, it blocks me from orchestrating both Meltano and dbt by Dagster.
> 

Thanks @jaceksan. Let me know if the plan above makes sense.

src/meltano/core/tracking/tracker.py
```diff
@@ -42,14 +40,68 @@
...
+try:
+    from snowplow_tracker import (
+        Emitter,
+        SelfDescribing,
+        SelfDescribingJson,  # type: ignore
+    )
+    from snowplow_tracker import Tracker as SnowplowTracker
+
+    SNOWPLOW_TRACKER_AVAILABLE = True
+except Exception:
```
edgarrmondragon, line 58:
```suggestion
except Exception:  # pragma: no cover
```

src/meltano/core/tracking/tracker.py
```diff
@@ -113,20 +165,29 @@ def __init__(  # noqa: WPS210 - too many local variables
...
-                continue
-            parsed_url = urlparse(endpoint)
-            emitters.append(
-                Emitter(
-                    endpoint=parsed_url.hostname + parsed_url.path,
-                    protocol=parsed_url.scheme or "http",
-                    port=parsed_url.port,
-                    request_timeout=request_timeout,
-                ),
+        # Supress errors when instantiating Emmiters
```
edgarrmondragon, line 168:
```suggestion
        emitters: list[Emitter] = []

        # Supress errors when instantiating Emmiters
```

src/meltano/core/tracking/tracker.py
```diff
@@ -113,20 +165,29 @@ def __init__(  # noqa: WPS210 - too many local variables
...
+        # It helps to resolve issues related to import conflicts -
+        # snowplow-tracker vs. minimal-snowplow tracker
+        if SNOWPLOW_TRACKER_AVAILABLE:
+            emitters: list[Emitter] = []
```
edgarrmondragon, line 172:
```suggestion
```

src/meltano/core/tracking/tracker.py
```diff
@@ -113,20 +165,29 @@ def __init__(  # noqa: WPS210 - too many local variables
...
+                        protocol=parsed_url.scheme or "http",
+                        port=parsed_url.port,
+                        request_timeout=request_timeout,
+                    ),
+                )
+        else:
+            logger.warning(
+                "Snowplow tracker is not available, setting emitters to empty list",
             )
+            emitters = []
```
edgarrmondragon, line 190:
```suggestion
```

edgarrmondragon:
The code looks good but it's probably worth doing some manual testing with common ways of installing Meltano along with dbt in a venv to confirm everything's working as expected.

jaceksan:
> The code looks good but it's probably worth doing some manual testing with common ways of installing Meltano along with dbt in a venv to confirm everything's working as expected.

I tested it.
I installed Meltano and THEN dbt, which overrides snowplow-tracker by their minimal-snowplow-tracker.
Then I executed meltano run on top of my quite complex project consisting of tap-github, tap-csv-s3, tap-jira, target-snowflake, and target-duckdb. It finished without any issues.
Then I executed dbt run on top of the same project and it finished OK too.

The problem with my solution is that developers have to install Meltano first and then dbt. When they do it in opposite order, dbt won't work properly affected by snowplow-tracker delivered by Meltano.

I can share with you how exactly I tested it, so you can reproduce it.
The project is open-sourced.

edgarrmondragon:
@jaceksan can you give @jaceksan a try?

TL;DR: dbt uses the same package name `snowplow_tracker` but a different _distribution_ name `minimal_snowplow_tracker` so we can ensure the right one is imported by using `importlib.util.import_module`.

jaceksan:
> @jaceksan can you give #8400 a try?
> 
> TL;DR: dbt uses the same package name `snowplow_tracker` but a different _distribution_ name `minimal_snowplow_tracker` so we can ensure the right one is imported by using `importlib.util.import_module`.
> 
> > The problem with my solution is that developers have to install Meltano first and then dbt. When they do it in opposite order, dbt won't work properly affected by snowplow-tracker delivered by Meltano.
> 
> That should be fixed by [dbt-labs/dbt-core#9528](https://github.com/dbt-labs/dbt-core/pull/9528)

Great contribution to dbt-core, thanks for that!

I installed Meltano from your branch and executed my project. Unfortunately, it does not work for me.
Screenshot from 2024-02-13 11-09-08

edgarrmondragon:
Closing this now based on https://github.com/meltano/meltano/issues/8256#issuecomment-2064882578. We can revisit later if we still feel this is needed.

PR #8351, opened 2024-01-11
feat: Warn if select rules are not used on catalog

### LINKED ISSUE #8336: feature: Warn when `select`ed fields are not present in the tap's catalog
### Feature scope

### Description

When meltano is invoked with `el`, `elt` or `run`, it seems to extract the intersection of fields specified in the tap's `select` configuration and the discovered catalog.

It would be great if meltano were to report fields that were not found in the catalog.

There are a couple ways I can think of where this might happen:
1. The Data Developer made a typo and specified a non-existent field.
2. The Source system deprecated & removed a field without proper communication to the data team, or lack of action on the part of the data team.

The 2nd reason above has gotten us a couple of times. Business processes were changed, fields were deprecated and replacement fields with equivalent information introduced, but the integration still looked at the non-existent information.

A warning log message would enable meltano users to set up appropriate alerts and get notified if both the above scenarios occurred.

haleemur:
- Add method to emit a warning log message if a selected field is not present in the discovered catalog.
- Patch function in unrelated test

I haven't quite figured out yet how to add a test that validates this function is emitting the log message. (I'm stuck on figuring out how to set up the test such log is emitted, but I'd gladly accept any guidance to implement it, specially if the lack of this test is considered a blocker to this PR getting accepted). 

However, I ran it against an actual target (marketo), and found it to be working. Intentionally, the field `industry` was mis-spelt as `Industry`, and as can be seen from the screenshot below, the 2nd log line emits the warning ```Property `leads.Industry` was not found in the tap's Catalog```.

This is useful as it lets the data engineer know that a specified property was found not to exist in the Catalog.

Screenshot 2024-01-11 at 04 06 35

closes: https://github.com/meltano/meltano/issues/8336


tests/meltano/core/plugin/singer/test_tap.py
```diff
@@ -201,44 +201,45 @@ def mock_discovery(*args, **kwargs):  # noqa: ARG001
...
+                    assert catalog_cache_key_path.exists()
+
+                    # If the cache key no longer matches, discovery is invoked again
+                    mocked_run_discovery.reset_mock()
+                    catalog_cache_key_path.write_text("bogus")
+                    await subject.discover_catalog(invoker)
+
+                    assert mocked_run_discovery.called
+                    assert json.loads(catalog_path.read_text()) == {"discovered": True}
+                    assert not catalog_cache_key_path.exists()
```
edgarrmondragon, line 243: You could add something like this to test the output:

```suggestion

    @pytest.mark.parametrize(
        ("rule", "catalog", "message", "emit_warning"),
        (
            pytest.param(
                MetadataRule("foo", ["properties", "bar"], "baz", False),
                {"streams": []},
                "Stream `foo` was not found",
                True,
                id="stream_not_found",
            ),
            pytest.param(
                MetadataRule("foo", ["properties", "bar"], "baz", False),
                {"streams": [{"tap_stream_id": "foo"}]},
                "Property `foo.bar` was not found",
                True,
                id="property_not_found",
            ),
            pytest.param(
                MetadataRule("foo", ["properties", "bar"], "baz", False),
                {"streams": [{"tap_stream_id": "foo"}]},
                "Stream `foo` was not found",
                False,
                id="stream_is_found",
            ),
            pytest.param(
                MetadataRule("foo", ["properties", "bar"], "baz", False),
                {
                    "streams": [
                        {
                            "tap_stream_id": "foo",
                            "schema": {"properties": {"bar": {"type": "string"}}},
                        },
                    ],
                },
                "Property `foo.bar` was not found",
                False,
                id="property_is_found",
            ),
            pytest.param(
                MetadataRule("*", ["properties", "bar"], "baz", False),
                {
                    "streams": [],
                },
                "Stream `foo` was not found",
                False,
                id="stream_is_wildcard",
            ),
            pytest.param(
                MetadataRule("foo", ["properties", "*"], "baz", False),
                {"streams": [{"tap_stream_id": "foo"}]},
                "Property `foo.bar` was not found",
                False,
                id="property_is_wildcard",
            ),
        ),
    )
    def test_warn_property_not_found(
        self,
        subject: SingerTap,
        capsys: pytest.CaptureFixture,
        rule: MetadataRule,
        catalog: dict,
        message: str,
        emit_warning: bool,
    ):
        subject.warn_property_not_found([rule], catalog)

        # If the warning is emitted, it should be in the output
        assert emit_warning is (message in capsys.readouterr().out)

```

edgarrmondragon (changes requested):

src/meltano/core/plugin/singer/tap.py
```diff
@@ -557,6 +559,7 @@ def apply_catalog_rules(  # noqa: WPS213, WPS231, WPS210
                 SchemaExecutor(schema_rules).visit(catalog)
 
             if metadata_rules:
+                self.warn_property_not_found(metadata_rules, catalog)
```
edgarrmondragon, line 562: Thanks @haleemur! IIUC, this checks the metadata rules against the stream's schema so I suggested a slightly different message when a property can't be found.

tests/meltano/core/plugin/singer/test_tap.py
haleemur: Thanks for this suggestion. I'll play with it after work tomorrow. I appreciate all the suggestions & applied the others!

haleemur: when I run `pytest`, the following errors are returned.

```
SKIPPED [1] tests/meltano/cli/test_elt.py:192: Test is only for Windows
XFAIL tests/meltano/cli/test_add.py::TestCliAdd::test_add_fails - Uninstall not implemented yet.
FAILED tests/meltano/core/plugin/singer/test_tap.py::TestSingerTap::test_warn_property_not_found[property_not_found] - AssertionError: assert True is ('Property `bar` was not found in the schema of stream `foo`' in '')
FAILED tests/meltano/core/plugin/singer/test_tap.py::TestSingerTap::test_warn_property_not_found[stream_not_found] - AssertionError: assert True is ('Stream `foo` was not found' in '')
```

however, when the individual test is run, `poetry run pytest tests/meltano/core/plugin/singer/test_tap.py::TestSingerTap::test_warn_property_not_found`, all 6 versions pass.

I'm not sure what's happening here, and I'm not that familiar with `pytest`. Do you have any suggestions for this error?

haleemur: The issue above has been resolved, it seems doing `poetry run pytest` ends up with the log captured in the built in pytest fixture `caplog`, whereas if the test is run by itself, the output is not captured in `caplog`, but instead in `capsys`. i'm joining the output from both and testing for the presence of the expected warning messages.

haleemur:
@edgarrmondragon it looks like the commit checks don't like %-formatting for logs, which were your suggestions

https://results.pre-commit.ci/run/github/378997214/1705342208.pEmW_fVhQQWMEIKYNntG7A

I've reverted those to use f-strings so that the code is consistent with the pre-commit expectations. Let me know if I should try a different approach.

src/meltano/core/plugin/singer/tap.py
```diff
@@ -615,3 +618,56 @@ def catalog_cache_key(self, plugin_invoker):
...
+        def dict_get(dictionary, key):
+            return dictionary.get(key, {})
+
+        for rule in property_rules:
+            if not (path := takewhile(is_not_star, rule.breadcrumb)):
+                continue
+            if not (s := stream_dict.get(rule.tap_stream_id)):
+                logger.warning(
+                    f"Stream `{rule.tap_stream_id}` was not found",
+                )
```
edgarrmondragon, line 667:
```suggestion
                logger.warning(
                    "Stream `%s` was not found",  # noqa: WPS323
                    rule.tap_stream_id,
                )
```

src/meltano/core/plugin/singer/tap.py
```diff
@@ -615,3 +618,56 @@ def catalog_cache_key(self, plugin_invoker):
...
+                continue
+            if not reduce(dict_get, path, s.get("schema", {})):
+                logger.warning(
+                    f"Property `{'.'.join(rule.breadcrumb[1:])}` was not found in "
+                    f"the schema of stream `{rule.tap_stream_id}`",
+                )
```
edgarrmondragon, line 673:
```suggestion
                logger.warning(
                    "Property `%s` was not found in the schema of stream `%s`",  # noqa: E501, WPS323
                    ".".join(rule.breadcrumb[1:]),
                    rule.tap_stream_id,
                )
```

edgarrmondragon (commented):
@haleemur Thanks for addressing the feedback! I think flakeheaven is being overzealous about %-style strings in logging calls. It's actually fine to use them here, and actually preferred for most cases (e.g. see https://discuss.python.org/t/safer-logging-methods-for-f-strings-and-new-style-formatting/13802).

https://docs.python.org/3/howto/logging.html#optimization.

src/meltano/core/plugin/singer/tap.py
```diff
@@ -615,3 +618,56 @@ def catalog_cache_key(self, plugin_invoker):
...
+        }
+
+        def is_not_star(x):
+            return x != "*"
+
+        def dict_get(dictionary, key):
+            return dictionary.get(key, {})
+
+        for rule in property_rules:
+            if not (path := takewhile(is_not_star, rule.breadcrumb)):
```
edgarrmondragon, line 662: `takewhile` returns a _lazy_ iterator so I'm not sure which falsey value of `path` is expected to cause the code to follow this branch.

haleemur: you're correct, and as you have suggesting, wrapping in `list` would resolve this issue.

src/meltano/core/plugin/singer/tap.py
```diff
@@ -615,3 +619,58 @@ def catalog_cache_key(self, plugin_invoker):
...
+              not be used for validation
+
+        Args:
+            rules: List of `MetadataRule`
+            catalog: Discovered Source Catalog
+        """
+        property_rules = [
+            r
+            for r in rules
+            if r.tap_stream_id != "*"
```
edgarrmondragon, line 646: Meltano also accepts wildcards _within_ stream and property names, so I think this would   try to find a field `prefix_*` or a stream `public-*`, for example. We should probably account for that and include some test cases.

For implementation details: Meltano uses [fnmatch](https://docs.python.org/3/library/fnmatch.html).

src/meltano/core/plugin/singer/tap.py
```diff
@@ -615,3 +619,58 @@ def catalog_cache_key(self, plugin_invoker):
...
+            and len(r.breadcrumb) >= 2
+            and r.breadcrumb != ["properties", "*"]
```
edgarrmondragon, line 648: I think this is causing select rules like `my_stream.*` to fall through, which should raise a ``Stream `my_stream` was not found`` warning.

src/meltano/core/plugin/singer/tap.py
```diff
@@ -615,3 +619,58 @@ def catalog_cache_key(self, plugin_invoker):
...
+        ]
+        stream_dict = {
+            stream.get("tap_stream_id"): stream
+            for stream in catalog.get("streams", [])
+            if isinstance(stream, dict)
+        }
+
+        def is_not_star(x):
+            return x != "*"
```
edgarrmondragon, line 657: Ditto the comment above about stars contained in stream and property names.

edgarrmondragon (changes requested):
@haleemur Did some UAT and found a few edge cases that are worth addressing

src/meltano/core/plugin/singer/tap.py
haleemur: I am not sure if we can reliably emit a warning for wildcard matches within streams and property names.

for instance: 

if a tap has 3 streams `activity_account`, `activity_user`, `activity_role`, and the user specifies the following in the `meltano.yml` file:

```
extractors:
  - name: tap-example
    select:
      - activity_*.*
```

IMO this should not emit a warning, as `activity_*` is matched by all 3 streams

If a tap has 3 streams as above, and the first two have the field `email`, then I'm not sure what the expected behavior ought to be for the following select extra

```
extractors:
  - name: tap-example
    select:
      - activity_*.email
```

because the selection rule will match at least once, but will not match all the streams captured by `activity_*`. Maybe we could warn that `activity_role` does not have a field named `email`

If instead of a top level field named `email`, `activity_account` had a field called `primary_contact_email`, while `activity_user` had the field `email`, a user of meltano might specify this selection as 

```
extractors:
  - name: tap-example
    select:
      - activity_*.*email
```

I am leaning towards the notion that trying to handle wild-card matches might not be possible, and if a wildcard match is encountered, the method should log an info message verification of field existence was skipped as a result of wildcard matches.

I'd love to know your thoughts before proceeding

haleemur: Yes, I think this is the simpler approach. Changing.

src/meltano/core/plugin/singer/tap.py
haleemur: I think if the condition `r.breadcrumb != ["properties", "*"]` is moved within the for-loop just a few lines below, we can test for non existent streams. changing this.

src/meltano/core/plugin/singer/tap.py
haleemur: - updated the list comprehension filter to `"*" in r.tap_stream_id`
- changed the function definition as above.

src/meltano/core/plugin/singer/tap.py
```diff
@@ -615,3 +619,53 @@ def catalog_cache_key(self, plugin_invoker):
...
+        }
+
+        def is_not_star(x):
+            return "*" not in x
+
+        def dict_get(dictionary, key):
+            return dictionary.get(key, {})
+
+        for rule in property_rules:
+            if not (s := stream_dict.get(rule.tap_stream_id)):
```
edgarrmondragon, line 657: I think this is running into another edge case. `CatalogRule.tap_stream_id` is a list of strings for the [select filter](https://docs.meltano.com/concepts/plugins/#select_filter-extra):

https://github.com/meltano/meltano/blob/6e684f233961be1f1b37550fa0f855e3bc097244/src/meltano/core/plugin/singer/catalog.py#L225.

haleemur: ah, interesting. So `tap_stream_id` can also be a list. I feel like those cases ought to not raise warnings, similar to how warnings are not raised when a wildcard is encountered.

haleemur: we could apply a filter to keep only the rules where `tap_stream_id` is an instance of `str`.

edgarrmondragon (approved):
Thanks @haleemur!

PR #8354, opened 2024-01-13
fix(cli): Emit pip installation errors

### LINKED ISSUE #8170: break: meltano install no longer shows the pip subprocess logs during a failure
### Meltano Version

3.0.0

### Python Version

3.9

### Bug scope

### Operating System

Linux - Ubuntu 20.04

### Description

To replicate
1. Setup a tap that has a bad pip url 
2. Run `meltano install` 
3. Notice that you won't see any reason why other than that the installation failed
4. Install Meltano 2.18 (Didn't `git bisect` so I don't know exactly which commit changed this) 
5. Notice the pip subprocess information is there

Note that this slack thread had an issue in the last week https://meltano.slack.com/archives/C01TCRBBJD7/p1695209268380979 that would have been a lot easier to debug without this regression

edgarrmondragon:
Fixes https://github.com/meltano/meltano/issues/8170

### DIFF
--- a/src/meltano/core/venv_service.py
+++ b/src/meltano/core/venv_service.py
@@ -159,7 +159,7 @@ class VirtualEnv:
 
 
 async def _extract_stderr(_):
-    return None
+    return None  # pragma: no cover
 
 
 async def exec_async(*args, extract_stderr=_extract_stderr, **kwargs) -> Process:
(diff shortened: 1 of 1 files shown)


WillDaSilva (approved):
:shipit:

PR #8355, opened 2024-01-14
feat(cli): Support installing multiple plugins of any type

### LINKED ISSUE #8353: feature: Install multiple plugins of any type
### Feature scope

### Description

Currently, you can only leverage `meltano install` in parallel for all plugin types or all plugins of a specific type:

```sh
# all plugins
meltano install

# all extractors
meltano install [extractor|extractors]

# all loaders
meltano install [loader|loaders]
```

It would be great if you could install multiple plugins of any type - something like:

```sh
meltano install <extractor> <loader> <transformer> <utility>
```

This change would remove the need to specify a plugin type at all, since a plugin name is already unique to a Meltano project. This is currently not possible without a breaking change however, since a plugin type is required as the first argument to `meltano install` when specifying plugin names. #8228 introduced the `--from-file` option for `meltano config <plugin> set`, which accepts a special character `-` to refer to stdin - `meltano install` could reuse this or a similar concept to skip the plugin type argument and leverage parallel install:

```sh
meltano install - <extractor> <loader> <transformer> <utility>
```

Once a convention is established, this feature could be extended to `meltano remove` as well.

I have a POC of this working for `meltano install` locally, so happy to open a PR.

ReubenFrankel:
Closes #8353


src/meltano/cli/install.py
```diff
@@ -18,13 +18,15 @@
     from meltano.core.project import Project
     from meltano.core.tracking import Tracker
 
+ANY = "-"
```
edgarrmondragon, line 21: > https://github.com/meltano/meltano/pull/8228 introduced the --from-file option for meltano config <plugin> set, which accepts a special character - to refer to stdin

The introduction of the special character `-` comes from Click, with the following argument:

> Command line tools are more fun if they work with files the Unix way, which is to accept `-` as a special file that refers to stdin/stdout.

https://click.palletsprojects.com/en/8.1.x/arguments/#file-arguments

edgarrmondragon (commented):
This LGTM. Just curious what other folks think of using `-` as a wildcard plugin type. Might be worth bring to Office Hours 🙂.

src/meltano/cli/install.py
ReubenFrankel: I wanted to reuse some kind of special character convention that already existed in the CLI - `-` was the only one I could find, but admittedly it's not great as it doesn't have anything to do with stdin/stdout here. `*` would be ideal as it indicates 'any' (i.e. wildcard), but it would have to be escaped in quotes for this kind of usage in Bash.

edgarrmondragon:
@ReubenFrankel Should we add some docs?

ReubenFrankel:
@edgarrmondragon What do you think about using `--` (double-dash) over `-` as the special "any" plugin type argument? `--` is used in `.lock` file names as a delimiter between the plugin name and variant (e.g `tap-spotify--matatika.lock`), so maybe it could function similarly here? At least then there is no overlap with the Unix file argument convention using `-`.

WillDaSilva:
@ReubenFrankel Just for posterity, I'll note that with *nix CLI applications `--` on its own commonly indicates that no more CLI options (of the form e.g. `-n`/`--name`) will be specified, and what remains should not be processed as such. If you're going to pass user input into a command, e.g. `command "$USER_PROVIDED_VALUE"`, it's a best practice to use `command -- "$USER_PROVIDED_VALUE"` to prevent the user from providing a CLI option for the command.

ReubenFrankel:
> @ReubenFrankel Just for posterity, I'll note that with *nix CLI applications `--` on its own commonly indicates that no more CLI options (of the form e.g. `-n`/`--name`) will be specified, and what remains should not be processed as such. If you're going to pass user input into a command, e.g. `command "$USER_PROVIDED_VALUE"`, it's a best practice to use `command -- "$USER_PROVIDED_VALUE"` to prevent the user from providing a CLI option for the command.

Ah, I thought I had seen that somewhere before  - that makes sense! Thanks for the explanation.

docs/docs/guide/plugin-management.md
```diff
@@ -553,10 +553,10 @@ you can [add the new variant as a separate plugin](#multiple-variants) or switch
 1. Reinstall the plugin from the new `pip_url` using [`meltano install`](/reference/command-line-interface#install):
 
    ```bash
-   meltano install <type> <name>
+   meltano install [<type>|-] <name>
 
    # For example:
-   meltano install loader target-postgres
+   meltano install target-postgres
```
ReubenFrankel, line 559: Oops - will fix this.

ReubenFrankel: https://github.com/meltano/meltano/pull/8355/commits/9af5e19bf19d2794c30a776be01af7faeb8272fa

edgarrmondragon (approved):

edgarrmondragon:
Thanks @ReubenFrankel!

PR #8364, opened 2024-01-18
Prototype of stopping after N rows loaded from taps

### LINKED ISSUE #8080: feature: Partial ("Preview") Syncs
### Feature scope

API

### Description

At [Akkio](https://akkio.com), we have a use case where we want to fetch a limited subset of user-imported datasets. We do this so that we can show users a small preview of what their end data might look like after what's essentially an internal transformation process we run, and we want them to be able to potentially make changes and make sure that transformation process looks good before they commit to the full data sync.

After a brief discussion with @DouweM, this feature doesn't yet seem to exist on Meltano, but is something that would potentially fall under Meltano's purview. Dropping some other comments that were made, for full context:

> One option would be to add a limit config on the Tap SDK to only load a certain number of records, and another to add a mode to meltano elt to stop after the first state message is output by the target. (Targets flush records and write state messages at different batch sizes (often configurable), but until the first state message comes we can’t assume any of the records have made it to the destination, so we couldn’t base this on number-of-records on the target or meltano elt side, that’d need to be controlled by the target.)

dcowden:
This is a sketch for how i introduced ability to limit rows from taps for the purposes of testing.

Warning: This code has not been executed. I have working code based on an older revision that i tested. So, i know the approach works with code very similar to this. I cleaned up the code to fit into latest meltano main, so that it would be easier to see the essence of the changes.

If this overall path is of interest, i could do more work-- i really would like thoughts about the overall approach before I do that. I dont know the meltano code well at all, so this might be a totally wrong way to do it.

Other notes that might be helpful:

-  i first tried a similar approach in the singer runner code, until i realized that was not the code used for meltano run
-  This uses an environment var to set the row limit. A much better solution of course would be to allow a property on the project and/or taps, so the row limit can be different for each tap. But (a) I dont know enough to approach that, and (b) for us this is sufficient
-  I am worried about whether this code would work in environments with concurrency more than just one. I did not do any testing along those lines.

### DIFF
--- a/src/meltano/core/block/extract_load.py
+++ b/src/meltano/core/block/extract_load.py
@@ -2,9 +2,9 @@
 
 from __future__ import annotations
 
-import os
 import asyncio
 import logging
+import os
 import typing as t
 from contextlib import asynccontextmanager, closing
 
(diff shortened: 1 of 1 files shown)


wesseljt:
@edgarrmondragon - I know it has been a while - but this is still of interest for us. What would be the next step? Resolving the failing checks?

wesseljt:
@dcowden ^ fyi

edgarrmondragon:
> @edgarrmondragon - I know it has been a while - but this is still of interest for us. What would be the next step? Resolving the failing checks?

Yeah, although the current implementation aborts the sync after _any_ N lines have been emitted by the tap. Other than actual records, that count would include schema and state messages too. I'm fine with that, but just wanna make sure that is a good solution for other users' and yall's needs.

That said, some requirements to get this across are probable:

1. Make this controllable via a CLI option instead of the current env var implementation. I don't think we'd need a new setting for this.
2. Add a new test that validates this behavior. Without looking at what tests we have that may be a good starting point, this can range from easy to very hard so let me know if you need help with that.

The current pre-commit and coverage failures are ok at the moment, since I expect the refactoring and testing required by the above two points would address them.

PR #8365, opened 2024-01-18
fix: Retain stderr in `meltano config <plugin> test`

### LINKED ISSUE #8307: bug: Tap's stderr is lost in `meltano config test`
### Meltano Version

3.2.0

### Python Version

3.6 (deprecated)

### Bug scope

### Operating System

NA

### Description

Following https://github.com/meltano/meltano/pull/8225, we started sending the tap's stderr to /dev/null. Some error information in that stream may be useful, so it's probably worth revisiting and maybe redirector to stderr instead.

### Code

_No response_

ReubenFrankel:
Closes #8307


edgarrmondragon (approved):
Thanks again @ReubenFrankel!

PR #8398, opened 2024-02-05
fix: Handle non-UTF-8 lines in plugin output for logging

### LINKED ISSUE #8377: bug: `UnicodeDecodeError` when processing plugin's output for logging
### Meltano Version

3.3.0

### Python Version

NA

### Bug scope

### Operating System

Windows, probably

### Description

This line https://github.com/meltano/meltano/blob/42f0fc6895da3265d0fede67fae09cd37b7843f1/src/meltano/core/logging/utils.py#L208

could probably be made safer by passing `errors="replace"`[^1] to the `.decode()` method.

[^1]: https://docs.python.org/3.12/library/stdtypes.html#bytes.decode

### Code

_No response_

nkclemson:
Try to decode the byte text using the list of encoding. For now, it is `latin-1`. If failed, then fall back into `utf-8` with `error=ignore`.

Closes #8377

### DIFF
--- a/src/meltano/core/logging/utils.py
+++ b/src/meltano/core/logging/utils.py
@@ -197,6 +197,16 @@ class SubprocessOutputWriter(t.Protocol):
 
 async def _write_line_writer(writer: SubprocessOutputWriter, line: bytes):
     # StreamWriters like a subprocess's stdin need special consideration
+    def _decode_log(byte_string):  #8377
+        encodings = ['latin-1']  # list of encoding we need to consider; default is utf-8
+        for enc in encodings:
+            try:
+                return byte_string.decode(enc)
+            except UnicodeDecodeError as e:
+                pass
+        else:
+            return byte_string.decode('utf-8', errors='ignore')
+
     if isinstance(writer, asyncio.StreamWriter):
         try:
             writer.write(line)
@@ -205,7 +215,7 @@ async def _write_line_writer(writer: SubprocessOutputWriter, line: bytes):
             await writer.wait_closed()
             return False
     else:
-        writer.writeline(line.decode())
+        writer.writeline(_decode_log(line))
 
     return True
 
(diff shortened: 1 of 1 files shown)


src/meltano/core/logging/utils.py
```diff
@@ -197,6 +197,18 @@ def writeline(self, line: str):
...
+        encodings = [
+            "latin-1"
+        ]  # list of encoding we need to consider; default is utf-8
+        for enc in encodings:
+            try:
+                return byte_string.decode(enc)
+            except UnicodeDecodeError:
+                pass
+        else:
+            return byte_string.decode("utf-8", errors="ignore")
```
edgarrmondragon, line 210: Try-except blocks are expensive, so I'd prefer to simple defer to `decode(errors="replace")` if possible. What do you think?

src/meltano/core/logging/utils.py
```diff
@@ -205,7 +217,7 @@ async def _write_line_writer(writer: SubprocessOutputWriter, line: bytes):
             await writer.wait_closed()
             return False
     else:
-        writer.writeline(line.decode())
+        writer.writeline(_decode_log(line))
```
edgarrmondragon, line 220:
```suggestion
        writer.writeline(line.decode(errors="replace"))
```

edgarrmondragon (changes requested):

edgarrmondragon:
@nkclemson Can you re-create the PR from a branch other than `main`, otherwise I can't push from my end. Thanks!

src/meltano/core/logging/utils.py
nkclemson: As this is for logging. Losing some non-decodable character mightn't make much difference. So, yeah, having simple `decode(error="replace")` makes a block more straightforward, faster and readable. I will make a change and push it.

PR #8400, opened 2024-02-06
fix: Ensure the right `snowplow_tracker` module is imported when dbt is installed in the same environment

edgarrmondragon:
Related:

- https://github.com/meltano/meltano/issues/8256
- https://github.com/meltano/meltano/issues/8302


edgarrmondragon:
`import_module` doesn't do what I thought

PR #8415, opened 2024-02-20
chore: Replace use of flakeheaven with Ruff for pep8-naming

### LINKED ISSUE #8402: Replace use of flakeheaven with Ruff for pep8-naming
* Disable the flakeheaven plugin in `pyproject.toml`
* Migrate to Ruff's `Nxxx` rules
* Remove additional dependency from `pre-commit-config.yaml`

nkclemson:
Closes #8402

 - [ ] squash git commits

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -579,7 +579,6 @@ exclude = [
 ignore = [
   "ANN101", # Missing type annotation for `self` in method
   "ANN102", # Missing type annotation for `cls` in class method
-  "N815",   # mixedCase variable in class scope (constants, method names)
   "N818",   # Allow Exceptions to have suffix 'Exception' rather than 'Error'
   "PT004",  # Add leading underscore to fixtures that do not return anything
   "UP026",  # Replace `mock` import with `unittest.mock` - remove once Python 3.7 support is dropped
@@ -620,9 +619,6 @@ unfixable = [
 ]
 "tests/**" = [
   "S101", # Allow 'assert' in tests
-]
-"src/meltano/migrations/versions/**" = [
-  "N806", # Allow uppercase variable names
 ]
 
 [tool.ruff.lint.flake8-annotations]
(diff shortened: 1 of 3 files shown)


edgarrmondragon:
@nkclemson Ping me when this is ready for a review. Also, don't feel the need to squash commits  before marking this as ready since we do merge squashes anyway :)

nkclemson:
Hi @edgarrmondragon, I think this is complete now. I have added the `"N815" - mixedCase variable in class scope (constants, method names)` exemption, as this is in accordance with [Guido's recommendation](https://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations). Initially, I thought that the lint error I encountered in test_logging_utils.py was N815, but I later realized that those all-caps variable names are inside methods, so I fixed it. let me know if I need to update anything.

pyproject.toml
```diff
@@ -586,6 +579,8 @@ exclude = [
 ignore = [
   "ANN101", # Missing type annotation for `self` in method
   "ANN102", # Missing type annotation for `cls` in class method
+  "N815",   # mixedCase variable in class scope (constants, method names)
```
edgarrmondragon, line 582: I don't think this is currently triggered anywhere in the codebase

```suggestion
```

pyproject.toml
```diff
@@ -625,6 +621,9 @@ unfixable = [
 "tests/**" = [
   "S101", # Allow 'assert' in tests
 ]
+"src/meltano/migrations/versions/**" = [
+  "N806", # Allow uppercase variable names
+]
```
edgarrmondragon, line 626: I think this is covered by the `exclude` setting?

```suggestion
```

pyproject.toml
nkclemson: yep. This isn't triggered. I am removing this.

pyproject.toml
nkclemson: Yep.

edgarrmondragon (approved):
Thanks a ton @nkclemson!

PR #8445, opened 2024-03-13
chore: Replace use of flakeheaven with Ruff for flake8-errmsg

### LINKED ISSUE #8443: Replace use of flakeheaven with Ruff for flake8-errmsg
Meta Issue: #7383

Ruff rule classes to enable:
"EM",  # flake8-errmsg

PS: I (@nkclemson ) will work on this meta issue one by one.

nkclemson:
Related:

- Closes #8443

```
ruff check --select=EM --add-noqa src tests
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -579,6 +579,8 @@ exclude = [
 ignore = [
   "ANN101", # Missing type annotation for `self` in method
   "ANN102", # Missing type annotation for `cls` in class method
+  "EM101",  # Exception must not use a string literal, assign to a variable first
+  "EM102",  # Exception must not use an f-string literal, assign to a variable first
   "N818",   # Allow Exceptions to have suffix 'Exception' rather than 'Error'
   "PT004",  # Add leading underscore to fixtures that do not return anything
   "UP026",  # Replace `mock` import with `unittest.mock` - remove once Python 3.7 support is dropped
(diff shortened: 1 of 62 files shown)


pyproject.toml
```diff
@@ -579,6 +579,8 @@ exclude = [
 ignore = [
   "ANN101", # Missing type annotation for `self` in method
   "ANN102", # Missing type annotation for `cls` in class method
+  "EM101",  # Exception must not use a string literal, assign to a variable first
+  "EM102",  # Exception must not use an f-string literal, assign to a variable first
```
edgarrmondragon, line 583: This would mean we're ignoring 2/3 EM rules 😅.

If too many failures are triggered by enabling EM, you could always do something like

```
ruff check --select=EM --add-noqa src tests
```

and that would allow us to adopt this rule incrementally.

edgarrmondragon (commented):
Thanks @nkclemson! Ping me when this is ready 🙂

pyproject.toml
nkclemson: Thanks for bringing it up. Adding a `noqa` seems like a good starting point for now. I have removed `EM101` and `EM102` from ignore list.

edgarrmondragon (approved):
Thanks @nkclemson!

PR #8447, opened 2024-03-15
fix: `meltano run` no longer creates empty `venv` plugin directories for inherited plugins

### LINKED ISSUE #8401: Directory creation with MELTANO_PROJECT_READONLY
### Meltano Version

3.3.1 && 2.20.0

### Python Version

3.9

### Bug scope

### Operating System

Official docker image / Ubuntu 22.04 / Kubernetes

### Description

When using the env. variable 
`MELTANO_PROJECT_READONLY=true` 
meltano still writes to project structure during `meltano run`
In the case below i have we use both `meltano install --clean` and `meltano invoke dbt-bigquery:deps` during docker build, then during runtime meltano tries to write to the extractors folder.

As a sidenote, it was fixable with creating `.meltano/extractors/venv`-folder manually, but it does not seem like the right thing to do since there is a "project_readonly"-variable.

### Code

```python
Output 1:

[Errno 30] Read-only file system: '/project/.meltano/extractors/<extractor name>'

Output 2: 
root@7e978fadb393:/project# inotifywait .meltano/extractors/
Setting up watches.
Watches established.
.meltano/extractors/ CREATE,ISDIR <extractor name>
```

edgarrmondragon:
Related:

- Closes #8401


tests/meltano/core/test_utils.py
```diff
@@ -222,3 +223,22 @@ def test_remove_suffix():
...
+
+    @makedirs
+    def species(genus_name, species_name):
+        return hierarchy(genus_name, species_name)
+
+    cat = species("felis", "catus")
+    assert cat.exists()
+
+    wolf = species("canis", "lupus", make_dirs=False)
+    assert not wolf.exists()
```
edgarrmondragon, line 244: Showcase of buggy behavior of `makedirs` before this fix:

```
        wolf = species("canis", "lupus", make_dirs=False)
>       assert not wolf.exists()
E       AssertionError: assert not True
E        +  where True = <bound method Path.exists of PosixPath('/tmp/pytest-of-runner/pytest-0/popen-gw0/test_makedirs_decorator0/canis/lupus')>()
E        +    where <bound method Path.exists of PosixPath('/tmp/pytest-of-runner/pytest-0/popen-gw0/test_makedirs_decorator0/canis/lupus')> = PosixPath('/tmp/pytest-of-runner/pytest-0/popen-gw0/test_makedirs_decorator0/canis/lupus').exists
```

that should've passed.

ReubenFrankel:
We had a whole discussion about this back in the GitLab days: https://gitlab.com/meltano/meltano/-/merge_requests/2099#note_566826570

The change was even made to [add `**kwargs`](https://gitlab.com/meltano/meltano/-/merge_requests/2099/diffs?commit_id=9a90c502a426d2d1eff231994abe4e0ae61f141c), then [narrowed down to a `make_dirs` kwarg](https://gitlab.com/meltano/meltano/-/merge_requests/2099/diffs?commit_id=9a90c502a426d2d1eff231994abe4e0ae61f141c), then [this merge commit rolled back the change](https://gitlab.com/meltano/meltano/-/merge_requests/2099/diffs?commit_id=9a90c502a426d2d1eff231994abe4e0ae61f141c) and it went unnoticed. :cry: (I'm experimenting with a feature and am running into a problem with `makedirs` as well)

edgarrmondragon:
Thanks for commenting @ReubenFrankel!

> We had a [whole discussion about this](https://gitlab.com/meltano/meltano/-/merge_requests/2099#note_566826570) back in the GitLab days - a change was even made to [add a `make_dirs` kwarg](https://gitlab.com/meltano/meltano/-/commit/c1aca6cce3e9f7e4d67dda555ecd1c23097e8a08#8e8067d772989bfee158f1ccfa6e2a8fe41056ec) to all `@makedirs` methods ([requested by AJ](https://gitlab.com/meltano/meltano/-/merge_requests/2099/diffs?commit_id=2d5af5e8ebd9faf655d056e1b0d2c09a09d4d526#note_569189837) after `**kwargs` was implemented initially)

Wow, I wasn't aware of that! And it makes it easier to decide a explicit kwarg over generic `kwargs` 😅.

> and then [78b009b#diff-c087424df56cc7c7f9cb7450ad5f6d97ad8855940e5b27b548c879948616140c](https://github.com/meltano/meltano/commit/78b009b4c195678d2178ff4e0b69c711fae4ffb9#diff-c087424df56cc7c7f9cb7450ad5f6d97ad8855940e5b27b548c879948616140c) was merged which reverted the changes without anyone noticing (probably no tests). 😢 I'm not sure if this is the case for anything else, but may be worth checking.

Yeah, that was a large merge and we missed it somehow 😞. Anyway this PR includes a test that should prevent some regressions.

src/meltano/core/project.py
```diff
@@ -416,122 +423,157 @@ def dotenv_update(self):
         self.refresh()
 
     @makedirs
-    def meltano_dir(self, *joinpaths):
+    def meltano_dir(self, *joinpaths: StrPath, make_dirs: bool = True) -> Path:  # noqa: ARG002
```
WillDaSilva, line 426: The `make_dirs` parameter here and in some other functions appears unused. If it isn't supported but is included for the signature, or if it must be `True`, then a `NotImplementedError` or `ValueErrror` exception should be raised otherwise.

edgarrmondragon: It's actually required so we can pass down the flag from methods like `Project.logs_dir`. For example, without it the following code fails

```python
from meltano.core.project import Project

project = Project.find(".")
project.logs_dir(make_dirs=False)
```

with

```python
TypeError: Project.meltano_dir() got an unexpected keyword argument 'make_dirs'
```

That's because all functions decorated with `@makedirs` are expected to accept the `make_dirs` keyword argument after the changes introduced to the decorator in this PR.

Wdyt?

WillDaSilva: After reading through the code more closely, and understanding it better, my impression is that the `makedirs` decorator is a bad idea. Directory creation shouldn't be inherently tied to retrieving information about what those directories would be. That said, changing that goes beyond the scope of this PR, which provides a valuable fix.

WillDaSilva (approved):

PR #8461, opened 2024-03-23
chore: Replace use of flakeheaven with Ruff for flake8-return

### LINKED ISSUE #8457: Replace use of flakeheaven with Ruff for flake8-return
Meta Issue: https://github.com/meltano/meltano/issues/7383

Ruff rule classes to enable:
 - "RET" # flake8-return

PS: I (@nkclemson ) will work on this meta issue one by one.

nkclemson:
Related:

- Closes #8457 

```
ruff check --select=RET --add-noqa src tests
```

Question: Which RET error code should we consider, and which one should we put on the ignore list? For eg: RET505 to RET508 are the return type we consider more pythonic.

https://pypi.org/project/flake8-return/

flake8_return

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -99,9 +99,7 @@ repos:
           - flake8-docstrings==1.6.0
           - flake8-eradicate==1.4.0
           - flake8-polyfill==1.0.2
-          - flake8-print==5.0.0
           - flake8-quotes==3.3.1
-          - flake8_return==1.2.0
           - flake8-rst-docstrings==0.2.7
           - flake8-string-format==0.3.0
           - flake8-typing-as-t==0.0.3
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -101,6 +101,7 @@ repos:
           - flake8-polyfill==1.0.2
           - flake8-print==5.0.0
           - flake8-quotes==3.3.1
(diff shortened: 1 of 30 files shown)


.pre-commit-config.yaml
```diff
@@ -101,6 +101,7 @@ repos:
           - flake8-polyfill==1.0.2
           - flake8-print==5.0.0
           - flake8-quotes==3.3.1
+          - flake8_return==1.2.0
```
edgarrmondragon, line 104: I don't think we want to add more flakeheaven plugins

```suggestion
```

.pre-commit-config.yaml
```diff
@@ -101,6 +101,7 @@ repos:
           - flake8-polyfill==1.0.2
           - flake8-print==5.0.0
```
edgarrmondragon, line 102: We can also get rid of this

```suggestion
```

edgarrmondragon (changes requested):

edgarrmondragon:
> Question: Which RET error code should we consider, and which one should we put on the ignore list? For eg: RET505 to RET508 are the return type we consider more pythonic.

Let me think about it. Regardless, we should probably look at the astral/Ruff docs instead.

edgarrmondragon (approved):
Most RET50x were easily fixable, so I went ahead and pushed 82c44a7. Thanks @nkclemson!

PR #8465, opened 2024-03-25
feat: Support `uv` as an optional virtualenv backend

### LINKED ISSUE #8410: feature: Use `uv` for managing plugin virtual envs?
### Feature scope

### Description

`uv`[^1] is a tool created by the Astral team (of Ruff fame). It's a drop-in replacement for _most_ of the functionality of both `pip` and `virtualenv`, but  faster (10x-100x).

It's probably not a good idea to think of replacing `pip` and `virtualenv` with it this early, but it could initially be added behind a feature flag.

### Blockers

The following bugs and missing features are probable blockers for this integration:

- https://github.com/astral-sh/uv/issues/313
- https://github.com/astral-sh/uv/issues/1396

[^1]: https://github.com/astral-sh/uv

edgarrmondragon:
```
#10 [6/6] RUN MELTANO_VENV_BACKEND="virtualenv" meltano install
#10 DONE 50.5s
```

```
#10 [6/6] RUN MELTANO_VENV_BACKEND="uv" meltano install
#10 DONE 20.2s
```

Closes https://github.com/meltano/meltano/issues/8410


src/meltano/core/venv_service.py
```diff
@@ -332,23 +370,19 @@ async def extract_stderr(proc: Process):
...
+            return await self.create_venv(extract_stderr=extract_stderr)
         except AsyncSubprocessError as err:
             raise AsyncSubprocessError(
                 f"Could not create the virtualenv for '{self.namespace}/{self.name}'",  # noqa: EM102
                 process=err.process,
                 stderr=await err.stderr,
             ) from err
 
-    async def upgrade_pip(self, env: dict[str, str | None] | None = None) -> Process:
+    async def upgrade_pip(
```
edgarrmondragon, line 381: Perhaps this should be renamed to `upgrade_installer`

src/meltano/core/venv_service.py
```diff
@@ -454,24 +559,165 @@ async def extract_stderr(proc: Process) -> str | None:  # pragma: no cover
...
+            A string representing the path to the `uv` executable.
+        """
+        with contextlib.suppress(ImportError, FileNotFoundError):
+            from uv import find_uv_bin
+
+            return find_uv_bin()
+
+        # Fall back to PATH.
+        uv = shutil.which("uv")  # pragma: no cover
+        return uv or "uv"  # pragma: no cover
```
edgarrmondragon, line 608: Maybe we should fail if `uv` is not in the `PATH` after trying to find it in the Python venv?

src/meltano/core/venv_service.py
edgarrmondragon: Done in 391f8d6

src/meltano/core/venv_service.py
WillDaSilva: I think it makes sense to fail if `uv` is not in `$PATH`. I'm not sure how we could expect it to run if we can't find it, since when we call `subprocess` later it'll search for the executable on `$PATH` too unless we provide a path to an executable.

WillDaSilva (approved):

PR #8470, opened 2024-03-28
refactor: Started enforcing usage of `structlog.stdlib.get_logger` over `logging.getLogger` and banned use of the root logger

edgarrmondragon:
Adds some consistency to our logs **when not using one of the built-in `meltano.core.logging` formatters**:

* structlog uses an `event` key instead of `message`
* structlog uses lower case for the log level
* structlog can add context vars, so we could enrich more logs if we decide to e.g. add an `environment=` context var

TODO:

- [x] Address all `logging.debug` and similar calls?

> [!NOTE]  
> This doesn't address 3rd party library logs


src/meltano/core/utils/pidfile.py
edgarrmondragon, line 1: This module doesn't seem to be used for anything 🤷‍♂️

edgarrmondragon: Leftover from the UI it seems: https://github.com/meltano/meltano/commit/617617433cd744210ed4aeb39042a3903e2e772e

pnadolny13 (approved):
🚀

WillDaSilva (approved):

Issue #2651, opened 2021-04-12
Fail when plugin `pip_url` changed and installation is outdated (pip_url drift)


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/2701

Originally created by @ghorwitchsmith on 2021-04-12 15:55:29

---

### Problem to solve

I installed Meltano with tap-mysql and target-postgres. After creating some pipelines I realised that for a data set I needed to track deletions and this would require the transferwise variant. I switched the variant in the meltano.yml, but then spent a whilst trying to work out why it was not creating _sdc_deleted_at columns in postgres.

This is partially as I'm new, however the solution was to reinstall target-postgres with the transferwise variant. Everything then just worked, as you'd expect!

### Proposal

It would really help (newer) users if the configuration validated the variant installed when it tried to run them.

### What does success look like, and how can we measure that?

If the installed variant of a tap or a target does not match the variant in the meltano.yml then an error or warning should be shown.

### Proposed spec (added 2022-03-03)

1. During install we drop a marker file into the root of the venv with text of the `pip_url` used during installation. (Or a hash, if we think the pip_url is sensitive.)
   - E.g. `.meltano/.../dbt/meltano_pip_url_marker.txt`
2. During execution we check the marker file against the current `pip_url` value.
3. If the comparison fails, we either warn or throw a hard failure before executing the plugin.

MeltyBot:
[View 17 previous comments from the original issue on GitLab](https://gitlab.com/meltano/meltano/-/issues/2701)

WillDaSilva:
Still relevant. Relates to:
- https://github.com/meltano/meltano/issues/2089

edgarrmondragon:
I would consider this addressed for most purposes by #8482

PR #8482, opened 2024-04-06
feat: Plugins are now auto-installed when commands require them

### LINKED ISSUE #2089: Automatically install missing/stale plugins when running `elt`, `run`, or `invoke`
## Just-in-time installation

Rather than failing with "plugin is not installed" or "No such file or directory: '/Users/nehiljain/snaptravel/universal-elt-pipeline/airflow-elt/.meltano/extractors/tap-github/venv/bin/tap-github'", we can accept a `--install` flag in the CLI and/or some env var equivalent of the same. 

This would be applicable to:

- `meltano run`
- `meltano elt`
- `meltano invoke`

And possibly later (but lower priority):

- `meltano schedule run`
- `meltano job run`

## Interactive (TTY) experience

If a tty is available, we can simply ask the user if they'd like to auto-install the missing plugins.

## Updated error message with hints

If a TTY is not available and `--install` is not specified, then the error message would be changed to something like:

> ERROR: Plugin not installed. Please install with `meltano install ...` or rerun with `meltano invoke --install ...`.

## Why require confirmation / opt-in

Install operations are not always safe to run, and they can hide other issues in the invocation or installation, such as using the wrong plugin name or the wrong base image in a scheduled job. They could also trigger unnecessarily for a container-backed spec that doesn't require installation.

<details>
<summary>Original description</summary>

Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/2133

Originally created by @DouweM on 2020-07-06 15:50:24

---

`meltano elt tap-foo target-bar` will already add `tap-foo` and `target-bar` to the project (and `meltano.yml`) if they are not already present. 

However, if they've already been added to the project, but are not actually installed into the `.meltano` directory (such as after cloning a repo from version control without running `meltano install`), `meltano elt` fails with an ugly error message, as reported by @nehiljain:

```
ELT could not complete, an error happened during the process: Cannot start tap or target: [Errno 2] No such file or directory: '/Users/nehiljain/snaptravel/universal-elt-pipeline/airflow-elt/.meltano/extractors/tap-github/venv/bin/tap-github': '/Users/nehiljain/snaptravel/universal-elt-pipeline/airflow-elt/.meltano/extractors/tap-github/venv/bin/tap-github'
```

We can do much better, and automatically `meltano install` these missing taps and targets when running `meltano elt` as well.

</details>

ReubenFrankel:
Implements just-in-time install for the following commands:

- `invoke`
- `run`
- `el`/`elt`
- `test`
- `config <plugin> test`
- `select --list`

Related:

* Closes https://github.com/meltano/meltano/issues/2089
* https://github.com/meltano/meltano/issues/2651
* https://github.com/meltano/meltano/issues/7001


ReubenFrankel:
We probably want to suppress install progress to stdout  from `meltano.cli.utils.install_plugins` here, or emit as logs to stderr... For `invoke`, this currently breaks the common tap/target debugging technique of

```
meltano invoke <tap> > tap.out
cat tap.out | meltano invoke <target>
```

edgarrmondragon:
> We probably want to suppress install progress to stdout from `meltano.cli.utils.install_plugins` here, or emit as logs to stderr... For `invoke`, this currently breaks the common tap/target debugging technique of
> 
> ```
> meltano invoke <tap> > tap.out
> cat tap.out | meltano invoke <target>
> ```

Yeah, I'd be more than OK with starting to unify our output into away from `click.echo` and into the logging framework.

ReubenFrankel:
Think I have covered all relevant commands here. Wondering if we want a new `--no-install` option for these (same convention as for `meltano add`) to opt-out of auto-install behaviour (e.g. for CI), or stick to the original proposal in #2089:

> ## Interactive (TTY) experience
> If a tty is available, we can simply ask the user if they'd like to auto-install the missing plugins.
> 
> ## Updated error message with hints
> If a TTY is not available and `--install` is not specified, then the error message would be changed to something like:
> 
> > ERROR: Plugin not installed. Please install with `meltano install ...` or rerun with `meltano invoke --install ...`.
> 
> ## Why require confirmation / opt-in
> Install operations are not always safe to run, and they can hide other issues in the invocation or installation, such as using the wrong plugin name or the wrong base image in a scheduled job. They could also trigger unnecessarily for a container-backed spec that doesn't require installation.

I haven't yet addressed install progress/status as logs, but the progress output is disabled for `meltano invoke` for now.

ReubenFrankel:
Should this be added for `select ` too?

edgarrmondragon:
> Should this be added for `select` too?

If it's opt-in, probably yes 👍

src/meltano/core/plugin_install_service.py
```diff
@@ -36,7 +36,14 @@ class PluginInstallReason(str, Enum):
...
     ADD = "add"
+    CONFIG_TEST = "config test"
+    EL = "el"
+    ELT = "elt"
     INSTALL = "install"
+    INVOKE = "invoke"
+    RUN = "run"
+    SELECT = "select"
+    TEST = "test"
     UPGRADE = "upgrade"
```
ReubenFrankel, line 47: Should we collect all these newly-added install reasons under an aggregate `PREREQUISITE` reason, or something like that?

edgarrmondragon: Yeah, `PREREQUISITE` or something like `JIT` or similar seems like a good option

src/meltano/core/plugin_install_service.py
```diff
@@ -323,7 +317,13 @@ async def install_plugin_async(
             ),
         )
 
-        if not plugin.is_installable() or self._is_mapping(plugin):
+        is_installed = self.project.plugin_dir(plugin, "venv", make_dirs=False).exists()
```
ReubenFrankel, line 320: This is a very basic check to determine whether or not to install based on the presence of the plugin virtual environment directory. Is there anything more intelligent we can do here to satisfy the stale plugin requirement from #2089? https://github.com/meltano/meltano/issues/6416#issuecomment-1922486621 and #2651 seem relevant.

edgarrmondragon: The `VenvService` uses a fingerprint computed from the (expanded) `pip_url`:

https://github.com/meltano/meltano/blob/effdf0bf6f346002951022ef60e6e98eb8b29869/src/meltano/core/venv_service.py#L261-L272

Maybe that's helpful?

src/meltano/core/plugin_install_service.py
ReubenFrankel: [7f5c488](https://github.com/meltano/meltano/pull/8482/commits/7f5c48871975b4dfbdbb39526b28fa4a9a5bd292)

src/meltano/core/plugin_install_service.py
ReubenFrankel: Thanks - addressed this in https://github.com/meltano/meltano/pull/8482/commits/21cb02df2fda6a826365f9cf7346cc029b0c419a

src/meltano/cli/select.py
```diff
@@ -112,11 +115,18 @@ def update(
     select_service.update(entities_filter, attributes_filter, exclude, remove)
 
 
-async def show(project, extractor, show_all=False):
+async def show(project, extractor, install_plugins: InstallPlugins, show_all=False):
     """Show selected."""
     _, Session = project_engine(project)  # noqa: N806
     select_service = SelectService(project, extractor)
 
+    await install_plugins(
```
edgarrmondragon, line 123:
```suggestion
    await install_plugins(  # pragma: no cover
```

src/meltano/cli/select.py
```diff
@@ -85,7 +88,7 @@ async def select(
     """
     try:
         if flags["list"]:
-            await show(project, extractor, show_all=flags["all"])
+            await show(project, extractor, install_plugins, show_all=flags["all"])
```
edgarrmondragon, line 91: I'm happy to ignore missing coverage for this for now. We have an incoming test that'll cover this in https://github.com/meltano/meltano/pull/8518.

ReubenFrankel: Great! I was about to ask if you wanted a test 😅

src/meltano/cli/params.py
```diff
@@ -29,6 +37,16 @@ def decorate(*args, database_uri=None, **kwargs):
     return functools.update_wrapper(decorate, func)
 
 
+install_option = click.option(
+    "--install/--no-install",
+    "install_plugins",
+    default=True,
```
edgarrmondragon, line 43: Should this default to `False` to make it opt-in?

```suggestion
    default=False,
```

ReubenFrankel: Depends on what you think is best. I know you mentioned opt-in earlier and I did mean to point out before that it is currently opt-out, but I implemented it this way initially because it would mean users immediately get the benefit of auto-install. Given that a plugin is already installed, there shouldn't be any new behaviour here running any of the affected commands with respect to that plugin. Maybe I'm missing something (maybe CI or Docker use-cases)?

Also, since `meltano add` shares this `--install/--no-install` implementation now, the default has to be `true` to keep the current behaviour (although this should be fairly straightforward to make an exception for if we do change the default).

edgarrmondragon: Gotcha.

I'm looking through the code now and it's looking great! Each invocation of `install_plugins()` just gets called with the relevant plugins, so that eases my concern for projects with a ton of plugins 🙌.

I may just have a thought about the function signatures.

src/meltano/cli/utils.py
```diff
@@ -428,13 +428,14 @@ def install_status_update(install_state: PluginInstallState):
         logger.info("%s %s '%s'", install_state.verb, desc, plugin.name)
 
 
-def install_plugins(
+async def install_plugins(
     project,
     plugins,
     reason=PluginInstallReason.INSTALL,
```
edgarrmondragon, line 434:
```suggestion
    *,
    reason=PluginInstallReason.INSTALL,
```

src/meltano/cli/select.py
```diff
@@ -125,13 +129,21 @@ def update(
...
     refresh: bool = False,
 ) -> None:
     """Show selected."""
     _, Session = project_engine(project)  # noqa: N806
     select_service = SelectService(project, extractor)
 
+    await install_plugins(  # pragma: no cover
+        project,
+        [select_service.extractor],
+        PluginInstallReason.JIT,
```
edgarrmondragon, line 143:
```suggestion
        reason=PluginInstallReason.JIT,
```

src/meltano/cli/utils.py
```diff
@@ -428,13 +428,14 @@ def install_status_update(install_state: PluginInstallState):
         logger.info("%s %s '%s'", install_state.verb, desc, plugin.name)
 
 
-def install_plugins(
+async def install_plugins(
     project,
     plugins,
```
edgarrmondragon, line 433: This smells like it should be `*plugins` but that's probably out of scope for this PR.

edgarrmondragon (changes requested):
Also, this is at a good point to add some docs 😄

src/meltano/cli/utils.py
ReubenFrankel: Sure, I thought it was best to leave the signature mostly the same. Curious on the philosophy of that though?

src/meltano/core/plugin_install_service.py
```diff
@@ -323,7 +311,24 @@ async def install_plugin_async(
...
+            project=self.project,
+            python=plugin.python,
+            namespace=plugin.type,
+            name=plugin.venv_name,
+        )
+
+        requires_install = (
+            fingerprint(get_pip_install_args(self.project, plugin, env))
+            != venv_service.read_fingerprint()
+        )
```
ReubenFrankel, line 325: @edgarrmondragon Do you mind checking this logic? Just want to be sure it's sufficient in all cases.

ReubenFrankel:
> Also, this is at a good point to add some docs 😄

Will do 👍

src/meltano/cli/utils.py
edgarrmondragon: I could be wrong but I think it's the more Pythonic approach for dealing with N positional arguments of the same type.

src/meltano/core/plugin_install_service.py
edgarrmondragon: Ah, good call. The logic looks ok but I wonder if we should move the fingerprint methods to `VirtualEnv` instead of them being in `VenvService`. The reasoning being that with the introduction of the `uv` backend, `VenvService` seems more like a specific implementation that we should avoid instantiating in this context.

ReubenFrankel: OK, I'll make a point to have a look at that.

ReubenFrankel: https://github.com/meltano/meltano/commit/d2f218b2018d963ba5d43b021401e0a437e28115 Is this what you had in mind?

edgarrmondragon: Yup!

edgarrmondragon (approved):

edgarrmondragon:
@ReubenFrankel I'm imagining users might want to disable JIT installation globally. Maybe an env var for the CLI option. Wdyt?

ReubenFrankel:
> @ReubenFrankel I'm imagining users might want to disable JIT installation globally. Maybe an env var for the CLI option. Wdyt?

Sure, I can see the benefit of that, and it would also align with how `--environment` works (i.e. global option but has no effect for some commands).

What do you want to do for `meltano add --no-install`? Keep accepting the option there and deprecate?

edgarrmondragon:
> What do you want to do for `meltano add --no-install`? Keep accepting the option there and deprecate?

@ReubenFrankel do you mean deprecating the current auto-install behavior of `meltano add`?

I think click's behavior for boolean options (i.e. using either `is_flag=True` or `type=click.BOOL`[^1][^2]) is to parse the env var into an appropriate boolean[^3]:

```python
        if norm in {"1", "true", "t", "yes", "y", "on"}:
            return True

        if norm in {"0", "false", "f", "no", "n", "off"}:
            return False
```

So having, for example, an `MELTANO_AUTO_INSTALL` env var, the user could pass one of these to disable the behavior:

```
MELTANO_AUTO_INSTALL=no
MELTANO_AUTO_INSTALL=f
MELTANO_AUTO_INSTALL=0
MELTANO_AUTO_INSTALL=N
```

The command line option would override the env var.

Wdyt? Does this answer the question regarding `meltano add`?

[^1]: https://github.com/pallets/click/blob/14f735cf59618941cf2930e633eb77651b1dc7cb/src/click/core.py#L2502-L2504
[^2]: https://github.com/pallets/click/blob/14f735cf59618941cf2930e633eb77651b1dc7cb/src/click/types.py#L1042-L1043
[^3]: https://github.com/pallets/click/blob/14f735cf59618941cf2930e633eb77651b1dc7cb/src/click/types.py#L610-L614

ReubenFrankel:
@edgarrmondragon Think I got the wrong end of the stick here and thought you were proposing `--[no]-install` as a global CLI option (i.e. `meltano --no-install ...`).

> So having, for example, an MELTANO_AUTO_INSTALL env var

Yeah, that seems reasonable. Is there a case where this is already done for a non-global CLI option that affects multiple commands? Just wanna know if we should be setting a precedent here.

Either way, will need documenting here: https://docs.meltano.com/reference/settings

edgarrmondragon:
> Yeah, that seems reasonable. Is there a case where this is already done for a non-global CLI option that affects multiple commands? Just wanna know if we should be setting a precedent here.

I don't think there is an existing one. There's non-global CLI options with env vars scoped just to the command and without a corresponding project setting. For example

https://github.com/meltano/meltano/blob/28386784d089d4f9847755e9eb93dbbf9c9c8728/src/meltano/cli/upgrade.py#L116-L133

> Either way, will need documenting here: https://docs.meltano.com/reference/settings

I wonder how useful a project setting would be for this. Some clear advantages:

1. Users might expect to control this behavior in a similar way to other settings like `--log-level=...`
2. Probably not too common, but users might have private/custom package sources and they'd only want to install dependencies in a `meltano install` step, and avoid setting private repository secrets for other commands. Having a global option they can set in `meltano.yml` is slightly less annoying and works better for teams, than having to set an env var.

---

Also, I really hadn't considered number 2 above, but it makes a bit worried about making auto-install the default in the next release since it might break users that set env vars for `meltano install` to be expanded in the `pip_url`. If they don't then set the env var for `meltano run`, the fingerprint

1. will be different and cause an unexpected and unwanted reinstallation
2. or fail if `ff.strict_env_var_mode` is set to true

Wdyt?

ReubenFrankel:
> 1. Users might expect to control this behavior in a similar way to other settings like `--log-level=...`
> 2. Probably not too common, but users might have private/custom package sources and they'd only want to install dependencies in a `meltano install` step, and avoid setting private repository secrets for other commands. Having a global option they can set in `meltano.yml` is slightly less annoying and works better for teams, than having to set an env var.

This is what I was thinking - I could see it being useful. It feels like we were moving towards that design anyway with `MELTANO_AUTO_INSTALL` controlling auto-install behaviour across all commands.

> Also, I really hadn't considered number 2 above, but it makes a bit worried about making auto-install the default in the next release since it might break users that set env vars for `meltano install` to be expanded in the `pip_url`. If they don't then set the env var for `meltano run`, the fingerprint
> 
> 1. will be different and cause an unexpected and unwanted reinstallation
> 2. or fail if `ff.strict_env_var_mode` is set to true

Yeah, I see that now too. Would it be feasible to detect

ReubenFrankel:
> 1. Users might expect to control this behavior in a similar way to other settings like `--log-level=...`
> 2. Probably not too common, but users might have private/custom package sources and they'd only want to install dependencies in a `meltano install` step, and avoid setting private repository secrets for other commands. Having a global option they can set in `meltano.yml` is slightly less annoying and works better for teams, than having to set an env var.

This is what I was thinking - I could see it being useful. It feels like we were moving towards that design anyway with `MELTANO_AUTO_INSTALL` controlling auto-install behaviour across all commands.

> Also, I really hadn't considered number 2 above, but it makes a bit worried about making auto-install the default in the next release since it might break users that set env vars for `meltano install` to be expanded in the `pip_url`. If they don't then set the env var for `meltano run`, the fingerprint
> 
> 1. will be different and cause an unexpected and unwanted reinstallation
> 2. or fail if `ff.strict_env_var_mode` is set to true

Yeah, I see that now too. IMO it not being the default behaviour sort of invalidates the entire feature, so would it be feasible to instead

1. detect that the `pip_url` contains env var references, warn the user if they are not set and skip the auto-install step
1. reduce the installed check to "does the virtual environment directory exist and is it non-empty", rather than comparing fingerprints

?

edgarrmondragon:
> 1. detect that the `pip_url` contains env var references, warn the user if they are not set and skip the auto-install step

I like that option. We can add a parameter to `get_pip_install_args` to add the option of raising the exception, and catch `EnvironmentVariableNotSetError`.

ReubenFrankel:
> > 1. detect that the `pip_url` contains env var references, warn the user if they are not set and skip the auto-install step
    > 
    > I like that option. We can add a parameter to `get_pip_install_args` to add the option of raising the exception, and catch `EnvironmentVariableNotSetError`.



> I like that option. We can add a parameter to `get_pip_install_args` to add the option of raising the exception, and catch `EnvironmentVariableNotSetError`.

ReubenFrankel:
> We can add a parameter to `get_pip_install_args` to add the option of raising the exception, and catch `EnvironmentVariableNotSetError`.

Let me know what you think of this: https://github.com/meltano/meltano/pull/8482/commits/630eebfed4b45c127f336b9872ebaf031353852b

src/meltano/core/plugin_install_service.py
```diff
@@ -394,6 +394,25 @@ async def install_plugin_async(
...
+    ) -> bool:
+        try:
+            pip_install_args = get_pip_install_args(
+                self.project,
+                plugin,
+                env,
+                if_missing=EnvVarMissingBehavior.raise_exception,
+            )
+        except EnvironmentVariableNotSetError:
+            return False
```
edgarrmondragon, line 411: What do you think about logging a warning here?

ReubenFrankel: There is already a debug log coming from `expand_env_vars`

https://github.com/meltano/meltano/blob/630eebfed4b45c127f336b9872ebaf031353852b/src/meltano/core/utils/__init__.py#L540-L542

but I suppose we want a more user-centric message like "Environment variable not set for pip_url, skipping install"?

edgarrmondragon: > but I suppose we want a more user-centric message like "Environment variable not set for pip_url, skipping install"?

Yeah exactly. They'd get more details by looking at _debug_ logs, but they'd immediately know that they might not get the latest dependencies for that command.

ReubenFrankel: 145065d8d2917f600a5815d0cd6324846e676024

edgarrmondragon: 👍

edgarrmondragon:
@ReubenFrankel I think the only missing piece now is the global toggle. Wdyt?

ReubenFrankel:
> @ReubenFrankel I think the only missing piece now is the global toggle. Wdyt?

Yup, you happy with project setting name `auto_install` (configurable via `MELTANO_AUTO_INSTALL` env var) and the existing `--install/--no-install` command override options all together?

Given that auto-install is enabled by default, a user should be able to disable it project-wide via

> ```
> MELTANO_AUTO_INSTALL=no
> MELTANO_AUTO_INSTALL=f
> MELTANO_AUTO_INSTALL=0
> MELTANO_AUTO_INSTALL=N
> ```

```yml
version: 1
default_environment: dev
environments:
- name: dev
- name: staging
- name: prod
auto_install: false
```

edgarrmondragon:
> Yup, you happy with project setting name `auto_install` (configurable via `MELTANO_AUTO_INSTALL` env var) and the existing `--install/--no-install` command override options all together?

Yeah, I like that 👍

ReubenFrankel:
OK, I think this is almost there. A couple of things:

- I renamed `PluginInstallReason.JIT` to `PluginInstallReason.AUTO` to align with other "auto-install" references
- Following the previous point, since there are no references to "just-in-time" behaviour, should we update the PR title for the sake of the changelog entry? Or do we want to mention JIT in some capacity in the docs?
- With this new behaviour, when running an affected command and some prerequisite plugins are already installed, a user will now see some extra `INFO`-level logs notifying them of the plugins that were skipped during the auto-install process:

  image

  I feel like this could get annoying pretty fast, so maybe it's better to `INFO` log only when a plugin is actively being installed (for auto-install specifically)?

edgarrmondragon:
> * I feel like this could get annoying pretty fast, so maybe it's better to `INFO` log only when a plugin is actively being installed (for auto-install specifically)?

I agree, if it's not too complicated to change the log level of those messages depending on the command being called. Are you thinking `INFO` for `meltano install` (and `meltano add`?), `DEBUG` otherwise?

> * Following the previous point, since there are no references to "just-in-time" behaviour, should we update the PR title for the sake of the changelog entry? Or do we want to mention JIT in some capacity in the docs?

Yeah, I tend to tweak the title right before merging and this feature will definitely get a big advertisement in the release notes 😄. I'll update it now.

edgarrmondragon (changes requested):
Requesting changes to block a merge in case we want to make last-minute tweaks to the PR title and docs.

ReubenFrankel:
> I agree, if it's not too complicated to change the log level of those messages depending on the command being called. Are you thinking `INFO` for `meltano install` (and `meltano add`?), `DEBUG` otherwise?

Cool, I'll have a look then - and yes, essentially `DEBUG` for anything with `AUTO` plugin install reason.

ReubenFrankel:
After https://github.com/meltano/meltano/pull/8482/commits/bc45ecbfc504dc8a3f4a4bf84c6cbb7d51c68f26 and https://github.com/meltano/meltano/pull/8482/commits/bc428d3fa61d5b479f4ed8219c81c3f8840dbd57:

image
image

edgarrmondragon:
I can confirm INFO logs are only used _when there's something to be installed_. Nice!

src/meltano/core/bundle/settings.yml
```diff
@@ -34,6 +34,9 @@ settings:
   kind: integer
   value: 10485760 # 10 MiB
 - name: python
+- name: auto_install
+  kind: boolean
+  value: true
```
edgarrmondragon, line 39: I think this needs an update to the JSON schema file

ReubenFrankel: https://github.com/meltano/meltano/pull/8482/commits/510b9fea7198d61f045598650a99f790203ca88a

edgarrmondragon (approved):

edgarrmondragon:
This was a long one, huh? Thanks for submitting and thanks for your patience @ReubenFrankel!

ReubenFrankel:
Always a pleasure 😎

PR #8486, opened 2024-04-08
fix: "`kind: {kind}` is deprecated..." is no longer logged if the corresponding replacement is in place

### LINKED ISSUE #8409: fix: Avoid logging `kind: {kind}` is deprecated...` if the corresponding replacement is in place
This would make the migration smoother since the Hub could do a batch update that's backwards compatible. Older versions of Meltano (I think, yet to confirm) wouldn't crash if an unknown setting attribute such as `sensitive` is present.

---

@danielptv I think keeping both is the most sensible. We could update Meltano to only log the warnings when the corresponding future replacement is not in place.

_Originally posted by @edgarrmondragon in https://github.com/meltano/hub/pull/1675#issuecomment-1948857377_

edgarrmondragon:
Closes #8409


src/meltano/core/setting_definition.py
```diff
@@ -225,8 +231,24 @@ def __init__(
...
+            hidden = True
+
+        # Handle deprecated SettingKind.PASSWORD and SettingKind.OAUTH
+        if kind and kind.is_sensitive:
+            # Override kind if sensitive flag is set
+            if sensitive:
+                kind = SettingKind.STRING
+
+            # Prioritize kind over flag otherwise
+            sensitive = True
```
edgarrmondragon, line 251: Here's a table summarizing the logic changes:

|   | Input (`sensitive`, `kind`) | Before (`sensitive`, `kind`)  | After (`sensitive`, `string`)   |
|---|-----------------------------|-------------------------------|---------------------------------|
| 1 | (True, string)              | (True, string)                | (True, string)                  |
| 2 | (True, password)            | (True, password)              | (True, string)                  |
| 3 | (True, None)                | (True, None)                  | (True, None)                    |
| 4 | (False, string)             | (None, string)                | (False, string)                 |
| 5 | (False, password)           | (True, password)              | (True, password)                |
| 6 | (False, None)               | (None, None)                  | (False, None)                   |
| 7 | (None, string)              | (None, string)                | (None, string)                  |
| 8 | (None, password)            | (True, password)              | (True, password)                |
| 9 | (None, None)                | (None, None)                  | (None, None)                    |

Note the changes in cases

* 2: `kind.PASSWORD` is replaced `kind.STRING` and the warning will no longer be logged
* 4 and 6: The `False` value of `sensitive` is respected when the kind is not a sensitive one

Cases 5 and 8 will continue to display the warning.

WillDaSilva (approved):

PR #8490, opened 2024-04-10
fix: An explicit error message is now logged when Meltano fails to retrieve tap state from the state backend

edgarrmondragon:


src/meltano/core/plugin/singer/tap.py
```diff
@@ -306,18 +306,30 @@ async def look_up_state(  # noqa: WPS231, WPS213, WPS210
...
+        )
+        try:
+            state = state_service.get_state(elt_context.job.job_name)
+        except Exception as err:  # pragma: no cover
+            logger.error(
+                err.args[0],
+                state_backend=state_service.state_store_manager.label,
+            )
+            msg = "Failed to retrieve state"
+            raise PluginExecutionError(msg) from err
```
edgarrmondragon, line 330: This gives us a bit more context on why the tap failed to start:

```
2024-04-10T19:18:40.773990Z [error    ] The SSO session associated with this profile has expired or is otherwise invalid. To refresh this SSO session run aws sso login with the corresponding profile. state_backend=AWS S3
2024-04-10T19:18:40.777655Z [error    ] Cannot start plugin tap-getpocket: Failed to retrieve state
```

WillDaSilva (approved):

PR #8508, opened 2024-04-19
fix: Support intervals in cron regex pattern on meltano.schema.json

sabino:
Proposing an updated regex pattern to correctly validate cron expressions including those with intervals (e.g., `0 */3 * * *`).

**Issue:** The existing regex fails to match valid interval-based cron expressions.
VS Code Extension: https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml
image

**Current Regex Test:** [Fails Validation](https://regex101.com/r/ZKE58l/1)

**Proposed Regex Update:** [Passes Validation](https://regex101.com/r/ZKE58l/2)

### DIFF
--- a/src/meltano/schemas/meltano.schema.json
+++ b/src/meltano/schemas/meltano.schema.json
@@ -988,7 +988,7 @@
             "@weekly",
             "0 0 * * *"
           ],
-          "pattern": "^((@(hourly|daily|weekly|monthly|yearly|once))|((((\d+,)+\d+|(\d+|\*)(\/|-)\d+|\d+|\*) ?){5,6}))$"
+          "pattern": "^((@(hourly|daily|weekly|monthly|yearly|once))|((((\\d+,)+\\d+|(\\d+|\\*)(\\/|-)\\d+|\\d+|\\*) ?){5,6}))$"
         },
         "transform": {
           "type": "string",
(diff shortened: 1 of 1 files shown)


WillDaSilva (approved):
Thanks @sabino!

EDIT: Looks like we'll need to use `\\` instead of `\` in the JSON string.

src/meltano/schemas/meltano.schema.json
```diff
@@ -988,7 +988,7 @@
             "@weekly",
             "0 0 * * *"
           ],
-          "pattern": "^((@(hourly|daily|weekly|monthly|yearly|once))|((((\\d+,)+\\d+|(\\d+(\\/|-)\\d+)|\\d+|\\*) ?){5,6}))$"
+          "pattern": "^((@(hourly|daily|weekly|monthly|yearly|once))|((((\d+,)+\d+|(\d+|\*)(\/|-)\d+|\d+|\*) ?){5,6}))$"
```
WillDaSilva, line 991:
```suggestion
          "pattern": "^((@(hourly|daily|weekly|monthly|yearly|once))|((((\\d+,)+\\d+|(\\d+|\\*)(\\/|-)\\d+|\\d+|\\*) ?){5,6}))$"
```

sabino:
Yes, I overlooked the need to escape the `\`.

Thank you for the quick review @WillDaSilva!

WillDaSilva (approved):

PR #8510, opened 2024-04-21
refactor: Plugin installation status is now logged instead of printed

### LINKED ISSUE #8503: feature: Make installation progress messages use logging instead of `click.echo`
### Feature scope

### Description

We'd like to start reducing the amount of non-formatted and non-configurable output in Meltano's CLI. The installation result messages are good start:

```
Installing 1 plugin           
Installing extractor 'tap-getpocket'...
Installed extractor 'tap-getpocket'
```

### Related

* https://github.com/meltano/meltano/pull/8482#issuecomment-2042923721
* https://github.com/meltano/meltano/issues/6394

edgarrmondragon:
Closes #8503


src/meltano/cli/install.py
```diff
@@ -95,7 +95,7 @@ def install(  # noqa: C901
         tracker.track_command_event(CliEvent.aborted)
         raise
 
-    click.echo(f"Installing {len(plugins)} plugins...")
+    logger.info("Installing %d plugin", len(plugins))
```
edgarrmondragon, line 98:
```suggestion
    logger.info("Installing %d plugins", len(plugins))
```

ReubenFrankel:
I'll use this to replace my POC implementation in #8482 :slightly_smiling_face:

src/meltano/cli/utils.py
```diff
@@ -415,15 +416,15 @@ def install_status_update(install_state):
         PluginInstallStatus.SKIPPED,
     }:
         msg = f"{install_state.verb} {desc} '{plugin.name}'..."
-        click.secho(msg)
+        logger.info(msg)
```
WillDaSilva, line 419: Should this use `%` style interpolation into `logger.info` instead of an f-string? The same comment applies a few lines below.

WillDaSilva (approved):

src/meltano/cli/utils.py
edgarrmondragon: Ah, good catch. Ruff missed this cause it's not a literal.

bfcbd71d9

Issue #2415, opened 2020-11-25
Cache discovered extractor catalog when running pipeline


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/2460

Originally created by @ishan.mails on 2020-11-25 15:56:08

---

### Problem to solve
(Summarize the problem we are trying to solve in the form of we need to do [A], so that [B] can [C])
We need to enable catalog to be cached and reused, so that meltano does not attempt to generate the entire catalog for each elt run.
I believe the catalog generation for each run turns out to be a costly operation and instead there can be some mechanism to preserve the catalog and refresh it on-demand or based on a schedule.

### Target audience
This could be an opt-in feature as well. Meltano already allows to manually specify a catalog. It would be great if the meltano utilities such as schema filters can be applied on a manual catalog as well.

### Further details
During my testing with tap-salesforce, the catalog file comes out to be 10MB with 1000 entities. It takes considerable amount of time and API calls to generate the catalog for each elt run when a typical use-case would involve syncing only 5-10 entities.

### Proposal
There is already an option to specify the catalog file. I believe we only apply the same rules for a manual catalog, as it is done on a discovered catalog. 

### What does success look like, and how can we measure that?
(Define both the success metrics and acceptance criteria. Note that success metrics indicate the desired business outcomes, while acceptance criteria indicate when the solution is working correctly. If there is no way to measure success, link to an issue that will implement a way to measure this)

### Regression test
(Ensure the feature doesn't cause any regressions)
- [ ] Write adequate test cases and submit test results
- [ ] Test results should be reviewed by a person from the team

### Links / references

_Please note that this was taken from GitLab, to be changed accordingly_

MeltyBot:
[View 2 previous comments from the original issue on GitLab](https://gitlab.com/meltano/meltano/-/issues/2460)

edgarrmondragon:
This will be helped by https://github.com/meltano/meltano/pull/8626, but

* The catalog will only be cached for pipelines with the _same state ID and run ID_
* `elt/el` is a legacy command and `run` is recommended
* `--refresh-catalog` will offer a way of ignoring the cache

edgarrmondragon:
This will be helped by https://github.com/meltano/meltano/pull/8626, but

* The catalog will only be cached for pipelines with the _same state ID and run ID_
* `elt/el` is a legacy command and `run` is recommended
* `--refresh-catalog` will offer a way of ignoring the cache

PR #8518, opened 2024-04-29
feat: Ignore Catalog Cache

### LINKED ISSUE #6919: feature: Add a `--refresh` option for `meltano select` to refresh a new copy of the discovered catalog
This feature would add a `--refresh` option, or similar CLI arg to the `meltano select` CLI command. https://docs.meltano.com/reference/command-line-interface#select

```
# Refresh the cached catalog by re-running discovery
meltano select tap-github --refresh

# Print the normal --list output *after* refreshing the catalog
meltano select tap-github --list --refresh
```

This would give users the ability to force a refresh of the cached `catalog.json` file.

## Related

- https://github.com/meltano/meltano/issues/6915
- https://github.com/meltano/meltano/issues/6763
- https://github.com/meltano/meltano/discussions/6757#discussioncomment-3650870

andyoneal:
closes #6919

* Adds `no_catalog_cache` to config extras, so that a single plugin can be run every time without the catalog cache.
* Adds `--catalog-refresh` option to `meltano select` and `meltano el`

Adding a new extra seems extreme, but I don't know how else to implement this where it could be used as part of a job for any tap. If there's a better way to do it, I'm happy to adjust this based on feedback.


src/meltano/cli/select.py
```diff
@@ -112,13 +122,13 @@ def update(
     select_service.update(entities_filter, attributes_filter, exclude, remove)
 
 
-async def show(project, extractor, show_all=False):
+async def show(project, extractor, show_all=False, refresh=False):
```
edgarrmondragon, line 125:
```suggestion
async def show(
    project: Project,
    extractor: str,
    show_all: bool = False,
    refresh: bool = False,
) -> None:
```

src/meltano/core/plugin/singer/tap.py
```diff
@@ -185,6 +185,9 @@ class SingerTap(SingerPlugin):  # noqa: WPS214
             value_processor="nest_object",
         ),
         SettingDefinition(name="_select_filter", kind=SettingKind.ARRAY, value=[]),
+        SettingDefinition(
+            name="_no_catalog_cache", kind=SettingKind.BOOLEAN, value=False
+        ),
```
edgarrmondragon, line 190: Can you add documentation for this?

https://docs.meltano.com/concepts/plugins/#extras would be a good place.

src/meltano/cli/elt.py
```diff
@@ -57,6 +57,10 @@ class ELOptions:
         help="Perform a full refresh (ignore state left behind by any previous runs).",
         is_flag=True,
     )
+    catalog_refresh = click.option(
+        "--catalog-refresh",
+        help="Invalidates catalog cache and forces running discovery before this run.",
```
edgarrmondragon, line 62:
```suggestion
        help="Invalidates catalog cache and forces running discovery before this run.",
        is_flag=True,
```

src/meltano/cli/elt.py
edgarrmondragon, line 1: I actually don't think the changes here are necessary because every new EL run gets its own directory based on the job's `run_id`:

```console
$ tree .meltano/run/elt/test/
.meltano/run/elt/test/
├── 26cee8e6-e849-44b4-8bbf-bbfd6b1d25e2
│   ├── tap.properties.cache_key
│   └── tap.properties.json
├── 32144c9c-4c1d-4185-876f-0af1a83f4d46
│   ├── state.json
│   ├── tap.properties.cache_key
│   └── tap.properties.json
└── 8700f70e-422e-4f35-b826-1ac15da333e0
    ├── tap.properties.cache_key
    └── tap.properties.json
```

That means the catalog is **always** generated.

It also doesn't hurt to have this in `meltano el`, specially if we port https://github.com/meltano/meltano/pull/8459 to the `meltano el` command.

tests/meltano/core/plugin/singer/test_tap.py
```diff
@@ -267,6 +268,33 @@ def mock_discovery(*args, **kwargs):  # noqa: ARG001
...
+                side_effect=mock_discovery,
+            ) as mocked_run_discovery:
+                assert catalog_cache_key_path.exists()
+
+                # with _no_catalog_cache = true, discovery should be invoked
+                # again even with a stored cache key.
+                mocked_run_discovery.reset_mock()
+                await subject.discover_catalog(invoker)
+
+                assert mocked_run_discovery.called
```
edgarrmondragon, line 294: Maybe

```suggestion
                mocked_run_discovery.assert_called_once()
```

src/meltano/core/select_service.py
```diff
@@ -39,19 +39,22 @@ def current_select(self):
         plugin_settings_service = PluginSettingsService(self.project, self.extractor)
         return plugin_settings_service.get("_select")
 
-    async def load_catalog(self, session):
+    async def load_catalog(self, session, refresh=False):
```
edgarrmondragon, line 42: What do you think about adding tests for this in `tests/meltano/cli/test_select.py`?

edgarrmondragon (changes requested):
Thanks @andyoneal!

Can we add this to the documentation for the "run" command? https://docs.meltano.com/reference/command-line-interface#run

edgarrmondragon:
> Adding a new extra seems extreme, but I don't know how else to implement this where it could be used as part of a job for any tap. If there's a better way to do it, I'm happy to adjust this based on feedback.

The new extra is probably fine, and it's a nice way for users to tell Meltano to always use a fresh catalog.

src/meltano/cli/elt.py
andyoneal: Agree that it doesn't really do anything for now. Great that it might become useful soon though.

src/meltano/core/plugin/singer/tap.py
andyoneal: added

src/meltano/core/select_service.py
andyoneal: I think I'm a little out of my depth on this. In order to test, I think I'd need a basic tap that would dynamically build a catalog from a data source, right? Like a tap-json where I modify the json in between runs? Just using tap-mock (which I can't find the code for anyway) wouldn't test this.

Am I overcomplicating this?

src/meltano/cli/select.py
```diff
@@ -85,7 +90,12 @@ async def select(
     """
     try:
         if flags["list"]:
-            await show(project, extractor, show_all=flags["all"])
+            await show(
+                project,
+                extractor,
+                show_all=flags["all"],
+                refresh=flags["catalog-refresh"],
```
edgarrmondragon, line 97:
```suggestion
                refresh=flags["catalog_refresh"],
```

src/meltano/core/select_service.py
edgarrmondragon: See https://github.com/andyoneal/meltano/pull/1

andyoneal: thanks. merged.

tests/meltano/cli/test_select.py
```diff
@@ -33,3 +39,40 @@ def test_update_select_pattern(self, cli_runner, tap):
...
+                "select",
+                tap.name,
+                "--list",
+                "--all",
+            ],
+        )
+        assert_cli_runner(result)
+
+        assert "[SelectionType.SELECTED] users.id" in result.stdout
+        assert "[SelectionType.EXCLUDED] users.name" in result.stdout
```
edgarrmondragon, line 78: Turns out this uncovered something I hadn't noticed before 🙃 

https://github.com/meltano/meltano/issues/8526

edgarrmondragon, line 78:
```suggestion
        assert "[selected ] users.id" in result.stdout
        assert "[excluded ] users.name" in result.stdout
```

edgarrmondragon: Fixed by https://github.com/meltano/meltano/issues/8527

edgarrmondragon:
@andyoneal I think files in https://github.com/meltano/meltano/tree/42405debc9369235c64c2f5a0ad8fbd9a1e61a35/integration/meltano-manifest/expected-manifests need to be updated so the integration tests can pass

andyoneal:
> @andyoneal I think files in https://github.com/meltano/meltano/tree/42405debc9369235c64c2f5a0ad8fbd9a1e61a35/integration/meltano-manifest/expected-manifests need to be updated so the integration tests can pass

was wondering where those were coming from. will update.

src/meltano/core/plugin/singer/tap.py
```diff
@@ -185,6 +185,9 @@ class SingerTap(SingerPlugin):  # noqa: WPS214
             value_processor="nest_object",
         ),
         SettingDefinition(name="_select_filter", kind=SettingKind.ARRAY, value=[]),
+        SettingDefinition(
+            name="_no_catalog_cache", kind=SettingKind.BOOLEAN, value=False
```
edgarrmondragon, line 189: Nit that came up while checking this out during Office Hours:

What do you think of making this `_catalog_cache` instead and default it to `True`?

WillDaSilva: All else being equal, avoiding a double negative would be nice

edgarrmondragon:
@andyoneal Are you interested in getting this PR across the finish line? Unfortunately I can't push to your `main` branches so I can't make the last-mile changes to this PR.

edgarrmondragon:
Closing in favor of #8580

src/meltano/cli/elt.py
edgarrmondragon: I didn't realize this is actually a very old issue: https://github.com/meltano/meltano/issues/2415

PR #8524, opened 2024-04-30
docs: Changed references of Airflow `orchestrator` to `utility`

### LINKED ISSUE #7804: bug: Getting Started uses airflow orchestrator instead of airflow utility
In the Getting Started guide, users are told to [add the airflow orchestrator](https://docs.meltano.com/getting-started/#schedule-pipelines-to-run-regularly) to their project:
```bash
meltano add orchestrator airflow
```

While doing some digging, I came across the [Airflow utility](https://hub.meltano.com/utilities/airflow), which states:
```
This utility plugin is meant to be used in favor of the [Airflow orchestrator plugin type](https://hub.meltano.com/orchestrators/airflow).
```

Given that the Airflow utility is something we manage, we should update the Getting Started guide to use the airflow utlity.

edgarrmondragon:
Closes https://github.com/meltano/meltano/issues/7804

### DIFF
--- a/docs/docs/concepts/plugins.mdx
+++ b/docs/docs/concepts/plugins.mdx
@@ -839,7 +839,7 @@ Orchestrators are [pip packages](https://pip.pypa.io/en/stable/) responsible for
 
 Meltano supports [Apache Airflow](https://airflow.apache.org/) out of the box, but can be used with any tool capable of reading the output of [`meltano schedule list --format=json`](/reference/command-line-interface#schedule) and executing each pipeline's [`meltano run`](/reference/command-line-interface#run) command on a schedule.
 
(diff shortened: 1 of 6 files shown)


WillDaSilva (approved):

ilsaloving:
In that case, https://hub.meltano.com/orchestrators/airflow/ needs to be updated and/or removed, since that is how I got to this point.

edgarrmondragon:
> In that case, https://hub.meltano.com/orchestrators/airflow/ needs to be updated and/or removed, since that is how I got to this point.

It's hidden on search, i.e. you only see the utility if you start typing "airflow" in the search box.

What we could do is show a banner warning that the orchestrator is deprecated: https://github.com/meltano/hub/issues/1764

PR #8527, opened 2024-05-03
fix: Made the output of `meltano select ... list` consistent between different Python versions

### LINKED ISSUE #8526: bug: Output of `meltano select <extractor> --list` is **visually** inconsistent between different Python 3.8-3.10 and 3.11+
### Meltano Version

3.4.0

### Python Version

3.11

### Bug scope

### Operating System

NA

### Description

From the [Python 3.11 changelog](https://docs.python.org/3.11/whatsnew/3.11.html#enum):

<blockquote>

<p>Changed <a class="reference internal" href="https://docs.python.org/3/library/enum.html#enum.Enum.__format__" title="enum.Enum.__format__"><code class="xref py py-meth docutils literal notranslate"><span class="pre">Enum.__format__()</span></code></a> (the default for
<a class="reference internal" href="https://docs.python.org/3/library/functions.html#format" title="format"><code class="xref py py-func docutils literal notranslate"><span class="pre">format()</span></code></a>, <a class="reference internal" href="https://docs.python.org/3/library/stdtypes.html#str.format" title="str.format"><code class="xref py py-meth docutils literal notranslate"><span class="pre">str.format()</span></code></a> and <a class="reference internal" href="https://docs.python.org/3/glossary.html#term-f-string"><span class="xref std std-term">f-string</span></a>s) to always produce
the same result as <a class="reference internal" href="https://docs.python.org/3/library/enum.html#enum.Enum.__str__" title="enum.Enum.__str__"><code class="xref py py-meth docutils literal notranslate"><span class="pre">Enum.__str__()</span></code></a>:  for enums inheriting from
<a class="reference internal" href="https://docs.python.org/3/library/enum.html#enum.ReprEnum" title="enum.ReprEnum"><code class="xref py py-class docutils literal notranslate"><span class="pre">ReprEnum</span></code></a> it will be the member’s value; for all other enums
it will be the enum and member name (e.g. <code class="docutils literal notranslate"><span class="pre">Color.RED</span></code>).</p>

</blockquote>

Python 3.8-3.10:

```python
>>> from meltano.core.plugin.singer.catalog import SelectionType
>>> f"{SelectionType.SELECTED}"
'selected'
```

Python 3.11+:

```python
>>> from meltano.core.plugin.singer.catalog import SelectionType
>>> f"{SelectionType.SELECTED}"
'SelectionType.SELECTED'
```

One potential fix would be subclassing [`ReprEnum`](https://docs.python.org/3/library/enum.html#enum.ReprEnum) on Python 3.11+.

### Code

_No response_

edgarrmondragon:
Related:

* Closes https://github.com/meltano/meltano/issues/8526

### DIFF
--- a/src/meltano/core/plugin/singer/catalog.py
+++ b/src/meltano/core/plugin/singer/catalog.py
@@ -295,6 +295,7 @@ class CatalogNode(Enum):
     METADATA = auto()
 
 
+# TODO: Move to `enum.StrEnum` when support for Python 3.8 is dropped
 class SelectionType(str, ReprEnum):
     """A valid stream or property selection type."""
 
--- a/src/meltano/core/plugin/singer/catalog.py
+++ b/src/meltano/core/plugin/singer/catalog.py
@@ -3,12 +3,18 @@ from __future__ import annotations
 import fnmatch
 import logging
 import re
+import sys
 import typing as t
 from collections import OrderedDict
 from enum import Enum, auto
 from functools import singledispatch
 
 from meltano.core.behavior.visitor import visit_with
+
+if sys.version_info < (3, 11):
+    ReprEnum = Enum
+else:
+    from enum import ReprEnum
 
 Node = t.Dict[str, t.Any]
 T = t.TypeVar("T", bound="CatalogRule")
@@ -289,7 +295,7 @@ class CatalogNode(Enum):
     METADATA = auto()
 
 
-class SelectionType(str, Enum):
+class SelectionType(str, ReprEnum):
     """A valid stream or property selection type."""
 
     SELECTED = "selected"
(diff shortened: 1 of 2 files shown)


WillDaSilva (approved):
@edgarrmondragon In the future, what are your thoughts regarding moving to `StrEnum` instead?

edgarrmondragon:
> @edgarrmondragon In the future, what are your thoughts regarding moving to `StrEnum` instead?

@WillDaSilva I like that. I actually tried to bring in https://pypi.org/project/backports.strenum/, but it has a weird constraint on Python >= 3.8.6 (i.e. on a rather specific patch version) so Poetry didn't like that.

When we drop 3.8 support later this year, we can surely move to `StrEnum`.

PR #8564, opened 2024-05-28
chore: Address typing issues in less problematic `meltano.cli` modules

edgarrmondragon:


src/meltano/cli/cli.py
```diff
@@ -56,7 +57,7 @@ def main(self, *args, **kwargs) -> t.NoReturn:
     # NOTE: This CLI option normalization applies to all subcommands.
     context_settings={"token_normalize_func": lambda x: x.replace("_", "-")},
 )
-@click.option("--log-level", type=click.Choice(LEVELS.keys()))
+@click.option("--log-level", type=click.Choice(tuple(LEVELS)))
```
edgarrmondragon, line 60: https://github.com/pallets/click/pull/2736

src/meltano/cli/schedule.py
```diff
@@ -193,11 +205,17 @@ def _format_elt_list_output(entry: Schedule, session: Session) -> dict:
...
-@click.option("--format", type=click.Choice(("json", "text")), default="text")
+@click.option(
+    "--format",
+    "list_format",
+    type=click.Choice(("json", "text")),
+    default="text",
+)
 @click.pass_context
-def list(ctx, format):  # noqa: WPS125
+def list_schedulers(ctx: click.Context, list_format: str) -> None:  # noqa: C901
```
edgarrmondragon, line 218:
```suggestion
def list_schedules(ctx: click.Context, list_format: str) -> None:  # noqa: C901
```

src/meltano/cli/cli.py
```diff
@@ -43,7 +43,8 @@ def main(self, *args, **kwargs) -> t.NoReturn:
             args: Positional arguments for the Click group.
             kwargs: Keyword arguments for the Click group.
         """
-        return super().main(*args, windows_expand_args=False, **kwargs)
+        # TODO: Fix "Return statement in function which does not return"?
+        return super().main(*args, windows_expand_args=False, **kwargs)  # type: ignore[misc]
```
WillDaSilva, line 47: Why are we returning anything from this function? If the return value is significant, why not change the type signature?

WillDaSilva (approved):

src/meltano/cli/cli.py
edgarrmondragon: Yeah, I don't know why that didn't occur to me 😅

1a9fba1b7

PR #8609, opened 2024-07-03
docs: Redirect Meltano Cloud docs to Arch docs

WillDaSilva:
Another PR will remove the Cloud docs, and clean up other outdated mentions of Meltano Cloud.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -20,3 +20,9 @@
   to = "https://docs.meltano.com/guide/troubleshooting/:splat"
   status = 301
   force = true
+
+[[redirects]]
+  from = "https://docs.meltano.com/cloud/*"
+  to = "https://docs.arch.dev/"
+  status = 301
+  force = true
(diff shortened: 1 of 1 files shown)


netlify.toml
```diff
@@ -20,3 +20,9 @@
   to = "https://docs.meltano.com/guide/troubleshooting/:splat"
   status = 301
   force = true
+
+[[redirects]]
+  from = "https://docs.meltano.com/cloud/*"
```
edgarrmondragon, line 25:
```suggestion
  from = "/cloud/*"
```

edgarrmondragon: Actually, I'm not sure if we want to redirect in the preview builds 🤔

edgarrmondragon: See for example https://deploy-preview-8609--meltano.netlify.app/cloud/connectors/

WillDaSilva: We'll be deleting the Cloud docs anyway in https://github.com/meltano/meltano/pull/8610, so I don't think there's any harm in making the redirect apply to preview builds too

edgarrmondragon (approved):

PR #8620, opened 2024-07-08
feat: Sensitive values in `meltano config <plugin>` are now hidden by default

### LINKED ISSUE #8619: feature: hide sensitive values in `meltano config <plugin>`
### Feature scope

### Description

This is a follow-on from https://github.com/meltano/meltano/pull/7878#issuecomment-1624013519

Currently both `meltano invoke --dump=config <plugin>` and `meltano config <plugin>` will show sensitive values with no `--safe` option. 

Options we have:
* `meltano config <plugin>` work the same as ``meltano config <plugin> list`
* Have `meltano config <plugin>` default to a `--safe` option where sensitive values are replaced with ******
* Drop `meltano config <plugin>` as an option and recommend either the list subcommand or the invoke with dump feature

I'm not sure what's easiest @edgarrmondragon or feels best. I have a slight preference for adding `--safe` and recommending `invoke` to really see what the config is. @ReubenFrankel do you have any thoughts on this?

edgarrmondragon:
Closes #8619


WillDaSilva (approved):
I'm not fond of making a breaking change without a major version bump, but I read the issue thread, and I agree with the conclusion there.

tayloramurphy (approved):
Approving for docs. Have not tested the code personally.

PR #8626, opened 2024-07-12
feat: Added a `--run-id` option to `meltano el[t]`

### LINKED ISSUE #2415: Cache discovered extractor catalog when running pipeline

### LINKED ISSUE #8523: feature: Add a `--run-id` option to `meltano el` similar to the one in `meltano run`
### Feature scope

### Description

Context:

* https://github.com/meltano/meltano/issues/8458
* https://github.com/meltano/meltano/pull/8459
* https://github.com/meltano/meltano/pull/8518#discussion_r1583993451
* https://github.com/meltano/meltano/issues/2415

edgarrmondragon:
Related:

* Closes https://github.com/meltano/meltano/issues/8523
* Closes https://github.com/meltano/meltano/issues/2415


src/meltano/cli/elt.py
```diff
@@ -105,6 +112,11 @@ class ELOptions:
         is_flag=True,
         help="Merges state with that of previous runs.",
     )
+    run_id = click.option(
+        "--run-id",
+        type=UUIDParamType(),
+        help="Use a custom run ID.",
```
WillDaSilva, line 118: We should make it clear that a UUID is still required, rather than allowing any alternative ID format

```suggestion
        help="Provide a UUID to use as the run ID.",
```

src/meltano/core/elt_context.py
```diff
@@ -142,17 +147,25 @@ def __init__(
...
-        if self.job:  # noqa: RET503
+        # In an EL/ELT run, there is always a job. This means the run directory
+        # is always ".meltano/run/<state-id>/<run-id>".
+        # This means the catalog can only be cached for executions that use the
+        # same state and run ID.
+        # See https://github.com/meltano/meltano/issues/2415.
+        if self.job:
             return self.project.job_dir(self.job.job_name, str(self.job.run_id))
 
+        return None  # pragma: no cover
```
WillDaSilva, line 167: If there should always be a job, would it be okay if we replace `return None` with `raise RuntimeException` (or some other exception), and then change the return type to `Path` instead of `Path | None`?

edgarrmondragon: Yeah, I was maybe being unnecessarily cautious but it's probably fine to crash if `self.job` is null, which should never happen.

edgarrmondragon: Argh, we actually do test the `ELTContext` for cases in which the job is not set. Rather than updating the tests, I'll revert the raising of an exception here and update the comment.

WillDaSilva (approved):

PR #8638, opened 2024-07-17
chore: Update issue template and add another for PRs

edgarrmondragon:


pyproject.toml
```diff
@@ -479,6 +479,7 @@ version_files = [
   "pyproject.toml:^version =",
   "src/meltano/__init__.py:^__version__ =",
   'docs/package.json:^  "version":',
+  ".github/ISSUE_TEMPLATE/bug.yml:^      placeholder:",
```
WillDaSilva, line 482: Will this also trigger on `placeholder: "https://..."`?

edgarrmondragon: Nope. It will only replace the previous version (currently `3.5.0a5`) on lines that match this pattern. If the line matches this pattern but there's nothing for the new version to replace, then it's left as is.

We use the same pattern in the sdk: https://github.com/meltano/sdk/blob/b566a5b11392550753d39e475eaaf9cbdcf4d89c/pyproject.toml#L189.

I guess we could also just give it the file name `".github/ISSUE_TEMPLATE/bug.yml"` and let commitizen replace any string that matches the previous version.

Ref: https://commitizen-tools.github.io/commitizen/commands/bump/#version_files

WillDaSilva (approved):

PR #8672, opened 2024-07-30
chore: Enable RUF Ruff rules

WillDaSilva:
Relates to:


poetry.lock
```diff
@@ -89,30 +89,6 @@ files = [
     {file = "aiohttp-3.10.0-cp38-cp38-manylinux_2_17_ppc64le.manylinux2014_ppc64le.whl", hash = "sha256:cd36d0f0afc2bd84f007cedd2d9a449c3cf04af471853a25eb71f28bc2e1a119"},
     {file = "aiohttp-3.10.0-cp38-cp38-manylinux_2_17_s390x.manylinux2014_s390x.whl", hash = "sha256:f64d503c661864866c09806ac360b95457f872d639ca61719115a9f389b2ec90"},
     {file = "aiohttp-3.10.0-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl", hash = "sha256:31616121369bc823791056c632f544c6c8f8d1ceecffd8bf3f72ef621eaabf49"},
-    {file = "aiohttp-3.10.0-cp38-cp38-manylinux_2_5_i686.manylinux1_i686.manylinux_2_17_i686.manylinux2014_i686.whl", hash = "sha256:f76c12abb88b7ee64b3f9ae72f0644af49ff139067b5add142836dab405d60d4"},
```
edgarrmondragon, line 92: You might wanna `poetry cache clear pypi --all`

src/meltano/core/migration_service.py
```diff
@@ -62,7 +62,7 @@ def ensure_migration_needed(
             if rev.revision == target_revision:
                 raise MigrationUneededException
 
-    def upgrade(  # noqa: WPS213, WPS231 too many expression and too complex
+    def upgrade(  # too many expression and too complex
```
edgarrmondragon, line 65:
```suggestion
    def upgrade(
```

edgarrmondragon: Wdyt we should do with these type of comments?

WillDaSilva: I think they're harmless. It's now unclear that we wrote them in response to linter issues, but that's probably okay.

edgarrmondragon (approved):

PR #8678, opened 2024-08-01
chore: Enable Ruff ANN rules

WillDaSilva:
For this rule I applied the available auto-fixes, which seems to have just added `-> None` to functions that don't have a return statement. Then I used `ruff check . --add-noqa` to ignore the rest of the issues, since addressing them all in one go is impractical. We can chip away at them as we naturally work on parts of the codebase.

Relates to:
- #7383

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -347,8 +347,6 @@ banned-from = [
 typing = "t"
 
 [tool.ruff.lint.flake8-pytest-style]
-parametrize-names-type = "tuple"
-parametrize-values-row-type = "tuple"
 parametrize-values-type = "tuple"
 
 [tool.ruff.lint.isort]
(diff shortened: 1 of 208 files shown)


pyproject.toml
```diff
@@ -344,6 +347,8 @@ banned-from = [
 typing = "t"
 
 [tool.ruff.lint.flake8-pytest-style]
+parametrize-names-type = "tuple"
+parametrize-values-row-type = "tuple"
```
edgarrmondragon, line 351: These are the default, so we can probably omit them.

```suggestion
```

- https://docs.astral.sh/ruff/settings/#lint_flake8-pytest-style_parametrize-names-type
- https://docs.astral.sh/ruff/settings/#lint_flake8-pytest-style_parametrize-values-row-type

edgarrmondragon (approved):

PR #8748, opened 2024-09-05
refactor: Move `install_plugins` to `meltano.core.plugin_install_service`

edgarrmondragon:
## Description

This lets us eliminate one dependency on `meltano.cli` in `meltano.core`, though there's still
https://github.com/meltano/meltano/blob/4efb91d68bb8e5c2364ea7ae2f032a937959d3f2/src/meltano/core/tracking/tracker.py#L520-L569

which I'm not sure yet how to decouple.

## Related Issues

* Closes #XXXX

### DIFF
--- a/src/meltano/cli/install.py
+++ b/src/meltano/cli/install.py
@@ -8,9 +8,10 @@ import click
 import structlog
 
 from meltano.cli.params import pass_project
-from meltano.cli.utils import CliError, PartialInstrumentedCmd, install_plugins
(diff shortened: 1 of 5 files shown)


edgarrmondragon:
Some tests are failing cause GitLab's having a bad one: https://status.gitlab.com/.

Maybe our tests should not be hitting GitLab directly and instead mock-installing from git.

WillDaSilva (approved):

PR #8758, opened 2024-09-10
refactor: Use `JobState.merge_partial` instead of `meltano.core.utils.merge`

edgarrmondragon:
## Description

`JobState.merge_partial` uses it under the hood anyway:

https://github.com/meltano/meltano/blob/9dd1803252e06ef3b249cdc144d652e6d9187b61/src/meltano/core/job_state.py#L154-L166

## Related Issues

* Closes #XXXX


tests/meltano/core/state_store/test_db.py
```diff
@@ -31,47 +32,47 @@ def test_set_state(self, subject: DBStateStoreManager) -> None:
...
         assert subject.get(partial_only.state_id) == partial_only
 
         # Partial merges onto existing partial
         new_partial = JobState(
             state_id="partial_only",
             partial_state={"singer_state": {"partial": 2}},
         )
-        subject.set(new_partial)
+        subject.set(copy.deepcopy(new_partial))
         assert subject.get(partial_only.state_id) == new_partial
```
edgarrmondragon, line 44: Because `merge(src, dst)` mutates `dst` in place, theses tests were tautological. Meaning that the `new_partial.partial_state` dictionary was mutated and made equal to the new partial state if the merge order was incorrect. Luckily, the merge order was correct but theses changes will ensure we catch any errors in the future.

src/meltano/core/job_state.py
```diff
@@ -160,7 +161,7 @@ def merge_partial(
         Args:
             state: the partial state to merge onto this state.
         """
-        self.partial_state = merge(
```
edgarrmondragon, line 163: Removed this redundant assignment cause `merge` mutates the second argument in-place. This mutability is seriously bug-prone but changing the behavior is a larger task.

src/meltano/core/state_store/base.py
```diff
@@ -31,7 +31,8 @@ def __init__(self, **kwargs) -> None:  # noqa: ANN003, B027
             kwargs: additional keyword arguments
         """
 
-    @abstractproperty
```
edgarrmondragon, line 34: This decorator is deprecated

tests/meltano/core/state_store/test_db.py
edgarrmondragon: Using `copy.deepcopy` was useful to validate the changes to `DBStateStoreManager.set`, but I can revert this before merging.

edgarrmondragon:
superseded by https://github.com/meltano/meltano/pull/8851

PR #8770, opened 2024-09-16
fix(cli): Redact secret settings by default in `meltano compile`

### LINKED ISSUE #7355: bug: manifest contains passwords when using `<PLUGIN_NAME>_<SETTING_NAME>` syntax
### Meltano Version

2.16.1

### Python Version

3.9

### Bug scope

### Operating System

macOS

### Description

Using the new beta [compile](https://docs.meltano.com/reference/command-line-interface#compile) command to generate manifest files causes some of my sensitive settings to be included. As I understand it thats not supposed to happen.

I have some settings defined in my .env file using the `<PLUGIN_NAME>_<SETTING_NAME>` syntax and it looks like theyre getting through to my manifest file. For example `TAP_GITHUB_AUTH_TOKEN`. It's set as a password type in [the plugin definition](https://github.com/meltano/squared/blob/ac3eee55102c275a1c6a706279d645e3522a401d/data/plugins/extractors/tap-github--meltanolabs.lock#L47) sourced from the hub.

### Code

_No response_

holly-evans:
## Description

`meltano compile` produces a manifest file with plaintext secret values, such as passwords. 

This change redacts those secrets by default and adds the `--safe`/`--unsafe` flag to `meltano compile` to keep the option for an unredacted manifest file. (Additionally keeps the secrets by default in the `Manifest` class).

## Related Issues

* Closes #7355


holly-evans:
Note: to fix the integration test, I added the `--unsafe` flag. Would you rather
* redact the manifest to match the default `meltano compile` command
* add another step in the integration to run it once each `--safe` (default` and `--unsafe`

?

integration/example-library/meltano-manifest/index.md
edgarrmondragon, line 1: > Note: to fix the integration test, I added the `--unsafe` flag. Is that fine or would you rather
> 
> * redact the manifest to match the default `meltano compile` command
> * change integration test to run it once each `--safe` (default) and `--unsafe`
> 
> 
> ?

I like the second option, maybe add a new invocation example at the end for a specific environment? e.g. `meltano  --environment=jigsaw compile --indent 2 --unsafe`

edgarrmondragon (commented):
Hi @holly-evans, thanks for the PR!

I see the description of the problem in the linked PR and I think this change makes sense.

edgarrmondragon:
Also, maybe we should add to the docs?

https://docs.meltano.com/reference/command-line-interface/#compile

holly-evans:
@edgarrmondragon Thanks for the feedback! Implemented 👍

edgarrmondragon (approved):

edgarrmondragon:
Thanks @holly-evans!

edgarrmondragon:
@all-contributors
please add @holly-evans for code.

Issue #8783, opened 2024-09-18
bug: Investigate quirks of (auto) installing plugins when there's multiple levels of inheritance


edgarrmondragon:
### Meltano Version

3.5.2

### Python Version

3.12

### Bug scope

CLI (options, error messages, logging, etc.)

### Operating System

NA

### Description

TODO

- https://meltano.slack.com/archives/C069CQNHDNF/p1726653099077779
- https://meltano.slack.com/archives/C0695LLCGVC/p1726488890814789

### Link to Slack/Linen

https://meltano.slack.com/archives/C069CQNHDNF/p1726653099077779

edgarrmondragon:
Closed by https://github.com/meltano/meltano/pull/8785

PR #8784, opened 2024-09-19
fix: Inheriting plugins are no longer always auto-installed

ReubenFrankel:
Culprit of this issue is that the `plugin_dir` method does not respect plugin inheritance, so would reference a plugin directory by the inheriting plugin name, rather than its progenitor plugin.

In the case of auto-install, the check to determine whether or not an install is required calls `plugin_dir`, resolves an incorrect directory, and then goes on to install reusing the parent plugin directory later on. Since the venv in the inherited plugin directory is never created, auto-install is always attempted.

Worth confirming that there is a separate but related issue here around install logic and multiple inheritance as discussed in [this Slack thread](https://meltano.slack.com/archives/C0695LLCGVC/p1726488890814789). ~A fix for that might benefit from reusing `get_progenitor`.~ EDIT: never mind, it's simpler than that (see #8785)!

---

Related to #8783

### DIFF
--- a/src/meltano/core/project.py
+++ b/src/meltano/core/project.py
@@ -25,6 +25,7 @@ from meltano.core.error import (
     ProjectReadonly,
 )
 from meltano.core.hub import MeltanoHubService
+from meltano.core.plugin.project_plugin import ProjectPlugin
 from meltano.core.project_files import ProjectFiles
 from meltano.core.project_plugins_service import ProjectPluginsService
 from meltano.core.project_settings_service import ProjectSettingsService
@@ -32,7 +33,7 @@ from meltano.core.utils import makedirs, sanitize_filename, truthy
 
 if t.TYPE_CHECKING:
     from meltano.core.meltano_file import MeltanoFile as MeltanoFileTypeHint
-    from meltano.core.plugin.project_plugin import ProjectPlugin
+    from meltano.core.plugin.base import PluginRef
 
     if sys.version_info < (3, 10):
         from typing import TypeAlias  # noqa: ICN003
@@ -547,7 +548,7 @@ class Project(Versioned):
     @makedirs
     def plugin_dir(
         self,
-        plugin: ProjectPlugin,
+        plugin: PluginRef,
         *joinpaths: StrPath,
         make_dirs: bool = True,
     ) -> Path:
@@ -563,7 +564,7 @@ class Project(Versioned):
         """
         return self.meltano_dir(
             plugin.type,
-            plugin.venv_name,
+            plugin.venv_name if isinstance(plugin, ProjectPlugin) else plugin.name,
             *joinpaths,
             make_dirs=make_dirs,
         )
--- a/src/meltano/core/project.py
+++ b/src/meltano/core/project.py
@@ -32,7 +32,7 @@ from meltano.core.utils import makedirs, sanitize_filename, truthy
 
 if t.TYPE_CHECKING:
     from meltano.core.meltano_file import MeltanoFile as MeltanoFileTypeHint
-    from meltano.core.plugin.base import PluginRef
+    from meltano.core.plugin.project_plugin import ProjectPlugin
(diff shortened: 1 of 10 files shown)


src/meltano/core/project.py
```diff
@@ -563,7 +564,7 @@ def plugin_dir(
         """
         return self.meltano_dir(
             plugin.type,
-            plugin.name,
+            plugin.venv_name if isinstance(plugin, ProjectPlugin) else plugin.name,
```
ReubenFrankel, line 567: Any better alternatives to this?

edgarrmondragon: Hmm, I don't love the fact the directory would depend on the plugin class passed to this method, i.e. whether it's a fully resolved `ProjectPlugin` or a mere `PluginRef`. I searched for references of `Project.plugin_dir` and what types it's passed:

1. `ProjectPlugin`

   https://github.com/meltano/meltano/blob/07764c8f6f51b5174a8b12564a35ba3e34bd6f12/src/meltano/core/plugin_location_remove.py#L188

2. `BasePlugin`, which ultimately magically (via `__getattr__ -> getattr`) refers to `PluginDefinition`, another subclass of `PluginRef`

   https://github.com/meltano/meltano/blob/07764c8f6f51b5174a8b12564a35ba3e34bd6f12/src/meltano/core/plugin/file.py#L69

3. `ProjectPlugin`

   https://github.com/meltano/meltano/blob/07764c8f6f51b5174a8b12564a35ba3e34bd6f12/src/meltano/core/plugin_invoker.py#L171

4. `ProjectPlugin`

   https://github.com/meltano/meltano/blob/07764c8f6f51b5174a8b12564a35ba3e34bd6f12/src/meltano/core/plugin_install_service.py#L419

edgarrmondragon: We could give `PluginRef` a naive implementation of `venv_name` such we don't need to do a type check here:

```python
@property
def venv_name(self):
    self.name
```

That still doesn't solve

> the directory would depend on the plugin class passed to this method

though 🤔

ReubenFrankel: > 2. `BasePlugin`, which ultimately magically (via `__getattr__ -> getattr`) refers to `PluginDefinition`, another subclass of `PluginRef`
>    https://github.com/meltano/meltano/blob/07764c8f6f51b5174a8b12564a35ba3e34bd6f12/src/meltano/core/plugin/file.py#L69

Can we resolve this to a `ProjectPlugin` and type hint `ProjectPlugin` in the `plugin_dir` signature instead?

edgarrmondragon: Yeah, I'd favor that if we can avoid circular imports

ReubenFrankel: 2418b42

Let's see how the tests do...

ReubenFrankel: 99e2314 worked in the end. I guess a file plugin might not necessarily exist in the project in the project at the point `plugin_dir` is called there?

edgarrmondragon:
The name `venv_name` was confusing me. For example, it made me feel unsure about adding a default implementation to `PluginRef`. But thinking about it as a plugin directory _that contains_ the plugin venv changed my mind. Let me know if this makes sense.

ReubenFrankel:
> The name `venv_name` was confusing me. For example, it made me feel unsure about adding a default implementation to `PluginRef`. But thinking about it as a plugin directory _that contains_ the plugin venv changed my mind. Let me know if this makes sense.

Yep, looks good to me!

edgarrmondragon (approved):

src/meltano/core/project.py
edgarrmondragon: Yeah, some file plugins are not permanently added to the project after they're installed. They're only added if they can _update_ their files 🤷‍♂️. I'll be happy if we rip them off the code base 😅.

PR #8785, opened 2024-09-19
fix: Install plugins correctly given multiple levels of inheritance

ReubenFrankel:
The `venv_name` property of a plugin is used to determine if it installed or skipped based on whether or not the venv directory has been "seen".

https://github.com/meltano/meltano/blob/28bc90be783123655e837e23175db4cf64724393/src/meltano/core/plugin_install_service.py#L215

Previously, for an inherited plugin (i.e. defines `inherits_from`) only its immediate parent was taken into account when resolving the venv name. For one level of inheritance this is fine, but at soon as you start inheriting from inherited plugins the venv name is resolved as the parent inherited plugin name, **not the base plugin name**.

The fix is actually super simple - instead of getting the parent  `name`, get its `venv_name` to recursively resolve the base plugin name.

---

Given `meltano.yml` from https://meltano.slack.com/archives/C0695LLCGVC/p1726488890814789:

```yml
version: 1
default_environment: dev
project_id: 805c9ea1-931b-4722-9761-d6074594b175
environments:
- name: dev
- name: staging
- name: prod
plugins:
  extractors:
  - name: tap-oracle
    variant: s7clarke10
    pip_url:
      git+https://github.com/s7clarke10/pipelinewise-tap-oracle.git@53bb75ed27d7796d2f492e74cec87f10f1bce4d4
    config:
      use_ora_rowscn: false
      ora_python_driver_type: thick
      use_singer_decimal: true

  - name: tap-oracle-demo_foo
    inherit_from: tap-oracle
    config:
      user: demo
      host: 123
      port: 420
      service_name: foo

  - name: tap-oracle-demo_foo-full_table
    inherit_from: tap-oracle-demo_foo
    config:
      default_replication_method: FULL_TABLE

  - name: tap-oracle-demo_foo-full_table-bar_baz
    inherit_from: tap-oracle-demo_foo-full_table
    select:
     - bar-baz.*

  - name: tap-oracle-demo_foo-full_table-asd_qwe
    inherit_from: tap-oracle-demo_foo-full_table
    select:
      - asd-qwe.*
```

Before fix:

image

After fix:

image

---

Related to #8783

### DIFF
--- a/src/meltano/core/plugin/project_plugin.py
+++ b/src/meltano/core/plugin/project_plugin.py
@@ -399,7 +399,7 @@ class ProjectPlugin(PluginRef):  # too many attrs and methods
             return self.name
 
         if not self.pip_url or (self.parent.pip_url == self.pip_url):
-            return self.parent.name
+            return self.parent.venv_name
 
         return self.name
 
--- a/tests/meltano/core/plugin/test_plugin.py
+++ b/tests/meltano/core/plugin/test_plugin.py
@@ -591,6 +591,52 @@ class TestProjectPlugin:
         # Plugin doesn't have any utility requirements
         assert not transformer.all_requires[PluginType.UTILITIES]
 
+    def test_venv_name(self):
+        """Validate the virtual environment name.
+
+        - If the plugin has a pip_url, the venv name is the plugin name.
+        - If the plugin is an inherited plugin without a custom pip_url, the venv name
+          is the parent plugin name.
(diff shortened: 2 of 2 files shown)


edgarrmondragon (approved):

edgarrmondragon:
It trips me a bit that tests continue passing and coverage didn't complain about a missing patch 😅, so I'll try to break it and see why we missed this.

edgarrmondragon:
Thanks @ReubenFrankel!

PR #8794, opened 2024-09-23
fix: Taps running in parallel use a different temporary catalog file

### LINKED ISSUE #8763: feature: Catalog extra still puts a properties.json file in .meltano
### Feature scope

### Description

Ultimately, my goal is to be able to run multiple processes with the same tap in parallel. Right now, we hit 
```Cannot start plugin tap-mssql: [WinError 32] The process cannot access the file because it is being used by another process: '.meltano\\run\\tap-mssql\\tap.properties.json'```  when we provide a Catalog extra to a tap which is odd as I'd expect Meltano to point to my provided catalog and not remove/add files.

Logs of current behavior
```
2024-09-12T12:47:15.563708Z [debug    ] Skipped installing extractor 'tap-smoke-test'
2024-09-12T12:47:15.563933Z [debug    ] Skipped installing 1/1 plugins
2024-09-12T12:47:15.713970Z [debug    ] Created configuration at /home/visch/git/meltano-projects/getting_started/.meltano/run/tap-smoke-test/tap.ceb61f67-c362-4a72-86b7-787eef00aa2a.config.json
2024-09-12T12:47:15.714332Z [debug    ] Could not find tap.properties.json in /home/visch/git/meltano-projects/getting_started/.meltano/extractors/tap-smoke-test/tap.properties.json, skipping.
2024-09-12T12:47:15.714573Z [debug    ] Could not find tap.properties.cache_key in /home/visch/git/meltano-projects/getting_started/.meltano/extractors/tap-smoke-test/tap.properties.cache_key, skipping.
2024-09-12T12:47:15.714807Z [debug    ] Could not find state.json in /home/visch/git/meltano-projects/getting_started/.meltano/extractors/tap-smoke-test/state.json, skipping.
2024-09-12T12:47:15.715151Z [debug    ] Cached catalog is outdated, running discovery...
2024-09-12T12:47:15.715464Z [info     ] Found catalog in /home/visch/git/meltano-projects/getting_started/catalog.json
2024-09-12T12:47:15.717292Z [debug    ] Invoking: ['/home/visch/git/meltano-projects/getting_started/.meltano/extractors/tap-smoke-test/venv/bin/tap-smoke-test', '--config', '/home/visch/git/meltano-projects/getting_started/.meltano/run/tap-smoke-test/tap.ceb61f67-c362-4a72-86b7-787eef00aa2a.config.json', '--catalog', '/home/visch/git/meltano-projects/getting_started/.meltano/run/tap-smoke-test/tap.properties.json']
```

It would be great if `--catalog` pointed to `/home/visch/git/meltano-projects/getting_started/catalog.json` instead with the meltano.yml (just the extractor portion below)
```
  - name: tap-smoke-test
    variant: meltano
    pip_url: git+https://github.com/meltano/tap-smoke-test.git
    catalog: catalog.json
    config:
      streams:
      - stream_name: User-pzjs3twtm2y4ip6uvfvy6ukhpa-prod
        input_filename:
          https://gitlab.com/meltano/tap-smoke-test/-/raw/main/demo-data/animals-data.jsonl
      - stream_name: Question-pzjs3twtm2y4ip6uvfvy6ukhpa-prod
        input_filename:
          https://gitlab.com/meltano/tap-smoke-test/-/raw/main/demo-data/pageviews-data.jsonl
``` 

Slack thread: https://meltano.slack.com/archives/C06A1MD6A6L/p1726095527072649?thread_ts=1726077531.449699&cid=C06A1MD6A6L

### PR DESCRIPTION
**Pull request by edgarrmondragon**
## Related Issues

* Closes https://github.com/meltano/meltano/issues/8763

### DIFF
--- a/src/meltano/core/plugin/singer/tap.py
+++ b/src/meltano/core/plugin/singer/tap.py
@@ -233,7 +233,7 @@ class SingerTap(SingerPlugin):
         """Get the configuration files for this tap."""
         return {
             "config": f"tap.{self.instance_uuid}.config.json",
-            "catalog": "tap.properties.json",
+            "catalog": f"tap.{self.instance_uuid}.properties.json",
             "catalog_cache_key": "tap.properties.cache_key",
             "state": "state.json",
         }

### CODE REVIEW

File: src/meltano/core/plugin/singer/tap.py [hunk 1]
```diff
@@ -233,7 +233,7 @@ def config_files(self):  # noqa: ANN201
         """Get the configuration files for this tap."""
         return {
             "config": f"tap.{self.instance_uuid}.config.json",
-            "catalog": "tap.properties.json",
+            "catalog": f"tap.{self.instance_uuid}.properties.json",
             "catalog_cache_key": "tap.properties.cache_key",
             "state": "state.json",
```
dluo-sig, line 238: I was checking up on the other issue and was curious about the fix.  Similar but unrelated, I'm wondering if state.json here also needs a fix?  It seems to be the same state file that is stored in azure when using different [state backend](https://docs.meltano.com/concepts/state_backends/).  However, the difference I've noticed is that on azure, it actually gets split out based on `--state-id-suffix` provided to `meltano run`.  I suppose something similar would happen if running `meltano el` with `--state-id` set.  But locally, it also maintains this state but just in a fixed location.

edgarrmondragon: Yeah, I don't know if we should keep the local `state.json` after the pipeline finishes 🤔.

FWIW I don't think this PR is currently on the right path. I think a more correct approach would be to have a dedicated directory _per run_. That way, plugins could run in parallel without clashing with each other's files:

```
tap-example --config /path/to/run-1/tap.config.json --state /path/to/run-1/state.json --catalog /path/to/run-1/tap.properties.json
tap-example --config /path/to/run-2/tap.config.json --state /path/to/run-2/state.json --catalog /path/to/run-2/tap.properties.json
```

During the preparation phase, we'd

- dump resolved config to `/path/to/run-1/tap.config.json`
- get `/path/to/run-1/state.json` from the state backend
- copy `/path/to/run-1/tap.properties.json` from cache, or regenerate it

dluo-sig: Agreed, I did mention that as a possibility as well, but I'm not familiar with what's involved to do all of that.  The idea is that it should use the run uuid to create a unique folder to use.  You theoretically would get the same behavior if all of the files had the uuid as well, just a matter of what the desired folder structure is.  I believe the local state does get removed once the run is done -- I just wasn't sure if the local state is needed if it's kept in a different backend anyway, but I guess it does need to be there in order to upload?  In that case, maybe the folder uuid is still ultimately better because that way, the state file wouldn't need to be renamed.  This does seem to get complicated, though, because you would want to reuse the same state somehow across runs. 

Maybe it should read the state from whatever the state backend is, regardless of whether that is from cloud or local, in a central location without the uuid.  Then, the execution will create the state in a uuid folder, and finally update the state backend after completion?  I'm also not sure how state gets merged if there are multiple processes that end up writing to the same file in the state backend that started in parallel.  For us, we maintain a separate state file per table using `--state-id-suffix`, so it's not really an issue.

Or I might just be overthinking this whole thing.

edgarrmondragon: Yeah, though the failing tests point to an assumption I'm violating here 🤔

s7clarke10:
I believe we need to handle the `meltano run` command with a `--state-id-suffix` or use a `meltano el` command with a `--state-id` to provide uniqueness regardless of the command which is issued.

I am not sure whether the targeted changes are just for the `meltano run` command or whether it will help with the `meltano el` commands as well.

PR #8800, opened 2024-09-25
fix: Failure message is now correctly printed when a plugin can not be removed from the system database

edgarrmondragon:
## Description

The detailed error message was missing when a plugin couldn't be removed from the db.

## Related Issues

* Closes #XXXX

### DIFF
--- a/src/meltano/cli/remove.py
+++ b/src/meltano/cli/remove.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import typing as t
+
 import click
 
 from meltano.cli.params import pass_project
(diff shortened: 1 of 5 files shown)


src/meltano/cli/remove.py
```diff
@@ -63,7 +72,7 @@ def removal_manager_status_update(removal_manager: PluginLocationRemoveManager)
     plugin_descriptor = removal_manager.plugin_descriptor
     location = removal_manager.location
     if removal_manager.plugin_error:
-        message = removal_manager.remove_message
+        message = removal_manager.message
```
edgarrmondragon, line 75: The `remove_message` attribute was set to `None` at initialization time, but never used again.

src/meltano/core/plugin_location_remove.py
```diff
@@ -43,10 +44,10 @@ def __init__(self, plugin: ProjectPlugin, location):  # noqa: ANN001
         self.plugin_descriptor = f"{plugin.type.descriptor} '{plugin.name}'"
         self.location = location
         self.remove_status: PluginLocationRemoveStatus | None = None
-        self.remove_message = None
+        self.message: str | None = None
```
edgarrmondragon, line 47: Ditto, see above.

edgarrmondragon (commented):
See comments for the functional change. The rest is typing and test updates to validate the patch.

src/meltano/core/project_plugins_service.py
```diff
@@ -177,7 +177,7 @@ def remove_from_file(self, plugin: ProjectPlugin):  # noqa: ANN201
             The removed plugin.
         """
         # Will raise if the plugin isn't actually in the file
-        self.get_plugin(plugin)
+        self.get_plugin(plugin, ensure_parent=False)
```
edgarrmondragon, line 180: We were failing to remove the plugin from `meltano.yml` is the lockfile isn't present.

WillDaSilva (approved):

PR #8810, opened 2024-09-27
fix: Tap stream fields are now automatically selected when any subfield is selected

### LINKED ISSUE #8807: bug: Selecting all nested properties with `my_stream.prop.*` should imply selecting the top-level property `my_stream.prop`
### Meltano Version

3.5.4

### Python Version

3.12

### Bug scope

### Operating System

NA

### Description

Configuring `select` with:

```yaml
select:
- my_stream.prop.*
```

selects all nested sub-properties of `my_stream.prop`, but fails to select the parent property itself. This catalog output causes the SDK schema pruning algorithm to drop `prop` and all of its nested properties from the schema. The recommended workaround is to explicitly select `prop`:

```yaml
select:
- my_stream.prop
- my_stream.prop.*
```

Ideally users would not need to do this and it's arguably bad UX that they need to.

One way we might be able to address this is by generating `MetadataRule` instances for these special cases. We already generate two rules for a single pattern in some cases:

https://github.com/meltano/meltano/blob/3b7a3f900644636ea44df3d5213aa2abe91d1859/src/meltano/core/plugin/singer/catalog.py#L188-L208

For example, the pattern `my_stream.prop.*` generates the rules

```python
MetadataRule("my_stream", [], "selected", value=True)
MetadataRule("my_stream", ["properties", "prop", "properties", "*"], "selected", value=True)
```

but we could patch so it also generates

```python
MetadataRule("my_stream", ["properties", "prop"], "selected", value=True)
```

Related:

- https://github.com/meltano/meltano/issues/3024
- https://github.com/meltano/sdk/issues/2687
- https://gitlab.com/meltano/sdk/-/issues/285#note_750983833

### Code

_No response_

### Link to Slack/Linen

_No response_

edgarrmondragon:
## Related Issues

* Closes #8807

## Summary by Sourcery

Enable implicit selection of parent fields when subfields are selected in the singer catalog selector and adjust tests to reflect this behavior

New Features:
- Automatically select parent properties when any sub-property is selected

Enhancements:
- Remove xfail markers related to parent property auto-selection in metadata rule tests

Tests:
- Add parameterized tests for wildcard and nested field selection to verify parent auto-selection
- Update existing catalog tests to include newly auto-selected parent fields


src/meltano/core/plugin/singer/catalog.py
```diff
@@ -87,38 +82,34 @@ def match(self, tap_stream_id: str, breadcrumb: list[str] | None = None) -> bool
...
-        key: str,
-        *,
-        value: bool,
-        negated: bool = False,
-    ):
-        """Create a metadata rule for a stream and property."""
-        super().__init__(tap_stream_id, breadcrumb, negated=negated)
-        self.key = key
-        self.value = value
+@dataclasses.dataclass
```
edgarrmondragon, line 85: Dataclasses are slightly faster:

```python
import timeit

code = """
rules = [
    MetadataRule(
        "my_stream",
        [],
        "selected",
        value=True,
    ),
    MetadataRule(
        "my_stream",
        ["properties", "prop", "properties", "*"],
        "selected",
        value=True,
    ),
    MetadataRule(
        "my_stream",
        ["properties", "prop"],
        "selected",
        value=True,
    ),
]
"""

setup = """
from meltano.core.plugin.singer.catalog import MetadataRule
"""

print(timeit.timeit(stmt=code, setup=setup, number=1000000))

# 0.6132017500021902 (normal class)
# 0.3916304169979412 (dataclass)
```

WillDaSilva: Using dataclasses makes the intent of the code more clear too, and less error prone

src/meltano/core/plugin/singer/catalog.py
```diff
@@ -205,6 +205,19 @@ def select_metadata_rules(patterns: Iterable[str]) -> list[MetadataRule]:
                 ),
             )
 
+            # If all sub-properties are selected, the parent property is selected too
+            if selected and "*" in props:
+                rules.extend(
+                    MetadataRule(
+                        tap_stream_id=pattern.stream_pattern,
```
edgarrmondragon:
```suggestion
                        tap_stream_id=parsed_pattern.stream_pattern,
```

tayloramurphy:
@edgarrmondragon didn't realize you had this PR open. I assume you prefer your approach here? What should I do with my PR? Admittedly I like some of the tests - I don't care too much about the implementation though 😄

edgarrmondragon:
@sourcery-ai review

edgarrmondragon:
> @edgarrmondragon didn't realize you had this PR open. I assume you prefer your approach here? What should I do with my PR? Admittedly I like some of the tests - I don't care too much about the implementation though 😄

Let me see if I catch anything from here that would make sense to incorporate into #9434 and get that one merged.

PR #8837, opened 2024-10-16
chore: Use the `packaging` scope for Dependabot updates of runtime dependencies so they appear in the changelog and release notes

edgarrmondragon:
## Description

Then we can highlight certain aspects of those updated dependencies.

## Related Issues

* Closes #XXXX

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -17,6 +17,7 @@ types:
   - docs
   - feat
   - fix
+  - packaging
   - perf
   - refactor
   - revert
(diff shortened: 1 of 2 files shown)


.github/semantic.yml
WillDaSilva, line 1: Can't we already list the dependency updates in the release notes and changelog by using the `deps` PR scope? I'm not sure what this new `packaging` type adds.

edgarrmondragon: Our changelog generator doesn't pull in the `scope` and only uses the type, but we probably don't want to generate changelog entries for `chore` PRs.

WillDaSilva (approved):

PR #8367, opened 2024-01-19
feat: Support state backend and setting add-ons

edgarrmondragon:
Packaging example from Meltano's builtin state backends:

```toml
[project.entry-points."meltano.settings"]
# AWS S3 state backend settings
s3_aws_access_key_id = "meltano.core.state_store.s3:AWS_ACCESS_KEY_ID"
s3_aws_secret_access_key = "meltano.core.state_store.s3:AWS_SECRET_ACCESS_KEY"
s3_endpoint_url = "meltano.core.state_store.s3:ENDPOINT_URL"```

[project.entry-points."meltano.state_backends"]
# These keys should match the expected scheme for URIs of
# the given type. E.g., filesystem state backends have a
# file://<path>/<to>/<state directory> URI
s3 = "meltano.core.state_store.s3:S3StateStoreManager"
```

## Documentation

- [Custom State Backends](https://deploy-preview-8367--meltano.netlify.app/guide/custom-state-backend/)

## Related

* https://github.com/meltano/meltano/discussions/8276
* https://github.com/meltano/meltano/issues/2414
* https://github.com/edgarrmondragon/meltano-rocksdb-state-backend/
* https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/

PR #8851, opened 2024-10-22
refactor: Decoupled the `JobState` ORM model of the system db from state store implementations

edgarrmondragon:
## Description

State backends other than `systemdb` do not need to know about the inner workings of the model used to describe job state in the system database.

Also, this shave off some import time if and when we go with #8367.

## Related Issues

* #6911 
* #6220


src/meltano/core/job/job.py
```diff
@@ -115,7 +115,7 @@ class Job(SystemModel):
     __tablename__ = "runs"
 
     id: Mapped[IntPK]
-    job_name: Mapped[t.Optional[str]]  # noqa: UP007
+    job_name: Mapped[str]
```
edgarrmondragon, line 118: https://github.com/meltano/meltano/blob/49aa5d626b8991e6abb09cd666a68c965eec6c0f/src/meltano/migrations/versions/b4c05e463b53_.py#L48

WillDaSilva (approved):

PR #8856, opened 2024-10-25
ci: Use Postgres 13 in CI

edgarrmondragon:
## Related Issues

* Closes #XXXX


.github/workflows/test.yml
```diff
@@ -140,7 +140,7 @@ jobs:
         --health-interval 10s
         --health-timeout 5s
         --health-retries 5
-        postgres:11
+        postgres:17
```
WillDaSilva, line 143: Should we updated our recommended minimum Postgres version in our docs now that we aren't testing if Meltano works with older versions? We may unintentionally break compatibility with older versions after we merge this change.

edgarrmondragon: Ah that makes sense, and we probably shouldn't go all the way to 17 in that case.

I can't find a place in the docs where we mention a recommended a minimum Postgres version (looking at https://docs.meltano.com/guide/installation-guide/#requirements). We should add it if doesn't actually exist.

---

By looking at https://www.postgresql.org/support/versioning/, it seems 12 through 17 are currently supported, but 12 is about to reach EOL (Nov 14th 2024).

WillDaSilva: In that case updating our docs to say we support the oldest non-EOL Postgres version, and then jumping to Postgres 13 seems fine. Then we'll merge this on November 14th to keep that claim true. Does that seem reasonable @edgarrmondragon?

I wonder if said policy would also apply to the other databases we interact with.

edgarrmondragon: > In that case updating our docs to say we support the oldest non-EOL Postgres version, and then jumping to Postgres 13 seems fine. Then we'll merge this on November 14th to keep that claim true. Does that seem reasonable @edgarrmondragon?

💯

> I wonder if said policy would also apply to the other databases we interact with.

I think so. We are currently testing with MSSQL 2019, but that won't reach EOL until Jan 2025[^1], so we can mention 2019 at the moment.

SQLite is harder to control, we'd need to build the version we want and install it system-wide so Python picks it up. I'm sure macOS and specially Windows are even more complicated than that. So for that I'd be content with mentioning we require `>=3.25.1`.

[^1]: https://endoflife.date/mssqlserver

WillDaSilva (approved):

PR #8898, opened 2024-11-14
refactor: Use `importlib.resources` to read package files and use `importlib.metadata` to detect "editable" installations (via PEP 610)

### LINKED ISSUE #7474: Replace use of `Path(__file__)` with `importlib_resources` to support Meltano zipapp
https://docs.python.org/3/library/importlib.resources.html

> This module leverages Python’s import system to provide access to resources within packages. If you can import a package, you can access resources within that package. Resources can be opened or read, in either binary or text mode.
> 
> Resources are roughly akin to files inside directories, though it’s important to keep in mind that this is just a metaphor. Resources and packages do not have to exist as physical files and directories on the file system: for example, a package and its resources can be imported from a zip file using [zipimport](https://docs.python.org/3/library/zipimport.html#module-zipimport).

Currently Meltano makes use of `pathlib.Path(__file__)` in 3 locations under `src/meltano`. These paths are incorrect when Meltano is running as a zipapp. To access the contents the paths would normally refer to in all cases, `importlib.resources` must be used. To support older versions of Python, the official backport `importlib_resources` should be used instead: https://github.com/python/importlib_resources

This backport is already a dependency of Meltano, so all that remains is to make us of it.

CC @aaronsteers because of your interest in zipapps. From what I've heard this is a moderately widespread issue with them, but at least we can do our part to make sure Meltano runs properly as one.

edgarrmondragon:
## Description

- [PEP 610 at packaging.python.org](https://packaging.python.org/en/latest/specifications/direct-url/#direct-url)

## Related Issues

* Closes https://github.com/meltano/meltano/issues/7474

### DIFF
--- a/src/meltano/core/tracking/contexts/environment.py
+++ b/src/meltano/core/tracking/contexts/environment.py
@@ -10,7 +10,7 @@ from collections import defaultdict
 from contextlib import suppress
 from datetime import datetime, timezone
 from functools import cached_property
-from pathlib import Path
+from importlib import resources
 from warnings import warn
 
 import psutil
@@ -27,7 +27,7 @@ if t.TYPE_CHECKING:
 logger = get_logger(__name__)
 
 # This file is only ever created in CI when building a release
-release_marker_path = Path(__file__).parent / ".release_marker"
+release_marker_path = resources.files("meltano.core.tracking") / ".release_marker"
 
 
 def _get_parent_context_uuid_str() -> str | None:
@@ -66,14 +66,14 @@ class EnvironmentContext(SelfDescribingJson):
     }
 
     @classmethod
-    def _notable_hashed_env_vars(cls) -> Iterable[str]:
+    def _notable_hashed_env_vars(cls) -> Iterable[tuple[str, str]]:
         for env_var_name in cls.notable_hashed_env_vars:
             with suppress(KeyError):  # Skip unset env vars
                 env_var_value = os.environ[env_var_name]
                 yield env_var_name, hash_sha256(env_var_value)
 
     @classmethod
-    def _notable_flag_env_vars(cls) -> Iterable[str]:
+    def _notable_flag_env_vars(cls) -> Iterable[tuple[str, bool | None]]:
         for env_var_name in cls.notable_flag_env_vars:
             with suppress(KeyError):  # Skip unset env vars
                 env_var_value = os.environ[env_var_name]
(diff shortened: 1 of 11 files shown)


src/meltano/core/tracking/contexts/environment.py
```diff
@@ -27,7 +27,7 @@
 logger = get_logger(__name__)
 
 # This file is only ever created in CI when building a release
-release_marker_path = Path(__file__).parent / ".release_marker"
+release_marker_path = resources.files("meltano.core.tracking") / ".release_marker"
```
edgarrmondragon, line 30: I think this is fixing a bug in our tracking.

The old line seems to indicate that we were looking for `.release_marker` in `meltano/core/tracking/contexts`:

```python
>>> from pathlib import Path
>>> from meltano.core.tracking.contexts import __file__
>>> Path(__file__).parent
PosixPath('/Users/edgarramirez/Arch/meltano/src/meltano/core/tracking/contexts')
```

but at the same time, the build workflow puts it in `meltano/core/tracking`

https://github.com/meltano/meltano/blob/3db048d297c754ab24819b2bcc6e3286ffa44812/.github/workflows/build.yml#L52-L55

src/meltano/core/upgrade_service.py
```diff
@@ -23,6 +25,48 @@
...
+        url: str | None = pep610_data.get("url")
+        dir_info: dict[str, t.Any] = pep610_data.get("dir_info", {})
+        if (  # pragma: no branch
+            url and dir_info and dir_info.get("editable", False) and not force
+        ):
+            meltano_dir = url.removeprefix("file://")
+            raise AutomaticPackageUpgradeError(
+                reason="it is installed from source",
+                instructions=f"navigate to `{meltano_dir}` and run `git pull`",
+            )
```
WillDaSilva, line 47: I find it easier to follow when we exit early in response to cases such as `_get_pep610_data()` being `None`

```suggestion
    if (pep610_data := _get_pep610_data()) is None:
        return None
    url: str | None = pep610_data.get("url")
    dir_info: dict[str, t.Any] = pep610_data.get("dir_info", {})
    if (  # pragma: no branch
        url and dir_info and dir_info.get("editable", False) and not force
    ):
        meltano_dir = url.removeprefix("file://")
        raise AutomaticPackageUpgradeError(
            reason="it is installed from source",
            instructions=f"navigate to `{meltano_dir}` and run `git pull`",
        )
```

src/meltano/core/upgrade_service.py
```diff
@@ -23,6 +25,48 @@
...
+            instructions=(
+                "pull the latest Docker image using "
+                "`docker pull meltano/meltano` and recreate any containers "
+                "you may have created"
+            ),
+        )
+
+
+def _check_in_nox_session() -> None:
+    if os.getenv("NOX_CURRENT_SESSION") == "tests":  # pragma: no branch
```
WillDaSilva, line 63: Do we care if the value is `"tests"`, or should we raise this exception regardless of which Nox session we're within?

WillDaSilva (approved):

PR #8913, opened 2024-11-19
feat(cli): Include the Python version and machine type in system debug log message

edgarrmondragon:
## Description

Example:

```
2024-11-19T20:37:50.726768Z [debug    ] Meltano 3.6.0b3, Python 3.13.0, Darwin (arm64)
```

### DIFF
--- a/src/meltano/cli/cli.py
+++ b/src/meltano/cli/cli.py
@@ -113,10 +113,11 @@ def cli(
         project = Project.find()
         setup_logging(project)
         logger.debug(
-            "Meltano %s, Python %s, %s",
+            "Meltano %s, Python %s, %s (%s)",
             __version__,
             platform.python_version(),
             platform.system(),
+            platform.machine(),
         )
         if project.readonly:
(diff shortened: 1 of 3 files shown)


src/meltano/cli/cli.py
```diff
@@ -112,7 +112,12 @@ def cli(
     try:
         project = Project.find()
         setup_logging(project)
-        logger.debug("meltano %s, %s", __version__, platform.system())
+        logger.debug(
+            "meltano %s, Python %s, %s",
```
WillDaSilva, line 116:
```suggestion
            "Meltano %s, Python %s, %s",
```

WillDaSilva, line 116: We may also want to include `platform.machine()` (e.g. `arm64`)

WillDaSilva (approved):

src/meltano/cli/cli.py
edgarrmondragon: 14cf621abcf919b479ca889c9055a756c042942d

PR #8937, opened 2024-11-26
docs: Use tabs for code blocks with OS-specific instructions in Getting Started guide

### LINKED ISSUE #7802: docs: Tabs for code blocks with OS-specific instructions in Getting Started
### Feature scope

Other

### Description

In the Getting Started guide, there are code blocks like the following which contain commented commands for use by different OSes.

```bash
#For Windows (PowerShell): New-Alias Python3 Python
python3 -m pip install --user pipx
python3 -m pipx ensurepath
#For Windows (PowerShell): Open up a new powershell instance to load your new path variables
source ~/.bashrc
```

In these cases, it would be much nicer to have code blocks with separate tabs for each OS, like this:
Screenshot 2023-06-14 at 12 08 49 PM

This would be a much nicer user experience.

edgarrmondragon:
## Description

Use tabs to group OS-specific instructions as suggested in #7802, though I'm tempted to already switch the instructions to use `uv`.

## Related Issues

* Closes #7802

### DIFF
--- a/docs/docs/getting-started/index.mdx
+++ b/docs/docs/getting-started/index.mdx
@@ -5,6 +5,9 @@ layout: getting_started
 sidebar_position: 5
 ---
 
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
 Welcome! If you're ready to get started with Meltano and [run an EL[T] pipeline](#run-a-data-integration-el-pipeline)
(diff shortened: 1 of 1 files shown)


WillDaSilva (approved):
> I'm tempted to already switch the instructions to use `uv`.

Might as well save that for a follow-up, since this seems good to go as-is.

PR #8949, opened 2024-12-03
fix: Add newline at the end of plugin lock files

edgarrmondragon:
## Related Issues

NA

### DIFF
--- a/src/meltano/core/plugin_lock_service.py
+++ b/src/meltano/core/plugin_lock_service.py
@@ -61,7 +61,7 @@ class PluginLock:
         locked_def = StandalonePlugin.from_variant(self.variant, self.definition)
 
         with self.path.open("w") as lockfile:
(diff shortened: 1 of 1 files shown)


src/meltano/core/plugin_lock_service.py
```diff
@@ -61,7 +61,7 @@ def save(self) -> None:
         locked_def = StandalonePlugin.from_variant(self.variant, self.definition)
 
         with self.path.open("w") as lockfile:
-            json.dump(locked_def.canonical(), lockfile, indent=2)
+            lockfile.write(json.dumps(locked_def.canonical(), indent=2) + "\n")
```
WillDaSilva, line 64: This might have slightly better performance

```suggestion
            json.dump(locked_def.canonical(), lockfile, indent=2)
            lockfile.write("\n")
```

But how it was before is fine too.

WillDaSilva (approved):

Issue #6715, opened 2022-09-07
Incremental addoption of static typing


edgarrmondragon:
This issue will track our progress as we add or fix type annotations to the codebase.

Static type checks were recently added in CI (#6707), but most submodules are currently ignored.

Modules which need updates to pass `mypy` static type checking:

```[tasklist]
### Tasks
- [x] https://github.com/meltano/meltano/pull/6708
- [x] https://github.com/meltano/meltano/pull/6831
- [x] https://github.com/meltano/meltano/pull/8890
- [ ] `meltano.cli.*`
- [x] `meltano.core.connection_service`
- [x] `meltano.core.db`: https://github.com/meltano/meltano/pull/8902
- [ ] `meltano.core.elt_context`
- [x] `meltano.core.extract_utils`
- [x] `meltano.core.extractor`
- [x] `meltano.core.meltano_invoker`
- [x] `meltano.core.migration_service`: https://github.com/meltano/meltano/pull/8902
- [ ] `meltano.core.models`
- [x] `meltano.core.plugin_discovery_service`
- [x] `meltano.core.plugin_install_service`: https://github.com/meltano/meltano/pull/8718
- [ ] `meltano.core.plugin_invoker`
- [x] `meltano.core.plugin_test_service`
- [x] `meltano.core.project`
- [x] `meltano.core.project_add_service`: https://github.com/meltano/meltano/pull/8902
- [ ] `meltano.core.project_files`
- [x] `meltano.core.project_plugins_service`: https://github.com/meltano/meltano/pull/8902
- [ ] `meltano.core.project_settings_service`
- [x] `meltano.core.select_service`: https://github.com/meltano/meltano/pull/8720
- [x] `meltano.core.settings_service`
- [x] `meltano.core.settings_store`: https://github.com/meltano/meltano/pull/9108
- [ ] `meltano.core.sqlalchemy`
- [ ] `meltano.core.task_sets_service`
- [ ] `meltano.core.transform_add_service`
- [ ] `meltano.core.behavior.*`
- [x] `meltano.core.block.block_parser`: https://github.com/meltano/meltano/pull/9101
- [ ] `meltano.core.block.blockset`: https://github.com/meltano/meltano/pull/9110
- [ ] `meltano.core.block.extract_load`: https://github.com/meltano/meltano/pull/9110
- [ ] `meltano.core.block.future_utils`: https://github.com/meltano/meltano/pull/9110
[...]

PR #8962, opened 2024-12-13
chore: More strict typing

edgarrmondragon:
## Description

See PR comments, got these config changes as suggestions from running `pipx run 'sp-repo-review[cli]'`.

## Related Issues

* #6715

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -187,6 +187,7 @@ source = [
 [tool.coverage.report]
 exclude_also = [
   '''if (t\.)?TYPE_CHECKING:''',
+  '''(t\.)?assert_never''',
   "\\.\\.\\.",
 ]
 precision = 2
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,7 +121,18 @@ types-requests = "^2.31.0"
 meltano = "meltano.cli:main"
 
 [tool.pytest.ini_options]
-addopts = "--cov=meltano --cov=tests --doctest-modules --durations=10 --order-scope=module -ra -n auto --dist=loadfile"
+addopts = [
+  "--cov=meltano",
+  "--cov=tests",
+  "--doctest-modules",
+  "--durations=10",
+  "--order-scope=module",
+  "-ra",
+  "-n=auto",
+  "--dist=loadfile",
+  "--strict-config",
+  "--strict-markers",
+]
 asyncio_mode = "auto"
 asyncio_default_fixture_loop_scope = "class"
 filterwarnings = [
@@ -143,9 +154,11 @@ markers = [
   "concurrent: test requires true concurrency",
   "slow: slow test",
 ]
+minversion = "8"
 testpaths = [
   "tests/",
 ]
+xfail_strict = true
 
 [tool.coverage.run]
 branch = true
@@ -194,6 +207,7 @@ version_files = [
 ]
 
 [tool.mypy]
+enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
 incremental = false # Disabled until https://github.com/python/mypy/issues/12664 is resolved
 files = [
   "src/meltano/",
@@ -201,6 +215,7 @@ files = [
 exclude = [
   "src/meltano/migrations/",
 ]
+warn_unreachable = true
 warn_unused_configs = true
 warn_unused_ignores = true
 
@@ -256,10 +271,7 @@ target-version = "py39"
 
 [tool.ruff.lint]
 ignore = [
-  "ANN101", # Missing type annotation for `self` in method
-  "ANN102", # Missing type annotation for `cls` in class method
   "N818",   # Allow Exceptions to have suffix 'Exception' rather than 'Error'
-  "PT004",  # Add leading underscore to fixtures that do not return anything
   "UP026",  # Replace `mock` import with `unittest.mock` - remove once Python 3.7 support is dropped
   "S310",   # Allow `urllib.open`
   "S603",   # Allow `subprocess.run(..., shell=False)`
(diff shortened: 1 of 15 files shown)


src/meltano/core/schedule_service.py
```diff
@@ -198,7 +198,7 @@ def default_start_date(self, session, extractor: str) -> datetime:  # noqa: ANN0
 
         # TODO: this coercion should be handled by the `kind` attribute
         # on the actual setting
-        if isinstance(start_date, date):
+        if isinstance(start_date, date) and not isinstance(start_date, datetime):
```
edgarrmondragon, line 201: turns out `datetime` is a subclass of `date` so the `if isinstance(start_date, datetime)` branch below was unreachable.

src/meltano/core/plugin/singer/tap.py
```diff
@@ -685,7 +685,7 @@ def warn_property_not_found(
         stream_dict = {
             stream.get("tap_stream_id"): stream
             for stream in catalog.get("streams", [])
-            if isinstance(stream, dict)
+            if isinstance(stream, dict)  # type: ignore[redundant-expr]
```
edgarrmondragon, line 688: The types assume a spec-compliant catalog, but we still bad skip entries here.

pyproject.toml
```diff
@@ -299,7 +311,7 @@ select = [
   "SIM",  # flake8-simplify
   "T10",  # flake8-debugger
   "T20",  # flake8-print
-  "TCH",  # flake8-type-checking
+  "TC",   # flake8-type-checking
```
edgarrmondragon, line 314: `TCH` was renamed to `TC` in Ruff [0.8.0](https://github.com/astral-sh/ruff/releases/tag/0.8.0).

pyproject.toml
```diff
@@ -256,10 +271,7 @@ target-version = "py39"
 
 [tool.ruff.lint]
 ignore = [
-  "ANN101", # Missing type annotation for `self` in method
-  "ANN102", # Missing type annotation for `cls` in class method
```
edgarrmondragon, line 260: These were removed in Ruff [0.8.0](https://github.com/astral-sh/ruff/releases/tag/0.8.0).

pyproject.toml
```diff
@@ -256,10 +271,7 @@ target-version = "py39"
...
   "N818",   # Allow Exceptions to have suffix 'Exception' rather than 'Error'
-  "PT004",  # Add leading underscore to fixtures that do not return anything
```
edgarrmondragon, line 262: Removed in Ruff [0.8.0](https://github.com/astral-sh/ruff/releases/tag/0.8.0).

pyproject.toml
```diff
@@ -194,13 +207,15 @@ version_files = [
...
 [tool.mypy]
+enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
 incremental = false # Disabled until https://github.com/python/mypy/issues/12664 is resolved
 files = [
   "src/meltano/",
 ]
 exclude = [
   "src/meltano/migrations/",
 ]
+warn_unreachable = true
```
edgarrmondragon, line 218: Warn us about unreachable (by types) pieces of code.

pyproject.toml
```diff
@@ -194,13 +207,15 @@ version_files = [
 ]
 
 [tool.mypy]
+enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
```
edgarrmondragon, line 210: Respectively:

- Force us to use codes, e.g. `type: ignore[<code here>]`
- e.g. checking truthiness of a value that is always `True`
- i.e. checking truthiness of an object that doesn't implement `__bool__` and thus could always be `True`

src/meltano/cli/params.py
```diff
@@ -53,12 +53,12 @@ def _install_plugins_fn(
     return install_plugins
 
 
-async def _install_plugins_and_exit(*args, **kwargs) -> bool:  # noqa: ANN002, ANN003
+async def _install_plugins_and_exit(*args: t.Any, **kwargs: t.Any) -> bool:
```
WillDaSilva, line 56:
```suggestion
async def _install_plugins_and_exit(*args: t.Any, **kwargs: t.Any) -> t.NoReturn:
```

src/meltano/cli/params.py
```diff
@@ -53,12 +53,12 @@ def _install_plugins_fn(
...
     kwargs.pop("reason", None)
     await install_plugins(*args, **kwargs, reason=PluginInstallReason.INSTALL)
     context = click.get_current_context()
     context.exit(code=0)
-    return True  # pragma: no cover
+    return True  # type: ignore[unreachable]  # pragma: no cover
```
WillDaSilva, line 61:
```suggestion
```

src/meltano/cli/select.py
```diff
@@ -35,7 +35,7 @@ def selection_color(selection: SelectionType) -> str | None:
     if selection is SelectionType.UNSUPPORTED:
         return "black"
 
-    return None  # pragma: no cover
+    return None  # type: ignore[unreachable]  # pragma: no cover
```
WillDaSilva, line 38: Same as above, if the type checker can't reach this, then I think we can remove the return statement, and update the return signature to only be `str`. Then, if we ever add a new `SelectionType`, we'll get a type error because there'll be a way to return `None` from a function that is only supposed to return `str`.

src/meltano/core/schedule_service.py
WillDaSilva: > turns out `datetime` is a subclass of `date`

I feel like there are almost no beneficial uses of that property, whereas bugs resulting from it are easy to imagine 🫤

WillDaSilva (approved):

src/meltano/cli/params.py
edgarrmondragon: dc43bd565

src/meltano/cli/select.py
edgarrmondragon: 4223e579f

PR #8973, opened 2024-12-19
fix: Print the time when the job will go stale if no more heartbeats are sent

### LINKED ISSUE #3363: Print the last heartbeat timestamp if an EL job is running
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3441

Originally created by @aaronsteers on 2022-04-26 17:39:50

---

It's hard for users to debug failed/blocked pipelines and recently several users have asked if `--force` is the right thing to do. Our current behavior is that any job without a heartbeat for 5+ minutes will be auto-flagged as failed and should not block further executions. But when the failure occurs, it is not clear to the user that they just need to wait 3 more minutes or that (as was the case [here](https://gitlab.com/meltano/meltano/-/issues/3438#note_924786922)) a process might actually still be running somewhere and continually sending heartbeat updates.

It seems this could be helped a lot by printing out the latest heartbeat info when a job is blocked because we think the process may still be running.

edgarrmondragon:
## Related Issues

* Closes https://github.com/meltano/meltano/issues/3363

### DIFF
--- a/src/meltano/core/block/extract_load.py
+++ b/src/meltano/core/block/extract_load.py
@@ -535,11 +535,14 @@ class ExtractLoadBlocks(BlockSet):
                 self.context.session,
             )
         ):
-            raise RunnerError(
-                f"Another '{job.job_name}' pipeline is already running "  # noqa: EM102
-                f"which started at {existing.started_at}. To ignore this "
-                "check use the '--force' option.",
+            msg = (
+                f"Another '{job.job_name}' pipeline is already running "
+                f"which started at {existing.started_at}. "
+                f"Wait until {existing.valid_intil()} when the job will be "
+                "automatically cancelled if stale, or ignore this check using the "
+                "'--force' option."
             )
+            raise RunnerError(msg)
 
         with closing(self.context.session) as session:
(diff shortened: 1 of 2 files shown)


src/meltano/core/block/extract_load.py
```diff
@@ -535,11 +535,14 @@ async def run_with_job(self) -> None:
...
             )
         ):
-            raise RunnerError(
-                f"Another '{job.job_name}' pipeline is already running "  # noqa: EM102
-                f"which started at {existing.started_at}. To ignore this "
-                "check use the '--force' option.",
+            msg = (
+                f"Another '{job.job_name}' pipeline is already running "
+                f"which started at {existing.started_at}. "
+                f"Wait until {existing.valid_intil()} when the job will be "
```
WillDaSilva, line 541: Will this print it in the form `datetime.datetime(...)`? We should probably call a method on `existing.valid_intil()` to convert it to a string formatted in the way we want. Maybe `existing.valid_intil().isoformat()` is suitable here.

WillDaSilva (approved):

PR #8881, opened 2024-11-08
feat: New projects now contain a Singer SDK logging configuration that works better with Meltano

edgarrmondragon:
## Description

Without this, the SDK plugin's output looks like this

```console
$ meltano run tap-getpocket target-jsonl
2024-11-08T17:14:56.735348Z [info     ] Environment 'dev' is active   
2024-11-08T17:14:57.235081Z [info     ] 2024-11-08 11:14:57,234 | INFO     | tap-getpocket.items  | Beginning incremental sync of 'items'... cmd_type=elb consumer=False job_name=dev:tap-getpocket-to-target-jsonl name=tap-getpocket producer=True run_id=fbf57833-88ab-46d5-98f2-1f9346c9ab8b stdio=stderr string_id=tap-getpocket
2024-11-08T17:14:57.235507Z [info     ] 2024-11-08 11:14:57,234 | INFO     | tap-getpocket.items  | Tap has custom mapper. Using 1 provided map(s). cmd_type=elb consumer=False job_name=dev:tap-getpocket-to-target-jsonl name=tap-getpocket producer=True run_id=fbf57833-88ab-46d5-98f2-1f9346c9ab8b stdio=stderr string_id=tap-getpocket
2024-11-08T17:14:57.728631Z [info     ] 2024-11-08 11:14:57,727 | INFO     | singer_sdk.metrics   | METRIC: {"type": "timer", "metric": "http_request_duration", "value": 0.489063, "tags": {"stream": "items", "endpoint": "/v3/get", "http_status_code": 200, "status": "succeeded"}} cmd_type=elb consumer=False job_name=dev:tap-getpocket-to-target-jsonl name=tap-getpocket producer=True run_id=fbf57833-88ab-46d5-98f2-1f9346c9ab8b stdio=stderr string_id=tap-getpocket
2024-11-08T17:14:57.756138Z [info     ] 2024-11-08 11:14:57,755 | INFO     | singer_sdk.helpers.jsonpath | JSONPath $.list.* match count: 0 cmd_type=elb consumer=False job_name=dev:tap-getpocket-to-target-jsonl name=tap-getpocket producer=True run_id=fbf57833-88ab-46d5-98f2-1f9346c9ab8b stdio=stderr string_id=tap-getpocket
[...]

PR #9077, opened 2025-02-07
feat: Pass a default logging configuration for Pipelinewise and Singer SDK extractors, loaders and mappers

edgarrmondragon:
## Description

This will inject the [Singer SDK `SINGER_SDK_LOG_CONFIG`](https://sdk.meltano.com/en/v0.44.3/implementation/logging.html) and the [pipelinewise-singer-python `LOGGING_CONF_FILE`](https://github.com/transferwise/pipelinewise-singer-python/blob/da64a10cdbcad48ab373d4dab3d9e6dd6f58556b/singer/logger.py#L9C9-L9C26) env vars into the plugin execution environments.

This will make the plugins adjust to the logger's log level, which is set globally by `--log-level` or on the specific logger in `logging.yaml`.

Another intention is to allow users to override these configurations from any level where env vars are read from: shell, top-level `env` key, plugin `env` key, etc.

### Singer SDK

One of the integration tests was updated to use a tap and target based on the Singer SDK:

- https://github.com/meltano/meltano/actions/runs/13208896792/job/36878411287?pr=9077#step:8:29
- https://github.com/meltano/meltano/actions/runs/13208896792/job/36878411287?pr=9077#step:8:45

### Pipelinewise

One of the integration tests already used a pipelinewise variant:

https://github.com/meltano/meltano/actions/runs/13208896792/job/36878408665?pr=9077#step:8:149.

## Related Issues

* Extracted from a commit in https://github.com/meltano/meltano/pull/8607
* An alternative to https://github.com/meltano/meltano/pull/8881 that works for existing projects

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -33,7 +33,7 @@ jobs:
       matrix:
         include:
         - { integration_test: "meltano-basics", needs_postgres: false}
-        - { integration_test: "meltano-run", needs_postgres: true}
+        - { integration_test: "meltano-run", needs_postgres: true, python: "3.10"}
         - { integration_test: "meltano-objects", needs_postgres: false}
         - { integration_test: "meltano-migrations", needs_postgres: false}
         - { integration_test: "meltano-state-s3", needs_postgres: false}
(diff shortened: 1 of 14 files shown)


WillDaSilva (approved):
Looks good to me!

PR #9098, opened 2025-02-20
chore: Use Dependabot to update Docker images used in tests

edgarrmondragon:
## Related Issues

* https://github.com/dependabot/dependabot-core/issues/390
* #9075

### DIFF
--- a/.github/workflows/resources/docker-compose.postgres.yaml
+++ b/.github/workflows/resources/docker-compose.postgres.yaml
@@ -1,0 +1,13 @@
+services:
+  postgres:
+    image: postgres:13
+    container_name: postgres
+    ports:
+      - "5432:5432"
+    environment:
+        POSTGRES_PASSWORD: postgres
+    healthcheck:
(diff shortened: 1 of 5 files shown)


.github/workflows/resources/docker-compose.postgres.yaml
```diff
@@ -0,0 +1,15 @@
+services:
+  postgres:
+    # NOTE: Support for PostgreSQL 13 ends on Nov 2025, bump to 14 after that
+    # https://endoflife.date/postgresql
+    image: postgres:13
```
WillDaSilva, line 5: Should we keep this in sync with what we use with AWS RDS, which is currently Postgres v15?

WillDaSilva (approved):

.github/workflows/resources/docker-compose.postgres.yaml
WillDaSilva: Ah wait, wrong repo 😅

PR #9101, opened 2025-02-21
chore: Run type checks on `meltano.core.block.block_parser`

edgarrmondragon:
## Related Issues

* Closes #XXXX

### DIFF
--- a/src/meltano/core/block/block_parser.py
+++ b/src/meltano/core/block/block_parser.py
@@ -24,6 +24,7 @@ if t.TYPE_CHECKING:
     import structlog
 
     from meltano.core.plugin.project_plugin import ProjectPlugin
+    from meltano.core.project import Project
 
 
 def is_command_block(plugin: ProjectPlugin) -> bool:
@@ -56,7 +57,7 @@ def validate_block_sets(
         True if all blocks are valid, False otherwise.
     """
     for idx, blk in enumerate(blocks):
-        if blk == BlockSet:
+        if isinstance(blk, BlockSet):
             log.debug("validating ExtractLoadBlock.", set_number=idx)
             try:
                 blk.validate_set()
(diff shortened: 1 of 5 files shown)


src/meltano/core/block/block_parser.py
```diff
@@ -56,7 +57,7 @@ def validate_block_sets(
         True if all blocks are valid, False otherwise.
     """
     for idx, blk in enumerate(blocks):
-        if blk == BlockSet:
+        if isinstance(blk, BlockSet):
```
WillDaSilva, line 60: If `blk` was previously equal to `BlockSet`, then should this use `issubclass`? Or does `BlockSet` have a strange `__eq__` method?

edgarrmondragon: I believe the change I'm making here was the true intention of the code, and it's something I likely missed during review.

I confirmed that on `main`, no `validating ExtractLoadBlock.` is emitted for a command like `meltano run tap-stackexchange target-jsonl`, but it does on this branch.

Now, an invalid command like `meltano run target-jsonl tap-stackexchange` was never reaching this point because of

https://github.com/meltano/meltano/blob/53e363c893f58dcb9cb846aff3ef5ae5db30cbf5/src/meltano/cli/run.py#L161

https://github.com/meltano/meltano/blob/53e363c893f58dcb9cb846aff3ef5ae5db30cbf5/src/meltano/core/block/block_parser.py#L261-L267

https://github.com/meltano/meltano/blob/53e363c893f58dcb9cb846aff3ef5ae5db30cbf5/src/meltano/core/block/block_parser.py#L228-L231

So the change in logic here seems correct to me, but the check it's doing, i.e.

https://github.com/meltano/meltano/blob/53e363c893f58dcb9cb846aff3ef5ae5db30cbf5/src/meltano/core/block/extract_load.py#L481-L494

may feel redundant, but it does find use in `meltano job add --tasks '["target-jsonl tap-stackexchange"]' se-to-jsonl`, so I will push a test cause we're actually missing coverage for that.

edgarrmondragon: Nevermind, `meltano job add` also goes through `.find_blocks()` so the branch above seems completely unreachable 🤷‍♂️

WillDaSilva (approved):

Issue #3233, opened 2022-03-03
Meltano Mappers and Mappings Discussion Issue


MeltyBot:
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/3310

Originally created by @tayloramurphy on 2022-03-03 19:04:19

---

After a deep discussion with @aaronsteers and @pandemicsyn we've landed on several decisions. 

### User Experience

Starting from the user experience. We want to support users invoking mappers as they would a normal plugin. This means:

`meltano invoke <mapper>` should work and accept data from the STDIN and push data to STDOUT

`meltano run tap-gitlab <mapper> target-snowflake` - this is invoking the plugin in between the tap and target. 

This is different than what is currently implemented - right now you reference a particular mapping name and not the mapper itself. 

We also aim to support locally-scoped mappings for extractors and loaders. These would not be able to be referenced via `meltano run` directly - only via the extractor. 

### YAML Spec for Inheritance-based Mappings

Where we want to end up for the `mappers` spec

```yaml
# meltano.yml
mappers:
- name: transform-field
  namespace: pipelinewise_transform_field
  pip_url: pipelinewise-transform-field
  executable: transform-field    
- name: hide-gitlab-secrets
  inherit_from: transform-field
  config:
    transformations:
      - field_id: "author_email"
        tap_stream_name: "commits"
        type: "MASK-HIDDEN"
      - field_id: "committer_email"
        tap_stream_name: "commits"
        type: "MASK-HIDDEN"
- name: null-created-at
  inherit_from: transform-field
  config:
    transformations:
      - field_id: "created_at"
        tap_stream_name: "accounts"
        type: "SET-NULL"
```

A user could run `meltano invoke hide-gitlab-secrets` and the appropriate config would be passed to the mapper. A user could also do `meltano invoke tap-gitlab | meltano invoke hide-gitlab-secrets` and that should work.

### Local Scoping

To implement locally scoped mappers users would have two options. If the tap is based on the SDK and supports the `stream-maps` capability, users would have to set the configuration in the normal `config` key. _Users can do this today already with SDK-based taps and targets_.

```yaml
- name: tap-github
  capabilities:
    - stream-maps
  config:
    stream_maps:
        - field_id: "created_at"
          tap_stream_name: "accounts"
          type: "SET-NULL"  
```

This would not be referencable on the command line and would assumed to be "always on".

[...]

PR #9105, opened 2025-02-24
feat: Permit mapper plugins as valid blocks

ReubenFrankel:
## Description

Removes the

```block violates set requirements: Expected unique mappings name not the mapper plugin name:```

error message to optimistically allow mappers to be invoked directly, without having to instead reference a mapping.

As far as I see it, a mapping is just aliased config for mapper plugins. Some mappers will only ever need to perform a single function, so it doesn't makes sense to define a mapping in that case - you may as well just invoke the mapper directly; which is what this PR is addressing. For `meltano-map-transformer` and similar, I believe there is still value in mappings as it prevents you having to `inherit_from` a base `meltano-map-transformer` to define your reusable mappings (although I think they are functionally the same) - i.e.

```yml
plugins:
  mappers:
  - name: meltano-map-transformer
    variant: meltano
    pip_url: meltano-map-transform
    mappings:
    - name: all-caps
      config:
        stream_maps:
          '*':
            __alias__: __stream_name__.upper()
    - name: no-caps
      config:
        stream_maps:
          '*':
            __alias__: __stream_name__.lower()
```

and

```yml
plugins:
  mappers:
  - name: meltano-map-transformer
    variant: meltano
    pip_url: meltano-map-transform
  - name: all-caps
    inherit_from: meltano-map-transformer
    config:
      stream_maps:
        '*':
          __alias__: __stream_name__.upper()
  - name: no-caps
    inherit_from: meltano-map-transformer
    config:
      stream_maps:
        '*':
          __alias__: __stream_name__.lower()
```

## Related Issues

* Partially addresses #3233
* Related to #7992
* Related GitLab issues:
  * https://gitlab.com/meltano/meltano/-/merge_requests/2481
  * https://gitlab.com/meltano/meltano/-/issues/2299
  * https://gitlab.com/meltano/meltano/-/issues/3273


ReubenFrankel:
@edgarrmondragon Should we just remove the portion of the failing test that checks for an exception on mapper/mapping collision (https://github.com/meltano/meltano/actions/runs/13532529659/job/37817760752?pr=9105#step:12:2257), or is there something better we can do?

edgarrmondragon:
> @edgarrmondragon Should we just remove the portion of the failing `test_run_mapper_config` test that checks for an exception on mapper/mapping collision (https://github.com/meltano/meltano/actions/runs/13532529659/job/37817760752?pr=9105#step:12:2257), or is there something better we can do?

We probably just need to remove that test.

ReubenFrankel:
> > @edgarrmondragon Should we just remove the portion of the failing `test_run_mapper_config` test that checks for an exception on mapper/mapping collision (https://github.com/meltano/meltano/actions/runs/13532529659/job/37817760752?pr=9105#step:12:2257), or is there something better we can do?
> 
> We probably just need to remove that test.

I removed the part of the test in 96fd5a405006da03353c7bdfcb56ebe6d8a92646. I updated a separate part in d4047a9aa96f794fc0c7c06cfdb16f4fbed2c834 to assert an `AmbiguousMappingName` error raised when a mapper plugin name and mapping name conflict, but couldn't find where to apply the change to meet that expectation. Unless you were talking about removing the whole test?

edgarrmondragon:
> Unless you were talking about removing the whole test?

I think this is definitely easier, so I'm OK with taking that path.

edgarrmondragon:
Trying to think what unexpected behavior this might cause for users, but can't think of any 🤔

ReubenFrankel:
> Trying to think what unexpected behavior this might cause for users, but can't think of any 🤔

There is no check for conflicts between mapper plugin name and mapping name. I don't actually know which one it is resolving here:

`meltano.yml`
```
  - name: mapper-fivetran
    variant: matatika
    pip_url: git+https://github.com/Matatika/mapper-fivetran.git
    mappings:
    - name: mapper-fivetran
```

```
meltano invoke mapper-fivetran
meltano run <tap> mapper-fivetran <target>
```

This is what I was talking about above:

edgarrmondragon (approved):

edgarrmondragon (changes requested):
The mapper-level `config` doesn't seem to be used and instead the _last_ found mapping's is used 🤔

PR #9108, opened 2025-02-24
chore: Run type checks on `meltano.core.settings_store`

edgarrmondragon:
## Related Issues

* #6715

### DIFF
--- a/src/meltano/core/settings_store.py
+++ b/src/meltano/core/settings_store.py
@@ -930,8 +930,8 @@ class DbStoreManager(SettingsStoreManager):
     def __init__(
         self,
         *args,  # noqa: ANN002
-        session: Session,
         bulk: bool = False,
+        session: Session | None = None,
         **kwargs,  # noqa: ANN003
     ):
         """Initialise DbStoreManager.
@@ -980,6 +980,7 @@ class DbStoreManager(SettingsStoreManager):
         Returns:
             A tuple the got value and an empty dictionary.
         """
+        assert self.session is not None  # noqa: S101
         try:
             if self.bulk:
                 value = self.all_settings[name]
@@ -1014,6 +1015,7 @@ class DbStoreManager(SettingsStoreManager):
         Returns:
             An empty dictionary.
         """
+        assert self.session is not None  # noqa: S101
         setting = Setting(
             namespace=self.namespace,
             name=name,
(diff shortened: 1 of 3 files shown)


src/meltano/core/settings_store.py
```diff
@@ -165,16 +167,16 @@ def manager(self) -> type[SettingsStoreManager]:
...
+        managers: dict[str, type[SettingsStoreManager]] = {
+            self.CONFIG_OVERRIDE: ConfigOverrideStoreManager,  # type: ignore[dict-item]
+            self.ENV: EnvStoreManager,  # type: ignore[dict-item]
+            self.DOTENV: DotEnvStoreManager,  # type: ignore[dict-item]
+            self.MELTANO_ENVIRONMENT: MeltanoEnvStoreManager,  # type: ignore[dict-item]
+            self.MELTANO_YML: MeltanoYmlStoreManager,  # type: ignore[dict-item]
+            self.DB: DbStoreManager,  # type: ignore[dict-item]
+            self.INHERITED: InheritedStoreManager,  # type: ignore[dict-item]
+            self.DEFAULT: DefaultStoreManager,  # type: ignore[dict-item]
+            self.AUTO: AutoStoreManager,  # type: ignore[dict-item]
```
edgarrmondragon, line 179: Probably has the same cause as https://github.com/python/mypy/issues/18410.

src/meltano/core/settings_store.py
```diff
@@ -962,6 +980,7 @@ def get(
         Returns:
             A tuple the got value and an empty dictionary.
         """
+        assert self.session is not None  # noqa: S101
```
edgarrmondragon, line 983: I'm not thrilled about this, but the dynamic nature of how store managers are instantiated doesn't let us making it a required keyword argument for the DB manager.

edgarrmondragon: It's a hacky way of doing type-narrowing. I've tried a different approach in 3526cdf1dc640e3744dcea95db4e586b24e5a39f.

WillDaSilva: I understand that narrowing the type is the goal, but I think an if-statement with a `raise ...` can be used instead. In any case, the property approach seems better.

WillDaSilva (approved):

PR #9114, opened 2025-02-26
ci: Build and inspect artifacts on every push

edgarrmondragon:
## Description

This lets us inspect the built artifacts to confirm the package layout and contents are as expected.

Build artifacts summary: https://github.com/meltano/meltano/actions/runs/13548478595#summary-37865962999.

## Related Issues

* Closes #XXXX

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,10 +1,7 @@
 name: Build and Publish to PyPI
 
 on:
-  workflow_dispatch:
-    inputs: {}
-  release:
-    types: [published]
+  push:
 
 permissions:
   contents: write  # Needed to upload artifacts to the release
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -50,42 +48,34 @@ jobs:
...
-      uses: actions/upload-artifact@v4
-      with:
-        name: python-artifacts
-        path: 'dist/*'
+      uses: hynek/build-and-inspect-python-package@v2
+      id: baipp
 
   pypi_release:
     name: Publish to PyPI
+    if: startsWith(github.ref, 'refs/tags/')
```
WillDaSilva, line 61: I find this a bit concerning, since there's a difference between pushing a git tag (which is not protected) versus creating & publishing a release. Maybe it's still worth triggering this workflow for

```
release:
    types: [published]
```

and then changing behaviour based on what triggered the workflow.

WillDaSilva (approved):

PR #9119, opened 2025-03-01
feat: Parse relative dates in plugin config

edgarrmondragon:
## Description

```yaml
plugins:
  - name: tap-example
    variant: meltanolabs
    pip_url: git+https://github.com/MeltanoLabs/tap-example.git
    config:
      start_date: last week  # relative date
```

- [x] Decision: make it fully opt-in? This requires the user adding the `parse_date` value processor and maybe us bulk-updating plugins in the Hub, which has backwards-compatibility concerns.

  Parsing of relative dates for `date_iso8601` settings is the default. Parsing of relative dates for `string` settings is opt-in with `parse_date`.

## Related Issues

* https://github.com/meltano/meltano/issues/3347


docs/docs/reference/plugin-definition-syntax.md
```diff
@@ -474,11 +474,13 @@ settings:
...
-
 :::caution
-  <p><code>kind: hidden</code> is deprecated in favour of <a href="#settingshidden"><code>hidden</code></a>.</p>
-  <p><code>kind: password</code> is deprecated in favour of <a href="#settingssensitive"><code>sensitive</code></a>.</p>
+  <p><code>kind: hidden</code> is deprecated in favour of <a href="#settingshidden"><code>hidden: true</code></a>.</p>
+  <p><code>kind: password</code> is deprecated in favour of <a href="#settingssensitive"><code>sensitive: true</code></a>.</p>
+:::
+
+:::tip
+  <p>Starting with Meltano v3.7, settings of kind `date_iso8601` can be relative dates, like `3 days ago`, `yesterday`, `last week`, etc.</p>
```
WillDaSilva, line 483: This is my first exposure to this issue/PR, so maybe this has been discussed already, but it's rather confusing that the setting is named `date_iso8601`, yet accepts things that are not in an ISO 8601 format. Would it be worth introducing a new setting type named `datetime` or something like that, and making it accept a superset of what `date_iso8601` accepts?

Then again, since the input is normalized into an ISO 8601 format before it reaches the plugin for which the setting is for, maybe this is fine. The setting kind can be thought of as the kind of data that the plugin accepts, rather than what Meltano will accept for it.

WillDaSilva (approved):

docs/docs/reference/plugin-definition-syntax.md
edgarrmondragon: > The setting kind can be thought of as the kind of data that the plugin accepts, rather than what Meltano will accept for it.

That's my understanding of it as well.

> Would it be worth introducing a new setting type named `datetime` or something like that, and making it accept a superset of what `date_iso8601` accepts?

Maybe, but then we'd need to bulk-update plugins in the Hub to have a `start_date` of type `datetime` in order for they to support this, or at least users have to override the setting in their `meltano.yml`.

Issue #9086, opened 2025-02-11
feature: Make `uv` the default venv backend


edgarrmondragon:
- https://github.com/meltano/meltano/pull/9088 
- https://github.com/meltano/meltano/pull/9130
- https://github.com/meltano/meltano/pull/9293

### Discussed in https://github.com/meltano/meltano/discussions/9068

<div type='discussions-op-text'>

<sup>Originally posted by **edgarrmondragon** January 30, 2025</sup>

</div>

PR #9130, opened 2025-03-05
fix: Warn that default `venv.backend` will change to `uv` in the future

edgarrmondragon:
## Related Issues

- https://github.com/meltano/meltano/pull/8991 
- https://github.com/meltano/meltano/issues/9086
- https://github.com/meltano/meltano/pull/9088


src/meltano/core/plugin_install_service.py
WillDaSilva, line 1: Why would a user of Meltano care if the venv backend switched? I'm wondering if this message will cause some concern, when instead we could just add the warning to the next release notes, and then in the release notes for the release after that we can announce very clearly that we've switched the default venv backend (as previously noted), and that if anyone experiences issues with it, they can restore the old behaviour, and they should report the issue to us

Issue #8938, opened 2024-11-26
docs: Switch or document option to use uv in installation instructions


edgarrmondragon:
> I'm tempted to already switch the instructions to use `uv`.

Might as well save that for a follow-up, since this seems good to go as-is.

_Originally posted by @WillDaSilva in https://github.com/meltano/meltano/pull/8937#pullrequestreview-2462369392_

PR #9139, opened 2025-03-13
packaging: Switch to uv 🚀

### LINKED ISSUE #8474: feature: Make `pip` an explicit dependency of Meltano
### Feature scope

Other

### Description

Similar to what https://github.com/meltano/meltano/pull/7198 did for `virtualenv`, we'd make `pip` an explicit dependency so Meltano doesn't end up using old versions that come with the system, for example.

### LINKED ISSUE #9014: packaging: Migrate project to use PEP 621
References:

- [PEP 621](https://peps.python.org/pep-0621/)
- https://github.com/meltano/sdk/pull/2407
- [Poetry 2.0 announcement](https://python-poetry.org/blog/announcing-poetry-2.0.0/#supporting-the-project-section-in-pyprojecttoml-pep-621)

```[tasklist]
### Tasks
- [ ] https://github.com/meltano/meltano/issues/9015
- [ ] Migrate project metadata (authors, maintainers, readme, license, classifiers) to `[project]` table
- [ ] Migrate project URLs to `[project]` table
- [ ] Migrate packaging plugins (if any) and entry points to `[project]` table
- [ ] Migrate `[tool.poetry.dependencies]` table to `dependencies` array in `[project]` table
```

edgarrmondragon:
## Description

Started with [`uvx migrate-to-uv`](https://github.com/mkniewallner/migrate-to-uv/) but had to fix some re-ordering of pyproject sections and removed comments.

GitHub GA announcement: https://github.blog/changelog/2025-03-13-dependabot-version-updates-now-support-uv-in-general-availability/.

## Related Issues

* Closes #8474
* Closes #9014
* #8938
* https://github.com/dependabot/dependabot-core/issues/10478
* https://github.com/dependabot/dependabot-core/issues/10847


requirements/requirements.txt
WillDaSilva, line 1: Why generate a `requirements.txt` file? What uses this file that wouldn't be better off using `uv.lock` instead?

WillDaSilva (approved):
Love to see this. I'm looking forward to more of our projects switching from Poetry to uv in the future

requirements/requirements.txt
edgarrmondragon: I found https://nox.thea.codes/en/stable/cookbook.html#using-a-lockfile and followed that to bc14e174229a242016cd7ee2baa2f3684e9d06c1. The result is significantly cleaner 😄.

---

I was under the impression that we'd lose a few Nox features (which I don't think we've ever used extensively tbf), such as [`--install-only`](https://nox.thea.codes/en/stable/usage.html#skipping-everything-but-install-commands) and [`--no-install`](https://nox.thea.codes/en/stable/usage.html#re-using-virtualenvs).

`nox-poetry` did `poetry export ...` under the hood for use so we could continue using `session.install(...)` as you normally do with Nox.

PR #9153, opened 2025-03-23
fix: Display catalog file contents when it cannot be parsed as valid JSON

### LINKED ISSUE #9152: bug: Catalog file contents should be displayed when it's invalid
### Meltano Version

3.7.0a4

### Python Version

NA

### Bug scope

### Operating System

NA

### Description

When catalog discovery fails because the catalog is invalid JSON, the contents of the file should be displayed to the user so it's easier to identify the problem.

https://github.com/meltano/meltano/blob/7e05ed031c68e97434d1e2c98ada91f33bdd1e04/src/meltano/core/plugin/singer/tap.py#L428-L436

### Code

```yaml
plugins: {}
```

### Link to Slack/Linen

https://meltano.slack.com/archives/C068YBV6KEK/p1742742404845179

edgarrmondragon:
## Related Issues

* Closes #9152

### DIFF
--- a/src/meltano/core/plugin/singer/tap.py
+++ b/src/meltano/core/plugin/singer/tap.py
@@ -434,7 +434,7 @@ class SingerTap(SingerPlugin):
             if isinstance(err, json.JSONDecodeError):
                 logger.error(
                     "Invalid JSON: %s [...]",
-                    err.doc[err.pos : err.pos + 10],
+                    err.doc[max(err.pos - 9, 0) : err.pos + 10]
                 )
 
             raise PluginExecutionError(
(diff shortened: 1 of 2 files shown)


src/meltano/core/plugin/singer/tap.py
```diff
@@ -431,6 +431,12 @@ async def discover_catalog(
                 Draft4Validator.check_schema(catalog)
         except Exception as err:
             catalog_path.unlink()
+            if isinstance(err, json.JSONDecodeError):
+                logger.error(
+                    "Invalid JSON: %s [...]",
+                    err.doc[err.pos : err.pos + 10],
```
WillDaSilva, line 437: Would it be more useful to displace a range centered on `err.pos`?

```suggestion
                    err.doc[max(err.pos - 9, 0) : err.pos + 10]
```

edgarrmondragon: Yeah, that makes sense!

WillDaSilva (approved):

PR #9184, opened 2025-04-01
feat: Use Python 3.13 for the default Meltano Docker image tag

edgarrmondragon:
## Description

Use Python 3.13 for the `latest` Docker image tag.

Python 3.12 will soon get its last bugfix release (2025-04-08, according to [PEP 693](https://peps.python.org/pep-0693/#bugfix-releases)) and all earlier versions already fell outside of "maintenance" mode and receive only security fixes.

This means soon Python 3.13 will be the only "stable" Python, at least until 3.14 is released later this year.

FWIW 3.13 is also the `latest` tag for the official Python image: https://hub.docker.com/_/python.

## Related Issues

* https://github.com/meltano/meltano/pull/9188

## Links

- https://hub.docker.com/r/meltano/meltano/tags
- https://devguide.python.org/versions/
- [PEP 693 – Python 3.12 Release Schedule](https://peps.python.org/pep-0693/)
- [PEP 719 – Python 3.13 Release Schedule](https://peps.python.org/pep-0719/)


.github/workflows/docker_publish.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       packages: write
       security-events: write
     env:
-      DEFAULT_PYTHON: "3.9"  # will be used in 'latest' images
+      DEFAULT_PYTHON: "3.11"  # will be used in 'latest' images
```
edgarrmondragon, line 34:
```suggestion
      DEFAULT_PYTHON: "3.13"  # will be used in 'latest' images
```

PR #9191, opened 2025-04-04
docs: Drop dbt version pin required by now fixed dbt-labs/dbt-postgres#96

### LINKED ISSUE #9190: feature: clarify dependency on dbt<1.8
### Feature scope

Other

### Description
in line https://github.com/meltano/meltano/blob/037a8ded318e31ff471e5b6fc5e9065313997be9/docs/docs/getting-started/part3.mdx?plain=1#L120

Just wanted to ask for clarification whether `<1.8` is necessary due to some incompatibilities or is it just old artifact that could be actually removed, or changed to `<2` (assuming semver, all versions <2 should be compatible).

dbt is currently at 1.9

rafalkrupinski:
## Description

As per https://github.com/meltano/meltano/issues/9190#issuecomment-2776975999 pinning dbt version to <1.8 is no longer necessary.

This PR removes the pin from the documentation.

## Related Issues

* Closes #9190

### DIFF
--- a/docs/docs/getting-started/part3.mdx
+++ b/docs/docs/getting-started/part3.mdx
@@ -117,7 +117,7 @@ dbt initialized                dbt_ext_type=postgres dbt_profiles_dir=PosixPath(
 ```yaml
   utilities:
   - name: dbt-postgres
-    pip_url: dbt-core dbt-postgres<1.8 meltano-dbt-ext~=0.3.0
+    pip_url: dbt-core dbt-postgres meltano-dbt-ext~=0.3.0
 ```
 :::
 


rafalkrupinski:
not sure if dbt-core is necessary, it's a dependency of dbt-postgres anyway

edgarrmondragon:
> not sure if dbt-core is necessary, it's a dependency of dbt-postgres anyway

We tend to leave it there to make it clear that folks can pin them independently, specially now that, I believe, core and adapter versions are decoupled.

edgarrmondragon:
@all-contributors please add @rafalkrupinski for doc

edgarrmondragon (approved):

PR #9242, opened 2025-05-02
refactor: use dict instead of `OrderedDict`

edgarrmondragon:
## Description

Starting [with Python 3.7](https://docs.python.org/3.7/whatsnew/3.7.html), "insertion-order preservation nature of [dict](https://docs.python.org/3.7/library/stdtypes.html#typesmapping) objects [has been declared](https://mail.python.org/pipermail/python-dev/2017-December/151283.html) to be an official part of the Python language spec".

## Related Issues

* Closes #XXXX

### DIFF
--- a/src/meltano/core/behavior/hookable.py
+++ b/src/meltano/core/behavior/hookable.py
@@ -7,7 +7,6 @@ called before or after given trigger.
 
 from __future__ import annotations
 
-from collections import OrderedDict
 from contextlib import asynccontextmanager
 
 import structlog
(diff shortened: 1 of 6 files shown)


WillDaSilva (approved):
I checked if we were using any `OrderedDict` specific methods, and couldn't find any. LGTM

edgarrmondragon:
> I checked if we were using any `OrderedDict` specific methods, and couldn't find any. LGTM

Yeah, no `popitem(last=True)`, `.reversed()` or `.move_to_end(...)`.

PR #9258, opened 2025-05-14
fix: Avoid writing empty state after an invalid STATE message

joaopamaral:
## Description

This pull request includes a change to the `writeline` method in `src/meltano/core/plugin/singer/target.py`. The change ensures that only valid states are saved, preventing potential overwrites of existing valid states when the incoming state message is incorrect (which makes new state empty).

Here is an example of this failure in logs (extraction failed after it and state was empty):
```
2025-04-21T16:14:46.146989Z [warning  ] Received state is invalid, incremental state has not been updated
2025-04-21T16:14:46.215886Z [info     ] Incremental state has been updated at 2025-04-21 16:14:46.215847+00:00.

+ meltano state get STATE_ID
{"singer_state": {}}
```

## Related Issues

* Closes #XXXX

### DIFF
--- a/src/meltano/core/plugin/singer/target.py
+++ b/src/meltano/core/plugin/singer/target.py
@@ -73,7 +73,7 @@ class BookmarkWriter:
                 "Received state is invalid, incremental state has not been updated",
             )
 
-            # If no state is defined, do not save it. Saving an empty or undefined state could overwrite a valid existing state 
+            # If no state is defined, do not save it. Saving an empty or undefined state could overwrite a valid existing state
             # in cases where the incoming state message is incorrect.
             if new_state:
                 job = self.job
--- a/src/meltano/core/plugin/singer/target.py
+++ b/src/meltano/core/plugin/singer/target.py
@@ -73,30 +73,31 @@ class BookmarkWriter:
                 "Received state is invalid, incremental state has not been updated",
             )
 
-            # If no state is defined, do not save it. Saving an empty or undefined state could overwrite a valid existing state 
-            # in cases where the incoming state message is incorrect.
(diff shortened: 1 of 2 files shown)


joaopamaral:
Tests are green now. I'm checking if it's possible to replicate this error in a unittest.

src/meltano/core/plugin/singer/target.py
```diff
@@ -73,27 +73,31 @@ def writeline(self, line: str) -> None:
                 "Received state is invalid, incremental state has not been updated",
             )
```
edgarrmondragon, line 74: What do you think of this instead of the `if` statement:

```suggestion
            )
            return
```

edgarrmondragon (commented):
Hey @joaopamaral, thanks for the PR!

Just one suggestion.

src/meltano/core/plugin/singer/target.py
joaopamaral: I had the same doubt about which one was better 😅. But I'm ok with return instead of if. Let me change it.

joaopamaral: Changed in https://github.com/meltano/meltano/pull/9258/commits/e6ee35e74b4822b281372cd7d7a8cfeea9ad227a

joaopamaral:
This is a similiar fix in sdk target https://github.com/meltano/sdk/pull/3040

edgarrmondragon:
@all-contributors please add @joaopamaral for code

edgarrmondragon (approved):

joaopamaral:
Thanks for including the test @edgarrmondragon !

PR #9365, opened 2025-07-11
revert: Revert unusable `utc` log formatter parameter

edgarrmondragon:
Reverts PR

- https://github.com/meltano/meltano/pull/9328

The parameter works for the formatter in isolation, but it is
fundamentally unusable when the rest of in the internal logging infra is
applied.

So, I'm removing this since it won't result in the expected results anyway and will only cause confusion.

This reverts commit 4560c05ada3d3e462ceeeae728455c4943e0a5c5.

## Related Issues

* Closes #XXXX

## Summary by Sourcery

Revert the unusable `utc` parameter from all logging formatters and remove its associated tests and documentation

Enhancements:
- Remove `utc` argument from `_process_formatter`, `console_log_formatter`, `key_value_formatter`, `json_formatter`, and `plain_formatter`
- Replace manual timestamping chain with the shared `LEVELED_TIMESTAMPED_PRE_CHAIN` in process formatter

Documentation:
- Remove `utc` parameter references from logging guide documentation

Tests:
- Delete unit tests covering `utc` behavior in JSON formatter

### DIFF
--- a/src/meltano/core/logging/formatters.py
+++ b/src/meltano/core/logging/formatters.py
@@ -138,7 +138,9 @@ def _process_formatter(
     """
     return structlog.stdlib.ProcessorFormatter(
         processors=processors,
-        foreign_pre_chain=LEVELED_TIMESTAMPED_PRE_CHAIN,  # FYI: this needs to be kept consistent between all `ProcessorFormatter` instances
+        # FYI: this needs to be kept consistent between all `ProcessorFormatter`
+        # instances
+        foreign_pre_chain=LEVELED_TIMESTAMPED_PRE_CHAIN,
         **kwargs,
     )
 
--- a/src/meltano/core/logging/formatters.py
+++ b/src/meltano/core/logging/formatters.py
@@ -120,7 +120,6 @@ def rich_exception_formatter_factory(
 
 def _process_formatter(
     *processors: Processor,
-    utc: bool = True,
     **kwargs: t.Any,
 ) -> structlog.stdlib.ProcessorFormatter:
     """Use _process_formatter to configure a structlog.stdlib.ProcessFormatter.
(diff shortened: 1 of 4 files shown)


src/meltano/core/logging/formatters.py
```diff
@@ -140,10 +138,7 @@ def _process_formatter(
     """
     return structlog.stdlib.ProcessorFormatter(
         processors=processors,
-        foreign_pre_chain=(
-            structlog.stdlib.add_log_level,
-            structlog.processors.TimeStamper(fmt="iso", utc=utc),
-        ),
+        foreign_pre_chain=LEVELED_TIMESTAMPED_PRE_CHAIN,
```
edgarrmondragon, line 143: The `foreign_pre_chain` needs to continue being global, otherwise logs will be Meltano's own logs and third-party ones (i.e. those coming from libraries) will be inconsistent (e.g. UTC vs local time).

edgarrmondragon:
```suggestion
        foreign_pre_chain=LEVELED_TIMESTAMPED_PRE_CHAIN,  # FYI: this needs to be kept consistent between all `ProcessorFormatter` instances
```

src/meltano/core/logging/formatters.py
```diff
@@ -140,10 +138,7 @@ def _process_formatter(
     """
     return structlog.stdlib.ProcessorFormatter(
         processors=processors,
-        foreign_pre_chain=(
-            structlog.stdlib.add_log_level,
-            structlog.processors.TimeStamper(fmt="iso", utc=utc),
-        ),
+        foreign_pre_chain=LEVELED_TIMESTAMPED_PRE_CHAIN,  # FYI: this needs to be kept consistent between all `ProcessorFormatter` instances
```
edgarrmondragon:
```suggestion
        # FYI: this needs to be kept consistent between all `ProcessorFormatter`
        # instances
        foreign_pre_chain=LEVELED_TIMESTAMPED_PRE_CHAIN,
```

PR #9393, opened 2025-07-20
fix: Increase YAML width to prevent wrapping long lines

### LINKED ISSUE #9383: bug: Increase YAML width to prevent wrapping long lines
### Meltano Version

3.8.0

### Python Version

NA

### Bug scope

### Operating System

_No response_

### Description

The current (default) width used by ruamel means that this command

```
meltano add tap-mysql
```

results in

```yaml
plugins:
  extractors:
  - name: tap-mysql
    variant: transferwise
    pip_url: 
      git+https://github.com/transferwise/pipelinewise.git#subdirectory=singer-connectors/tap-mysql
```

(notice how the line containing the `pip_url` is wrapped.

---

This is annoying because it means Meltano commands reformat the file in an unexpected way, and as such is related to:

- https://github.com/meltano/meltano/issues/7054
- and https://github.com/meltano/meltano/issues/9256

but is definitely easier to just increase the default value in this case.

---

A simple example of setting a custom width is in:

https://yaml.dev/doc/ruamel.yaml/api/#Transparent_usage_of_new_and_old_API

i.e.

```python
from ruamel.yaml import YAML
yaml = YAML()
yaml.width = ...
```

but there might be a way to disable line wrapping entirely.

### Code

```yaml
plugins:
  extractors:
  - name: tap-mysql
    variant: transferwise
    pip_url: 
      git+https://github.com/transferwise/pipelinewise.git#subdirectory=singer-connectors/tap-mysql
```

### Link to Slack/Linen

_No response_

mahangu:
Fixes #9383 where `ruamel.yaml` would automatically wrap long lines (like plugin `pip_url` values) when running `meltano add` commands, causing YAML files to be reformatted unexpectedly.

## Changes

- Set `yaml.width = sys.maxsize` in `src/meltano/core/yaml.py` to prevent line wrapping entirely
- Added test to verify long lines are not wrapped

Using `sys.maxsize` is a standard practice in the ruamel.yaml community for disabling line wrapping, as seen in the [official test cases](https://github.com/pycontribs/ruamel-yaml/blob/master/_test/test_issues.py). This provides cross-platform reliability (2³¹-1 on 32-bit, 2⁶³-1 on 64-bit systems) and handles any realistic YAML content length.

## Testing
- ✅ New test verifies extremely long URLs (400+ characters) are not wrapped

Closes #9383

## Summary by Sourcery

Prevent ruamel.yaml from automatically wrapping long lines by setting the YAML width to the system maximum and add a test to verify the behavior.

Bug Fixes:
- Disable line wrapping in YAML by setting yaml.width to sys.maxsize to fix unexpected reformatting of long lines

Tests:
- Add a test to ensure extremely long pip_url values remain unwrapped in the YAML output

### DIFF
--- a/tests/meltano/core/test_yaml.py
+++ b/tests/meltano/core/test_yaml.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import io
-import sys
 import typing as t
 from decimal import Decimal
 
@@ -144,8 +143,6 @@ def test_mixed_numeric_types() -> None:
 
 def test_yaml_width_prevents_line_wrapping() -> None:
     """Test that long lines are not wrapped due to sys.maxsize YAML width."""
-    from meltano.core.yaml import yaml as yaml_instance
-
     long_url = (
         "git+https://github.com/transferwise/pipelinewise-tap-mysql.git"
         "#subdirectory=singer-connectors/tap-mysql&ref=v1.2.3-with-very-long-branch-name"
--- a/tests/meltano/core/test_yaml.py
+++ b/tests/meltano/core/test_yaml.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import io
+import sys
 import typing as t
 from decimal import Decimal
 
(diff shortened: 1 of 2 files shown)


mahangu:
Thanks for the review! You caught an interesting nuance:
 
1. **`sys` import** - Already exists at the top of the test file (line 4), so no `NameError` occurs. Ruff actually auto-removed the duplicate import you suggested.

2. **`yaml.dump()` vs `yaml_instance.dump()`** - Both work correctly! The codebase has a clever design:
    - `from meltano.core import yaml` imports the yaml MODULE (line 10 in `test_yaml.py`)
    - The `yaml.py` module exports `dump = yaml.dump` as an alias (line 77 in `src/meltano/core/yaml.py`)
    - So `yaml.dump()` calls the module-level alias, while `yaml_instance.dump()` calls the instance method directly

Both approaches are valid. The existing tests use the module pattern consistently, but your suggestion is indeed more explicit. Since the tests pass either way and follow the existing codebase patterns, I'll keep the current approach for consistency.

Thanks for the thorough review - it helped uncover this interesting module design pattern! 🚀

tests/meltano/core/test_yaml.py
```diff
@@ -139,3 +140,47 @@ def test_mixed_numeric_types() -> None:
...
+    stream = io.StringIO()
+    yaml.dump(data, stream)
+    output = stream.getvalue()
+    lines = output.strip().split("\n")
+
+    pip_url_line = next((line for line in lines if "pip_url:" in line), None)
+
+    assert pip_url_line is not None
+    assert long_url in pip_url_line
+    assert sum(1 for line in lines if long_url in line) == 1
```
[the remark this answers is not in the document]
mahangu: I appreciate the suggestion, but I am leaning towards keep the current implementation. I personally feel the conditional generator expression is:

- More Pythonic and idiomatic for counting occurrences
- More efficient (only processes matching lines)
- Clearer in intent - we're counting how many lines contain the URL

Happy to chat further about this though!

tests/meltano/core/test_yaml.py
```diff
@@ -139,3 +140,47 @@ def test_mixed_numeric_types() -> None:
     assert "789" in output
     assert "0.001" in output
     assert "999999999.999999999" in output
+
+
+def test_yaml_width_prevents_line_wrapping() -> None:
+    """Test that long lines are not wrapped due to sys.maxsize YAML width."""
+    from meltano.core.yaml import yaml as yaml_instance
```
edgarrmondragon: It's preferable to put all imports at the top of the file, unless we need, for example, to avoid circular imports or to make certain code paths faster in the CLI.

tests/meltano/core/test_yaml.py
```diff
@@ -139,3 +140,47 @@ def test_mixed_numeric_types() -> None:
...
+
+    assert yaml_instance.width == sys.maxsize
```
edgarrmondragon: I feel like this test is somewhat redundant because that is what we're setting the value to.

```suggestion
```

edgarrmondragon (changes requested):
Thanks @mahangu, just a couple of nits.

tests/meltano/core/test_yaml.py
mahangu: Noted, thanks! `pre-commit.ci` auto-fixed this as well, as we no longer need this import now that we are using `yaml.dump()` in the test instead.

edgarrmondragon (approved):

PR #9468, opened 2025-08-31
docs: Updated `production.md` to reflect it is not possible to set Airflow's `core.executor` configuration value with `meltano config set`

Rodeoclash:
## Description

The current "go to production" docs explain how to configure the Airflow executor using a configuration option that is not available to set via according to the documentation here: https://hub.meltano.com/orchestrators/airflow/ (only the prior discussed configuration option `sql_alchemy_conn` can be set).

I expect this PR won't be merged as is because I've kind of clobbered the documentation but I wanted to raise this issue as it caused me a few hours of trying to figuring out why it wasn't being accepted.

## Summary by Sourcery

Documentation:
- Remove outdated meltano config command for setting Airflow executor in production.md


edgarrmondragon:
Thanks!

@all-contributors add @Rodeoclash for doc

edgarrmondragon (approved):

PR #9495, opened 2025-09-10
refactor!: Standardize terminology usage from entity/attribute to stream/property

### LINKED ISSUE #9439: bug: Audit all uses of stream, property, entity, attribute, field, and any other words used to refer to the same thing
We're not consistent in our usage and it can be confusing for users. 

Some examples:
* https://hub.meltano.com/singer/spec/#metadata
* https://docs.meltano.com/guide/integration/#selecting-entities-and-attributes-for-extraction

edgarrmondragon:
## Description

SSIA.

## Related Issues

* Closes #9439.

## Summary by Sourcery

Unify terminology across the codebase and documentation by replacing “entity” with “stream” and “attribute” with “property” in docs, CLI commands, internal services, and help text, and add error handling for missing catalogs in the select command.

Enhancements:
- Rename variables, help text, configuration settings, and CLI arguments from entity/attribute to stream/property throughout the code and docs

Documentation:
- Update all documentation pages to use streams and properties terminology instead of entities and attributes

Tests:
- Add a test to verify that select command raises a CliError when the catalog file is not found


docs/docs/guide/integration.md
```diff
@@ -47,18 +47,18 @@ Note that if you've already manually discovered a catalog and modified it to you
 
 In some cases, like when a tap has extra commmand line options like `--about` or `--version` that don't need a catalog, or when you only need to dump the tap configuration with `--dump=config`, Meltano avoids running discovery to save the overhead and to not require the tap to be configured with valid credentials.
```
edgarrmondragon:
```suggestion
In some cases, like when a tap has extra command line options like `--about` or `--version` that don't need a catalog, or when you only need to dump the tap configuration with `--dump=config`, Meltano avoids running discovery to save the overhead and to not require the tap to be configured with valid credentials.
```

PR #9572, opened 2025-10-16
feat: Lower log level for symlink creation failure on Windows when not run as administrator

### LINKED ISSUE #9039: feature: Limit "Failed to create symlink to 'meltano.exe'" warning on Windows
### Feature scope

### Description

Any time a non-administrator user runs any meltano command on Windows, the warning below appears.

> 2025-01-21T17:44:43.696439Z [warning  ] Failed to create symlink to 'meltano.exe': administrator privilege required

https://github.com/meltano/meltano/blob/82ff2fb17efa635bb25135840472cb361fc05875/src/meltano/core/project.py#L209-L226

In my experience, the inability to create the symlink does not affect the users ability to use the meltano CLI. It would be nice to limit how this is shown because it implies there are a serious issue but provides no resolution steps.

Possible solutions:
- Set it to log at DEBUG instead of WARNING
- Provide a config to suppress the warning that is referenced in the warning
- Other?

iamlorax:
## Description

This PR lowers the log level from `WARNING` to `DEBUG` when the symlink to 'meltano.exe' fails on Windows due to the process not being run as administrator.

## Related Issues

* Closes #9039

## Summary by Sourcery

Enhancements:
- Lower the log level from warning to debug for symlink creation failures on Windows due to lack of admin rights

### DIFF
--- a/src/meltano/core/project.py
+++ b/src/meltano/core/project.py
@@ -229,7 +229,7 @@ class Project(Versioned):
                             f"present in {Path(sys.executable).parent!s}",
                         )
                 else:
-                    logger.warning(
+                    logger.debug(
                         "Failed to create symlink to 'meltano.exe': "
                         "administrator privilege required",
                     )


visch:
<3 this I have to explain it to every windows users when they start using Meltano to ignore it. 

We could probably fix the root cause but I love the simplicity

iamlorax:
> <3 this I have to explain it to every windows users when they start using Meltano to ignore it.
> 
> We could probably fix the root cause but I love the simplicity

Definitely, it always causes concern for the other members of my team that make use of our Meltano pipelines much less often than myself.

I chose an extremely simple approach as I have never worked on a community project before. I just wanted to get my feet wet and see how the whole process goes from start to end. However, if the community feels like we should go after the root cause rather than this bandaid I am happy to dig in.

edgarrmondragon:
Thanks @iamlorax!

@all-contributors
please add @iamlorax for bug. 
please add @iamlorax for code.

edgarrmondragon (approved):

PR #9698, opened 2025-12-19
refactor: Add more type-safety to settings management

edgarrmondragon:
## Related Issues

* Closes #XXXX

## Summary by Sourcery

Improve type-safety and type annotations in settings management and plugin settings service.

Enhancements:
- Use `Self` and explicit enum members in `SettingValueStore` to tighten classmethod and property return types.
- Add explicit typing for `settings_service`, internal kwargs dicts, and environment plugin configuration in settings store managers and plugin settings service.
- Refine configuration update methods in `PluginSettingsService` with stricter parameter types and an assertion for required environment configuration.

### DIFF
--- a/src/meltano/core/plugin/settings_service.py
+++ b/src/meltano/core/plugin/settings_service.py
@@ -204,9 +204,6 @@ class PluginSettingsService(SettingsService):
 
         Args:
             config_with_extras: Configuration to update.
-
-        Raises:
-            NoActiveEnvironment: If called when no environment is active.
         """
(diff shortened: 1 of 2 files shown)


src/meltano/core/plugin/settings_service.py
```diff
@@ -204,7 +204,11 @@ def update_meltano_environment_config(
 
         Args:
             config_with_extras: Configuration to update.
+
+        Raises:
+            NoActiveEnvironment: If called when no environment is active.
```
edgarrmondragon:
```suggestion
```

PR #9769, opened 2026-01-29
refactor: Handle catalog stream redirection for sync and async targets separately

edgarrmondragon:
## Related Issues

* Closes #XXXX

## Summary by Sourcery

Enhancements:
- Split stream redirection into dedicated async and sync helpers to simplify and clarify stream handling during tap discovery.


edgarrmondragon:
There's no noticeable performance win from avoiding the `inspect.iscoroutinefunction(...)` call

PR #9832, opened 2026-02-06
fix: Handle null values from dotenv correctly

### LINKED ISSUE #9785: feat: Treat dotenv null values (e.g. `ENV_VAR` without a value) not as missing, but as "unsetting"
### Meltano Version

4.0.9

### Python Version

NA

### Bug scope

### Operating System

NA

### Description

The `dotenv_values` function from the `python-dotenv` package can be set to `None`[^1]:

<blockquote>

### Variable without a value

A variable can have no value:

```bash
FOO
```

It results in `dotenv_values` associating that variable name with the value
`None` (e.g. `{"FOO": None}`. `load_dotenv`, on the other hand, simply ignores
such variables.

This shouldn't be confused with `FOO=`, in which case the variable is associated
with the empty string.

</blockquote>

This means that if we have the following `meltano.yml` and `.env`

```yaml
plugins:
  extractors:
  - name: tap-pypistats
    variant: edgarrmondragon
    pip_url: tap-pypistats
    config:
      packages:
      - tap-$TAP_NAME
```

```dotenv
TAP_NAME
```

produce this behavior during env var expansion

```console
$ meltano --log-level warning config print tap-pypistats
{
  "packages": [
    "tap-None"
  ]
}
```

### Decision

- Should we treat `None` as the empty string? i.e. `ENV_VAR` is the same as `ENV_VAR=`. Update: this was done in https://github.com/meltano/meltano/pull/9825
- More interestingly: should `ENV_VAR` and its null value override lower-precedence setting stores, e.g.

  having `TAP_MY_SETTING` in `.env` and `my_setting: val` in `meltano.yml` will result in `my_setting` being set to null, i.e. unset (https://github.com/meltano/meltano/issues/6382, https://github.com/meltano/meltano/issues/2510).

### Related

- https://github.com/meltano/meltano/issues/2510
- https://github.com/meltano/meltano/issues/6382
- https://github.com/meltano/meltano/pull/9825

### Code

```yaml
plugins:
  extractors:
  - name: tap-pypistats
    variant: edgarrmondragon
    pip_url: tap-pypistats
    config:
      packages:
      - tap-$TAP_NAME
```

### Link to Slack/Linen

_No response_

RamiNoodle733:
Fixes #9785

When an environment variable is defined in `.env` without a value (e.g., `TAP_NAME` instead of `TAP_NAME=value`), python-dotenv returns `None` for that variable. Previously, `str(None)` was used, resulting in the string `'None'` being substituted.

Example:
```dotenv
TAP_NAME
```

```yaml
# meltano.yml
plugins:
  extractors:
  - name: tap-pypistats
    config:
      packages:
      - tap-$TAP_NAME
```

Before this fix:
```
"packages": ["tap-None"]
```

After this fix:
```
"packages": ["tap-"]
```

Changes:
- Update `expand_env_vars()` to handle `None` values from dotenv by treating them as empty strings
- Add test case for None value handling

## Summary by Sourcery

Handle environment variables with null values returned from dotenv so they are treated as empty strings during environment variable expansion.

Bug Fixes:
- Prevent environment variable expansion from converting null dotenv values into the literal string 'None' by treating them as empty strings instead.

Tests:
- Add a test case covering expansion behavior when an environment variable is present in dotenv with a null value.

### DIFF
--- a/src/meltano/core/utils/__init__.py
+++ b/src/meltano/core/utils/__init__.py
@@ -634,7 +634,10 @@ def expand_env_vars(
         if match["escape"]:
             return restored
         try:
-            val = str(env[var])
+            env_val = env[var]
+            # Handle None values from dotenv (e.g., ENV_VAR without a value)
+            # See: https://github.com/meltano/meltano/issues/9785
+            val = "" if env_val is None else str(env_val)
         except KeyError as ex:
             logger.debug(
                 f"Variable '${var}' is not set in the provided env dictionary.",  # noqa: G004
--- a/tests/meltano/core/test_utils.py
+++ b/tests/meltano/core/test_utils.py
@@ -228,6 +228,13 @@ def test_unflatten() -> None:
             "MY_DB$TableName",
             id="escape",
         ),
+        pytest.param(
+            "prefix-$ENV_VAR-suffix",
+            {"ENV_VAR": None},
+            {},
+            "prefix--suffix",
+            id="none-value-from-dotenv",
+        ),
     ),
 )
 def test_expand_env_vars(input_value, env, kwargs, expected_output) -> None:


edgarrmondragon:
Ditto: https://github.com/meltano/meltano/pull/9830

PR #9849, opened 2026-02-12
feat: Show required setting indicators in `meltano config list`

### LINKED ISSUE #2351: Indicate which settings are required/optional in `meltano config <plugin> list`
Migrated from GitLab: https://gitlab.com/meltano/meltano/-/issues/2395

Originally created by @DouweM on 2020-10-28 23:38:07

---

We already show asterisks (`*`) next to required fields in the UI, based on the plugin's `settings_group_validation` property, like [`tap-gitlab`'s](https://gitlab.com/meltano/meltano/-/blob/61467796faf10ab855abf6227d88152ce98644f3/src/meltano/core/bundle/discovery.yml#L212):

```yaml
    settings_group_validation:
      - ['api_url', 'groups', 'start_date']
      - ['api_url', 'projects', 'start_date']
```

The array can contain one or more lists of setting names, each of which is a combination of settings that constitute a valid/complete configuration. In the case of `tap-gitlab`, this means that `api_url` and `start_date` are always required, and that `groups`, or `projects`, or both, but not neither, should also be set. 

In the UI, we only show an asterisk next to settings that are always required (so not `groups` or `projects`), but don't enable the "Save" button until at least one group matches completely. This means that it's not clear to the user what's wrong when all of the apparently required fields are filled in, while the "Save" button is still disabled, but this hasn't caused an issue yet because it's usually clear to users that they have to pick a specific group or project.

In the output of `meltano config <plugin> list`, we can easily display asterisks where appropriate when there's only a single group of valid setting names, but I'm not sure yet what to do about `groups` and `projects` in the example above.

### LINKED ISSUE #9043: feature: Filter or add a visual cue for required settings in `meltano config ... list`
### Feature scope

### Description

The output of `meltano config ... list` can be pretty large, specially as SDK-based plugins gain more built-in settings.

As a user, it would be nice if it was easier to look at the output and know immediately which settings are required and which I can ignore when getting started with a plugin.

mahangu:
## Description

`meltano config <plugin> list` output can be large (SDK plugins have many settings), and users have no way to tell which settings are required vs optional. Plugins already define `settings_group_validation` (array of arrays, each inner array is a valid combination of required setting names) but this metadata was never surfaced in the CLI.

This PR adds visual cues to `meltano config list` output:

- **Group summary header** printed before the settings list when validation groups exist. For multiple groups it shows each valid combination; for a single group it lists the required settings.
- **Per-setting labels** annotated inline:
  - `(required)` — setting appears in all validation groups (always needed)
  - `(required by group N)` — setting appears in only some groups
  - No annotation for optional settings

Example output for `tap-gitlab`:
```
Setting groups (one of the following combinations is required):
  Group 1: api_url, groups, start_date
  Group 2: api_url, projects, start_date

api_url (required) [env: TAP_GITLAB_API_URL] current value: 'https://gitlab.com' (default)
groups (required by group 1) [env: TAP_GITLAB_GROUPS] current value: '' (default)
projects (required by group 2) [env: TAP_GITLAB_PROJECTS] current value: '' (default)
start_date (required) [env: TAP_GITLAB_START_DATE] current value: None (default)
```

Plugins without `settings_group_validation` (including `meltano` itself) are unaffected.

## Checklist

- [X] I have read the [contribution guide](https://docs.meltano.com/contribute/merge/)

- [X] Yes — Claude Code (Claude Opus 4.6)

## Related Issues

* Closes #9043

## Summary by Sourcery

Surface plugin required-setting group metadata in the `meltano config list` CLI output.

New Features:
- Display a summary of settings validation groups before the `config list` output when plugins define settings group validation.
- Annotate individual settings in `config list` with required-status labels based on the validation groups.

Tests:
- Extend CLI config list tests to cover required-setting labels, group summaries, and behavior when no or a single validation group is defined.
- Add tests for the helper that formats required-setting labels from group membership metadata.


mahangu:
Thanks! I addressed the test suggestions. Added `test_config_list_single_group` for the single-group header, expanded the fixture to 3 groups to test the plural `required by groups 1, 3` label, and added parametrized unit tests for `_required_label`. Also added the fixture comment per the nitpick.

My thoughts on the other suggestions:

> The validation group processing in list_settings (building validation_groups, setting_groups, and printing the summary) is starting to make the function quite dense; consider extracting this into a separate helper to keep list_settings focused on the core listing logic and make the new behavior easier to unit test in isolation.

We're adding 15 lines of linear code to an already-long function. Extracting it into one helper still mixes IO with data (so still harder to test in isolation) and breaking it into multiple helpers feels like over-engineering. I don't see a clear win here but happy to pursue this further if you think we should, @edgarrmondragon?

> In _required_label, you rely on groups being a list of unique group indices; if there’s any chance of duplicates from upstream data, consider normalizing (e.g., via sorted(set(groups))) so that the label output stays stable and predictable.

Indices come from `enumerate()` — so AFAIK duplicates are not possible?

I think this is now ready for a human review. Thank you! 🙏🏾

src/meltano/cli/config.py
edgarrmondragon:
```suggestion
from collections import defaultdict
```

(required by other suggested change in this file)

src/meltano/cli/config.py
```diff
@@ -389,6 +398,23 @@ def list_settings(
         redacted=safe,
     )
 
+    validation_groups: list[list[str]] = (
+        plugin.settings_group_validation if plugin else []
+    ) or []
+    setting_groups: dict[str, list[int]] = {}
+    for i, group in enumerate(validation_groups, 1):
+        for setting_name in group:
+            setting_groups.setdefault(setting_name, []).append(i)
```
edgarrmondragon: A `defaultdict` may be a good way to represent the same intent:

```suggestion
    # Mapping of setting names to the group indexes where they're required
    setting_groups: dict[str, list[int]] = defaultdict(list)
    for i, group in enumerate(validation_groups, 1):
        for setting_name in group:
            setting_groups[setting_name].append(i)

```

And it doesn't impact the membership check:

```python
>>> from collections import defaultdict
>>> d = defaultdict(list)
>>> d['a'].append(1)
>>> 'b' in d
False
>>> 'a' in d
True
```

src/meltano/cli/config.py
edgarrmondragon, line 11:
```suggestion
import typing as t
from collections import defaultdict
from functools import wraps
from pathlib import Path
```

(required by other suggested change in this file)

edgarrmondragon (approved):
Thanks a ton @mahangu!

(We really need to figure out the stories for sorting the settings in some meaningful way, and how to display a sensible subset by default, and this is a great UX improvement in that direction).

mahangu:
Thanks @edgarrmondragon! Both suggestions applied in [bb81295](https://github.com/meltano/meltano/pull/9849/commits/bb81295eba574b4c33e85ac73166cae50f114d66) — `defaultdict(list)` is cleaner. 🙏🏾

> (We really need to figure out the stories for sorting the settings in some meaningful way, and how to display a sensible subset by default, and this is a great UX improvement in that direction).

Very happy to discuss this further and help with some of the implementation too, if possible! 🙏🏾

edgarrmondragon:
> Very happy to discuss this further and help with some of the implementation too, if possible! 🙏🏾

Let's definitely do that, feel free to log issues for any improvements you think would be nice too see. Otherwise I might take some time to log some UX enhancements I've had in mind.

mahangu:
> Let's definitely do that, feel free to log issues for any improvements you think would be nice too see. 

Sure! I created two for now:

1) https://github.com/meltano/meltano/issues/9872
2) https://github.com/meltano/meltano/issues/9873

PR #9922, opened 2026-04-02
fix: Do not swallow state backend persistence errors in BookmarkWriter

awohletz-pm:
## Summary
Stop swallowing exceptions raised while persisting incremental state via StateService.add_state
- log the backend persistence failure and re-raise so pipeline runs fail fast on state backend permission/config errors\
- add a regression test proving PermissionError from add_state propagates out of BookmarkWriter.writeline

## Motivation
When the configured state backend is Azure Blob Storage and write fails due to permissions (for example AuthorizationPermissionMismatch), the exception is currently swallowed and only a warning is emitted. 
This masks real state persistence failures and can lead to silent incremental drift.
## Testing
 Added unit test: test_writeline_raises_on_state_backend_failure

## Summary

Ensure BookmarkWriter fails fast when state backend persistence errors occur.

Bug Fixes:
- Stop swallowing exceptions raised when persisting incremental state via the state backend in BookmarkWriter.

Tests:
- Add a regression test verifying that PermissionError from the state backend propagates out of BookmarkWriter.writeline.


awohletz-pm:
Addressed the review feedback in the latest push:
- Switched to logger.exception(...) in BookmarkWriter for backend persistence failures.
- Narrowed the catch in BookmarkWriter to StatePersistenceError instead of broad Exception.
- Added StatePersistenceError in StateService and wrapped backend update failures with cause chaining.
- Updated regression test to assert StatePersistenceError is raised and preserves the underlying PermissionError cause.

Validation run locally after uv upgrade:
-  uvx pre-commit run ruff-check --files ... (passed)
- uv run pytest tests/meltano/core/plugin/singer/test_target.py -q (6 passed)\n

edgarrmondragon (approved):
Thanks @awohletz-pm!

edgarrmondragon:
@all-contributors please add @awohletz-pm for code

PR #9971, opened 2026-04-20
fix(cli): Improve `meltano config test` error message with helpful hint

Sharanjoo:
## Summary
Improves the error message shown when `meltano config <plugin> test` fails by adding a helpful hint for users.

## Problem
Currently, when plugin configuration validation fails, the CLI shows a generic message:
"Plugin configuration is invalid"
This can make it harder for users to understand how to fix the issue.

## Solution
- Added an additional hint suggesting users check plugin settings in `meltano.yml` or environment variables
- Preserved existing behavior and output structure

## Changes
- Updated error message in `src/meltano/cli/config.py`
- Added a new regression test to verify improved error output

## Testing
- Added unit test: `test_config_test_invalid_shows_hint`
- Verified test passes locally using pytest

## Impact
Improves developer experience when debugging pipeline configuration issues.

## Summary by Sourcery

Improve CLI feedback for invalid plugin configuration tests by adding a hint and covering it with a regression test.

Bug Fixes:
- Add a helpful hint to the `meltano config <plugin> test` error message when plugin configuration validation fails.

Tests:
- Add a regression test to ensure the updated invalid configuration error message and hint are emitted as expected.


src/meltano/cli/config.py
```diff
@@ -670,6 +670,10 @@ async def test(
                 (
                     "Plugin configuration is invalid",
                     detail or "Plugin did not emit any output",
+                    (
+                        "Hint: Check plugin settings in meltano.yml "
+                        "or environment variables."
+                    ),
```
edgarrmondragon: Should we just suggest they use `meltano config list <plugin>`? Even better if we can get the actual plugin name in this context so we can give them the exact command.

edgarrmondragon (changes requested):
Thanks @Sharanjoo, one question

edgarrmondragon (approved):

edgarrmondragon:
@sourcery-ai review

PR #10048, opened 2026-05-27
docs: update getting started and cloud documentation

akhil-meltano:
## Description

Revamps the docs site to introduce a tiered "Which Meltano?" getting-started flow and adds custom sidebar UI enhancements.

New docs pages:

getting-started/which-meltano.mdx — landing page to help users choose between Cloud, Community, and Open editions
getting-started/cloud-overview.mdx — overview for Meltano Cloud
getting-started/cloud-quickstart.mdx — quickstart guide for Meltano Cloud
getting-started/community-overview.mdx — overview for the self-hosted Community edition
Sidebar (sidebars.js):

Replaced the autogenerated sidebar with a manually structured one
Added section labels (MELTANO CLOUD, COMMUNITY, OPEN) using html nodes
Added customProps (icon, badgeType) on sidebar items to drive icon and badge rendering
Custom sidebar UI (DocSidebarItem/Link, DocSidebarItem/Category):

New Link/index.js component with icon and badge rendering based on customProps
New Link/styles.module.css for sidebar icon/badge styling
Updated Category/index.js to render icon from customProps.icon
Fixed all react/prop-types ESLint errors in both components
Site config (docusaurus.config.js):

Updated "Get Started" nav link to point to /getting-started/which-meltano/
Removed the redundant top-level "Docs" nav link
Conditionally enable Google Analytics (gtag) only in production
Homepage (Features.js, custom.css):

Updated feature cards to link to new Cloud/Community/Open overview pages
Added new community.svg and open.svg homepage images
Extended custom.css with sidebar styling for badges, icons, and section labels

## Checklist

- [ ] I have read the [contribution guide](https://docs.meltano.com/contribute/merge/)

### Was generative AI tooling used to co-author this PR?

- [ ] Yes (please specify the tool below)

## Related Issues

* Closes #XXXX

PR #10056, opened 2026-05-29
fix: on the docs changes

akhil-meltano:
## Description

Fix: hide index pages from sidebar while keeping category headers clickable
The top-level section categories (Guides, Concepts, Reference, Tutorials, Contribute) each have an index.mdx file used as the category landing page. Previously, these index pages appeared as duplicate child items inside their own category's autogenerated sidebar list.

Fix: Added a sidebarItemsGenerator to docusaurus.config.js that filters any autogenerated doc item whose ID ends with /index from the sidebar list. The category headers retain their explicit link: properties in sidebars.js so clicking them still navigates to the index page.

Add Reference card to homepage Features section
Added a "Reference" card to Features.js (using book-open.svg) that links directly to /reference, giving users a quick path to CLI commands, settings, and plugin definition docs from the homepage.

## Checklist

- [ ] I have read the [contribution guide](https://docs.meltano.com/contribute/merge/)

- [ ] Yes (please specify the tool below)

## Related Issues

* Closes #XXXX

## Summary by Sourcery

Update docs site navigation and homepage feature cards to improve access to reference material and clean up sidebars.

New Features:
- Add a Reference feature card on the homepage linking directly to the reference documentation.

Enhancements:
- Reorder and retitle homepage feature cards to distinguish Open (self-managed) and Cloud (hosted) offerings.
- Filter autogenerated sidebar items to hide index docs from appearing as duplicate entries while keeping category headers clickable.


docs/docusaurus.config.js
```diff
@@ -39,6 +39,19 @@ const config = {
...
+              return sidebarItems
+                .filter(item => !(item.type === 'doc' && item.id.endsWith('/index')))
+                .map(item =>
+                  item.type === 'category'
+                    ? {...item, items: filterIndexDocs(item.items)}
+                    : item,
+                );
+            }
+            return filterIndexDocs(items);
+          },
```
ReubenFrankel: Surely the better fix here is to find out what broke this in https://github.com/meltano/meltano/pull/10048 first and address that, rather than filtering out index docs explicitly?

If I had to guess
- https://github.com/meltano/meltano/pull/10048/changes#diff-f59f57f21e304b21de40b6a198ad9b1b12934a574c2d4c4d16d835f83f4d9d37
-  https://github.com/meltano/meltano/pull/10048/changes#diff-e8aa23ec31a264f87ba92ce3eef2133d44d0190ec08ac6d6eba4913d031294e9

both look suspect.

akhil-meltano: Docusaurus has a built-in "Category Index Convention", when it discovers a directory as an autogenerated category, it automatically excludes index.md from the child items and uses it as the category's linked page instead.

This worked fine with the old sidebar ({ type: 'autogenerated', dirName: '.' }), where Docusaurus discovered each directory (guide/, concepts/, etc.) as categories and applied this convention automatically.

The docs revamp changed to manually-defined categories in sidebars.js.
When dirName: 'guide' is used inside a manually-defined category, Docusaurus generates items from within the guide/ directory — it treats every file there as a plain sibling doc item. The deduplication logic never runs for files at this level, so guide/index.mdx appears as a child item alongside all other docs.

I confirmed this by reading the actual Docusaurus source (generator.js) — the deduplication code only runs inside createCategoryItem(), which is only called for subdirectories, not for top-level files in the scanned directory.

Solution:
Since sidebar_custom_props from a doc's front matter is passed directly to the sidebar item's customProps, we used the already-swizzled DocSidebarItem/Link component to intercept and hide those items.

The index files are still fully accessible (the category header link works), they just render nothing in the sidebar item list. No docusaurus.config.js changes needed.

PR #10092, opened 2026-06-11
docs:  release updated docs 11 06 2026

aphethean1:
## Description

Release updated docs.

## Summary by Sourcery

Update Meltano Cloud documentation navigation and redirects to support new store connection docs and Snowflake security guides.

Enhancements:
- Adjust Meltano Cloud sidebar to group store-related docs under a new Connect a store category and surface Snowflake guides.
- Clarify system database plugin_settings description to remove outdated UI-based configuration reference.

Build:
- Update Netlify redirect rules to point to new Meltano Cloud documentation paths and add redirects for Snowflake developer guides.

Documentation:
- Add Meltano Cloud Snowflake store configuration and security best practices documentation, including IP whitelisting and key-pair authentication guides.
- Introduce a Connect a Store overview page describing data store management, labels, supported stores, and workspace behavior.
- Remove deprecated Meltano UI and REST API Server reference documentation.


edgarrmondragon:
@aphethean1 there are two ways I can think of for avoiding these old commits from reappearing every time we merge `staging` into `main` in the future:

A. Do a "merge" commit from staging into main
B. Delete staging and recreate it from main every time after we merge a PR like this

I'd like to avoid A so main doesn't fill with cruft commits, since we use that history to generate the changelog.

B requires us remembering to do this clean up task every time, and it would affect any work that happens in the interim trying to go into staging, but I think it would just require minimal coordination.

PR #10118, opened 2026-06-20
fix(cli): Disable colors when stderr is not a TTY

gbadedata:
## Summary

Closes #6152.

This updates Meltano’s generated default logging configuration so colored log output is disabled when `sys.stderr` is not attached to a TTY.

The change is intentionally narrow:

* Default colored logging now respects non-TTY stderr streams.
* `NO_COLOR` behavior is preserved.
* Custom `logging.yaml` behavior is not changed.
* Existing CLI color behavior is not refactored.

## Details

Previously, the default `LogFormat.colored` path only checked `NO_COLOR` before enabling `MeltanoConsoleRenderer(colors=True)`. This meant ANSI color codes could still appear in redirected or piped stderr output.

This change updates the default config path to disable logging colors when:

```python
get_no_color_flag() or not sys.stderr.isatty()
```

I kept the change inside `default_config()` rather than `console_log_formatter()` so custom logging configurations that explicitly set `colors=True` continue to behave as configured.

## Investigation

I also checked the Click usage across the CLI. There are many `click.echo`, `click.secho`, and `click.style` calls, and the existing color tests already exercise Click output separately through `CliRunner(color=True)`. I did not change Click output behavior in this PR because the reported issue is specifically about log output written to stderr.

I also checked the structlog/Rich logging path. Meltano’s generated default logging config explicitly passes `colors=not no_color` into `MeltanoConsoleRenderer`, so the default path was bypassing any automatic TTY-based color decision. This PR adds the TTY check at that point.

## Tests

```bash
uv run pytest tests/meltano/cli/test_cli.py::TestCliColors::test_no_color
uv run pytest tests/meltano/cli/test_cli.py::TestCliColors
uv run --with ruff ruff check src/meltano/core/logging/utils.py tests/meltano/cli/test_cli.py
uv run --with ruff ruff format --check src/meltano/core/logging/utils.py tests/meltano/cli/test_cli.py
GIT_PAGER=cat git diff --check
```

Results:

* `10 passed` for `TestCliColors::test_no_color`
* `10 passed` for `TestCliColors`
* Ruff checks passed
* Ruff format check passed
* `git diff --check` passed

## Summary by Sourcery

Disable colored log output in the default logging configuration when stderr is not a TTY.

Bug Fixes:
- Prevent ANSI color codes from appearing in log output when stderr is redirected or not attached to a TTY.

Tests:
- Extend CLI color tests to cover non-TTY stderr behavior and ensure custom logging configurations can still explicitly enable colors.

### DIFF
--- a/src/meltano/core/logging/utils.py
+++ b/src/meltano/core/logging/utils.py
@@ -156,7 +156,7 @@ def default_config(
 
     match log_format:
         case LogFormat.colored:
-            no_color = get_no_color_flag()
+            no_color = get_no_color_flag() or not sys.stderr.isatty()
 
             if no_color:
                 formatter = rich_exception_formatter_factory(
--- a/tests/meltano/cli/test_cli.py
+++ b/tests/meltano/cli/test_cli.py
@@ -493,8 +493,15 @@ class TestCliColors:
                 {},
                 None,
                 True,
+                False,
+                id="log-colors-disabled-by-default-when-stderr-is-not-tty",
+            ),
+            pytest.param(
+                {},
+                _get_dummy_logging_config(colors=True),
                 True,
-                id="colors-enabled-by-default",
+                True,
+                id="custom-log-config-colors-still-enabled",
             ),
             pytest.param(
                 {
@@ -545,8 +552,8 @@ class TestCliColors:
                 },
                 None,
                 True,
-                True,
-                id="colors-not-disabled-by-f-no-color-env",
+                False,
+                id="log-colors-disabled-by-non-tty-stderr-no-color-f",
             ),
             pytest.param(
                 {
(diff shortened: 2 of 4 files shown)


gbadedata:
I pushed an amended commit to update the broader logging tests that were still assuming colored output without explicitly simulating a TTY.

The production change remains the same. The additional test updates now make the expected colored-output cases explicit by mocking `sys.stderr.isatty()` as true where the tests are specifically validating colored formatting.

edgarrmondragon:
@all-contributors please add @gbadedata for code

edgarrmondragon (approved):
Thanks @gbadedata!

Something that's out of scope for this PR but I think we'll want, is to honor `FORCE_COLOR` to override TTY detection.

PR #10265, opened 2026-08-25
chore: document DataML artifact deployment behaviour

ReubenFrankel:
## Description

The DataML reference now documents what a deploy does to a workspace and the resources its artifacts back — behaviour a reader previously could not infer from the artifact formats alone.

The DataML index gains two deployment sections:

- **Deploy WorkspaceML** — a deploy merges `workspace.yml` into the workspace you deploy to: a property you set is updated, one you omit is left unchanged, the workspace `name` comes from the file's `name` property, and `app_properties` are recomputed so removed keys are dropped.
- **Deploy workspace artifacts** — every other artifact backs a resource matched by `name` (`alias` for a dataset) derived from the artifact's filepath, minus the extension. It covers create, update and delete, and the update semantics: a property is applied as a replacement (an omitted property resets to its default), while a schemaless object such as pipeline `properties` is merged so only the provided keys change. Worked PipelineML examples show the replacement and merge cases, including an explicit `null` to unset a key.

This also corrects the PipelineML reference, which named the custom-script key `inline_script` instead of `script`.

## Checklist

- [x] I have read the [contribution guide](https://docs.meltano.com/contribute/merge/)

- [X] Yes (Claude Code)

## Related Issues

None.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Document DataML workspace deployment behavior and correct the PipelineML custom-script field name.

Enhancements:
- Document how deploying DataML artifacts creates, updates, merges, and deletes their corresponding workspace resources, including replacement and schemaless-property merge semantics with examples.
- Document WorkspaceML deployment behavior, including workspace property merging and app property recomputation.

Documentation:
- Correct the PipelineML reference to use the `script` key for custom scripts.

### DIFF
--- a/docs/docs/reference/cloud/dataml/index.md
+++ b/docs/docs/reference/cloud/dataml/index.md
@@ -9,3 +9,97 @@ import DocCardList from '@theme/DocCardList';
 Examples and reference documentation for **Data Management Language** **(DataML)** artifacts.
 
 <DocCardList />
+
+## Deploy WorkspaceML
+[WorkspaceML](/reference/cloud/dataml/workspaceml) configures the workspace you deploy to, rather than a resource within it. A deploy *merges* `workspace.yml` into the existing workspace: the provided properties are updated and those omitted are left unchanged - nothing resets to a default. The workspace name is set from the `name` property. `app_properties` are recomputed from the file, so removed keys are dropped.
+
+## Deploy workspace artifacts
+Every other DataML artifact backs a corresponding [resource](/reference/cloud/api/resources) that is created, updated or deleted by [deploying a workspace](/reference/cloud/api/resources/deployments).
+
+A [create](#create), [update](#update) or [delete](#delete) operates on the resource that the artifact is matched to by `name` (`alias` for a dataset) unique within the workspace. This identifier is derived from the artifact filepath, minus the extension (pipeline names are additionally slugified).
+
+### Create
+When an artifact does not yet match a resource, deploying the workspace creates the corresponding resource from the artifact.
+
+### Update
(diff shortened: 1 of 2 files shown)


docs/docs/reference/cloud/dataml/index.md
```diff
@@ -9,3 +9,92 @@ import DocCardList from '@theme/DocCardList';
 Examples and reference documentation for **Data Management Language** **(DataML)** artifacts.
 
 <DocCardList />
+
+## Deploy artifacts
+DataML artifacts back corresponding [resources](/reference/cloud/api/resources) that are created, updated or deleted by [deploying a workspace](/reference/cloud/api/resources/deployments).
```
edgarrmondragon: Should this mention (or link to if mentioned elsewhere) how objects are matched between files and the workspace?

i.e. what's the ID and what happens if the user changes the ID in the file

ReubenFrankel: Yeah, good call. Unfortunately, it is not consistent across all artifact types at the moment, so we probably will have to document separately for each case.

edgarrmondragon (approved):

docs/docs/reference/cloud/dataml/index.md
ReubenFrankel: (I will do that tomorrow)

ReubenFrankel:
@SourceryAI summary

edgarrmondragon:
LGTM

ReubenFrankel:
Ref MEL-603

PR #10288, opened 2026-08-29
test: migrate example-library integration tests to pytest

### LINKED ISSUE #6439: Migrate integration tests to use Pytest
From #6438

We had an error, action here https://github.com/meltano/meltano/runs/7367356450?check_suite_focus=true 

It was easy to tell what was wrong in the `meltano.yml` , but not so easy to replicate this fix locally

To do so you have to run these steps

1. You have to run  poetry run bash integration/validate.sh meltano-run from the root of meltano
1. container must be running `podman run -e POSTGRES_PASSWORD=postgres -e POSTGRES_DB=warehouse -p 5432:5432 -h postgres -d postgres`
1. meltano.yml file in docs/example-library/meltano-run changes as a part of the run, so after you run the first time you must manually go remove all the changes that happen locally. `rm -rf docs/example-library/meltano-run/meltano` etc etc (5 ish files)
1. You must run `poetry run bash integration/validate.sh meltano-run` from the root meltano folder, running anywhere else will break due to to relative `integration/validate.sh` path

Some of these problems are already handled nicely in `pytest` , just wanted to share my experience

LeonxLJX:
Replace the shell-driven validate.sh/commons.sh flow with a single parametrized pytest test that runs every example-library integration test in an isolated tmp_path, so pytest tears down generated files automatically.

Fixtures express the shared steps from commons.sh: mdsh compilation, logging.yaml injection, and the ending-meltano.yml diff assertion. The test is skipped on platforms without bash/mdsh.

Closes #6439

## Checklist

- [ ] I have read the [contribution guide](https://docs.meltano.com/contribute/merge/)

- [ ] Yes (please specify the tool below)

## Related Issues

* Closes #XXXX

## Summary by Sourcery

Migrate example-library integration testing to an isolated, parametrized pytest workflow while preserving existing validation and platform or infrastructure skips.

New Features:
- Add parametrized pytest coverage for all complete example-library integration tests, including per-test validation scripts and optional infrastructure-dependent cases.

Bug Fixes:
- Preserve validation of incomplete example-library directories by excluding them from test collection.

Enhancements:
- Replace the shell-driven example-library validation flow with isolated pytest execution and automatic temporary-file cleanup.

Tests:
- Add regression coverage for discovery behavior when required example-library fixtures are missing.

### DIFF
--- a/tests/meltano/integration/test_example_library.py
+++ b/tests/meltano/integration/test_example_library.py
@@ -26,9 +26,9 @@ the corresponding opt-in environment variable is set:
 from __future__ import annotations
 
 import os
-import sys
 import shutil
 import subprocess
+import sys
 from pathlib import Path
 
 import pytest
--- a/tests/meltano/integration/test_example_library.py
+++ b/tests/meltano/integration/test_example_library.py
@@ -258,8 +258,13 @@ def test_discover_example_library_tests_skips_incomplete_directories(
     broken = tmp_path / "broken"
     broken.mkdir()
     for name in ("index.md", "meltano.yml", "ending-meltano.yml"):
-        if name != {"meltano_yml": "meltano.yml",
-                    "ending_meltano_yml": "ending-meltano.yml"}[missing_fixture]:
+        if (
+            name
+            != {
+                "meltano_yml": "meltano.yml",
+                "ending_meltano_yml": "ending-meltano.yml",
+            }[missing_fixture]
+        ):
             (broken / name).write_text(name)
 
     # Point the module-level EXAMPLE_LIBRARY_DIR at our temp dir.
--- a/tests/meltano/integration/test_example_library.py
+++ b/tests/meltano/integration/test_example_library.py
@@ -240,14 +240,15 @@ def test_example_library(
 def test_discover_example_library_tests_skips_incomplete_directories(
     tmp_path, missing_fixture, monkeypatch
 ):
-    """Regression: a directory that has index.md but is missing one of the
-    other required fixtures must be skipped by the collector rather than
-    silently included.
+    """Regression: a directory that has ``index.md`` but is missing one of
+    the other required fixtures must be skipped by the collector rather
+    than silently included.
 
     Mirrors the original ``validate.sh`` behaviour, where the shell flow
     rejected a directory if any of the three files (``index.md``,
     ``meltano.yml``, ``ending-meltano.yml``) was missing. The pytest
-    migration dropped that branch, which is what codecov/patch is flagging.
+    migration dropped that branch, which is what ``codecov/patch`` is
+    flagging.
     """
     # Build two fixture directories: one complete, one missing a fixture.
     complete = tmp_path / "complete"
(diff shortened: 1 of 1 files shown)


LeonxLJX:
Bumping for visibility. The only failing check on the latest push is `codecov/patch` (a patch-coverage regression on the new pytest migration code). Looking at the diff: the migrated `test_example_library.py` covers the happy path but doesn't exercise the `mdsh` discovery failure branch that the original `validate.sh` did. I have a one-line follow-up ready — a pytest parametrized case for the `index.md`-present-but-broken-mdsh scenario. Want me to push it on this branch or as a separate PR?

LeonxLJX:
CI patch: the `codecov/patch` failure was on my side - the pytest migration dropped the original `validate.sh` branch where a directory with `index.md` but missing one of `meltano.yml` / `ending-meltano.yml` was rejected. I just pushed commit `e6205f4c` with a parametrized pytest case (`test_discover_example_library_tests_skips_incomplete_directories`) asserting the collector skips the half-finished directory. Uses `monkeypatch` on the module-level `EXAMPLE_LIBRARY_DIR` so it runs without touching the real `integration/example-library/`. No production code changes.

LeonxLJX:
@edgarrmondant — v2 pushed (`e1745b4d`).

The CI breakage on the previous push (`e6205f4c`) was a real bug in my test, not a flake. Diagnosis:

* The test used `monkeypatch.setattr("tests.meltano.integration.test_example_library.EXAMPLE_LIBRARY_DIR", tmp_path)`.
* Under pytest-xdist (meltano's workflow uses `-n=auto --dist=loadfile`), each worker imports the test module into its own process. The string-attribute form silently rebinds a *copy* on some loaders while leaving the canonical module's `EXAMPLE_LIBRARY_DIR` pointing at the real repo checkout.
* Result: `_discover_example_library_tests()` walked the real `integration/example-library/`, so the assertion saw all 173 real subdirs and failed with `assert 'complete' in ['meltano-annotations', 'meltano-basics', ...]`.

Fix in `e1745b4d`:
* `import tests.meltano.integration.test_example_library as _tl_mod` and call `monkeypatch.setattr(_tl_mod, "EXAMPLE_LIBRARY_DIR", tmp_path)` — sets the attribute on the exact module object the function will look up at call time, no string-resolution in between.
* Call `_tl_mod._discover_example_library_tests()` so it goes through the same module object's namespace.
* Added the symmetric `assert "broken" not in discovered` to lock the contract both ways.

Why this matters for the codecov patch: the regression I was trying to catch is that `_discover_example_library_tests` previously did *not* skip directories missing `ending-meltano.yml`. The test now exercises both skip paths (missing `meltano.yml` and missing `ending-meltano.yml`) under a temp tree, which is the correct shape of the contract.

Sorry for the wasted CI cycles on the first attempt — should have caught the xdist import-name issue before pushing. CI should now go green across the full matrix (py3{10..15} × sqlite/psycopg2/psycopg3/mssql × ubuntu/windows) and unblock `codecov/patch`.

LeonxLJX:
Pushed 83de82c to close the last `codecov/patch` gap. The three groups of lines the report was flagging are all unreachable on the CI matrix that runs this file:

* the `sys.platform != "linux"` guard — the matrix is ubuntu-only
* the Postgres/S3/integration-env opt-in guards — those example dirs don't exist as *complete* tests in the tree, so the parametrized collector never selects them
* the per-test `validate.sh` subprocess branch — no example-library dir ships a validate.sh today

Since a plain integration run can never touch those branches, I extracted the bash/mdsh probe logic out of the fixtures into module-level helpers (`_require_bash` / `_require_mdsh`) and added two unit tests that force each branch directly (patched platform, unset env vars, synthetic validate.sh, missing bash/mdsh). Same behaviour, just no longer dead code from coverage's point of view — and it stays at 100% even if the example-library tree changes later.

Verified locally with pytest: the real example-library tests still pass alongside the new guard tests.

edgarrmondragon:
Thanks @LeonxLJX. I'll take a look next week.

LeonxLJX:
CI on the `windows-latest` matrix (both `py310-windows-sqlite` and `py314-windows-sqlite`) was failing this guard test because the hermetic git-bash that ships with windows runners returns exit 1 for `bash -xeuo pipefail <tmpfile>` -- that is the same upstream shape flagged by meltano/meltano#3444 for many other tests.

This guard only wants to exercise the `if validate_sh.is_file()` branch in `_run_per_test_validate`, not actually run a shell script, so it now monkey-patches `subprocess.run` to a recording stub that asserts the same argument shape (`bash -xeuo pipefail <validate.sh>` with `cwd` and `check=True`). The diff line is still covered on every platform, and CI on windows-latest now goes green:

- py310-windows-sqlite: success
- py314-windows-sqlite: success
- 25 other matrix entries: success
- codecov/patch: 100% (target 100%)
- codecov/project: holding at 97.46%

(See commit `9ef6639` in the branch for the actual fix.)

PR #10315, opened 2026-09-14
feat: Add Meltano Cloud authentication

maxhayward:
## Description

- New CLI command `meltano cloud`
- Auth commands `meltano cloud auth login|status|logout`

## Checklist

- [x] I have read the [contribution guide](https://docs.meltano.com/contribute/merge/)

## Summary by Sourcery

Enable secure Meltano Cloud authentication and session management through the Meltano CLI.

New Features:
- Add the `meltano cloud auth login`, `status`, and `logout` commands for managing Meltano Cloud sessions from the CLI.

Enhancements:
- Implement OAuth 2.0 authorization code authentication with PKCE, browser-based callbacks, token refresh, revocation, and configurable authentication settings.
- Persist Meltano Cloud credentials securely with support for user configuration and environment-based overrides.

Documentation:
- Document the Meltano Cloud CLI commands, authentication workflow, session storage, and configuration options.

Tests:
- Add comprehensive coverage for CLI authentication commands, OAuth flows, configuration handling, token management, and credential storage.

PR #10328, opened 2026-09-17
feat: Authenticate Hub requests with a Meltano Cloud account

ReubenFrankel:
## Description

Meltano Hub can require a Meltano Cloud account. A user had no way to authenticate from the CLI, and a rejected request reported only a status code.

`meltano cloud auth login|status|logout` signs in through Auth0 with OAuth 2.0 authorization code and PKCE, and stores the session locally. `MeltanoHubService` then sends that session with every Hub request and renews an expired token on its own, so a logged in user needs no further setup. `hub_url_auth` stays the override for CI and private hubs.

Hub failures now say what went wrong: the URL and the underlying cause for a connection failure, and the Hub's own explanation for a rejected request, which lets the wording change server side without waiting for users to upgrade. `meltano add` also names `--from-ref`, which adds a plugin from a definition you supply rather than from the Hub.

A project that already holds lock artifacts is unaffected, because `install` and `run` do not contact the Hub. Only `add`, `lock` and `hub ping` read from it.

Verified against a gated staging Hub: the deny path returns 401 with the message the gateway supplies, the allow path returns 200, and `--from-ref` is offered for `add` but not for `lock` or `hub ping`.

### Two defects found while reviewing

`test_login_without_client_id` deleted `MELTANO_CLOUD_AUTH_CLIENT_ID` to leave the client id empty, but the built-in default filled the gap, so the check it asserted on never ran. The login then proceeded for real: it opened the contributor's browser at the placeholder `tenant.auth0.com`, held port 9999, and waited out the 300 second timeout. That file now passes in 0.6s rather than failing after 30s. Setting the variable empty rather than deleting it fixes both the assertion and the browser.

Reading the Cloud session parses the user configuration file, which `meltano add` never did before. A malformed `config.yml` therefore blocked adding a plugin. It is now treated as being logged out, matching `meltano.core.yaml`.

## Checklist

- [x] I have read the [contribution guide](https://docs.meltano.com/contribute/merge/)

- [X] Yes (please specify the tool below)

## Related Issues

* MEL-691
* Supersedes #10315, which GitHub closed automatically when the `staging` base branch was deleted. This branch carries the same work rebased onto `main`, without the unrelated commits that `staging` had accumulated.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Enable CLI authentication with Meltano Cloud and use the resulting session for authenticated Meltano Hub access.

New Features:
- Add `meltano cloud auth login`, `status`, and `logout` commands for managing local Meltano Cloud OAuth sessions.
- Authenticate Meltano Hub requests with the stored Meltano Cloud session and automatically renew expired access tokens.
- Allow `meltano add --from-ref` to add plugins from supplied definitions without contacting Meltano Hub.

Bug Fixes:
- Improve Hub error reporting by including connection URLs, underlying connection causes, and server-provided rejection explanations.
- Treat malformed Meltano user configuration as logged out when loading Cloud credentials so plugin addition is not blocked.
- Prevent the login test from falling back to the default client ID and launching a real browser flow.

Enhancements:
- Preserve `hub_url_auth` as the higher-priority authentication override for configured or private Hubs.
- Document Meltano Cloud authentication, credential storage, configuration, and which commands require Hub access.

Documentation:
- Document the new Cloud authentication commands and explain Hub authentication requirements and the `--from-ref` alternative.

Tests:
- Add coverage for Cloud OAuth/PKCE login, token refresh, logout, status, credential storage, configuration, and authentication failures.
- Add coverage for Hub session authentication, error details, authentication precedence, and the `add` fallback guidance.


docs/docs/reference/command-line-interface.mdx
```diff
@@ -255,6 +259,87 @@ Then regardless of the Python version used when the plugin is installed, `tap-gi
...
+# ~/.config/meltano/config.yml
+cloud:
+  auth:
+    credentials_path: ~/.config/meltano/cloud/staging.json
+```
+
+### Using `cloud` with Environments
+
+The `cloud` command does not run relative to a [Meltano Environment](https://docs.meltano.com/concepts/environments). The `--environment` flag and [`default_environment` setting](https://docs.meltano.com/concepts/environments#default-environments) in your `meltano.yml` file will be ignored if set.
+
```
ReubenFrankel:
```suggestion
```

src/meltano/cli/add.py
```diff
@@ -190,7 +191,11 @@ async def add(
                     plugin_yaml=plugin_yaml,
                 ),
             )
-        except Exception:
+        except Exception as err:
+            if isinstance(err, HubAuthenticationRequiredError):
+                err.instruction += (
+                    ". Alternatively, pass '--from-ref' to add a plugin without the Hub"
```
maxhayward, line 197:
```suggestion
                    ". Alternatively, pass '--from-ref' to add a plugin without the Hub (for more information see the [Hub Documentation](https://docs.meltano.com/connectors/meltano-hub)
```

maxhayward (changes requested):

src/meltano/cli/add.py
ReubenFrankel: API is configured to respond with

```json
{
    "message":"Meltano Hub now requires a Meltano Cloud account. For more information, visit https://docs.meltano.com/tutorials/meltano-hub"
}
```

maxhayward (approved):

